PR #14, opened 2016-10-13
Add rudimentary ExplicitImplicit semantic rewrite.

olafurpg:
# \- Use (super experimental) 2.0 version of scalameta
- Use (super experimental) 4.0 version of scala.meta paradise
- SBT plugin injects the 4.0 paradise compiler plugin and compiles with -Dpersist.enable
- ExplicitImplicit inserts the fully qualified type name, no fancy imports yet
- Everything is tested, including the SBT plugin interaction!

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 # Trigger build 8
 build:
-  image: olafurpg/scalafix:0.0.1
+  image: ensime/ensime:v2.x-cache
   commands:
     - apt-get install nailgun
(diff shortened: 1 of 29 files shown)


olafurpg:
Superseded by #22. The approach in this PR had several problems:

* the scala.meta semantic in its current for is incomplete.
* the scala.meta semantic api is more ambitious than what scalafix needs. 
* we can move faster with our own custom semantic api.

Issue #19, opened 2016-10-26
Current documentation for sbt plugin refers to a version that can not be resolved


pedrorijo91:
After adding the plugin as referred in [the docs](https://scalacenter.github.io/scalafix/#SBT) I get

```
$ sbt scalafix
(...)
[warn]  module not found: ch.epfl.scala#sbt-scalafix;0.1.0
[warn] ==== typesafe-ivy-releases: tried
[warn]   https://repo.typesafe.com/typesafe/ivy-releases/ch.epfl.scala/sbt-scalafix/scala_2.10/sbt_0.13/0.1.0/ivys/ivy.xml
[warn] ==== sbt-plugin-releases: tried
[warn]   https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/ch.epfl.scala/sbt-scalafix/scala_2.10/sbt_0.13/0.1.0/ivys/ivy.xml
[warn] ==== local: tried
[warn]   /Users/pedrorijo/.ivy2/local/ch.epfl.scala/sbt-scalafix/scala_2.10/sbt_0.13/0.1.0/ivys/ivy.xml
[warn] ==== public: tried
[warn]   https://repo1.maven.org/maven2/ch/epfl/scala/sbt-scalafix_2.10_0.13/0.1.0/sbt-scalafix-0.1.0.pom
[warn] ==== Typesafe repository: tried
[warn]   https://repo.typesafe.com/typesafe/releases/ch/epfl/scala/sbt-scalafix_2.10_0.13/0.1.0/sbt-scalafix-0.1.0.pom
[warn] ==== Era7 maven releases: tried
[warn]   http://releases.era7.com.s3.amazonaws.com/ch/epfl/scala/sbt-scalafix_2.10_0.13/0.1.0/sbt-scalafix-0.1.0.pom
[warn] ==== Sbt in bintray releases: tried
[warn]   https://dl.bintray.com/sbt/sbt-plugin-releases/ch/epfl/scala/sbt-scalafix_2.10_0.13/0.1.0/sbt-scalafix-0.1.0.pom
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::          UNRESOLVED DEPENDENCIES         ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: ch.epfl.scala#sbt-scalafix;0.1.0: not found
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[...]

PR #20, opened 2016-10-26
Fix documentation for sbt plugin

pedrorijo91:
Current documentation for sbt plugin refers to a version that can not be resolved.

Solves #19


olafurpg:
I see version 0.1.0 in the docs: https://scalacenter.github.io/scalafix/#SBT

screen shot 2016-10-26 at 16 48 17

Could you have a cached version?

pedrorijo91:
yes, documentation referes to version 0.1.0, but that version cannot be resolved. Through [scala index](https://index.scala-lang.org/scalacenter/scalafix/scalafix) I can only see a 0.1.0-RC1 version

olafurpg:
I see, it appears I forgot to close the sonatype release. I re-published just now and 0.1.0 should sync to maven in the next 30 minutes.

The best place to update the website is from the `master` branch. All files in the ghpages branch are removed on every update, so any changes to the generated html files in the ghpages branch will disappear.

Thanks a lot for reporting! : )

PR #22, opened 2016-12-01
First semantic rewrite: ExplicitImplicit

olafurpg:
This PR implements
- a scalafix compiler plugin which exposes a custom semantic api for scalafix rewrite rules.
- a rewrite `ExplicitImplicit` that uses the semantic api provided by the compiler plugin.
- an sbt plugin that creates a `scalafix` task to run semantic rewrite rules.
- various clean ups in the scalafix repo. 

I went with implementing my own compiler plugin instead of waiting for the scala.meta semantic
API to become ready. The scala.meta semantic api is an ambitious project, which aims to solve a more general problem than scalafix aspires to solve. My plan is to use this new compiler plugin to build a custom scalafix semantic api. We can later see if some utilities/knowledge/experience learned from this approach can be later transferred into the scala.meta semantic api.

The rewrite is executed during compilation and fixes are written directly into the source files. This is a temporary solution. Ideally, we should persist the rewrites somewhere in a format that can be consumed by for example IDEs.

The sbt plugin injects the compiler plugin only for the `scalafix` command, but should otherwise be dormant and not effect regular compilation.

Review @xeno-by


scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -0,0 +1,60 @@
...
+        "Scalafix rewrite rules are executed from within the compiler " +
+        "plugin during compilation. " +
+        "Once scalafix has completed, the session is clear."
+  ) { state =>
+    // TODO(olafur) Is there a cleaner way to accomplish the same?
+    // Requirements: the plugin must be enabled only during the scalafix task/command.
+    val addScalafixCompilerPluginSetting: String =
+      s"""libraryDependencies in ThisBuild +=
+         |  compilerPlugin("ch.epfl.scala" %% "scalafix-nsc" % "$nightlyVersion")""".stripMargin
+    s"set $addScalafixCompilerPluginSetting" ::
```
olafurpg: @dwijnand any idea about a better approach? I only want to use a compiler plugin when compiling from the `scalafix` command/task. Context from the sbt gitter channel:
- https://gitter.im/sbt/sbt?at=584051871dec193f1414b318
- https://gitter.im/sbt/sbt?at=5840815a73abd79c55bff866

dwijnand: Alternatives are making a secondary project or making a new configuration. I would recommend trying the first as it's easier to get right and it's IDE-friendly (at least IntelliJ believes on "compile" and "test" scopes/configurations exist in the world).

xeno-by:
```scala
// SemanticApi.scala
package scalafix.rewrite

// The scala.meta semantic api is  not ready yet. For time being, we
// can implement our own simplified version of the semantic api to meet
// scalafix's custom needs.
trait SemanticApi {
  def typeSignature(defn: Defn): Option[Type]
}
```

So far, within `scalameta/paradise`, we've been exploring the top-down approach of first defining an API and then trying to implement its requirements. The approach explored in this repository is bottom-up: first, we start with use cases, and only then define an API that supports them. I like this, and I'm happy that our yesterday's discussion inspired you to explore this strategy!

I'm looking forward to seeing future progress in this direction and potentially adopting your efforts as `semantic.jar` in scala.meta.

core/src/main/scala/scalafix/rewrite/Rewrite.scala
```diff
@@ -1,23 +1,26 @@
...
   }
 
-  val name2rewrite: Map[String, Rewrite] = nameMap[Rewrite](
-    ProcedureSyntax,
-    VolatileLazyVal
-  )
-
-  val default: Seq[Rewrite] = name2rewrite.values.toSeq
+  val syntaxRewrites: Seq[Rewrite] = Seq(ProcedureSyntax, VolatileLazyVal)
+  val semanticRewrites: Seq[Rewrite] = Seq(ExplicitImplicit)
```
xeno-by: Why do you need to separate these rewrites?

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
dwijnand: I don't have examples at hand, sadly. I'd've loved to be able to provide you with something.

What I can provide is a suggestion to use 0.13.13's `derivedProjects` to define your derived scalafix-using project.

olafurpg: Any pointer is perfect, I'll take a look at `derivedProjects`. Thanks a lot! :)

dwijnand: And the link I looked up for you and then forgot to link :)

http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html#Synthetic+subprojects

olafurpg: Wow, that might be exactly what I'm looking for, 0.13.13 is awesome!

dwijnand: @eed3si9n is a god, I agree.

olafurpg:
Hallelujah!!! 🎉  The scripted tests passed after spending upwards of 3 hrs trying to get the ivy.home correct. Merging!

PR #23, opened 2016-12-02
Change design of sbt plugin, again.

olafurpg:
The previous implementation was problematic because the compiler plugin
didn't always load. This approach uses a similar design as scoverage,
enabling/disabling the plugin via scalacOptions.


olafurpg:
Superseded by #26

PR #28, opened 2016-12-08
Cross-build to 2.12

olafurpg:


olafurpg:
Hallelujah ⛪️ 🎉  Lesson learned, don't configure sbt.ivy.home and more generally, don't use system properties!

PR #32, opened 2016-12-13
Support hocon configuration.

olafurpg:
My experience from scalafmt is that the configuration quickly grows and using a robust configuration from the start is useful. I went with HOCON because

- HOCON appears to be the least controversial choice in the Scala community
- I can automatically read HOCON into a case class using metaconfig from scalafmt (not implemented yet since config class is small)


olafurpg:
BTW, @ShaneDelmore I setup build-info so that IntelliJ should better understand the structure of the project. Please give it a try and see if it helps. I'm happy to iterate further to lower the barrier of entry for contributors

PR #35, opened 2016-12-17
WIP: Issue 29 Xor

ShaneDelmore:
This isn't even close to ready yet, I am just making a pr to get some visibility and feedback on early trials.

### DIFF
--- a/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
+++ b/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
@@ -1,0 +1,53 @@
+package scalafix.rewrite
+
+import scala.collection.immutable.Seq
+import scala.{meta => m}
+import scalafix.util._
+import scala.meta._
+
+case object Xor2Either extends Rewrite {
+  override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
+    implicit val semanticApi: SemanticApi = getSemanticApi(ctx)
+
+    //Create a sequence of type replacements
+    val replacementTypes = List(
+      ReplaceType(t"cats.data.XorT", t"cats.data.EitherT", "EitherT"),
+      ReplaceType(t"cats.data.Xor", t"scala.util.Either", "Either"),
+      ReplaceType(t"cats.data.Xor.Left", t"scala.util.Left", "Left"),
+      ReplaceType(t"cats.data.Xor.Right", t"scala.util.Either.Right", "Right")
+    )
+
+    //Add in some method replacements
+    val replacementTerms = List(
+      ReplaceTerm(q"cats.data.Xor.Right.apply",
+                  q"scala.util.Right.apply",
+                  q"scala.util"),
+      ReplaceTerm(q"cats.data.Xor.Left.apply",
+                  q"scala.util.Left.apply",
+                  q"scala.util")
+    )
+
+    //Then add needed imports.
+    //todo - derive this from patches created, types
+    //and terms replaced
+    //Only add if they are not already imported
+    val additionalImports = List(
+      "cats.data.EitherT",
+      "cats.implicits._",
+      "scala.util.Either"
+    )
+
+    val typeReplacements =
+      new ChangeType(ast).gatherPatches(replacementTypes)
+
+    val termReplacements =
+      new ChangeMethod(ast).gatherPatches(replacementTerms)
+
+    //Make this additional imports, beyond what can be derived from the types
+    val addedImports =
+      if (typeReplacements.isEmpty && termReplacements.isEmpty) Seq[Patch]()
+      else new AddImport(ast).gatherPatches(additionalImports)
+
+    addedImports ++ typeReplacements ++ termReplacements
+  }
+}
--- a/core/src/main/scala/scalafix/util/ChangeMethod.scala
+++ b/core/src/main/scala/scalafix/util/ChangeMethod.scala
@@ -1,0 +1,53 @@
+package scalafix.util
+
+import scala.collection.immutable.Seq
+import scala.{meta => m}
+import scala.meta._
+import scalafix.rewrite._
+import syntax._
+
+case class ReplaceTerm(
+    original: m.Term,
+    replacement: m.Term,
+    importSegment: m.Term
+) {
+  require(replacement.syntax.contains(importSegment.syntax),
+          "Specify the portion of the new Term to be imported")
+}
+
+class ChangeMethod(ast: m.Tree)(implicit sApi: SemanticApi) {
+
+  import sApi._
+
+  def partialTermMatch(rt: ReplaceTerm)
+    : PartialFunction[m.Tree, (scala.meta.Term, ReplaceTerm)] = {
+    // only desugar selections
+    case t @ q"$expr.$name" & DTerm(desugared)
+        if desugared.termNames.map(_.syntax) == rt.original.termNames.map(
+          _.syntax) =>
+      //I am not sure how to get rid of this asInstanceOf
+      //But it should be safe because I have matched on DTerm above
+      t.asInstanceOf[m.Term] -> rt
+  }
+
+  def partialTermMatches(replacementTerms: Seq[ReplaceTerm])
+    : PartialFunction[m.Tree, (m.Term, ReplaceTerm)] =
+    replacementTerms.map(partialTermMatch).reduce(_ orElse _)
+
+  def terms(ptm: PartialFunction[m.Tree, (m.Term, ReplaceTerm)]) =
+    ast.collect(ptm)
+
+  def termReplacements(trms: Seq[(m.Term, ReplaceTerm)]): Seq[Patch] =
+    trms.map {
+      case (t, ReplaceTerm(oldTerm, newTerm, imported)) =>
+        val replacement = newTerm.termNames
+          .map(_.syntax)
+          .diff(imported.termNames.map(_.syntax)) //Strip off imported portion
+          .filterNot(_ == "apply") //special handling for apply, suppress as it will be inferred
+          .mkString(".")
+        Patch.replace(t, replacement)
+    }
+
+  def gatherPatches(tr: Seq[ReplaceTerm]): Seq[Patch] =
+    termReplacements(terms(partialTermMatches(tr)))
+}
--- a/core/src/main/scala/scalafix/util/ChangeType.scala
+++ b/core/src/main/scala/scalafix/util/ChangeType.scala
@@ -1,0 +1,39 @@
+package scalafix.util
+
+import scala.collection.immutable.Seq
+import scala.{meta => m}
+import scalafix.rewrite._
+
+//Provide a little structure to the replacements we will be performing
+case class ReplaceType(original: m.Type,
+                       replacement: m.Type,
+                       newString: String) {
+  def toPatch(t: m.Type): Patch = Patch.replace(t, newString)
+}
+
+class ChangeType(ast: m.Tree)(implicit sApi: SemanticApi) {
+  import sApi._
+
+  def partialTypeMatch(
+      rt: ReplaceType): PartialFunction[m.Tree, (m.Type, ReplaceType)] = {
+    case t @ DType(desugared)
+        if StructurallyEqual(desugared, rt.original).isRight =>
+      t -> rt
+  }
+
+  def partialTypeMatches(replacementTypes: Seq[ReplaceType])
+    : PartialFunction[m.Tree, (m.Type, ReplaceType)] =
+    replacementTypes.map(partialTypeMatch).reduce(_ orElse _)
+
+  def tpes(ptm: PartialFunction[m.Tree, (m.Type, ReplaceType)])
+    : Seq[(m.Type, ReplaceType)] = ast.collect { ptm }
+
+  //This is unsafe, come up with something better
+  def typeReplacements(tpes: Seq[(m.Type, ReplaceType)]): Seq[Patch] =
+    tpes.map {
+      case (tree, rt) => rt.toPatch(tree)
+    }
+
+  def gatherPatches(tr: Seq[ReplaceType]): Seq[Patch] =
+    typeReplacements(tpes(partialTypeMatches(tr)))
+}
--- a/core/src/main/scala/scalafix/util/Patch.scala
+++ b/core/src/main/scala/scalafix/util/Patch.scala
@@ -35,4 +35,24 @@ object Patch {
       .map(_.syntax)
       .mkString("")
   }
+
+  def replace(token: Token, replacement: String): Patch =
+    Patch(token, token, replacement)
+
+  def replace(tree: Tree, replacement: String): Patch =
+    Patch(tree.tokens.head, tree.tokens.last, replacement)
+
+  def insertBefore(token: Token, toPrepend: String) =
+    replace(token, s"$toPrepend${token.syntax}")
+
+  def insertBefore(tree: Tree, toPrepend: String): Patch =
+    replace(tree, s"$toPrepend${tree.syntax}")
+
+  def insertAfter(token: Token, toAppend: String) =
+    replace(token, s"$toAppend${token.syntax}")
+
+  def insertAfter(tree: Tree, toAppend: String): Patch =
+    replace(tree, s"${tree.syntax}$toAppend")
+
+  def delete(tree: Tree): Patch = replace(tree, "")
 }
--- a/core/src/test/resources/Xor/basic.source
+++ b/core/src/test/resources/Xor/basic.source
@@ -1,0 +1,36 @@
+rewrites = [Xor2Either]
+<<< xor 1
+import scala.concurrent.Future
+import cats.data.{ Xor, XorT }
+trait A {
+type MyDisjunction = Xor[Int, String]
+  val r: MyDisjunction = Xor.Right.apply("")
+  val s: Xor[Int, String] = cats.data.Xor.Left(1 /* comment */)
+  val t: Xor[Int, String] = r.map(_ + "!")
+  val nest: Seq[Xor[Int, cats.data.Xor[String, Int]]]
+  val u: XorT[Future, Int, String] = ???
+}
+>>>
+import cats.data.EitherT
+import cats.implicits._
+import scala.util.Either
+import scala.concurrent.Future
+import cats.data.{ Xor, XorT }
+trait A {
+type MyDisjunction = Either[Int, String]
+  val r: MyDisjunction = Right("")
+  val s: Either[Int, String] = Left(1 /* comment */)
+  val t: Either[Int, String] = r.map(_ + "!")
+  val nest: Seq[Either[Int, Either[String, Int]]]
+  val u: EitherT[Future, Int, String] = ???
+}
+<<< xor do not modify when not present
+import scala.concurrent.Future
+trait A {
+  val num = 1
+}
+>>>
+import scala.concurrent.Future
+trait A {
+  val num = 1
+}
--- a/scalafix-nsc/src/test/scala/cats/data/Xor.scala
+++ b/scalafix-nsc/src/test/scala/cats/data/Xor.scala
@@ -1,0 +1,17 @@
+package cats.data
+import scala.language.higherKinds
+
+sealed abstract class Xor[+A, +B] extends Product with Serializable {
+  def map[C](f: B => C) = ???
+}
+
+object Xor {
+  def left[A, B](a: A): A Xor B = Xor.Left(a)
+  def right[A, B](b: B): A Xor B = Xor.Right(b)
+  final case class Left[+A](a: A) extends (A Xor Nothing)
+  final case class Right[+B](b: B) extends (Nothing Xor B)
+}
+
+sealed abstract class XorT[F[_], A, B](value: F[A Xor B])
+
+sealed abstract class EitherT[F[_], A, B](value: F[Either[A, B]])
--- a/scalafix-nsc/src/test/scala/cats/implicits/package.scala
+++ b/scalafix-nsc/src/test/scala/cats/implicits/package.scala
@@ -1,0 +1,9 @@
+package cats
+
+import scala.language.implicitConversions
+
+package object implicits {
+  implicit class EitherOps[A, B](from: Either[A, B]) {
+    def map[C](f: B => C): Either[A, C] = ???
+  }
+}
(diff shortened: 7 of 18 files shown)


core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -13,16 +14,90 @@ case object Xor2Either extends Rewrite {
...
     // syntax structure we want and then run semantic.desugar.
-    ast.collect {
-      case t @ DType(t"cats.data.Xor") =>
-        Patch(t.tokens.head, t.tokens.last, s"Either")
+    val tpes: Seq[(Type, ReplaceType)] = ast.collect {
+      case t @ DType(t"cats.data.XorT") => t -> xorT
+      case t @ DType(t"cats.data.Xor") => t -> xor
+      case t @ DType(t"cats.data.Xor.Left") => t -> xorLeft
+      case t @ DType(t"cats.data.Xor.Right") => t -> xorRight
+    }
```
ShaneDelmore: Here I am playing around with the idea of collecting interesting nodes and information about them and generating patches later.  It doesn't add much in this case but I thought it could be useful if we attempted to do some sort of fusion later, for example the same tree might have multiple sets of appends performed on it, in which case we could merge them.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -13,16 +14,90 @@ case object Xor2Either extends Rewrite {
...
+
+    val allImports = ast.collect {
+      case t @ q"import ..$importersnel" => t -> importersnel
     }
+
+    val firstImport = allImports.head._1
+    val firstImportFirstToken = firstImport.tokens.head
+    val tokenBeforeFirstImport =
+      ast.tokens.takeWhile(_ != firstImportFirstToken).lastOption
```
ShaneDelmore: This is terrible.  Anyone have ideas for better ways to find the top level import block?  Currently I just grab the first, then the token before the first as I found that if the first becomes deleted (Patched with "") then I can not insert before it.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -13,16 +14,90 @@ case object Xor2Either extends Rewrite {
...
+      kv =>
+        kv._1 -> kv._2
+          .flatMap(_.importees)
+          .partition(imp => List("Xor", "XorT").contains(imp.syntax)))
+
+    //If no importees to be kept, delete the whole importer
+    val removedImports = partitionedImports.collect {
+      case (t, (remove, keep)) if keep.isEmpty => Seq(Patch.delete(t))
+      case (t, (remove, keep)) => remove.map(Patch.delete)
+    }.flatten
```
ShaneDelmore: This is a better example of where merging actions on a tree may be useful.  I was considering the example where I had separated Xor2Either and XorT2EitherT into two rewrites, in that case if I collected the removals of imports I may see that all importees below a given importer are removed, and then remove the whole importer in one shot.  Again, not really sure if this is worth the complexity yet.

core/src/main/scala/scalafix/util/Patch.scala
```diff
@@ -35,4 +35,24 @@ object Patch {
...
+  def insertBefore(tree: Tree, toPrepend: String): Patch =
+    replace(tree, s"$toPrepend${tree.syntax}")
+
+  def insertAfter(token: Token, toAppend: String) =
+    replace(token, s"$toAppend${token.syntax}")
+
+  def insertAfter(tree: Tree, toAppend: String): Patch =
+    replace(tree, s"${tree.syntax}$toAppend")
+
+  def delete(tree: Tree): Patch = replace(tree, "")
```
ShaneDelmore: These builders feel useful but I need to find a better way to insert before and after without modifying the actual tree as I am currently doing.  With this definition if I insertBefore a tree, then later replace the tree, the insertBefore is lost.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
ShaneDelmore: I need to figure out what to do if there are no imports, and also what to do if there is a wildcard import.

ShaneDelmore: I also need to be smarter about existing imports in general.  Since I am collecting up all imports and patches, I could just find imports needed per patch that are not already imported.  Maybe I need to switch to a structure with more information than just patch to make that convenient.

core/src/main/scala/scalafix/util/Patch.scala
olafurpg: These are useful indeed. I had thought of making `Patch` an adt of `Add`/`Delete`/`Replace` operations. The `Add` and `Delete` operations could resolve conflicts on patches for the same token, but care needs to be taken for the order of things.

olafurpg: I've discussed with Eugene about merging  `Patch` into scalameta/scalameta once we've figured out the details on how to make it robust and useful

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
olafurpg: Imports are still an open question, and they will require semantic information to handle hairy cases like relative imports etc. We had a brief discussion on gitter https://gitter.im/scalacenter/scalafix?at=58526907058ca967374c48b3

I think that imports will eventually be handled by a custom import rewrite. That rewrite will probably take liberty to force a particular import style

```scala
import a.b
import a.c

// OR
import a.{b, c}
```

Users might need to run the import rewrite once before running other scalafix rewrites in order to minimize the diff. Similar to formatting.

I suggest you don't worry too much about imports for now. Just let them be.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
olafurpg: See comment below on imports.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -13,16 +14,90 @@ case object Xor2Either extends Rewrite {
...
+    val xor: ReplaceType =
+      ReplaceType(t"cats.data.Xor", t"scala.util.Either", "Either")
+    val xorLeft: ReplaceType =
+      ReplaceType(t"cats.data.Xor.Left", t"scala.util.Left", "Left")
+    val xorRight: ReplaceType =
+      ReplaceType(t"cats.data.Xor.Right", t"scala.util.Either.Right", "Right")
+
+    //experimenting to see if I can create lists of types to try to match.  No luck so far.
+    val replacementTypes: List[ReplaceType] =
+      List(xorT, xor, xorLeft, xorRight)
```
olafurpg: I think you can use the `unapply` method on `Type` nodes

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -13,16 +14,90 @@ case object Xor2Either extends Rewrite {
...
+                           replacement: Type,
+                           newString: String)
+    val xorT: ReplaceType =
+      ReplaceType(t"cats.data.XorT", t"scala.util.EitherT", "EitherT")
+    val xor: ReplaceType =
+      ReplaceType(t"cats.data.Xor", t"scala.util.Either", "Either")
+    val xorLeft: ReplaceType =
+      ReplaceType(t"cats.data.Xor.Left", t"scala.util.Left", "Left")
+    val xorRight: ReplaceType =
+      ReplaceType(t"cats.data.Xor.Right", t"scala.util.Either.Right", "Right")
```
olafurpg: Down the road, we may be able to generalize rewrites so they can be implement from a `Seq[(Tree, Tree)]`. Very exciting.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -1,28 +1,55 @@
...
+      "cats.data.EitherT",
+      "cats.implicits._",
+      "scala.util.Either"
+    )
+
+    val typeReplacements = typeChanger.gatherPatches(replacementTypes)
+    val termReplacements = methodChanger.gatherPatches(replacementTerms)
+    val addedImports = importAdder.gatherPatches(additionalImports)
+
+    addedImports ++ typeReplacements ++ termReplacements
```
ShaneDelmore: This class is now a rough approximation of what I think a nice api for creating type refactorings might look like.  List some types to replace, list some methods to replace, and either derive new imports needed or list those as well.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -1,28 +1,55 @@
...
+
+    //Then add needed imports.
+    //todo - derive this from patches created, types
+    //and terms replaced
+    //Only add if they are not already imported
+    val additionalImports = List(
+      "cats.data.EitherT",
+      "cats.implicits._",
+      "scala.util.Either"
+    )
```
ShaneDelmore: I ran out of time tonight to create more strongly typed versions of imports.

core/src/main/scala/scalafix/util/ChangeMethod.scala
```diff
@@ -0,0 +1,41 @@
...
+  object DType extends Desugared[m.Type]
+  object DTerm extends Desugared[m.Term]
+
+  //This isn't very satisfying.  I am trying to match on a method, without including
+  //it's arguments
+  def partialTermMatch(rt: ReplaceTerm)
+    : PartialFunction[m.Tree, (scala.meta.Term, ReplaceTerm)] = {
+    case t @ DTerm(desugared)
+        if desugared.syntax.startsWith(rt.original.syntax) &&
+          desugared.syntax.endsWith("]") =>
```
ShaneDelmore: I had to resort to string matching here. If anyone has better ideas I would love the feedback.

core/src/main/scala/scalafix/util/ChangeType.scala
```diff
@@ -0,0 +1,43 @@
...
+}
+
+class ChangeType(ast: m.Tree, ctx: RewriteCtx)(implicit sApi: SemanticApi) {
+  object DType extends Desugared[m.Type]
+  object DTerm extends Desugared[m.Term]
+
+  def partialTypeMatch(
+      rt: ReplaceType): PartialFunction[m.Tree, (m.Type, ReplaceType)] = {
+    case t @ DType(desugared)
+        if StructurallyEqual(desugared, rt.original).isRight =>
```
ShaneDelmore: I was able to use StructurallyEqual for types, but haven't figured out how to do it for methods yet without losing the arguments when I create patches.  I end up matching both the method, and the method application with args, which overwrites them when I would like to not match on application.

scalafix-nsc/src/test/scala/cats/data/Xor.scala
```diff
@@ -1,10 +1,17 @@
 package cats.data
+import scala.language.higherKinds
 
-sealed abstract class Xor[+A, +B] extends Product with Serializable
+sealed abstract class Xor[+A, +B] extends Product with Serializable {
```
ShaneDelmore: Stub out enough of cats.data.Xor and XorT interface to be able to test rewrites without needing to add a dependency on cats

scalafix-nsc/src/test/scala/cats/implicits/package.scala
```diff
@@ -0,0 +1,9 @@
+package cats
+
+import scala.language.implicitConversions
+
+package object implicits {
```
ShaneDelmore: More stubbing of cats interfaces for testing purposes.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -1,28 +1,53 @@
...
+
+    val typeReplacements = typeChanger.gatherPatches(replacementTypes)
+    val termReplacements = methodChanger.gatherPatches(replacementTerms)
+    val addedImports = importAdder.gatherPatches(additionalImports)
```
ShaneDelmore: I am not removing imports yet.  I plan to work on this later if @olafurpg doesn't beat me to it.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -0,0 +1,55 @@
...
+import scala.meta._
+
+class Desugared[T <: Tree: Parse](implicit semantic: SemanticApi) {
+  def unapply(original: T): Option[T] = semantic.desugared(original)
+}
+
+case object Xor2Either extends Rewrite {
+  override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
+    implicit val semanticApi: SemanticApi = getSemanticApi(ctx)
+    object DType extends Desugared[Type]
```
olafurpg: I think we can move these to the `SemanticApi` trait instead of copying them around.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -1,28 +1,55 @@
...
+    val methodChanger = new ChangeMethod(ast, ctx)
+    val importAdder = new AddImport(ast, ctx)
+
+    //Create a sequence of type replacements
+    val replacementTypes = List(
+      ReplaceType(t"cats.data.XorT", t"cats.data.EitherT", "EitherT"),
+      ReplaceType(t"cats.data.Xor", t"scala.util.Either", "Either"),
+      ReplaceType(t"cats.data.Xor.Left", t"scala.util.Left", "Left"),
+      ReplaceType(t"cats.data.Xor.Right", t"scala.util.Either.Right", "Right")
+    )
```
olafurpg: Could we extract the last string like this?
```scala
@ t"scala.util.Either.Right" match {
    case t"$_.$name" => name
  }
res3: Type.Name = Right
```

core/src/main/scala/scalafix/util/ChangeMethod.scala
```diff
@@ -0,0 +1,41 @@
+package scalafix.util
+
+import scala.collection.immutable.Seq
+import scala.{meta => m}
+import scalafix.rewrite.{Desugared, RewriteCtx, SemanticApi}
+
+case class ReplaceTerm(original: m.Type, newString: String)
+
+class ChangeMethod(ast: m.Tree, ctx: RewriteCtx)(implicit sApi: SemanticApi) {
```
olafurpg: are you using `ctx`? btw, the semantic api is part of `ctx`

core/src/main/scala/scalafix/util/ChangeType.scala
```diff
@@ -0,0 +1,43 @@
...
+      t -> rt
+  }
+
+  def partialTypeMatches(replacementTypes: Seq[ReplaceType])
+    : PartialFunction[m.Tree, (m.Type, ReplaceType)] =
+    replacementTypes.map(partialTypeMatch).reduce(_ orElse _)
+
+  // NOTE. This approach is super inefficient, since we run semantic.desugar on
```
olafurpg: Ditto, this is really inefficient. But I say we don't worry about it until it's a problem.

core/src/test/resources/Xor/basic.source
```diff
@@ -0,0 +1,36 @@
...
+import scala.concurrent.Future
+import cats.data.{ Xor, XorT }
+trait A {
+type MyDisjunction = Either[Int, String]
+  val r: MyDisjunction = Right("")
+  val s: Either[Int, String] = Left(1 /* comment */)
+  val t: Either[Int, String] = r.map(_ + "!")
+  val nest: Seq[Either[Int, Either[String, Int]]]
+  val u: EitherT[Future, Int, String] = ???
+}
```
olafurpg: What about for comprehensions? Is it necessary to add `.right` on `Xor.flatmap` or does `cats.implicits` pimp up `Either`?

ShaneDelmore: flatMap and map it turns out are going to be a problem.  I have run on a few external repositories but it is not ready to merge, I need to handle imports better and more importantly I need to deal with the differences in variance between Either.map/flatMap and Xor.map/flatMap.  Do to issues in the left side variance, In some cases I will need to replace them with pattern matching.  Not looking forward to it but I don't believe there is another solution.

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
ShaneDelmore: Any ideas how I would pull "Right" out of "scala.util.Right.apply"?  I am stumped, spent a few hours on it tonight and getting nowhere. 

Quasiquotes are confusing me.  I can pull the prefix from t"cats.data.Xor.Right.apply" using t"$prefix.$" (cats.data.Xor.Right) and I can pull the name from t"cats.data.Xor.Right" using t"$.$name" but I cannot pull the prefix out, then pull the name from it (I get a match error) and also can't combine them as in t"$preprefix.$prefixname.$_" ?  I'm trying to do things the "right" way but keep running into issues that take hours with quasiquotes and would take a few minutes with string splitting.

ShaneDelmore: The biggest issue being, now the heck to I make them compose?  Say I used some sort of path to represent "scala.util.Right.apply".  With /scala/util/Right/apply it's trivial to get the name, the fully qualified name, the parent, the parents name alone, or full path, etc.  And if I make a function to get "name" which just does path.takeRight(1) and another to get parent which is path.init, it's easy to compose them and get parent's name.  With partial functions, quasiquotes, and pattern matching I am finding the few bits of scala syntax I have run into that really resist composition.  Pointers hugely appreciated.

olafurpg: I'm sad to hear this is a productivity killer.

You don't have to use quasiquotes. I'm guilty of using strings to strip of a prefix package name just because I just wanted to see something work, see [here](https://github.com/scalacenter/scalafix/blob/e2f97d651cbba38a287d86543fd36f9dbddbadda/scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala#L46). However, string manipulation is almost like programming with void pointers in C, try to avoid it when possible.

An alternative to quasiquotes that's not string manipulation is using the regular Tree ADT. I do this myself all the time ([example](https://github.com/scalacenter/scalafix/blob/e2f97d651cbba38a287d86543fd36f9dbddbadda/scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala#L31)). Quasiquotes tend to hide the fact that you're working with tree structures. With scalafmt, I found the relationship counter-intuitive between the code on the surface and the underlying parsed AST, for example

```scala
@ q"a.b.c.d".structure
res20: String = """
Term.Select(Term.Select(Term.Select(Term.Name("a"), Term.Name("b")), Term.Name("c")), Term.Name("d"))
"""
// note `d` is at the top and `a` is at the bottom
```

If you want to use quasiquotes, you might be looking for `transform` in combination with `StructurallyEqual` or `.structure`
```scala
@ val prefix = q"a.b.c"
prefix: Term.Select = a.b.c
@ val fqn = q"$prefix.d.e"
fqn: Term.Select = a.b.c.d.e
@ fqn.transform {
  case q"$somePrefix.$name" if somePrefix.structure == prefix.structure  => name
}
res15: Tree = d.e
```

core/src/main/scala/scalafix/util/ChangeMethod.scala
olafurpg: Are you doing this to overcome the inserted `.apply[T]` in the desugared output? I matched on the `.apply` in the rewrite: https://github.com/olafurpg/scalafix/blob/445b2cd783ef9d1e041198663ed335747133dfcc/core/src/main/scala/scalafix/rewrite/Xor2Either.scala#L22
Maybe I was too optimistic about the capabilities of this `desugared` approach. I am afraid this `desugared` approach will always require awkward workarounds like this.

In the future, I'm aiming for a user experience where the user can configure in `.scalafix.conf`

```
rename = [
  { from = "cats.data.Xor", to = "scala.util.Either }
]
```
and everything just works™️. But I suspect we'll need to get type equality `=:=`  in the semantic api to do that.

ShaneDelmore:
There are merge conflicts and I am not able to successfully resolve them yet for reasons I don't yet understand.  My current theory is that the desugar implementation might have changed, I am no longer able to match anything in the desugaring step from what I can see.

ShaneDelmore:
@olafurpg I tried to do the squash you requested but seem to have messed it up and cannot yet figure out how.  I will be leaving town on vacation soon, if I have a chance I will try to get to this but it may get neglected for a while.

ShaneDelmore:
I don't know the fix yet but it looks like the Desugar api is broken.  Nothing matches, if i turn on the logging statement in Desugar it only returns:
NscSemanticApi.scala:152  [result]: List()

olafurpg:
I experimented with simplifying the `desugar` implementation on my branch but that didn't work out as I had hoped to. I pushed the branch to gh out of habit but I didn't intend for that work to be used.

I'm sorry but I'm afraid I don't time to take a closer look at this until next week when I return to work. I have a feeling we may be able to add a `tpe` method to the semantic api to get the `m.Type` for most `m.Tree`. That might solve a lot of the shortcomings of the desugared approach: simpler pattern matching, no weird `.apply[$_]`, for-comprehensions.

Until then, Merry Christmas and a Happy New Year! 🎅

ShaneDelmore:
Got the merge figured out, it had nothing to do with the changes to Desugared, desugared wasn't working because I messed up the merge and lost the scalacOption -Yrangedpos which desugaring relies on.

olafurpg:
Aah, that makes sense. Yes, the current preliminary semantic api in scala.meta (which is mostly copy-paste from scalafix) asserts that yrangepos is enabled in settings: https://github.com/scalameta/paradise/blob/281484d6f5f06575af2139e00e871643af8268f4/plugin/src/main/scala/org/scalameta/paradise/mirrors/Mirrors.scala#L123

ShaneDelmore:
Added Circe test.  I am not able to get this to even run for me locally, it always passes with no changes.  Will keep working on it but pushed it to see if the CI system would run it correctly.

olafurpg:
Seems a bug in integrations was the reason why there were no diffs. Fix in https://github.com/scalacenter/scalafix/pull/37

ShaneDelmore:
I am closing this pr.  It was useful to investigate the experience of doing code rewrites with a patch based mechanism and I have now done enough to have learned I do not want to continue with this approach.  I will try the exercise again using either a transform or prism based approach next.

olafurpg:
This pr has been a valuable learning experience for me. Thank you for going
trying out this out at such an early stage. I agree, we need a higher level
patch api. I am drafting the design in my head as we speak and I hope to
implement it as soon as my right hand recovers.

PR #42, opened 2017-01-23
Add symbol-based type shortener

jvican:
This PR introduces a symbol-based type shortener to replace the string-based one. The benefits of this approach are:
* Provides more consistent behaviour with the Scala compiler
* Allows fine-grained control over what's in scope and what's not
* Helps future customization (as the whole information about scopes and symbols is available)
  
I think that there are some small details to flesh out before we nail down the desired behaviour. I'm commenting on some lines to make it clearer. I ran this on cats and I received pretty diffs, but for some reasons I was not successful at running it in scalajs.

### DIFF
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -23,6 +23,17 @@ case object ExplicitImplicit extends Rewrite {
       case _ => accum
     }
   }
+  /** Get the last sequential top level package from a source file. */
+  def getLastTopLevelPkg(potentialPkg: Stat): Stat = {
+    potentialPkg match {
+      case p: Pkg =>
+        val stats = p.stats
+        val first = stats.head
+        if (stats.size != 1) first
+        else getLastTopLevelPkg(first)
+      case _ => potentialPkg
+    }
+  }
   override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
     import scala.meta._
     val semantic = getSemanticApi(ctx)
@@ -39,7 +50,8 @@ case object ExplicitImplicit extends Rewrite {
         typ <- semantic.typeSignature(defn)
         importToken = parents(defn)
           .collectFirst {
-            case p: Pkg => p.stats.head.tokens.head
+            case p: Pkg =>
+              getLastTopLevelPkg(p).tokens.head
           }
           .getOrElse(ast.tokens.head)
       } yield {
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -43,8 +43,6 @@ case object ExplicitImplicit extends Rewrite {
           }
           .getOrElse(ast.tokens.head)
       } yield {
-        logger.elem(typ)
-
         val (shortenedTpe, missingImports) = semantic.shortenType(typ, defn)
         Patch(replace, replace, s"$replace: ${shortenedTpe.syntax}") +:
           missingImports.map(missingImport =>
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -23,17 +23,6 @@ case object ExplicitImplicit extends Rewrite {
       case _ => accum
     }
   }
-  /** Get the last sequential top level package from a source file. */
-  def getLastTopLevelPkg(potentialPkg: Stat): Stat = {
-    potentialPkg match {
-      case p: Pkg =>
-        val stats = p.stats
-        val first = stats.head
-        if (stats.size != 1) first
-        else getLastTopLevelPkg(first)
-      case _ => potentialPkg
-    }
-  }
   override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
     import scala.meta._
     val semantic = getSemanticApi(ctx)
@@ -48,19 +37,11 @@ case object ExplicitImplicit extends Rewrite {
         replace <- lhsTokens.reverseIterator.find(x =>
           !x.is[Token.Equals] && !x.is[Whitespace])
         typ <- semantic.typeSignature(defn)
-        importToken = parents(defn)
-          .collectFirst {
-            case p: Pkg =>
-              getLastTopLevelPkg(p).tokens.head
-          }
-          .getOrElse(ast.tokens.head)
       } yield {
-        val (shortenedTpe, missingImports) = semantic.shortenType(typ, defn)
-        Patch(replace, replace, s"$replace: ${shortenedTpe.syntax}") +:
-          missingImports.map(missingImport =>
-            Patch.AddLeft(importToken, s"import ${missingImport}\n"))
+        val shortenedTpe = semantic.shortenType(typ, defn)
+        Patch(replace, replace, s"$replace: ${shortenedTpe.syntax}")
       }
-    }.toSeq.flatten
+    }.toSeq
     ast.collect {
       case t @ m.Defn.Val(mods, _, None, body)
           if mods.exists(_.syntax == "implicit") &&
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -15,14 +15,6 @@ case object ExplicitImplicit extends Rewrite {
     case m.Term.ApplyType(m.Term.Name("implicitly"), _) => true
     case _ => false
   }
-  @scala.annotation.tailrec
-  final def parents(tree: Tree,
-                    accum: Seq[Tree] = Seq.empty[Tree]): Seq[Tree] = {
-    tree.parent match {
-      case Some(parent) => parents(parent, parent +: accum)
-      case _ => accum
-    }
-  }
   override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
     import scala.meta._
     val semantic = getSemanticApi(ctx)
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -3,6 +3,7 @@ package scalafix.rewrite
 import scala.{meta => m}
 import scalafix.util.Patch
 import scalafix.util.Whitespace
+import scalafix.util.logger
 
 case object ExplicitImplicit extends Rewrite {
   // Don't explicitly annotate vals when the right-hand body is a single call
@@ -27,7 +28,10 @@ case object ExplicitImplicit extends Rewrite {
         replace <- lhsTokens.reverseIterator.find(x =>
           !x.is[Token.Equals] && !x.is[Whitespace])
         typ <- semantic.typeSignature(defn)
-      } yield Patch(replace, replace, s"$replace: ${typ.syntax}")
+      } yield {
+        logger.elem(typ)
+        Patch(replace, replace, s"$replace: ${typ.syntax}")
+      }
     }.toSeq
     ast.collect {
       case t @ m.Defn.Val(mods, _, None, body)
--- a/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
+++ b/core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
@@ -1,5 +1,6 @@
 package scalafix.rewrite
 
+import scala.meta._
 import scala.{meta => m}
 import scalafix.util.Patch
 import scalafix.util.Whitespace
@@ -13,6 +14,14 @@ case object ExplicitImplicit extends Rewrite {
   private def isImplicitly(term: m.Term): Boolean = term match {
     case m.Term.ApplyType(m.Term.Name("implicitly"), _) => true
     case _ => false
+  }
+  @scala.annotation.tailrec
+  final def parents(tree: Tree,
+                    accum: Seq[Tree] = Seq.empty[Tree]): Seq[Tree] = {
+    tree.parent match {
+      case Some(parent) => parents(parent, parent +: accum)
+      case _ => accum
+    }
   }
   override def rewrite(ast: m.Tree, ctx: RewriteCtx): Seq[Patch] = {
     import scala.meta._
@@ -28,11 +37,20 @@ case object ExplicitImplicit extends Rewrite {
         replace <- lhsTokens.reverseIterator.find(x =>
           !x.is[Token.Equals] && !x.is[Whitespace])
         typ <- semantic.typeSignature(defn)
+        importToken = parents(defn)
+          .collectFirst {
+            case p: Pkg => p.stats.head.tokens.head
+          }
+          .getOrElse(ast.tokens.head)
       } yield {
         logger.elem(typ)
-        Patch(replace, replace, s"$replace: ${typ.syntax}")
+
+        val (shortenedTpe, missingImports) = semantic.shortenType(typ, defn)
+        Patch(replace, replace, s"$replace: ${shortenedTpe.syntax}") +:
+          missingImports.map(missingImport =>
+            Patch.AddLeft(importToken, s"import ${missingImport}\n"))
       }
-    }.toSeq
+    }.toSeq.flatten
     ast.collect {
       case t @ m.Defn.Val(mods, _, None, body)
           if mods.exists(_.syntax == "implicit") &&
--- a/core/src/main/scala/scalafix/util/Patch.scala
+++ b/core/src/main/scala/scalafix/util/Patch.scala
@@ -4,26 +4,34 @@ import scala.meta._
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token
 
-/**
-  * A patch replaces all tokens between [[from]] and [[to]] with [[replace]].
-  */
-case class Patch(from: Token, to: Token, replace: String) {
-  def insideRange(token: Token): Boolean =
-    token.input.eq(from.input) &&
-      token.end <= to.end &&
-      token.start >= from.start
-
-  val tokens = replace.tokenize.get.tokens.toSeq
-  def runOn(str: Seq[Token]): Seq[Token] = {
-    str.flatMap {
-      case `from` => tokens
-      case x if insideRange(x) => Nil
-      case x => Seq(x)
-    }
-  }
+sealed abstract class Patch {
+  def runOn(str: Seq[Token]): Seq[Token]
 }
 
 object Patch {
+  case class AddLeft(tok: Token, toAdd: String) extends Patch {
+    override def runOn(str: Seq[Token]): Seq[Token] = str.flatMap {
+      case `tok` => toAdd.tokenize.get.toSeq :+ tok
+      case t => List(t)
+    }
+  }
+  case class Replace(from: Token, to: Token, replace: String) extends Patch {
+    def insideRange(token: Token): Boolean =
+      token.input.eq(from.input) &&
+        token.end <= to.end &&
+        token.start >= from.start
+
+    val tokens = replace.tokenize.get.tokens.toSeq
+    def runOn(str: Seq[Token]): Seq[Token] = {
+      str.flatMap {
+        case `from` => tokens
+        case x if insideRange(x) => Nil
+        case x => Seq(x)
+      }
+    }
+  }
+  def apply(from: Token, to: Token, replace: String): Patch =
+    Replace(from, to, replace)
   def verifyPatches(patches: Seq[Patch]): Unit = {
     // TODO(olafur) assert there's no conflicts.
   }
--- a/core/src/main/scala/scalafix/util/Patch.scala
+++ b/core/src/main/scala/scalafix/util/Patch.scala
@@ -4,34 +4,26 @@ import scala.meta._
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token
 
-sealed abstract class Patch {
-  def runOn(str: Seq[Token]): Seq[Token]
+/**
+  * A patch replaces all tokens between [[from]] and [[to]] with [[replace]].
+  */
+case class Patch(from: Token, to: Token, replace: String) {
+  def insideRange(token: Token): Boolean =
+    token.input.eq(from.input) &&
+      token.end <= to.end &&
+      token.start >= from.start
+
+  val tokens = replace.tokenize.get.tokens.toSeq
+  def runOn(str: Seq[Token]): Seq[Token] = {
+    str.flatMap {
+      case `from` => tokens
+      case x if insideRange(x) => Nil
+      case x => Seq(x)
+    }
+  }
 }
 
 object Patch {
-  case class AddLeft(tok: Token, toAdd: String) extends Patch {
-    override def runOn(str: Seq[Token]): Seq[Token] = str.flatMap {
-      case `tok` => toAdd.tokenize.get.toSeq :+ tok
-      case t => List(t)
-    }
-  }
-  case class Replace(from: Token, to: Token, replace: String) extends Patch {
-    def insideRange(token: Token): Boolean =
-      token.input.eq(from.input) &&
-        token.end <= to.end &&
-        token.start >= from.start
-
-    val tokens = replace.tokenize.get.tokens.toSeq
-    def runOn(str: Seq[Token]): Seq[Token] = {
-      str.flatMap {
-        case `from` => tokens
-        case x if insideRange(x) => Nil
-        case x => Seq(x)
-      }
-    }
-  }
-  def apply(from: Token, to: Token, replace: String): Patch =
-    Replace(from, to, replace)
   def verifyPatches(patches: Seq[Patch]): Unit = {
     // TODO(olafur) assert there's no conflicts.
   }
--- a/core/src/test/resources/ExplicitImplicit/basic.source
+++ b/core/src/test/resources/ExplicitImplicit/basic.source
@@ -185,6 +185,52 @@ class A {
   }
   implicit val x: D.c.B = new D.c.B
 }
+<<< ONLY renamed term in type selection
+package E {
+  object D {
+    class C
+  }
+}
+class A {
+  import E.{D => d}
+  implicit val x = new d.C
+}
+>>>
+package E {
+  object D {
+    class C
+  }
+}
+class A {
+  import E.{D => d}
+  implicit val x: d.C = new d.C
+}
+<<< ONLY renamed term in deeper type selection
+package E {
+  object D {
+    object C {
+      class B
+    }
+  }
+}
+class A {
+  import E.{D => d}
+  import d.{C => c}
+  implicit val x = new c.B
+}
+>>>
+package E {
+  object D {
+    object C {
+      class B
+    }
+  }
+}
+class A {
+  import E.{D => d}
+  import d.{C => c}
+  implicit val x: c.B = new c.B
+}
 <<< two classes
 class A[T](e: T)
 class B {
--- a/core/src/test/resources/ExplicitImplicit/basic.source
+++ b/core/src/test/resources/ExplicitImplicit/basic.source
@@ -8,12 +8,10 @@ class A {
   implicit val x: List[Int] = List(1)
 }
 <<< map
-import scala.concurrent.Future
 class A {
   implicit val x = Map(1 -> "STRING")
 }
 >>>
-import scala.concurrent.Future
 class A {
   implicit val x: Map[Int, String] =
     Map(1 -> "STRING")
@@ -185,7 +183,7 @@ class A {
   }
   implicit val x: D.c.B = new D.c.B
 }
-<<< ONLY renamed term in type selection
+<<< renamed term in type selection
 package E {
   object D {
     class C
@@ -205,7 +203,7 @@ class A {
   import E.{D => d}
   implicit val x: d.C = new d.C
 }
-<<< ONLY renamed term in deeper type selection
+<<< renamed term in deeper type selection
 package E {
   object D {
     object C {
--- a/core/src/test/resources/ExplicitImplicit/basic.source
+++ b/core/src/test/resources/ExplicitImplicit/basic.source
@@ -32,16 +32,15 @@ class A {
   implicit val x: MyMap[Int, String] =
     MyMap(1 -> "STRING")
 }
-<<< removed map
-import scala.collection.immutable.{Map => _}
+<<< local map definition
 class A {
+  case class Map[K](elem: (K, String))
   implicit val x = Map(1 -> "STRING")
 }
 >>>
-import scala.collection.immutable.{Map => _}
 class A {
-  implicit val x: Map[Int, String] =
-    Map(1 -> "STRING")
+  case class Map[K](elem: (K, String))
+  implicit val x: Map[Int] = Map(1 -> "STRING")
 }
 <<< def works
 class A {
--- a/core/src/test/resources/ExplicitImplicit/basic.source
+++ b/core/src/test/resources/ExplicitImplicit/basic.source
@@ -7,13 +7,14 @@ class A {
 class A {
   implicit val x: List[Int] = List(1)
 }
-<<< map
+<<< ONLY map
 class A {
   implicit val x = Map(1 -> "STRING")
 }
 >>>
+import scala.collection.immutable.Map
 class A {
-  implicit val x: scala.collection.immutable.Map[Int, String] =
+  implicit val x: Map[Int, String] =
     Map(1 -> "STRING")
 }
 <<< def works
--- a/core/src/test/resources/ExplicitImplicit/basic.source
+++ b/core/src/test/resources/ExplicitImplicit/basic.source
@@ -49,7 +49,7 @@ class A {
 class A {
   class B { class C }
   implicit val x = new B
-  implicit val y = new x.C
+  implicit val y: x.C = new x.C
 }
 >>>
 class A {
@@ -89,11 +89,12 @@ class B {
   implicit val x = new foo.A(10)
 }
 >>>
+import foo.A
 package foo {
   class A[T](e: T)
 }
 class B {
-  implicit val x: foo.A[Int] = new foo.A(10)
+  implicit val x: A[Int] = new foo.A(10)
 }
 <<< implicitly 2712 trick
 class A {
@@ -169,15 +170,16 @@ object A {
   }
 }
 >>>
+import D.B
 object D {
   class B
 }
 object A {
   class C {
-    implicit val x: List[D.B] = List(new D.B)
+    implicit val x: List[B] = List(new D.B)
   }
 }
-<<< SKIP slick tuple
+<<< slick tuple
 object slick {
   case class Supplier(id: Int, name: String)
   implicit val supplierGetter = (arg: (Int, String)) => Supplier(arg._1, arg._2)
@@ -185,7 +187,7 @@ object slick {
 >>>
 object slick {
   case class Supplier(id: Int, name: String)
-  implicit val supplierGetter: ((Int, String)) => slick.Supplier = (arg: (Int, String)) => Supplier(arg._1, arg._2)
+  implicit val supplierGetter: ((Int, String)) => Supplier = (arg: (Int, String)) => Supplier(arg._1, arg._2)
 }
 <<< NOWRAP package import
 package scala.concurrent {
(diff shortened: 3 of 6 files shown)


core/src/main/scala/scalafix/util/Patch.scala
```diff
@@ -4,26 +4,34 @@ import scala.meta._
...
 
-/**
-  * A patch replaces all tokens between [[from]] and [[to]] with [[replace]].
-  */
-case class Patch(from: Token, to: Token, replace: String) {
-  def insideRange(token: Token): Boolean =
-    token.input.eq(from.input) &&
-      token.end <= to.end &&
-      token.start >= from.start
+sealed abstract class Patch {
```
jvican: These changes were imported from previous @olafurpg's commits in other branch.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
-          stripEnclosingPackage
-
-      val parsed = dialect(gtree.toString()).parse[m.Type]
-      parsed match {
-        case m.Parsed.Success(ast) =>
-          builder(gtree.pos.point) = cleanUp(ast)
-        case _ =>
-      }
+  /** The compiler sets different symbols for `PackageDef`s
+		* and term names pointing to that package. Get the
```
jvican: For some reason this is not aligned, sorry about that.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+    }
+
+    /** Look up a Meta name for both Reflect Term and Type names.
+      *
+      * Unfortunately, meta selection chains are term refs while they could
+      * be type refs (e.g. `A` in `A.this.B` is a term). For this reason,
+      * we need to check both names to guarantee whether a name is in scope.
+      */
+    @inline
+    def lookupBothNames(name: String,
```
jvican: I'm extending this method to return both the term symbol and type names if both exist. This can happen if types with same names are defined in a trait and its companion object, for example.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+        if (typeNameLookup.exists) typeNameLookup
+        else {
+          val termName = g.TermName(name)
+          val termNameLookup = in.lookup(termName)
+          if (termNameLookup.exists) termNameLookup
+          else g.NoSymbol
+        }
+
+      // Disambiguate overloaded symbols caused by name shadowing
+      if (symbol.isOverloaded) {
```
jvican: We try to disambiguate the overloaded symbol with the owner. If there's not luck, we fallback to a string-based approach: get the symbol whose full name is the longest common subexpression of our `tpe`.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+      val names = refNoThis.collect {
+        case tn: m.Term.Name => tn
+        case tn: m.Type.Name => tn
+      }
+
+      // Mix local scope with root scope for FQN and non-FQN lookups
+      val wholeScope = mixScopes(inScope, realRootScope)
+      val disambiguatingSyntax = ref.syntax
+      val (_, reversedSymbols) = {
+        names.iterator.foldLeft(wholeScope -> List.empty[g.Symbol]) {
```
jvican: The `iterator` here can be removed.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+            metaName -> st.renames.getOrElse(symbol, symbol).name
+        }.toMap
+
+        val localSym = inScope.lookup(lastSymbol.name)
+        if (lastSymbol.exists &&
+            (isPathDependent || localSym.exists)) {
+          // Return shortened type for names already in scope
+          val onlyNames = onlyPaths.map(_._1)
+          val removed = removePrefixes(refNoThis, onlyNames)
+          val shortenedAndRenamedType = removed.transform {
```
jvican: The renaming logic can go into an independent method for easing readability.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+        } else {
+          // Remove unnecessary packages from type name
+          val noRedundantPaths = {
+            val paths = onlyPaths.dropWhile(path =>
+              getUnderlyingPkgSymbol(path._2) != enclosingTerm)
+            if (paths.isEmpty) onlyPaths.map(_._1)
+            else paths.tail.map(_._1)
+          }
+
+          // Shortened names must be just one if no PDT
```
jvican: What's the idiomatic way of failing in scalafix? Here we're failing fast since these cases should never happen. I wonder if this is ok or we should go for another solution.

jvican:
So far, imports inside blocks are not recognised (and we should):
```scala
val foo = {
  import scala.concurrent.mutable.List
  List.empty[Int]
}
```
  
It could be done by changing the `ScopeTraverser` to visit `val`s and `def`s and then changing the logic in the `shortenType` method to get the enclosing scope. I'm happy to take a look at it in another PR, since these cases are rare and should not block any progress on scalafix's front.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: why do you have it?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+          if (removed.isEmpty) typeName
+          else m.Type.Select(removed.head, typeName)
+        case r => r
+      }
+
+      transformedRef.asInstanceOf[m.Ref]
+    }
+
+    /** Missing import and shortened name */
+    type Missing = (m.Ref, m.Ref)
```
olafurpg: Seeing how this is used, below I would go for a plain case class and remove the `Hit` type alias

```scala
case class ShortenedName(name: m.Ref, toImport: Option[m.Ref])
```

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: I think that is a good idea. I'm not religious about it, but I think the rule of 30 is nice to keep in mind https://github.com/databricks/scala-style-guide#rule_of_30

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: Are these exceptional/unexpected cases? If so, then throw exception with clear error message about what assumption was broken. If not exceptional or unexpected, then encode in type signature of `shortenType`

olafurpg: btw, I suspect this will eventually become a part of the scala.meta semantic api, once we have gained some experience with it in scalafix.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+          // Remove unnecessary packages from type name
+          val noRedundantPaths = {
+            val paths = onlyPaths.dropWhile(path =>
+              getUnderlyingPkgSymbol(path._2) != enclosingTerm)
+            if (paths.isEmpty) onlyPaths.map(_._1)
+            else paths.tail.map(_._1)
+          }
+
+          // Shortened names must be just one if no PDT
+          assert(shortenedNames.size == 1)
```
olafurpg: Include at least a clarifying message.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+    def mixScopes(sc: g.Scope, sc2: g.Scope) = {
+      val mixedScope = sc.cloneScope
+      sc2.foreach(s => mixedScope.enterIfNew(s))
+      mixedScope
+    }
+
     new SemanticApi {
+      override def shortenType(tpe: m.Type,
+                               owner: m.Tree): (m.Type, Seq[m.Ref]) = {
+        logger.elem(tpe)
```
olafurpg: Remove

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+        // Get only the type Select chains (that inside have terms)
+        val typeRefs = tpe.collect { case ts: m.Type.Select => ts }
+
+        // Collect rewrites to be applied on type refs
+        val typeRewrites =
+          typeRefs.map(tr => getMissingOrHit(tr, globalScope, enclosingPkg))
+        val typeRefsAndRewrites = typeRefs.zip(typeRewrites)
+        val mappedRewrites = typeRefsAndRewrites.map {
+          case (typeRef, missingOrHit) =>
+            typeRef -> missingOrHit.fold(_._2, identity)
```
olafurpg: `._.2, identity` is a sign that a case class might be a better encoding.

scalafix-tests/src/test/scala/scalafix/tests/IntegrationPropertyTest.scala
```diff
@@ -190,7 +190,7 @@ class ScalaJs
           Command("examples/test:compile")
         )
       ),
-      skip = true // GenJsCode is hard: import renames + dependent types
+      skip = false // GenJsCode is hard: import renames + dependent types
```
olafurpg: Did you run clean/compile after disabling scalafix? I think I disabled it for some reason while debugging.

jvican:
Diff from cats:
```patch
diff --git a/.jvmopts b/.jvmopts
deleted file mode 100644
index 83ed6fb4..00000000
--- a/.jvmopts
+++ /dev/null
@@ -1,12 +0,0 @@
-# see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
--Dfile.encoding=UTF8
--Xms1G
--Xmx6G
--XX:MaxPermSize=512M
--XX:ReservedCodeCacheSize=250M
--XX:+TieredCompilation
--XX:-UseGCOverheadLimit
-# effectively adds GC to Perm space
--XX:+CMSClassUnloadingEnabled
-# must be enabled for CMSClassUnloadingEnabled to work
--XX:+UseConcMarkSweepGC
diff --git a/js/src/test/scala/cats/tests/FutureTests.scala b/js/src/test/scala/cats/tests/FutureTests.scala
index e248189d..d81a4746 100644
--- a/js/src/test/scala/cats/tests/FutureTests.scala
+++ b/js/src/test/scala/cats/tests/FutureTests.scala
@@ -2,6 +2,7 @@ package cats
 package js
 package tests
 
+import scala.concurrent.ExecutionContextExecutor
 import cats.laws.discipline._
 import cats.js.instances.Await
 import cats.js.instances.future.futureComonad
@@ -17,7 +18,7 @@ import cats.laws.discipline.arbitrary._
 // https://issues.scala-lang.org/browse/SI-7934
 @deprecated("", "")
 class DeprecatedForwarder {
-  implicit def runNow = scala.scalajs.concurrent.JSExecutionContext.Implicits.runNow
+  implicit def runNow: ExecutionContextExecutor = scala.scalajs.concurrent.JSExecutionContext.Implicits.runNow
 }
 object DeprecatedForwarder extends DeprecatedForwarder
 import DeprecatedForwarder.runNow
diff --git a/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala b/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
index 769970c6..addac01e 100644
--- a/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -228,7 +228,7 @@ class LawTests extends FunSuite with Discipline {
           .forall { case (x, y) => a.eqv(x, y) == b.eqv(x, y) }
     }
 
-    implicit val monoidOrderN = Order.whenEqualMonoid[N]
+    implicit val monoidOrderN: Monoid[Order[N]] with Band[Order[N]] = Order.whenEqualMonoid[N]
     laws[GroupLaws, Order[N]].check(_.monoid)
     laws[GroupLaws, Order[N]].check(_.band)
 
diff --git a/kernel/src/main/scala/cats/kernel/instances/char.scala b/kernel/src/main/scala/cats/kernel/instances/char.scala
index 4238fa84..456afbe4 100644
--- a/kernel/src/main/scala/cats/kernel/instances/char.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/char.scala
@@ -4,7 +4,7 @@ package instances
 package object char extends CharInstances
 
 trait CharInstances {
-  implicit val catsKernelStdOrderForChar = new CharOrder
+  implicit val catsKernelStdOrderForChar: CharOrder = new CharOrder
 }
 
 class CharOrder extends Order[Char] {
diff --git a/project/plugins.sbt b/project/plugins.sbt
index 10bb5da4..e8db97b3 100644
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -11,3 +11,5 @@ addSbtPlugin("org.scala-js"         % "sbt-scalajs"           % "0.6.13")
 addSbtPlugin("com.github.tkawachi"  % "sbt-doctest"           % "0.4.1")
 addSbtPlugin("org.xerial.sbt"       % "sbt-sonatype"          % "1.1")
 addSbtPlugin("com.fortysevendeg"    % "sbt-microsites"        % "0.3.2")
+
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.2.2")
diff --git a/tests/src/test/scala/cats/tests/EvalTests.scala b/tests/src/test/scala/cats/tests/EvalTests.scala
index d7c41002..9a9af542 100644
--- a/tests/src/test/scala/cats/tests/EvalTests.scala
+++ b/tests/src/test/scala/cats/tests/EvalTests.scala
@@ -1,6 +1,11 @@
 package cats
 package tests
 
+import cats.Monoid
+import cats.Semigroup
+import cats.Order
+import cats.PartialOrder
+import cats.Eq
 import scala.math.min
 import cats.laws.ComonadLaws
 import cats.laws.discipline.{BimonadTests, CartesianTests, MonadTests, SerializableTests}
@@ -92,7 +97,7 @@ class EvalTests extends CatsSuite {
   }
 
   {
-    implicit val iso = CartesianTests.Isomorphisms.invariant[Eval]
+    implicit val iso: CartesianTests.Isomorphisms[Eval] = CartesianTests.Isomorphisms.invariant[Eval]
     checkAll("Eval[Int]", BimonadTests[Eval].bimonad[Int, Int, Int])
     checkAll("Eval[Int]", MonadTests[Eval].monad[Int, Int, Int])
   }
@@ -102,27 +107,27 @@ class EvalTests extends CatsSuite {
   checkAll("Eval[Int]", GroupLaws[Eval[Int]].group)
 
   {
-    implicit val A = ListWrapper.monoid[Int]
+    implicit val A: Monoid[ListWrapper[Int]] = ListWrapper.monoid[Int]
     checkAll("Eval[ListWrapper[Int]]", GroupLaws[Eval[ListWrapper[Int]]].monoid)
   }
 
   {
-    implicit val A = ListWrapper.semigroup[Int]
+    implicit val A: Semigroup[ListWrapper[Int]] = ListWrapper.semigroup[Int]
     checkAll("Eval[ListWrapper[Int]]", GroupLaws[Eval[ListWrapper[Int]]].semigroup)
   }
 
   {
-    implicit val A = ListWrapper.order[Int]
+    implicit val A: Order[ListWrapper[Int]] = ListWrapper.order[Int]
     checkAll("Eval[ListWrapper[Int]]", OrderLaws[Eval[ListWrapper[Int]]].order)
   }
 
   {
-    implicit val A = ListWrapper.partialOrder[Int]
+    implicit val A: PartialOrder[ListWrapper[Int]] = ListWrapper.partialOrder[Int]
     checkAll("Eval[ListWrapper[Int]]", OrderLaws[Eval[ListWrapper[Int]]].partialOrder)
   }
 
   {
-    implicit val A = ListWrapper.eqv[Int]
+    implicit val A: Eq[ListWrapper[Int]] = ListWrapper.eqv[Int]
     checkAll("Eval[ListWrapper[Int]]", OrderLaws[Eval[ListWrapper[Int]]].eqv)
   }
 
diff --git a/tests/src/test/scala/cats/tests/IdTests.scala b/tests/src/test/scala/cats/tests/IdTests.scala
index 0465dd34..a7b12751 100644
--- a/tests/src/test/scala/cats/tests/IdTests.scala
+++ b/tests/src/test/scala/cats/tests/IdTests.scala
@@ -1,10 +1,11 @@
 package cats
 package tests
 
+import cats.Id
 import cats.laws.discipline._
 
 class IdTests extends CatsSuite {
-  implicit val iso = CartesianTests.Isomorphisms.invariant[Id]
+  implicit val iso: CartesianTests.Isomorphisms[Id] = CartesianTests.Isomorphisms.invariant[Id]
 
   checkAll("Id[Int]", BimonadTests[Id].bimonad[Int, Int, Int])
   checkAll("Bimonad[Id]", SerializableTests.serializable(Bimonad[Id]))
diff --git a/tests/src/test/scala/cats/tests/NonEmptyListTests.scala b/tests/src/test/scala/cats/tests/NonEmptyListTests.scala
index c913a8f8..c59a86e5 100644
--- a/tests/src/test/scala/cats/tests/NonEmptyListTests.scala
+++ b/tests/src/test/scala/cats/tests/NonEmptyListTests.scala
@@ -1,6 +1,8 @@
 package cats
 package tests
 
+import cats.PartialOrder
+import cats.Order
 import cats.kernel.laws.{GroupLaws, OrderLaws}
 
 import cats.data.NonEmptyList
@@ -36,7 +38,7 @@ class NonEmptyListTests extends CatsSuite {
   checkAll("Eq[NonEmptyList[ListWrapper[Int]]]", SerializableTests.serializable(Eq[NonEmptyList[ListWrapper[Int]]]))
 
   {
-    implicit val A = ListWrapper.partialOrder[Int]
+    implicit val A: PartialOrder[ListWrapper[Int]] = ListWrapper.partialOrder[Int]
     checkAll("NonEmptyList[ListWrapper[Int]]", OrderLaws[NonEmptyList[ListWrapper[Int]]].partialOrder)
     checkAll("PartialOrder[NonEmptyList[ListWrapper[Int]]]", SerializableTests.serializable(PartialOrder[NonEmptyList[ListWrapper[Int]]]))
 
@@ -44,7 +46,7 @@ class NonEmptyListTests extends CatsSuite {
   }
 
   {
-    implicit val A = ListWrapper.order[Int]
+    implicit val A: Order[ListWrapper[Int]] = ListWrapper.order[Int]
     checkAll("NonEmptyList[ListWrapper[Int]]", OrderLaws[NonEmptyList[ListWrapper[Int]]].order)
     checkAll("Order[NonEmptyList[ListWrapper[Int]]]", SerializableTests.serializable(Order[NonEmptyList[ListWrapper[Int]]]))
```

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,104 +11,402 @@ import scalafix.ScalafixConfig
...
+    val rootPkg = st.topLevelPkg
+    val rootImported = st.scopes(rootPkg)
+    val realRootScope = rootPkg.info.members
+
+    // Compute offsets for the whole compilation unit
+    val traverser = new OffsetTraverser
+    traverser.traverse(unit.body)
+
+    def toMetaType(tp: g.Tree) =
+      config.dialect(tp.toString).parse[m.Type].get
```
olafurpg: This is causing the errors.

core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
```diff
@@ -13,6 +15,14 @@ case object ExplicitImplicit extends Rewrite {
     case m.Term.ApplyType(m.Term.Name("implicitly"), _) => true
     case _ => false
   }
+  @scala.annotation.tailrec
+  final def parents(tree: Tree,
```
olafurpg: remove

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -8,14 +8,37 @@ class A {
   implicit val x: List[Int] = List(1)
 }
 <<< map
+import scala.concurrent.Future
```
olafurpg: remove

core/src/main/scala/scalafix/util/Patch.scala
olafurpg: Are these changes necessary now? If not remove

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -44,14 +67,124 @@ class A {
 class A {
   class B { class C }
   implicit val x = new B
-  implicit val y = new x.C
+  implicit val y: x.C = new x.C
```
olafurpg: This is the input, remove?

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -44,14 +71,124 @@ class A {
...
+package cats {
+  package tests {
+    import cats.laws.discipline._
+    class IdTests {
+      implicit val iso: CartesianTests.Isomorphisms[Id] = CartesianTests.Isomorphisms.id[Id]
+    }
+  }
+}
+<<< higher kinded
+import scala.concurrent.Future
```
olafurpg: Remove?

core/src/test/resources/ExplicitImplicit/basic.source
jvican: Removed.

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -8,12 +8,39 @@ class A {
...
+}
+<<< renamed map
+import scala.collection.immutable.{Map => MyMap}
+class A {
+  implicit val x = MyMap(1 -> "STRING")
+}
+>>>
+import scala.collection.immutable.{Map => MyMap}
+class A {
+  implicit val x: MyMap[Int, String] =
```
jvican: Have a look at the other two tests checking renames.

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -44,14 +71,124 @@ class A {
...
+  implicit val y: x.C = new x.C
 }
 >>>
 class A {
   class B { class C }
   implicit val x: B = new B
   implicit val y: x.C = new x.C
 }
+<<< NOWRAP higher kinded cats
+package cats {
```
jvican: This is just a reproduction of an intricate scenario. Even though it has the names cats, I would leave it here. If you disagree, please point to concrete file you'd like to see have it.

core/src/test/resources/ExplicitImplicit/basic.source
jvican: No, this line is needed. Have a look at the shortened `Future` in the implicit val type result.

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -44,14 +71,124 @@ class A {
...
+}
+>>>
+class A {
+  object D {
+    val c = new C {}
+  }
+  trait C {
+    class B
+  }
+  implicit val x: D.c.B = new D.c.B
```
jvican: 💃

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+            case (qual: m.Term, path: m.Term.Name) =>
+              m.Term.Select(qual, renameType(path, renames))
+          }
+          m.Type.Select(termSelects, renamedTypeName)
+        }
+      }
+    }
+
     new SemanticApi {
+      override def shortenType(tpe: m.Type, owner: m.Tree): m.Type = {
```
xeno-by: If owner is only used for its position, why not pass a position here in the first place?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+          else st.scopes.getOrElse(enclosingPkg, rootImported)
+        }
+
+        // Prune owners and use them to create a local bottom up scope
+        val interestingOwners =
+          contextOwnerChain.takeWhile(_ != enclosingPkg)
+        val bottomUpScope = interestingOwners.iterator
+          .map(_.info.members.filter(keepRelevantMembers))
+          .reduce(mixScopes _)
+        val globalScope = mixScopes(bottomUpScope, userImportsScope)
```
xeno-by: Does this take import priority into account?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+    def toTypeName(name: m.Name): m.Type.Name =
+      if (name.is[m.Type.Name]) name.asInstanceOf[m.Type.Name]
+      else m.Type.Name(name.value)
+
+    /** Get the shortened type `ref` at a concrete spot. */
+    def getMissingOrHit(toShorten: m.Ref,
+                        inScope: g.Scope,
+                        enclosingTerm: g.Symbol): m.Ref = {
+
+      val refNoThis = stripThis(toShorten)
```
xeno-by: Will this strip things like `C.this.foo` or only `some.pkg.this.foo`?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
jvican: It doesn't yet, thanks for pointing out.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
jvican: Yes and yes. We decided to pass `owner` just in case we could need it afterwards, but it's unnecessary IIRC.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
jvican: Both, actually. They will be converted to `C.foo` and `some.pkg.foo`.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+    val rootPkg = st.topLevelPkg
+    val rootImported = st.scopes(rootPkg)
+    val realRootScope = rootPkg.info.members
+
+    // Compute offsets for the whole compilation unit
+    val traverser = new OffsetTraverser
+    traverser.traverse(unit.body)
+
+    def toMetaType(tp: g.Tree) = {
+      config.dialect(tp.toString).parse[m.Type].get
```
jvican: This is actually the cause of a problem with cats. We have no general solution so far, `tp` may not have a tree representation (inferred) and it's not fetchable from sources. Any suggestion @xeno-by?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+    val allImports = List(PredefModule, ScalaPackage, JavaLangPackage)
+  }
+
+  /** Return a scope with the default packages imported by Scala.
+    * Follow the same semantics as `rootImports` in the global Context. */
+  private def importDefaultPackages(scope: g.Scope, unit: g.CompilationUnit) = {
+    // Handle packages to import from user-defined compiler flags
+    val packagesToImportFrom = {
+      if (g.settings.noimports) Nil
+      else if (unit.isJava) RootImports.javaImports
```
jvican: I got this logic from nsc `Context` just in case this can in some weird cases be true. I would expect it to be false.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
+  private def keepRelevantMembers(s: g.Symbol): Boolean =
+    (s.isModule || s.isValue || s.isAccessor || s.isType) &&
+      !(s.isMethod || s.isSynthetic || s.hasPackageFlag)
+
+  @inline
+  private def mixScopes(sc: g.Scope, sc2: g.Scope) = {
+    val mixedScope = sc.cloneScope
+    sc2.foreach(s => mixedScope.enterIfNew(s))
+    mixedScope
+  }
```
jvican: There is indeed a performance cost attached to it. I'll study a better data structure to manipulate scopes in the future, right now we're focusing more on correctness. I've also realised that I should not be using `enterIfNew` in some concrete parts of the code. Fixing tomorrow.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
```diff
@@ -11,110 +11,466 @@ import scalafix.ScalafixConfig
...
-      val parsed = dialect(gtree.toString()).parse[m.Type]
-      parsed match {
-        case m.Parsed.Success(ast) =>
-          builder(gtree.pos.point) = cleanUp(ast)
-        case _ =>
-      }
+  /** The compiler sets different symbols for `PackageDef`s
+    * and term names pointing to that package. Get the
+    * underlying symbol of `moduleClass` for them to be equal. */
+  @inline
```
jvican: They can be removed.

core/src/test/resources/ExplicitImplicit/basic.source
```diff
@@ -172,15 +305,15 @@ object A {
...
   implicit val supplierGetter = (arg: (Int, String)) => Supplier(arg._1, arg._2)
 }
 >>>
 object slick {
   case class Supplier(id: Int, name: String)
-  implicit val supplierGetter: ((Int, String)) => slick.Supplier = (arg: (Int, String)) => Supplier(arg._1, arg._2)
+  implicit val supplierGetter: ((Int, String)) => Supplier = (arg: (Int, String)) => Supplier(arg._1, arg._2)
 }
 <<< NOWRAP package import
 package scala.concurrent {
```
jvican: Thanks for the code snippets, will add to tests. Forgot about our beloved `locally` :smile:.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: Anything trivial, e.g. `case class Scope(syms: List[Symbol])`, would do. The implementation of `Scope` is quite intricate, so it'll be just an unnecessary complication.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: Wait for a Type converter that is coming to 2.x along with nirvana and friends.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: What if `C` is an enclosing class? Then `C.foo` doesn't make sense.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: This signature is tricky because the current impl only works for defn.{val,var,def}, so  the `owner` should really be a `Defn`, if anything. With a position, it wouldn't be clear which position you should pass in e.g. the defn's name or the defn itself.

IIRC, the `owner` needs to map to some tree with an ownerchain.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: I'm not sure if the type converter will solve the problem either. Kind-projector synthesizes unique type names for each appearance of `Type.Name(?)`. The synthesized types look like this

```scala
cats.laws.discipline.CartesianTests.Isomorphisms[[β$0$]cats.free.Free[Option,β$0$]]
```

I'm not 100% sure, but I think those should be pretty-printed as `?`. Maybe we'll need to special case kind-projector since it's a very popular compiler plugin.

The best solution for now is to gracefully handle the parse error and not shorten that type signature. The type signature of `shortenType` indicates that it's always a safe operation, and indeed, you can always choose not to shorten the type.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: scalafix-nsc guards against `unit.isJava`, so this will always be false. I don't mind keeping it though for consistency with the compiler impl.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
olafurpg: We should optimize for correctness and simplicity for now. We can worry about performance later.

core/src/test/resources/ExplicitImplicit/basic.source
olafurpg: Awesome tests! Very nice 👍

core/src/test/resources/ExplicitImplicit/basic.source
olafurpg: Aha, gotcha. Maybe update the test name

core/src/test/resources/ExplicitImplicit/basic.source
olafurpg: No worries, let's keep it here. It's just such a big input, would be nice to have it smaller.

olafurpg:
BTW. Maybe move the next test cases to a `ShortenType.source`. We've gone waay beyond `basic.source` :D

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: Can the disambiguation between defn and defn's name be done automatically?

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: I think so, yes.

scalafix-nsc/src/main/scala/scalafix/nsc/NscSemanticApi.scala
xeno-by: We can specialcase kind-projector, yeah.

The `[T]Foo[T]` type doesn't have surface syntax in Scala, so it'll have to be emulated - either with kind-projector or with the `{ type λ[T] = Foo[T] }#λ` trick.

jvican:
I'll rebase this on top of the semantic API soon.

olafurpg:
With the introduction of scalahost and the scala.meta semantic API, I believe context aware scope would be a great feature to have in scala.meta instead of scalafix. Also, scalafix-nsc is being phased out in favor of the offline Mirror so I think that is the best way forward.

jvican:
@xeno-by I'm planning to follow up on this next weekend. Will you accept a PR fully implementing a symbol-based type shortener with the new semantic API?

olafurpg:
I think a good start would be to open a ticket in scalameta listing the details of this feature, what methods/signatures it would have etc.

PR #44, opened 2017-01-25
Implement new patch infrastructure.

olafurpg:
This commit is too big, I'm sorry. Main changes:

    - Patch(from, to, replace) replaced with TokenPatch(token, newToken)
      with helpful combinators such as Add{Left,Right}
    - New high-level TreePatch datatype with can be converted to TokenPatch.
    - Implemented two tree patches, AddGlobalImport and RemoveGlobalImport.
    - Implemented OrganizeImport, which is a prerequisite to use import tree
      patches.
    - OrganizeImports
      - orders imports by configurable groups
      - removes unused imports using scalac -Ywarn-unused-import
        infrastructure. The implementation is requires hijacking a few
        private fields in g.analyzer using evil reflection hackery.
      - option to expand relative imports
      - handles renames
      - configurable "always used" to force keeping an import such as
        acyclic.file

    Known bugs:

    - introduces false import on non-existing cats.data.UUID in circe
    - makes an import unused by expanding relative imports to
      fully-qualified imports, breaking compilation in projects
      with -Ywarn-unused-import (like cats) unless you run scalafix twice
    - crashes on scala.meta parser bug (in akka), not scalafix problem
      really.


olafurpg:
I'm struggling to locally reproduce the cats it test. Will need to leave it for tomorrow.

olafurpg:
Seems ci is broken for some reason, opening another PR

core/src/main/scala/scalafix/util/OrganizeImports.scala
```diff
@@ -0,0 +1,200 @@
...
+  def getLastTopLevelPkg(potPkg: Stat): Stat = potPkg match {
+    case Pkg(_, head +: Nil) => getLastTopLevelPkg(head)
+    case Pkg(_, head +: _) => head
+    case _ => potPkg
+  }
+
+  def getGlobalImports(ast: Tree)(
+      implicit ctx: RewriteCtx): (Seq[Import], Seq[CanonicalImport]) =
+    ast match {
+      case Pkg(_, Seq(pkg: Pkg)) => getGlobalImports(pkg)
```
ShaneDelmore: This whole file is great, I love how readable you have made it. I did one read through and immediately understood the code and how to use it 👍

PR #45, opened 2017-01-27
Implement new patch infrastructure.

olafurpg:
This commit is too big, I'm sorry. Main changes:

- Patch(from, to, replace) replaced with TokenPatch(token, newToken)
  with helpful combinators such as Add{Left,Right}
- New high-level TreePatch datatype with can be converted to TokenPatch.
- Implemented two tree patches, AddGlobalImport and RemoveGlobalImport.
- Implemented OrganizeImport, which is a prerequisite to use import tree
  patches.
- OrganizeImports
  - orders imports by configurable groups
  - removes unused imports using scalac -Ywarn-unused-import
    infrastructure. The implementation is requires hijacking a few
    private fields in g.analyzer using evil reflection hackery.
  - option to expand relative imports
  - handles renames
  - configurable "always used" to force keeping an import such as
    acyclic.file

Known bugs:

- introduces false import on non-existing cats.data.UUID in circe
- makes an import unused by expanding relative imports to
  fully-qualified imports, requiring you to run scalafix
  twice to fully remove unused imports (e.g. in cats with
  -Ywarn-unused-import)
- crashes on scala.meta parser bug (in akka), not scalafix problem
  really.

Expand relative

Test that patch is unchanged


olafurpg:
Drone seems to be pulling out the wrong commit, it ran against hash 9e5ec2bd6739aaa0eaa2e6fd34671a39be6fccb2 when it should have run against fc616f02a00b82471cfedee28e1b0cbbf03ef16f. Gonna close and reopen pr.

olafurpg:
Can't reopen this PR, new #46

PR #46, opened 2017-01-28
Organize imports

olafurpg:


olafurpg:
Organize imports is still a work-in-progress, main issues are

- running it on cats/circe/akka fails/crashes
- little documentation on how to use new tree patches 

I'm still gonna go ahead and merge this now so those who want to experiment with this new API don't have to work off a fork branch.

PR #48, opened 2017-01-28
Simplifify api for import patches

olafurpg:


scalafix-nsc/src/test/scala/scalafix/ImportPatches.sc
```diff
@@ -0,0 +1,23 @@
+// Small worksheet to demonstrate usage of import patches
```
olafurpg: @ShaneDelmore I got this working on my laptop
screen shot 2017-01-28 at 21 30 54

core/src/main/scala/scalafix/rewrite/RewriteCtx.scala
```diff
@@ -1,11 +1,29 @@
...
+    tokens: Tokens,
     tokenList: TokenList,
     comments: AssociatedComments,
     semantic: Option[SemanticApi]
 )
+
+object RewriteCtx {
+  def fromCode(ast: Tree,
+               config: ScalafixConfig = ScalafixConfig(),
+               semanticApi: Option[SemanticApi] = None): RewriteCtx = {
```
ShaneDelmore: What happens if no SemanticApi is supplied?  Is a new one created, or is functionality diminished in some way?

core/src/main/scala/scalafix/util/OrganizeImports.scala
```diff
@@ -200,14 +199,13 @@ private[this] class OrganizeImports private (implicit ctx: RewriteCtx) {
       val oldImports = getGlobalImports(code)
       val globalImports = oldImports.flatMap(getCanonicalImports)
       val cleanedUpImports = cleanUpImports(globalImports, patches)
-      val tokens = code.tokens
       val tokenToEdit =
         oldImports.headOption
           .map(_.tokens.head)
-          .getOrElse(tokens.head)
+          .getOrElse(ctx.tokens.head)
```
ShaneDelmore: Do we know ctx.tokens is non-empty here?

core/src/main/scala/scalafix/rewrite/RewriteCtx.scala
olafurpg: Good question. It depends on what rewrite is used, some rewrites like procedure syntax work 100% without the semantic api, organize imports works without it but can't handle relative imports in that case, explicit implicit will error.

I guess this could be better documented.

core/src/main/scala/scalafix/util/OrganizeImports.scala
olafurpg: The tree nodes that return empty tokens are incredibly few (if any) and I would say it was a truly exceptional case if those were passed in here. Top level trees always have a beginning of file token.

There's an interesting ongoing discussion in scalameta about validating trees for a given dialect. I would be interested to start a similar discussion about the possibility to use a non empty list type Where applicable. The challenge is try and encode as much as possible on the type level without burdening syntax too heavily. Quasi quotes help a lot here.

PR #51, opened 2017-02-06
Integrate scala.meta.semantic.v1.Mirror into scalafix

olafurpg:
The second commit in this PR bundles the bleeding edge (merged this morning!) scala.meta.semantic.v1.Mirror api together with scalafix.SemanticApi. The semantic mirror exposes useful methods such as `symbol(m.Ref): m.Ref`, which can be used to get fully qualified names of Refs.

scalafix.SemanticApi is still necessary to provide custom methods such as isUnusedImport that may be outside the scope of the scala.meta semantic api.

cc/ @xeno-by


build.sbt
```diff
@@ -160,12 +161,12 @@ lazy val `scalafix-nsc` = project
     crossScalaVersions := crossVersions,
     libraryDependencies ++= Seq(
       "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-      "org.scalameta"  %% "scalameta"     % Build.metaV % "provided",
-      "org.scalatest"  %% "scalatest"     % Build.testV % Test,
-      "com.lihaoyi"    %% "ammonite-ops"  % Build.ammoniteV % Test,
+      scalameta,
+      scalahost(scalaVersion.value),
```
xeno-by: scalahost includes scalameta, so the previous line is unnecessary.

core/src/main/scala/scalafix/rewrite/SemanticApi.scala
```diff
@@ -13,17 +19,21 @@ import scala.meta._
   *
   * See [[ExplicitImplicit]] for an example usage of this semantic api.
   */
-trait SemanticApi {
+trait SemanticApi extends Mirror {
```
xeno-by: How about we rename `SemanticApi` into `Mirror` (or `ScalafixMirror`) in order not to confuse people with multiple way of saying the same thing?

core/src/main/scala/scalafix/rewrite/SemanticApi.scala
```diff
@@ -13,17 +19,21 @@ import scala.meta._
...
   /** Returns the fully qualified name of this name, or none if unable to find it*/
   def fqn(name: Ref): Option[Ref]
 
-  /** Returns all used refs in this compilation unit */
-  def usedFqns: Seq[Ref]
-
   /** Returns true if importee is not used in this compilation unit, false otherwise */
   def isUnusedImport(importee: Importee): Boolean
+
+  // TODO(olafur) more elegant way to combine two interfaces
```
xeno-by: How about moving this part to the implementation rather than an interface?

core/src/main/scala/scalafix/rewrite/SemanticApi.scala
```diff
@@ -13,17 +19,21 @@ import scala.meta._
...
   * See [[ExplicitImplicit]] for an example usage of this semantic api.
   */
-trait SemanticApi {
+trait SemanticApi extends Mirror {
 
   /** Returns the type annotation for given val/def. */
   def typeSignature(defn: Defn): Option[Type]
 
   /** Returns the fully qualified name of this name, or none if unable to find it*/
   def fqn(name: Ref): Option[Ref]
```
xeno-by: Could you elaborate a bit more on why a mirror-based `fqn` didn't work for you?

scalafix-nsc/src/test/scala/scalafix/SemanticTests.scala
```diff
@@ -49,41 +55,66 @@ class SemanticTests extends FunSuite with DiffAssertions { self =>
...
-        fail(s"""scalac ${phase.name} error: ${error.msg} at ${error.pos}
-                |$code""".stripMargin))
+        fail(s"scalac ${phase.name} error: ${error.msg} at ${error.pos}"))
     })
-    unit
+    g.phase = run.phaseNamed("patmat")
+    g.globalPhase = run.phaseNamed("patmat")
+
+    unit -> unit.asInstanceOf[mirror.g.CompilationUnit].toDatabase
   }
```
xeno-by: I think you may be interested in depending on scalahost tests directly.

xeno-by: Not at the moment iirc. We could think about doing that.

olafurpg: scalahost-testkit?

core/src/main/scala/scalafix/rewrite/SemanticApi.scala
xeno-by: Any news about this? Do you think something needs to be fixed in `Ref.symbol`?

olafurpg:
Yes, it's blocked by my latest reported issue in scalameta which I bumped
into while running the mirror against slick.

Is it possible to run the mirror without fail fast mode?

Issue #29, opened 2016-12-11
Create rewrite  rule to upgrade from cats.data.Xor to Either


ShaneDelmore:
Cats has removed Xor in 0.8 series and migrated to Either now that Either is right biased in Scala 2.12.

- Rewrite usages of Xor to use Scala Either
- Rewrite usages of Xor.Right/Left to Scala Right/Left
- Rewrite usages of Xor methods that do not exist on Either to use Cats either enrichments
- Rewrite usages of XorT to EitherT
- Remove unused Xor imports
- Add imports as needed for Either enrichments

olafurpg:
Quick pointers on how to get started with semantic rewrites:

- `~scalafix-nsc/test` runs the rewrite unit tests
- you might want to start from this branch of my scalafix fork: https://github.com/olafurpg/scalafix/tree/no-postfix I've setup infrastructure there to run unit tests on a single rewrite. Just add a directory named `Xor2Either` (or whatever name you prefer) in `core/src/main/resources` and then you can write unit tests just like for the other rewrites.
- if you need a method in SemanticApi, feel free to use add it and implement in NscSemanticApi. Don't hesitate to ask me about this part in particular as I have some ideas for useful methods that are not there yet. I might also be willing to implement the methods it if you're missing some functionality.
- I would recommend you take a look at these three first, in that order:
  1. Rewrite usages of Xor to use Scala Either
  2. Rewrite usages of XorT to EitherT
  3.  Rewrite usages of Xor methods that do not exist on Either to use Cats either enrichments

Unfortunately, I don't have more time tonight to take a look at this. But I'm happy to answer any questions tomorrow and in the coming week.

ShaneDelmore:
After doing some initial work yesterday I am leaning towards implementing base rewrites first, then combining them for specific refactorings.  In this case the following seem helpful and re-usable:

- RewriteType
- RewriteMethod
- IdempotentImport //naming things is hard
- RemoveImportIfUnused

olafurpg:
Prototype implementation of this rewrite: https://github.com/scalacenter/scalafix/compare/master...olafurpg:xor

The trick is mainly twofold:

- enable compiler option `-Yrangepos` to get start+end offsets in gtrees.
- match against desugared version of each tree node

This approach seems to decently handle tricky cases like nested type parameters and it's agnostic to whether the user typed `cats.data.Xor` or `Xor`

milessabin:
Bravo! Very keen to see how this turns out :-)

[...]

PR #57, opened 2017-02-09
Experimenting with scala.meta mirror

olafurpg:
This PR integrates the scala.meta mirror even more tightly into scalafix. It includes an example Xor2Either rewrite using the Symbol/Signature API.

cc/ #29 @ShaneDelmore

### DIFF
--- a/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
+++ b/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
@@ -1,0 +1,25 @@
+package scalafix.rewrite
+import scala.collection.immutable.Seq
+import scala.meta._
+import scalafix.util.Patch
+import scala.meta.internal.ast.Helpers._
+
+import org.scalameta.logger
+
+case object Xor2Either extends Rewrite {
+  override def rewrite(code: Tree, rewriteCtx: RewriteCtx): Seq[Patch] = {
+    implicit val semantic = getSemanticApi(rewriteCtx)
+    object traverser extends Traverser {
+      override def apply(tree: Tree): Unit = {
+        Importee
+        tree match {
+          case ref: Term.Ref if ref.isStableId =>
+            logger.elem(ref)
+          case _ => super.apply(tree)
+        }
+      }
+    }
+    traverser(code)
+    Nil
+  }
+}
--- a/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
+++ b/core/src/main/scala/scalafix/rewrite/Xor2Either.scala
@@ -1,25 +1,37 @@
 package scalafix.rewrite
 import scala.collection.immutable.Seq
-import scala.meta._
-import scalafix.util.Patch
+import scala.meta.contrib._
 import scala.meta.internal.ast.Helpers._
-
-import org.scalameta.logger
+import scala.meta.{Symbol => _, _}
+import scala.meta.semantic.v1._
+import scalafix.syntax._
+import scalafix.util.TreeExtractors._
+import scalafix.util.Patch
+import scalafix.util.TreePatch._
+import scalafix.util.logger
 
 case object Xor2Either extends Rewrite {
   override def rewrite(code: Tree, rewriteCtx: RewriteCtx): Seq[Patch] = {
     implicit val semantic = getSemanticApi(rewriteCtx)
-    object traverser extends Traverser {
-      override def apply(tree: Tree): Unit = {
-        Importee
-        tree match {
-          case ref: Term.Ref if ref.isStableId =>
-            logger.elem(ref)
-          case _ => super.apply(tree)
-        }
-      }
-    }
-    traverser(code)
-    Nil
+    code.collectFirst {
+      case t: Term.Name
+          if semantic
+            .symbol(t)
+            .toOption
+            .exists(_.normalized == Symbol("_root_.cats.data.Xor.map.")) =>
+        AddGlobalImport(importer"cats.implicits._")
+    }.toList ++
+      Seq(
+        Replace(Symbol("_root_.cats.data.XorT."),
+                q"EitherT",
+                List(importer"cats.data.EitherT")),
+        Replace(Symbol("_root_.cats.data.Xor."), q"Either"),
+        Replace(Symbol("_root_.cats.data.Xor.Left."), q"Left"),
+        Replace(Symbol("_root_.cats.data.Xor.Right."), q"Right"),
+        Replace(Symbol("_root_.cats.data.XorFunctions.left."), q"Left"),
+        Replace(Symbol("_root_.cats.data.XorFunctions.right."), q"Right"),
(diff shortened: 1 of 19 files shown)


jvican:
👍 🎉

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
```diff
@@ -0,0 +1,37 @@
...
+        Replace(Symbol("_root_.cats.data.XorT."),
+                q"EitherT",
+                List(importer"cats.data.EitherT")),
+        Replace(Symbol("_root_.cats.data.Xor."), q"Either"),
+        Replace(Symbol("_root_.cats.data.Xor.Left."), q"Left"),
+        Replace(Symbol("_root_.cats.data.Xor.Right."), q"Right"),
+        Replace(Symbol("_root_.cats.data.XorFunctions.left."), q"Left"),
+        Replace(Symbol("_root_.cats.data.XorFunctions.right."), q"Right"),
+        RemoveGlobalImport(importer"cats.data.Xor"),
+        RemoveGlobalImport(importer"cats.data.XorT")
```
xeno-by: Wow. This looks amazingly concise.

core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -15,6 +19,14 @@ package object syntax {
...
           i.importee.structure == patch.importee.structure)
   }
+  implicit class XtensionTermRef(ref: Term.Ref) {
+    def toTypeRef: Type.Ref = ref match {
+      case Term.Name(name) => Type.Name(name)
+      case Term.Select(qual: Term.Ref, Term.Name(name)) =>
+        Type.Select(qual, Type.Name(name))
+      case _ => ref.syntax.parse[Type].get.asInstanceOf[Type.Ref]
+    }
+  }
```
xeno-by: Is this method used anywhere?

core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -25,6 +37,51 @@ package object syntax {
...
+    def toEither: Either[Exception, T] = completed match {
+      case Completed.Success(e) => Right(e)
+      case Completed.Error(e) => Left(e)
+    }
+    def toTry: Try[T] = Try(completed.get)
+  }
+  implicit class XtensionSymbol(symbol: Symbol) {
+    /** Returns simplified version of this Symbol.
+      *
+      * - No Symbol.Multi
```
xeno-by: What if there are other non-trivial symbols in the multi?

core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -25,6 +37,51 @@ package object syntax {
...
+        case Symbol.Global(Symbol.None, Signature.Term(name)) =>
+          Term.Name(name)
+        case Symbol.Global(sym, Signature.Type(name)) =>
+          Type.Select(sym.to[Term.Ref].get, Type.Name(name))
+        case Symbol.Global(sym, Signature.Term(name)) =>
+          Term.Select(sym.to[Term].get, Term.Name(name))
+      }
+      tree.asInstanceOf[T] // should crash
+    }
+  }
```
xeno-by: How about we add `Symbol.name` and `Symbol.fullName`, where the former always returns an `XXX.Name` and the latter does what you implemented?

project/Build.scala
```diff
@@ -5,12 +5,13 @@ style = defaultWithAlign
 }*/
 
 object Build {
-  val scalametaV = "1.6.0-641"
+  val scalametaV = "1.6.0-652.1486659769696"
```
xeno-by: What does this version do?

core/src/main/scala/scalafix/util/TreeExtractors.scala
```diff
@@ -0,0 +1,8 @@
+package scalafix.util
+import scala.meta._
+object TreeExtractors {
+  object `:WithParent:` {
+    def unapply(tree: Tree): Option[(Tree, Tree)] =
+      tree.parent.map(parent => tree -> parent)
+  }
+}
```
xeno-by: Is this used?

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
olafurpg: Yes! This is pretty amazing. And it can easily be parsed from config files, making it easy for library authors to ship custom (simple) rewrites

xeno-by: Can we also make this interactive, i.e have a CLI app with a prompt where you can enter your refactorings?

olafurpg: Yes, that would be doable. It would be cool to offer a preview of the diff of a single rewrite, so that you can reuse the same semantic db to run multiple rewrites.

core/src/main/scala/scalafix/syntax/package.scala
olafurpg: Then you need to write a custom rewrite. I made this normalized symbol to simplify the interface for authors writing basic rewrites, basically so it can be done via simple config file.

core/src/main/scala/scalafix/syntax/package.scala
olafurpg: That would be perfect 👍 I opened a ticket in scala.meta https://github.com/scalameta/scalameta/issues/675

core/src/main/scala/scalafix/util/TreeExtractors.scala
olafurpg: I was using it, and I have a similar one in scalafmt. Should probably be in contrib.

project/Build.scala
olafurpg: Configurable fix for https://github.com/scalameta/scalameta/issues/666

olafurpg:
I tracked down a performance issue with running the `Replace` patches on slick. The CI was running >1h when it should complete in 5min. It appears that calling `.symbol` on every Ref in a tree is slow.

This provides a valuable lesson: scalafix should not run rewrites during compilation, scalafix should run rewrites against a pre-made semanticdb. Why?

- simpler fixture for testing rewrites, load a Mirror and go!
- easier to experiment with semantic rewrites in worksheets
- easier to iterate on rewrite implementations, no need to re-compile large projects (slick/circe) all the time
- compilation is faster because we don't run rewrites
- rewrites can run in parallel! (big win for large projects)

core/src/main/scala/scalafix/rewrite/Xor2Either.scala
ShaneDelmore: Next step, let's tie this into the @deprecated annotation.  If my project has scalafix enabled, and I use a lib that has annotated a method with @deprecated while pointing it to a scalafix rewrite bundled with it, then offer to run the rewrite to remove the deprecated usage.

PR #68, opened 2017-02-21
Refactor rewrite api.

olafurpg:
Rewrite now accepts a type parameter. Why?

- Principle of least power, if you don't need semantic information,
  don't require it.
- Rewrite requirements are documented in the type.
- Possible to construct RewriteCtx from scala.meta.Mirror.
- Paves to way to contribute Rewrite upstream to scala.meta

### DIFF
--- a/core/src/main/scala/scalafix/rewrite/ProcedureSyntax.scala
+++ b/core/src/main/scala/scalafix/rewrite/ProcedureSyntax.scala
@@ -10,10 +10,10 @@ import scalafix.util.Patch
 import scala.collection.immutable.Seq
 import scalafix.util.TokenPatch
 
-case object ProcedureSyntax extends Rewrite {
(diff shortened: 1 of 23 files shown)


core/src/main/scala/scalafix/rewrite/ProcedureSyntax.scala
```diff
@@ -36,5 +36,6 @@ class ProcedureSyntax[T]() extends Rewrite[T] {
 }
 
 object ProcedureSyntax {
+  val default: ScalafixRewrite = apply()
```
ShaneDelmore: This is subtle.  I think it says that calling apply through default fixes an equality bug.  If so, I would add a comment here to prevent someone from inlining it in the future.

olafurpg: I agree. I plan to reorganize the configuration in an upcoming pr and there I will sort out better equality for rewrites.

PR #73, opened 2017-02-23
Replace ScalafixMirror with Mirror where possible.

olafurpg:
This makes is possible to use OrganizeImports (without remove unused
imports) and Replacer with scala.meta.Mirror instead of ScalafixMirror.

### DIFF
--- a/core/src/main/scala/scalafix/util/OrganizeImports.scala
+++ b/core/src/main/scala/scalafix/util/OrganizeImports.scala
@@ -7,14 +7,27 @@ import scala.meta._
 import scala.meta.semantic.v1.Completed
 import scala.meta.tokens.Token.Comment
 import scalafix.config.FilterMatcher
-import scalafix.rewrite.ScalafixCtx
 import scalafix.rewrite.RewriteCtx
 import scalafix.syntax._
 import scalafix.util.TreePatch.AddGlobalImport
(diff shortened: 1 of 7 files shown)


olafurpg:
@ShaneDelmore This enables catz-cradle to use import patches and `Replace` with scala.meta.Mirror instead of ScalafixMirror.

core/src/main/scala/scalafix/util/OrganizeImports.scala
```diff
@@ -7,14 +7,27 @@ import scala.meta._
...
+trait OrganizeImportsMirror {
+
+  /** Returns true if this importee is never used and can be removed. */
+  def isUnused(importee: Importee): Boolean
+
+  /** Returns fully qualified name of this reference
+    *
+    * For example scala.collection.immutable.List for List.
+    **/
+  def fullyQualifiedName(ref: Ref): Option[Ref]
```
xeno-by: I remember there was a problem with fqn not working correctly if implemented via scala.meta's mirror. What's its current status? Is there something I can do on my side?

olafurpg: The problem was unrelated to Mirror, I suspect it was caused by the incorrect setup with scalafix-nsc and scalahost.

PR #87, opened 2017-03-11
Fix #86

### LINKED ISSUE #86: Empty `imports.groups` trigger `NoSuchElementException`
Invoking scalafix using
```
rewrites = []
imports.organize = true
imports.alwaysUsed = []
imports.expandRelative = false
imports.groups = []
imports.removeUnused = true
imports.groupByPrefix = false
imports.spaceAroundCurlyBrace = false
```
triggers
```
[error] Failed to fix Test.scala. Error: null. java.util.NoSuchElementException
[error] scala.collection.LinearSeqOptimized$class.last(LinearSeqOptimized.scala:148)
[error] scala.collection.immutable.List.last(List.scala:84)
[error] scalafix.util.OrganizeImports$$anonfun$9$$anonfun$apply$6.apply(OrganizeImports.scala:142)
[error] scalafix.util.OrganizeImports$$anonfun$9$$anonfun$apply$6.apply(OrganizeImports.scala:142)
```

mlangc:
Fixes #86

### DIFF
--- a/scalafix-nsc/src/test/resources/checkSyntax/OrganizeImportsNoGroups.source
+++ b/scalafix-nsc/src/test/resources/checkSyntax/OrganizeImportsNoGroups.source
@@ -1,0 +1,22 @@
+rewrites = []
+imports.groups = []
+<<< Simple test case (see #86)
+import java.util.HashMap
+import java.util.Map
+import scala.collection.immutable.HashSet
+
+import java.util.LinkedHashMap
+
+object Test {
+  val x: Map[String, String] = new HashMap
+  val y: Set[String] = new HashSet
+}
+>>>
+import java.util.HashMap
+import java.util.Map
+import scala.collection.immutable.HashSet
+
+object Test {
+  val x: Map[String, String] = new HashMap
+  val y: Set[String] = new HashSet
+}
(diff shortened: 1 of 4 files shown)


scalafix-nsc/src/test/resources/checkSyntax/OrganizeImportsNoGroups.source
```diff
@@ -0,0 +1,28 @@
+rewrites = []
+imports.organize = true
+imports.alwaysUsed = []
```
olafurpg: Can we remove unnecessary lines from this configuration to highlight that `imports.groups = []` is the line being tested?

mlangc: I just did, and also verified that the test still triggers the bug. Note however, that there is also a downside on relying heavily on default values in tests:

1) a reader of the test has to consult them, to see what is actually tested
2) defaults might change in the future

olafurpg: Good point. The defaults are likely to change. I try to keep unit tests as small as possible because large unit test have a tendency to require unnecessary maintenance.

PR #98, opened 2017-03-24
Add SBT test that demonstrates problem with ooimports

mlangc:
Organize imports seems not to remove unused imports when used from the SBT plugin, as demonstrated by the added SBT test. Interestingly, an analogous test case for scalafix-nsc, that also comes with this commit, passes flawlessly.

My suspicion is that the "hijacking" performed by scalafix-nsc does not work so well in the context of the SBT plugin.

PR #99, opened 2017-03-25
Upgrade drone

olafurpg:
Unblock #98


olafurpg:
Finally got CI green, lets hope it stays that way ^^

PR #107, opened 2017-03-30
Add CLI options for classpath and sourcepath

### LINKED ISSUE #106: Expose cli options to construct offline mirror
It should be possible to run semantic rewrites from the cli with `--sourcepath` and `--classpath` options. Those options are necessary to construct an offline scala.meta.Mirror, see here: https://github.com/scalameta/scalameta/blob/baf4bc3c3b77d692750c7aec89c6ff66b1f1cf9e/scalahost/src/main/scala/scala/meta/Mirror.scala#L12

stuhood:
Adds CLI options for classpath and sourcepath, and constructs a mirror using them.

Fixes #106.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -180,7 +180,7 @@ lazy val `scalafix-nsc` = project
     allSettings,
     libraryDependencies ++= Seq(
       "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-      scalahost_nsc(scalaVersion.value),
+      scalahostNsc(scalaVersion.value),
       ammonite % Test,
       // integration property tests
       "org.typelevel"      %% "catalysts-platform" % "0.0.5"    % Test,
@@ -229,6 +229,10 @@ lazy val cli = project
   .settings(
     allSettings,
     packSettings,
+    libraryDependencies ++= Seq(
+      // NB: The Mirror has an undeclared dependency on the scalahost-nsc package.
+      scalahostNsc(scalaVersion.value)
+    ),
     moduleName := "scalafix-cli",
     packJvmOpts := Map(
       "scalafix" -> jvmOptions,
--- a/build.sbt
+++ b/build.sbt
@@ -180,7 +180,7 @@ lazy val `scalafix-nsc` = project
     allSettings,
     libraryDependencies ++= Seq(
       "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-      scalahost(scalaVersion.value),
+      scalahost_nsc(scalaVersion.value),
       ammonite % Test,
       // integration property tests
       "org.typelevel"      %% "catalysts-platform" % "0.0.5"    % Test,
--- a/cli/src/main/scala/scalafix/cli/Cli.scala
+++ b/cli/src/main/scala/scalafix/cli/Cli.scala
@@ -9,6 +9,7 @@ import scalafix.Scalafix
 import scalafix.cli.termdisplay.TermDisplay
 import scalafix.config.ScalafixConfig
 import scalafix.rewrite.ProcedureSyntax
+import scalafix.rewrite.ScalafixMirror
 import scalafix.rewrite.ScalafixRewrite
 import scalafix.rewrite.ScalafixRewrites
 import scalafix.util.FileOps
@@ -44,6 +45,16 @@ case class ScalafixOptions(
       ".scalafix.conf OR imports.organize=false"
     ) @ExtraName("c") config: Option[ScalafixConfig] = None,
     @HelpMessage(
+      "The classpath for the input files, to support semantic fixes."
+    ) @ValueDescription(
+      "entry1.jar:entry2.jar"
+    ) classpath: Option[String] = None,
+    @HelpMessage(
+      "The sourcepath for the input files, to support semantic fixes. May contain either sources or directories."
+    ) @ValueDescription(
+      "File2.scala:File1.scala:directory/one"
+    ) sourcepath: Option[String] = None,
+    @HelpMessage(
       s"Additional rewrite rules to run. NOTE. rewrite.rules = [ .. ] from --config will also run."
     ) @ValueDescription(
       s"$ProcedureSyntax OR file:LocalFile.scala OR scala:full.Name OR https://gist.com/.../Rewrite.scala"
@@ -77,6 +88,16 @@ case class ScalafixOptions(
   lazy val resolvedConfig: ScalafixConfig = config match {
     case None => ScalafixConfig.auto(common.workingDirectoryFile)
     case Some(x) => x
+  }
+
+  lazy val resolvedMirror: Option[ScalafixMirror] = (classpath, sourcepath) match {
+    case (Some(cp), Some(sp)) =>
+      Some(ScalafixMirror.fromMirror(scala.meta.Mirror(cp, sp)))
+    case (None, None) => None
+    case _ =>
+      // FIXME: improve during review.
+      throw new Exception(
+        "The semantic API was partially configured: both a classpath and sourcepath are required.")
   }
 
   lazy val outFromPattern: Pattern = Pattern.compile(outFrom)
@@ -121,7 +142,7 @@ object Cli {
   }
 
   def handleFile(file: File, options: ScalafixOptions): ExitStatus = {
-    val fixed = Scalafix.fix(Input.File(file), options.resolvedConfig)
+    val fixed = Scalafix.fix(Input.File(file), options.resolvedConfig, options.resolvedMirror)
     fixed match {
       case Fixed.Success(code) =>
         if (options.inPlace) {
--- a/cli/src/main/scala/scalafix/cli/Cli.scala
+++ b/cli/src/main/scala/scalafix/cli/Cli.scala
@@ -45,14 +45,22 @@ case class ScalafixOptions(
       ".scalafix.conf OR imports.organize=false"
     ) @ExtraName("c") config: Option[ScalafixConfig] = None,
     @HelpMessage(
-      "The classpath for the input files, to support semantic fixes."
+      "java.io.File.pathSeparator separated list of jar files" +
+      "or directories containing classfiles and `semanticdb` files." +
+      "The `semanticdb` files are emitted by the scalahost-nsc" +
+      "compiler plugin and are necessary for the semantic API to" +
+      "function. The classfiles + jar files are necessary for" +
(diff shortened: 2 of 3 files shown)


stuhood:
NB: This hasn't received very thorough testing yet, as I'm fighting a battle with ivy in order to reliably fetch the scalameta dependencies from bintray for https://github.com/pantsbuild/pants/compare/master...twitter:stuhood/scalafix

project/Dependencies.scala
```diff
@@ -5,10 +5,11 @@ style = defaultWithAlign
 }*/
 
 object Dependencies {
-  val scalametaV = "1.6.0"
+  val scalametaV = "1.7.0-485-086c4ac9"
   val paradiseV  = "3.0.0-M7"
   var testClasspath: String = "empty"
   def scalahost(scalaVersion: String): ModuleID = "org.scalameta" % s"scalahost_$scalaVersion" % scalametaV
+  def scalahost_nsc(scalaVersion: String): ModuleID = "org.scalameta" % s"scalahost-nsc_$scalaVersion" % scalametaV
```
olafurpg: `s/scalahost_nsc/scalahostNsc/`

cli/src/main/scala/scalafix/cli/Cli.scala
```diff
@@ -79,6 +90,16 @@ case class ScalafixOptions(
     case Some(x) => x
   }
 
+  lazy val resolvedMirror: Option[ScalafixMirror] = (classpath, sourcepath) match {
+    case (Some(cp), Some(sp)) =>
+      Some(ScalafixMirror.fromMirror(scala.meta.Mirror(cp, sp)))
+    case (None, None) => None
+    case _ =>
+      // FIXME: improve during review.
+      throw new Exception(
```
olafurpg: We can move this into the `parse` method, returning `Left(errorMessage)` . See https://github.com/scalacenter/scalafix/blob/9cbc036d0fa8fd8f0f388d9d3ea8909e8463bcb2/cli/src/main/scala/scalafix/cli/Cli.scala#L179

cli/src/main/scala/scalafix/cli/Cli.scala
```diff
@@ -44,6 +45,16 @@ case class ScalafixOptions(
       ".scalafix.conf OR imports.organize=false"
     ) @ExtraName("c") config: Option[ScalafixConfig] = None,
     @HelpMessage(
+      "The classpath for the input files, to support semantic fixes."
+    ) @ValueDescription(
+      "entry1.jar:entry2.jar"
```
olafurpg: Can also be a directory like `core/target/classes`, we can reuse the docstring from `Mirror.apply`

olafurpg:
> fighting a battle with ivy in order to reliably fetch the scalameta dependencies from bintray

@stuhood I think the bintray artifacts  are `publishMavenStyle := false`, so they use a different pattern from the maven artifacts, see

```
// 404
https://dl.bintray.com/scalameta/maven/org/scalameta/contrib_2.11/1.7.0-485-086c4ac9/contrib_2.11-1.7.0-485-086c4ac9.pom
// OK
https://dl.bintray.com/scalameta/maven/org.scalameta/contrib_2.12/1.7.0-485-086c4ac9/jars/contrib_2.12.jar
```

cli/src/main/scala/scalafix/cli/Cli.scala
stuhood: Since this requires two options to be set in order to succeed, it's not clear how to do that here.

stuhood:
@olafurpg , @xeno-by: It looks like `scala.meta.Mirror` has an undeclared dependency on `scalahost-nsc`: while [creating the OnlineMirror](https://github.com/scalameta/scalameta/blob/9d646038aca486dac8f9eb09c339c3ab8bc20900/scalahost/core/src/main/scala/scala/meta/internal/scalahost/v1/offline/Mirror.scala#L68-L92), if `scalahost` is present rather than `scalahost-nsc`, it will fail to load (unsurprisingly).

By declaring a dep on `scalahost-nsc` for the CLI tool, I was able to finally get the CLI tool to fail in a new and potentially interesting way (using [this branch](https://github.com/pantsbuild/pants/compare/master...twitter:stuhood/scalafix#diff-cc6674e164892f2b7d8cbbdeabacc6e4R34) and a locally published version of this change):
```
Error fixing file: /Users/stuhood/src/pants/.pants.d/scalafix/scalafix/252d64521cf9/src.scala.org.pantsbuild.zinc.logging.logging/current/Loggers.scala
Cause: /Users/stuhood/src/pants/.pants.d/scalafix/scalafix/252d64521cf9/src.scala.org.pantsbuild.zinc.logging.logging/current/Loggers.scala:63: error: implementation restriction: semantic API doesn't support this abstract syntax tree.
For more information, visit https://github.com/scalameta/scalameta/issues/621.
class FileLogger(file: File, append: Boolean) extends Logger {
^
/Users/stuhood/src/pants/.pants.d/scalafix/scalafix/252d64521cf9/src.scala.org.pantsbuild.zinc.logging.logging/current/Loggers.scala:63: error: implementation restriction: semantic API doesn't support this abstract syntax tree.
For more information, visit https://github.com/scalameta/scalameta/issues/621.
class FileLogger(file: File, append: Boolean) extends Logger {
^
at scala.meta.internal.scalahost.v1.online.Mirror.typecheck(Mirror.scala:156)
at scala.meta.internal.scalahost.v1.offline.Mirror.typecheck(Mirror.scala:63)
at scala.meta.internal.scalahost.v1.Mirror$class.symbol(Mirror.scala:34)
at scala.meta.internal.scalahost.v1.offline.Mirror.symbol(Mirror.scala:21)
at scalafix.rewrite.ScalafixMirror$$anon$1.symbol(ScalafixMirror.scala:40)
at scala.meta.semantic.v1.Api$XtensionRefSymbol.symbol(Api.scala:17)
at scalafix.rewrite.ScalaJsRewrites$$anonfun$1$$anonfun$hasGlobalScopeParent$1$1$$anonfun$apply$2.apply(ScalaJsRewrites.scala:57)
at scalafix.rewrite.ScalaJsRewrites$$anonfun$1$$anonfun$hasGlobalScopeParent$1$1$$anonfun$apply$2.apply(ScalaJsRewrites.scala:56)
at scala.meta.contrib.TreeOps$$anonfun$exists$1.apply(TreeOps.scala:25)
at scala.meta.contrib.TreeOps$$anonfun$exists$1.apply(TreeOps.scala:25)
at scala.meta.contrib.TreeOps$$anonfun$find$1.isDefinedAt(TreeOps.scala:19)
at scala.meta.contrib.TreeOps$$anonfun$find$1.isDefinedAt(TreeOps.scala:19)
at scala.meta.contrib.TreeOps$traverser$3$.apply(TreeOps.scala:31)
at scala.meta.contrib.TreeOps$.collectFirst(TreeOps.scala:38)
at scala.meta.contrib.TreeOps$.find(TreeOps.scala:19)
at scala.meta.contrib.TreeOps$.exists(TreeOps.scala:25)
at scala.meta.contrib.implicits.TreeExtensions$XtensionTreeOps.exists(TreeExtensions.scala:38)
at scalafix.rewrite.ScalaJsRewrites$$anonfun$1$$anonfun$hasGlobalScopeParent$1$1.apply(ScalaJsRewrites.scala:56)
at scalafix.rewrite.ScalaJsRewrites$$anonfun$1$$anonfun$hasGlobalScopeParent$1$1.apply(ScalaJsRewrites.scala:56)
at scala.collection.LinearSeqOptimized$class.exists(LinearSeqOptimized.scala:93)
```

So, assuming that that issue is an issue in github.com/scalameta/scalameta rather than here, then I think this PR is ready to land.

stuhood:
One other potential issue: `scalahost` (and `scalahost-nsc`) are set to `isFullCrossVersion`, but scalafix currently isn't. I believe that that means that scalafix must also be `isFullCrossVersion`?

cli/src/main/scala/scalafix/cli/Cli.scala
olafurpg: I'll reorganize this validation in a followup PR, this is good for now.

cli/src/main/scala/scalafix/cli/Cli.scala
```diff
@@ -79,6 +98,16 @@ case class ScalafixOptions(
     case Some(x) => x
   }
 
+  lazy val resolvedMirror: Option[ScalafixMirror] = (classpath, sourcepath) match {
```
olafurpg: Heads up, this will become `Mirror` once I've updated the `Scalafix.fix` method.

build.sbt
```diff
@@ -229,6 +229,10 @@ lazy val cli = project
   .settings(
     allSettings,
     packSettings,
+    libraryDependencies ++= Seq(
+      // NB: The Mirror has an undeclared dependency on the scalahost-nsc package.
+      scalahostNsc(scalaVersion.value)
```
olafurpg: scalahost-nsc is a fatjar that's only meant to be used as a compiler plugin via `-Xplugin:/scalahost-nsc.jar`. The third argument to `Mirror.apply` is the file path to the fatjar, it can be automatically detected if it's already on the classpath. I think for now, this is OK, but eventually we may want force scalafix-cli users to pass in this file path explicitly to avoid bloating the scalafix-cli classpath.

olafurpg:
> One other potential issue: scalahost (and scalahost-nsc) are set to isFullCrossVersion, but scalafix currently isn't. I believe that that means that scalafix must also be isFullCrossVersion?

Good question. I think that's right, scalafix depends on scala-compiler (not only for scalahost, also runtime compilation of rewrites) so it ought to be full-cross version like scalahost.

olafurpg:
@stuhood as for the last cryptic error message " semantic API doesn't support this abstract syntax tree", it seems to have been caused by the fact that several rewrites were enabled by default in `ScalafixConfig`, and one of those rewrites did an invalid `q"a".symbol` call. In #112, I removed all rewrites from the base settings so this should not happen in the next release.

stuhood:
Great. Thanks!

olafurpg:
v0.3.3 is out! Note that it's full-cross versioned now http://search.maven.org/#artifactdetails%7Cch.epfl.scala%7Cscalafix-core_2.12.1%7C0.3.3%7Cjar

olafurpg:
The changelog https://scalacenter.github.io/scalafix/#0.3.3

PR #112, opened 2017-04-03
Miscellaneous cli improvements

olafurpg:


olafurpg:
@stuhood FYI,

- scalafix  and scalafix-cli are now full-cross versioned.
- I removed the scalahost-nsc dependency on the `cli` module in favor of another `fatcli` module which is identical to `cli` except it has a dependency on scalahost-nsc.

jvican:
LGTM, :ship:.

PR #115, opened 2017-04-04
Setup auto-publish on merge to master.

olafurpg:
- Use scalameta/maven repository to avoid need for two custom resolvers
- Publish gh-pages docs


olafurpg:
Fingers cross that publishing work after merge 😄

Issue #92, opened 2017-03-16
Support linters, rewrites that only emit info/warn/error messages


SethTisue:
without the rewrite part, a rewrite rule becomes a linter rule, and Scalafix becomes a linter too.

I started a discussion on this possibility at https://contributors.scala-lang.org/t/whats-the-status-of-abide/609

ShaneDelmore:
Interesting thought. Not sure if this is in scope but to continue with the idea, if we separated out generating rewrites from applying rewrites then it becomes a linter with a possible auto-fix. If someone catches exception, we could warn, and if they ran autofix --rewriteName we could automatically rewrite to case NonFatal... for example.

olafurpg:
@SethTisue I just opened https://github.com/scalacenter/scalafix/pull/118 which adds support for rewrites to report info/warn/error messages to users. I think the next steps to add support for linters would be to

- Add a `Linter` class which returns `Seq[Message(msg, severity, position)]` instead of `Seq[Patch]`
- Change cli exit code depending on linter warnings/errors
- Add a `fatalWarnings: Boolean` configuration option, which promotes warning messages to error, just like the Scala compiler.

ShaneDelmore:
@olafurpg Regarding the linter, what do you think about having linters return  Seq[Message -> Seq[Patch]] to allow pairing patches with rewrites in case the user wants to fix them with Scalafix?  Say for the case of a deprecation warning or missing import?

ShaneDelmore:
Linter rule authors would not have to write patches, but if they decided to much of the code might be shared.

olafurpg:
Good idea @ShaneDelmore 

We might even want use such a data structure (name is open) for normal rewrites

```scala
Observation(message: String, patches: Seq[Patch])
```

since I think it's useful to attach messages to rewrites as well. It's been awhile since I looked at what they do in similar tools such as clang-tidy, rustfix. I'm sure we can get inspiration from them.

olafurpg:
It should be possible to accommodate linters with the current API. Rewrites have access to `ctx.reporter` which is a similar api as `g.reporter` in scalac, see https://github.com/scalacenter/scalafix/blob/db6152304aaef19c4a8ba4fc77b732264bf52fec/scalafix-core/src/main/scala/scalafix/config/ScalafixReporter.scala#L36-L40

Info, warning and error messages are printed to the console. The remaining parts to close this issue is to make sure that the scalafix cli exits with a non-zero exit status in case errors are reported. 

[...]

PR #118, opened 2017-04-05
Expose reporter api for rewrites

olafurpg:
This commit enables rewrites to report messages to scalafix users
through the RewriteCtx. This opens opportunities for scalafix to be
used as a linting tool (#92).

I went for a clean-room implementation of the logger instead of slf4j or
j.u.Logger because

1. info/warn/error messages can take an optional scala.meta.Position,
   messages with a custom position are formatted message with the source
   context like in scalac.
2. no runtime reflection, paving the way for Scala.js rewrites.

Example usage from rewrite author:
```scala
ctx.reporter.info/warn/error("bad name!", tree.pos)
```

In the cli integration, messages are printed out like this
```
/MyCode.scala:10: info: bad name!
object Tree
       ^
```
IDEs can use the position to highlight code with squiggly marks.

Example configuration for scalafix user:
```scala
reporter.minSeverity = Trace               # default=Info
reporter.filter.include = ["MyRewrite"]    # only emit messages from MyRewrite
reporter.filter.exclude = ["NoisyRewrite"] # skip messages from NoisyRewrite
```


scalafix-core/src/main/scala/scalafix/util/Severity.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import scalafix.config.ReaderUtil
+
+import metaconfig.Reader
+
+sealed abstract class Severity(color: String, val order: Int)(
+    implicit name: sourcecode.Name)
+    extends Ordered[Severity] {
+  override val toString: String =
+    color + name.value.toLowerCase + Console.RESET
```
fommil: this kind of thing can break (dumb) terminals that don't understand escape codes.

olafurpg: Good catch. It should be possible to disable color through .scalafix.conf, I opened #120 to track this.

PR #123, opened 2017-04-12
Meta 1.7

olafurpg:


scalafix-core/src/main/scala/scalafix/util/Patch.scala
```diff
@@ -30,7 +32,16 @@ abstract class ImportPatch(val importer: Importer) extends TreePatch {
 }
 
 object TreePatch {
-  case class Rename(from: Name, to: Name) extends TreePatch
+  trait RenamePatch
+  case class Rename(from: Name, to: Name) extends TreePatch with RenamePatch
+  case class RenameSymbol(from: Symbol, to: Name, normalize: Boolean = false)
```
olafurpg: @ShaneDelmore FYI, this tree patch works similarly to "Rename" in IDEs. I'm thinking of renaming `Rename` to `RenameReference` or `RenameThisInstance`, what do you think?

I plan to document the patches better soon. First, I want to focus on getting rid of the ScalafixMirror and improve build integrations.

scalafix-core/src/main/scala/scalafix/rewrite/Scalameta17.scala
```diff
@@ -0,0 +1,40 @@
...
+          case "(S)Lscala/meta/Lit;"                         => AddRight(tok, ".Short")
+          case "(Z)Lscala/meta/Lit;"                         => AddRight(tok, ".Boolean")
+          case "(Ljava/lang/String;)Lscala/meta/Lit;"        => AddRight(tok, ".String")
+          case "(Lscala/Symbol;)Lscala/meta/Lit;"            => AddRight(tok, ".Symbol")
+          case "(Lscala/runtime/BoxedUnit;)Lscala/meta/Lit;" => AddRight(tok, ".Unit")
+          case _                                             => AddRight(tok, "") // nothing
+        }
+    }
+    RenameSymbol(Symbol("_root_.scala.meta.io.AbsolutePath#getAbsolutePath()Ljava/lang/String;."), q"absolute") +:
+      litPatches
```
xeno-by: Shall we update the changelog for this?

scalafix-core/src/main/scala/scalafix/rewrite/Scalameta17.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.meta._
+import scala.meta.semantic.v1.Completed
+import scala.meta.semantic.v1.Mirror
+import scala.meta.semantic.v1.Signature
+import scala.meta.semantic.v1.Symbol
+import scalafix.util.Patch
+import scalafix.util.TokenPatch._
+import scalafix.util.TreePatch.RenameSymbol
+
+case object Scalameta17 extends Rewrite[Mirror] {
```
xeno-by: It's amazing that we can achieve this functionality in such a concise manner!

olafurpg: I agree! However, I still think the API is too complex. In particular, it's annoying to 

- set up test suite
- set up rewrite boilerplate (`import scalafix._`)
- copy-paste symbol syntax from mirror.database

I believe some form of boilerplate generator + simpler UI to find symbols (instead of copy-pasting from mirror.database) would go a long  way.

scalafix-core/src/main/scala/scalafix/rewrite/Scalameta17.scala
olafurpg: I'll send a PR to scala.meta with instructions once I release scalafix. This rewrite uses new features.

olafurpg:
CI is green! I'm gonna go ahead and publish scalafix 0.3.4 to unblock the scala.meta 1.7 release announcement.

scalafix-core/src/main/scala/scalafix/util/Patch.scala
ShaneDelmore: What would the scope of RenameReference be?  Is one renameSyntactic and one renameSemantic?  i still don't clearly understand what rename this instance does. Is it just replacing one single sequence of consecutive tokens?

olafurpg: Actually, thinking about it, I think the current `Rename` patch (RenameReference/ThisInstance) can be replaced with a 

```scala
def rename(name: Name, to: Name): Seq[TokenPatch] = {
  val tok = name.tokens.head
  Seq(
    Remove(tok),
    AddLeft(tok, to.syntax)
  )
}

```

 helper method. There's really no reason to make it a TreePatch.

Issue #124, opened 2017-04-13
Explore ways to consolidate Patch and Rewrite


olafurpg:
From https://gitter.im/scalacenter/scalafix?at=58eecfd28bb56c2d11a16ebd

> Shane Delmore @ShaneDelmore 03:09
> @olafurpg I still have some confusion about Patch vs Rewrite. I have a sample AddGlobalImport, and don't fully understand how it is a Patch and not a Rewrite. https://github.com/ShaneDelmore/cbt/blob/10ef4c952149d1a9f52a707f36fafcdcbcc66d28/examples/cross-rewrite-example/build/build.scala#L44
> In this case, I really want to add it only if it is needed, for example if Either.map is used in the file, but in that case I need a Rewrite, not a Patch, and it seems they cannot be used in the same places.
> 
> Shane Delmore @ShaneDelmore 03:27
> How exactly is AddGlobalImport a patch? It seems like a rewrite, it must need a context right?
> Or some sort of PseudoPatch? It doesn't feel like quite the same thing as "replace these tokens with some other tokens".
> 
> Ólafur Páll Geirsson @olafurpg 09:54
> @ShaneDelmore Great question. I've thought a lot about how to consolidate Patch and Rewrite.
> AddGlobalImport is in one form a rewrite that only produces TokenPatches
> While a rewrite can return TreePatches
> The reason AddGlobalImport is not a rewrite is becuase 1. it needs to run last and 2. it requires a rewrite ctx + all import patches from other rewrites

Some current issues:

- Too many concepts: Patch/TreePatch/TokenPatch, Rewrite, RewriteCtx, Mirror, RewriteCtx[Any], ...
- Still easy to do dangerous stuff, like create RemoveGlobalImport (which require semantic info) from a syntactic rewrite (triggers runtime error in `Patch.apply`)

olafurpg:
OK, here's a draft of potential approach to simplify things. It still keeps  `Patch` but

- gets rid of all `T <: Patch` in the public facing
- requires semantic patches like AddGlobalImport to prove they have a `RewriteCtx[Mirror]` in scope
- automatically handles flattening of `Seq[Seq[Patch]]`/`Option[Patch]`/...

```scala
class Rewrite[T] {
  def rewrite(ctx: RewriteCtx[T]): Patch // not Seq[Patch], use Patch.++ to combine patches
}
sealed abstract class Patch {
  def syntax: String // code after applying this patch
[...]

PR #127, opened 2017-04-18
Reorganize scalafix rewrite/patch api

olafurpg:
From experience of implementing several rewrites
Related #124 

TODOs
- [x] private tree/token patches, users only see `Patch` now
- [x] `Patch.+`
- [x] Rewrite.rewrite returns instead of `Seq[Patch]`
- [x] `import scalafix._`
- [x] deprecate ScalafixMirror
- [x] ctx.addLeft/addRight/removeImport/...
- [x] s/Rewrite[T]/Rewrite/
- [x] get rid of CanOrganizeImports (remove ability to organize imports syntactically, it's unsafe anyways)
- [x] clean up type signature of `Patch.applied` (no more implicit RewriteCtx needed, solved with `InCtx(patch, ctx, mirror)`)
- [x] remove scalafix-nsc and ScalafixMirror in favor of scalahost-nsc and Mirror
- [x] migrate sbt-scalafix from scalafix-nsc to scalafix-cli + Mirror


olafurpg:
Copied from https://github.com/scalacenter/scalafix/issues/124#issuecomment-294895217, quoting myself 

> The breaking changes are quite large so I'm happy to personally migrate any existing code using the current rewrite api. Don't hesitate to ping me about this.

> It's possible and quite easy to provide an automatic migration for rewrites that extend Rewrite[T]. However, some code I've seen for example in the cbt scalafix plugin doesn't use the Rewrite[T]. I believe it's better to manually migrate in those cases.

olafurpg:
This PR took an interesting swing today when I realized that the Rewrite and RewriteCtx type parameter is problematic for safely compiling+loading rewrites from source/classpath/url/file. Type erasure means that once we classload a `Rewrite[T]` we can't idiomatically check whether that rewrite is semantic or syntactic. It's probably doable with hacks, but I prefer not to go that route.

At this realization, I was confronted with two choices:

1. add constraint on `T` in `Rewrite[T]`, for example classtag, this change would ripple throughout the entire scalafix codebase and onto scalafix users.
2. remove the type parameter and figure out another safe way to ensure that semantic rewrites only access.

After some prototyping and getting help from the guys at the lab I think I've found a pretty awesome workaround based on option 2. The trick is to let semantic rewrites take the mirror as a constructor argument

```scala
// class style
case class MyRewrite(implicit mirror: Mirror) extends SemanticRewrite(mirror) {
  def rewrite(ctx: ...): Patch = ???
}

// lambda style
val myRewrite = Rewrite.semantic { implicit mirror => ctx =>
  ...
}
```

With this change, we can safely classload rewrites as long as we have a `Mirror` available when loading configuration. My WIP branch for this change is here https://github.com/olafurpg/scalafix/tree/no-tparams everything seems to work except the configuration classloading is remaining. The really nice part of this change is that

- the scalafix public api is much simpler now, there's no need to pass around a type parameter anymore
- library users keep full ability to abstract over semantic/syntactic rewrites 
- we can provide helpful error messages when trying to load semantic rewrites without an available `Mirror`.
- This change introduces complexity on the scalafix configuration implementation, but that complexity does not leak onto the user!

olafurpg:
For anyone who wants to read a summary of what I've been working on for 4 days now, see https://github.com/scalacenter/scalafix/pull/127/commits/f18bd69a5ffaf083a9a580114c9b09b074f19cf6

xeno-by:
I wonder if there could be an implicit conversion from ctx to mirror, so that instead of `implicit mirror => ctx =>` you can simply write `implicit ctx =>`.

olafurpg:
How would you limit that implicit conversion to semantic rewrites? Syntactic rewrites should not have access to a mirror. 

I would prefer not to hide away the concept of mirror to rewrite authors, since it's fairly fundamental to how the semantic api works.

xeno-by:
It's a very good point, haven't thought of that.

olafurpg:
The major changes from today were 2347e0361eb8ac35d0314e70573c20f15f50d198 and e1d507c7c2eaabf4220c7d66a24b8e064d9e909c I hope and believe this was the last remaining bit to get this PR merged.

scalafix-core/src/main/scala/scalafix/Scalafix.scala
```diff
@@ -1,50 +1,23 @@
...
     try {
-      val patches = config.rewrites.flatMap(_.rewrite(ctx))
-      val patched =
-        Patch.apply(patches)(CanOrganizeImports.ScalafixMirror, ctx)
-      Fixed.Success(patched)
+      Fixed.Success(ctx.config.rewrite.wrappedRewrite(ctx).applied)
     } catch {
-      case e: Failure =>
-        Fixed.Failed(e)
+      case NonFatal(e) =>
```
olafurpg: I don't like this at all. However, there are quite many things that can go wrong here since we are calling user-implemented Rewrites and I don't want to crash the scalafix-cli when fixing hundreds of files only because one file failed.

scalafix-core/src/main/scala/scalafix/Scalafix.scala
```diff
@@ -1,50 +1,23 @@
...
-  }
-  def fix(ast: Tree,
-          config: ScalafixConfig,
-          semanticApi: Option[ScalafixMirror]): Fixed = {
-    implicit val ctx: ScalafixCtx = RewriteCtx(
-      ast,
-      config,
-      semanticApi.getOrElse(ScalafixMirror.failingMirror(config.dialect)))
+
+  def fix(ctx: RewriteCtx): Fixed =
```
olafurpg: I'm not really sure what to do with this method. I like having one single entry point for Scalafix, but at the point where you have a RewriteCtx you have already done all of the work required to run the fix. Everything this method does is handle the wrappedRewrite call which is an internal implementation detail, which may change in the future.

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -0,0 +1,186 @@
...
+    case _ =>
+      sys.error(s"""Can't merge token patches:
+                   |1. $a
+                   |2. $b""".stripMargin)
+  }
+  def apply(patch: Patch): String = patch match {
+    case InCtx(p, ctx, Some(mirror)) =>
+      semanticApply(underlying(p))(ctx, mirror)
+    case InCtx(p, ctx, None) =>
+      syntaxApply(ctx, underlying(p).collect { case tp: TokenPatch => tp })
```
olafurpg: TODO: report errors on TreePatches instead of silently ignoring them

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -0,0 +1,186 @@
...
+          else InCtx(ap + bp, ac, am.orElse(bm))
+        case (_, InCtx(p, ctx, m)) => InCtx(p + this, ctx, m)
+        case (InCtx(p, ctx, m), _) => InCtx(p + other, ctx, m)
+        case _ => Concat(this, other)
+      }
+    }
+  def ++(other: Seq[Patch]): Patch = other.foldLeft(this)(_ + _)
+
+  def appliedDiff: String = {
+    ???
```
olafurpg: TODO

scalafix-core/src/main/scala/scalafix/rewrite/SemanticPatchOps.scala
```diff
@@ -0,0 +1,30 @@
+// scalafmt: {maxColumn = 100}
+package scalafix
+package patch
+
+import scala.meta._
+import scalafix.patch.TokenPatch._
+import scalafix.patch.TreePatch._
+
+class SyntacticPatchOps(ctx: RewriteCtx) {
```
olafurpg: Rename file to match syntax + semantic ops.

scalafix-core/src/main/scala/scalafix/Scalafix.scala
xeno-by: What would you do otherwise?

scalafix-core/src/main/scala/scalafix/Scalafix.scala
xeno-by: In what situation would you already have a RewriteCtx?

xeno-by: Not sure about the difference between `rewrite` and `wrappedRewrite`.

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -0,0 +1,186 @@
...
+
+sealed abstract class Patch {
+  // NOTE: potential bottle-neck, this might be very slow for large
+  // patches. We might want to group related patches and enforce some ordering.
+  def +(other: Patch): Patch =
+    if (this eq other) this
+    else {
+      (this, other) match {
+        case (InCtx(ap, ac, am), b @ InCtx(bp, bc, bm)) =>
+          if (ac ne bc) ???
```
olafurpg: This should not happen if you use Scalafix public api correctly, but it's very easy to manually construct a patch that breaks this invariant. I've replaced all `???` with custom exceptions and explanations why that was unexpected. I want to add the ability to encode partial failures (for example conflict error when merging token patches) inside a single patch later down the road.

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -0,0 +1,186 @@
...
+          else InCtx(ap + bp, ac, am.orElse(bm))
```
olafurpg: Good catch, I now fail-fast with a detailed error message if `am ne bm`. This is an invariant that should never happen.

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
olafurpg: Done

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -0,0 +1,186 @@
...
+  def applied: String = Patch(this)
+//    Patch.apply[T](this)
+}
+
+private[scalafix] case class Concat(a: Patch, b: Patch) extends Patch
+private[scalafix] case object EmptyPatch extends Patch
+private[scalafix] case class InCtx(patch: Patch,
+                                   ctx: RewriteCtx,
+                                   mirror: Option[Mirror])
+    extends Patch
```
olafurpg: I've added a long comment why this was needed. It's subject to change in the future, but that should not affect the public api.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitImplicit.scala
```diff
@@ -1,21 +1,20 @@
...
 
-import scala.{meta => m}
-import scalafix.util.Patch
+import scala.meta._
 import scalafix.util.Whitespace
 import scala.collection.immutable.Seq
-import scalafix.util.TokenPatch
 
-case object ExplicitImplicit extends Rewrite[ScalafixMirror] {
+case class ExplicitImplicit(mirror: Mirror) extends SemanticRewrite(mirror) {
```
olafurpg: It used to be a parameter of `rewrite`. One of the major motivations for this PR was to remove `mirror` from the `rewrite`. The short answer why is that type erasure makes it hard to do this safely, the long explanation is here f18bd69a5ffaf083a9a580114c9b09b074f19cf6

scalafix-core/src/main/scala/scalafix/rewrite/Rewrite.scala
```diff
@@ -1,31 +1,59 @@
...
-    apply(f)(sourcecode.Name(name))
-  def apply[T](f: RewriteCtx[T] => Seq[Patch])(
-      implicit name: sourcecode.Name): Rewrite[T] = new Rewrite[T]() {
-    override def rewrite[B <: T](ctx: RewriteCtx[B]): Seq[Patch] = f(ctx)
+  val syntaxRewriteConfDecoder = config.rewriteConfDecoder(None)
+  def empty: Rewrite = syntactic(_ => Patch.empty)
+  def combine(rewrites: Seq[Rewrite], mirror: Option[Mirror]): Rewrite =
+    rewrites.foldLeft(mirror.fold(empty)(emptySemantic))(_ andThen _)
+  def emptySemantic(mirror: Mirror): Rewrite =
+    semantic(x => y => Patch.empty)(Name("empty"))(mirror)
```
olafurpg: Good question! I thought this was not necessary either so I removed it in one of the commits only to find out this is actually necessary to for example run the rewrites from .scalafix.conf `patches.renameSymbol=s/from/to`. In these cases we need an `InCtx(Some(mirror))` for symbols but the patch can be empty. 

I think there's a way to get rid of this one. I've added a comment explaining why it's here and a propsal on how to get rid of it eventually.

scalafix-reflect/src/main/scala/scalafix/reflect/RewriteInstrumentation.scala
```diff
@@ -21,6 +21,7 @@ object RewriteInstrumentation {
       def unapply(arg: Term): Boolean = arg match {
         case q"Rewrite.syntactic($_)" => true
         case q"Rewrite.semantic($_)" => true
+        case q"Rewrite.semantic($_)" => true
```
ShaneDelmore: Lines 23 and 24 are identical

olafurpg: Good catch! That was totally accidental, I was playing around with that line last night while prototyping on logo for scala.meta, it should never have been commited 😅 

Removed.

olafurpg:
OK I believe this PR is ready to be merged 🎉

PR #128, opened 2017-04-25
Add GitHub shorthand for URL rewrites

### LINKED ISSUE #125: Shorter URI syntax for rewrites
Following up on a discussion with @xeno-by. It would be nice if it was possible to configure rewrite with a short memorable syntax. For example,

```
github:olafurpg/scalafmt/1.0 => http://github.com/olafurpg/scalafix-scalafmt/blob/master/scalafmt/rewrites/src/main/scala/Scalafmt10.scala
```

The `github:` protocol could maybe even be automatically inferred, if this syntax becomes popular. I believe such a short syntax would help library authors give concise migration instruction for their users on how to run the scalafix migration script.

Related:

- https://www.scala-js.org/news/2017/03/21/announcing-scalajs-0.6.15/
- https://github.com/scalameta/scalameta/blob/master/changelog/1.7.0.md

gabro:
Closes #125 

First working draft. I've setup some integration tests but they fail because the file referenced is not there. Not sure whether these tests make sense, or whether we should just unit test the shorthand expansion. Anyway, the `FileNotFound` exception shows a correct value, so 🎉.

@olafurpg take a look and let me know what you think!

### DIFF
--- a/scalafix-reflect/src/main/scala/scalafix/reflect/ScalafixCompilerDecoder.scala
+++ b/scalafix-reflect/src/main/scala/scalafix/reflect/ScalafixCompilerDecoder.scala
@@ -28,6 +28,32 @@ object ScalafixCompilerDecoder {
     def unapply(arg: Conf.Str): Option[URL] = arg match {
       case UriRewrite("http" | "https", uri) if uri.isAbsolute =>
         Option(uri.toURL)
+      case GitHubUrlRewrite(url) => Option(url)
+      case _ => None
+    }
+  }
+
+  object GitHubUrlRewrite {
+    private[this] val GitHubShorthand =
+      """github:([^\/]+)\/([^\/]+)\/([^\/]+)""".r
+    private[this] val GitHubShorthandWithSha =
+      """github:([^\/]+)\/([^\/]+)\/([^\/]+)\?sha=(.+)""".r
+
+    private[this] def expandGitHubURL(org: String,
+                                      repo: String,
+                                      version: String,
+                                      sha: String): URL = {
+      val normVersion = version.replaceAll("[^\\d]", "_")
+      val fileName = s"${repo.toLowerCase.capitalize}_$normVersion.scala"
+      new URL(
+        s"https://github.com/$org/$repo/blob/$sha/scalafix-rewrites/src/main/scala/$repo/scalafix/$fileName")
+    }
+
+    def unapply(arg: Conf.Str): Option[URL] = arg.value match {
+      case GitHubShorthandWithSha(org, repo, version, sha) =>
+        Option(expandGitHubURL(org, repo, version, sha))
+      case GitHubShorthand(org, repo, version) =>
+        Option(expandGitHubURL(org, repo, version, "master"))
       case _ => None
     }
   }
--- a/scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
+++ b/scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
@@ -1,0 +1,23 @@
+package scalafix
+package tests
+
+import scalafix.reflect.ScalafixCompilerDecoder.GitHubUrlRewrite
+
+import org.scalatest.{FunSuiteLike, Matchers}
+import metaconfig.Conf
+
+class GitHubUrlRewriteSuite extends FunSuiteLike with Matchers {
+  test("it expands a GitHub shorthand with a default sha") {
+    Conf.Str("github:someorg/somerepo/1.2.3") match {
+      case GitHubUrlRewrite(url) =>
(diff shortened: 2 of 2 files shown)


gabro:
Hu, also, I haven't ran scalafmt on the patch yet, so the CI is probably going to complain about that too.

gabro:
Ok, I've run scalafmt, that part should be ok now

scalafix-tests/src/it/scala/scalafix/tests/URLConfiguration.scala
```diff
@@ -20,4 +20,37 @@ class URLConfiguration
        |
     """.stripMargin
   DiffTest("url", code, "UrlRewrite.stat").foreach(runDiffTest)
+
+  val githubURL =
+    "github:scalacenter/scalafix/1.0"
```
olafurpg: Actually, I just realised that we of course just want to unit test this feature. One integration test for the actual http request should be enough. We can create a new tests suite named `GitHubUrlRewriteTest` inside scalafix-tests/src/test/scala/...

scalafix-tests/src/it/scala/scalafix/tests/URLConfiguration.scala
```diff
@@ -20,4 +20,37 @@ class URLConfiguration
...
+  val githubCode =
+    s"""
+       |rewrites = [
+       |  "$githubURL"
+       |]
+       |<<< NOWRAP GitHub shorthand
```
olafurpg: Since we only need to test the String => URL expansion, there's no need to test this.

scalafix-tests/src/it/scala/scalafix/tests/URLConfiguration.scala
gabro: yep, I realized it while writing it as well. I'll add a unit test and remove this.

gabro:
@olafurpg done! I've force-pushed the version of the tests, no more integration tests, just unit tests checking the shorthand expansion.

scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
```diff
@@ -0,0 +1,23 @@
+package scalafix
+package tests
+
+import scalafix.reflect.ScalafixCompilerDecoder.GitHubUrlRewrite
+
+import org.scalatest.{ FunSuiteLike, Matchers }
+import metaconfig.Conf
+
+class GitHubUrlRewriteSuite extends FunSuiteLike with Matchers {
```
gabro: I've used `FunSuiteLike` directly since no scalafix specific features are needed.

scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
```diff
@@ -0,0 +1,23 @@
+package scalafix
+package tests
+
+import scalafix.reflect.ScalafixCompilerDecoder.GitHubUrlRewrite
+
+import org.scalatest.{FunSuiteLike, Matchers}
+import metaconfig.Conf
+
+class GitHubUrlRewriteSuite extends FunSuiteLike with Matchers {
```
gabro: I've used `FunSuiteLike` directly since no scalafix-specific features are needed.

olafurpg: Great!

gabro:
🎉

scalafix-reflect/src/main/scala/scalafix/reflect/ScalafixCompilerDecoder.scala
```diff
@@ -28,6 +28,32 @@ object ScalafixCompilerDecoder {
...
+      """github:([^\/]+)\/([^\/]+)\/([^\/]+)\?sha=(.+)""".r
+
+    private[this] def expandGitHubURL(org: String,
+                                      repo: String,
+                                      version: String,
+                                      sha: String): URL = {
+      val normVersion = version.replaceAll("[^\\d]", "_")
+      val fileName = s"${repo.toLowerCase.capitalize}_$normVersion.scala"
+      new URL(
+        s"https://github.com/$org/$repo/blob/$sha/scalafix-rewrites/src/main/scala/$repo/scalafix/$fileName")
```
olafurpg: I just realized that `repo` can contain non-alphanumerical characters like `-` which are unsupported in package names. What do you think @gabro about doing a `replaceAll("[^a-zA-Z0-9]", "_")`?

gabro: Right, tracking it here: https://github.com/scalacenter/scalafix/issues/131

I'll fix it soon.

PR #140, opened 2017-04-30
Remove more complexity from public api

olafurpg:
I watched https://vimeo.com/215325495 yesterday and got inspired to follow up on #127 and polish the rewrite api even in hope to make it more intuitive 

- rewrite apply/diff is safe by construction now
- No more scalafix.Scalafix object, Rewrite is the main api.
- No more InCtx or wrappedRewrite hacks
- ScalafixConfig.rewrite is now gone, which means `ScalafixConfig.default` will be what 99% of users will want.
- Because a sane ScalafixConfig is trivial to construct now, a sane RewriteCtx can now be constructed from an `Input`

Example motivation for this PR

```scala
val myRewrite = Rewrite.syntactic { ??? }
// before
Scalafix.fix(Input.String("code"),
            ScalafixConfig.default.copy(rewrite = myRewrite))
// after
val myRewrite = Rewrite.syntactic { ??? }
myRewrite(Input.String("code")
```

Left for future:
- Rewrite.apply/diff return String and throw exception on failure. Wait until https://github.com/scalameta/scalameta/issues/821 is resolved to figure out error handling strategy.
- RewriteCtx feels kind of clunky, could we get rid of it somehow?


scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRewriteSuite.scala
```diff
@@ -16,17 +15,14 @@ class SyntacticRewriteSuite(rewrite: Rewrite)
   def check(name: String, original: String, expected: String): Unit = {
     test(name) {
       import scala.meta._
-      val obtained =
-        Scalafix
-          .fix(Input.String(original), ScalafixConfig(rewrite = rewrite))
-          .get
+      val obtained = rewrite.apply(Input.String(original))
```
olafurpg: In a nutshell, this diff right here is the motivation for this PR.

olafurpg:
cc/ @ShaneDelmore more breaking changes in the public api, I hope only for the better.

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRewriteSuite.scala
ShaneDelmore: This looks great!  Nothing left you can remove at this point without overloading rewrite.apply for strings vs files and given that can be contentious this seems about as close as you can possibly get to:

val obtained = rewrite(original)

Big fan of these changes 👍

olafurpg: Awesome, I'm glad you like it. Yeah, I'm open to adding overloads for strings and java.io.File.

xeno-by: 😍

scalafix-cli/src/main/scala/scalafix/cli/Cli.scala
```diff
@@ -289,36 +299,28 @@ object Cli {
...
-  }
-
   def handleFile(file: File, options: ScalafixOptions): ExitStatus = {
     val config =
       if (file.getAbsolutePath.endsWith(".sbt")) options.resolvedSbtConfig
       else options.resolvedConfig
-
-    val fixed = Scalafix.fix(Input.File(file), config.get)
+    val fixed =
+      Try(options.resolvedRewrite.get.apply(Input.File(file), config.get))
```
xeno-by: What do you think about the necessity to use multiple gets here?

scalafix-core/src/main/scala/scalafix/rewrite/Rewrite.scala
```diff
@@ -2,38 +2,69 @@ package scalafix
...
+
+  /** Returns unified diff from applying this patch */
+  final def diff(ctx: RewriteCtx): String =
+    diff(ctx, rewrite(ctx))
+  final protected def diff(ctx: RewriteCtx, patch: Patch): String = {
+    val original = ctx.tree.input
+    Patch.unifiedDiff(original,
+                      Input.LabeledString(original.label, apply(ctx, patch)))
+
+  }
```
xeno-by: I like this API a lot!

scalafix-cli/src/main/scala/scalafix/cli/Cli.scala
olafurpg: It's just sloppy coding, been meaning to refactor the cli a bit to get rid of those

PR #141, opened 2017-05-02
Add remove/replace token patches.

olafurpg:
These were useful for this rewrite here:
https://github.com/scalacenter/scalafix/commit/5865d681bb6c91313dd1039b529c72c1c126baad


ShaneDelmore:
I had my own version of these as well, nice to have them standard now.

PR #142, opened 2017-05-02
Safer derivation of package name from repo name

### LINKED ISSUE #131: Safer normalization of package-name in GitHub shorthand expansion
Currently we use the repo name as package name, but some repo names can contain invalid characters for a package (like `-`).

A simple solution is to replace all the invalid characters with a `_`. As per @olafurpg proposal a `replaceAll("[^a-zA-Z0-9]", "_")` should be enough.

gabro:
Closes #131 and also handles the case in which a repo name starts with a digit.

See unit tests, but in short

`some-repo` => `some_repo`

`42some-repo` => `_42some_repo`

Ugly but consistent (naming is hard)

### DIFF
--- a/scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
+++ b/scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
@@ -21,18 +21,18 @@ class GitHubUrlRewriteSuite extends FunSuiteLike with Matchers {
     }
   }
 
-  test("it replaces invalid characters in package name with _") {
+  test("it replaces invalid characters in package name and file name with _") {
     Conf.Str("github:someorg/some-repo/1.2.3") match {
       case GitHubUrlRewrite(url) =>
-        url.toString shouldBe "https://github.com/someorg/some-repo/blob/master/scalafix-rewrites/src/main/scala/some_repo/scalafix/Some-repo_1_2_3.scala"
+        url.toString shouldBe "https://github.com/someorg/some-repo/blob/master/scalafix-rewrites/src/main/scala/some_repo/scalafix/Some_repo_1_2_3.scala"
     }
   }
 
   test(
-    "it adds an underscore to the package name if the repo name begins with a digit") {
+    "it adds an underscore to the package name and to the file name if the repo name begins with a digit") {
     Conf.Str("github:someorg/42some-repo/1.2.3") match {
       case GitHubUrlRewrite(url) =>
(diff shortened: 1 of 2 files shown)


scalafix-tests/src/test/scala/scalafix/tests/GitHubUrlRewriteSuite.scala
```diff
@@ -20,4 +20,19 @@ class GitHubUrlRewriteSuite extends FunSuiteLike with Matchers {
         url.toString shouldBe "https://github.com/someorg/somerepo/blob/master~1/scalafix-rewrites/src/main/scala/somerepo/scalafix/Somerepo_1_2_3.scala"
     }
   }
+
+  test("it replaces invalid characters in package name with _") {
+    Conf.Str("github:someorg/some-repo/1.2.3") match {
+      case GitHubUrlRewrite(url) =>
+        url.toString shouldBe "https://github.com/someorg/some-repo/blob/master/scalafix-rewrites/src/main/scala/some_repo/scalafix/Some-repo_1_2_3.scala"
```
olafurpg: I just noticed that the filename can be `Some-repo_1_2_3.scala`, what do you think about using `packageName_normVersion.scala` instead?

gabro: Yeah, I noticed that too but I wasn't sure how to address it. Probably underscores are more consistent with the filename, so it makes sense.

Also, are you ok with keeping the capitalization? So it would be

```scala
s"${packageName.capitalize}_${normVersion}.scala"
```

olafurpg: I agree, I think it's a good idea to capitalize the package(name in the filename.

gabro: Sure, I'll do it.

Also, do you think it makes sense to make the `packageName` lowercase?

olafurpg: Aha, I'm not sure. For example for https://github.com/ThoughtWorksInc/Binding.scala you would get package thoughtworksinc, if the org uses uppercase letters it's usually for a reason. WDYT?

gabro: Right. Even though, in that specific example, they use a lowercase package themselves.
I'll leave it as is for now, I guess edge cases will pop up.

Right now instead the filename is lowercased, then capitalized. Should we keep the strategy or do we revert to now lowercasing it?

olafurpg:
Good catch! Let's lowercase it then.

gabro:
Done! @olafurpg

Issue #66, opened 2017-02-13
Auto-tupling


olafurpg:
See https://github.com/samuelgruetter/srewrite-test/compare/diffdemo1original...diffdemo1#diff-9c0d11181aa5ea69a38f9de1ec4fffd6

gabro:
I'm still wrapping my head around scalafix (as you may have noticed :) ). Any pointers on how to approach this? I'd like to give it a try!

olafurpg:
I noticed :) 

The way I would approach this one would be to start writing a small unit test


```
<<< ONLY auto-tupling
object tup {
  def fooo(t: (Int, Int)): Int = ???
  fooo(1, 2)
}
>>>
object tup {
  def fooo(t: (Int, Int)): Int = ???
  fooo((1, 2))
}
```

and then create a semantic Rewrite

```
case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
  override def rewrite(ctx: RewriteCtx): Patch = {
    logger.elem(mirror.database)
    ???
  }
}
```

And then you should see something like

```
ExplicitArgs:11 [mirror.database]:
/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/paradise3423812327028985194.scala
----------------------------------------------------------------------------------

Names:
[8..19): autotupling => _root_.autotupling.
[29..32): tup => _root_.autotupling.tup.
[41..45): fooo => _root_.autotupling.tup.fooo(Lscala/Tuple2;)I.
[46..47): t => _root_.autotupling.tup.fooo(Lscala/Tuple2;)I.(t)
[50..53): Int => _root_.scala.Int#
[55..58): Int => _root_.scala.Int#
[62..65): Int => _root_.scala.Int#
[68..71): ??? => _root_.scala.Predef.`???`()Lscala/Nothing;.
[74..78): fooo => _root_.autotupling.tup.fooo(Lscala/Tuple2;)I.
```

The interesting bit is the last line which shows that `fooo` takes a single argument of Tuple2. That should be a good indicator that auto-tupling is happening. You can query for the symbol with the extension method `Ref.symbol` as long as an implicit Mirror is in scope. `Ref` is a supertype of `Term.Name` like `foo` in `foo(1, 2)`.


[...]

PR #143, opened 2017-05-04
Add NoAutotupling semantic rewrite

gabro:
This is a first attempt at #66.

Trivial cases work fine, but it fails with multiple parameter lists since the tree is matched multiple times by `q"$fun(..$argss)"`.

E.g. when traversing the tree of `foo(1, 2)("a", "b")`, the pattern `q"$fun(..$argss)` matches

```scala
foo(1, 2)
```

and 

```scala
foo(1, 2)("a", "b")
```

so we end up with

```scala
foo(((1, 2)))(("a", "b"))
```

@olafurpg any hint to solve this?

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -2,11 +2,8 @@ package scalafix
 package rewrite
 
 import scala.meta._
-import scala.meta.contrib._
 import scalafix.syntax._
-import scala.meta.semantic.v1.Signature
-
-import org.scalameta.logger
+import scala.meta.semantic.Signature
 
 case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
 
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -1,0 +1,51 @@
+package scalafix
+package rewrite
+
+import scala.meta._
+import scala.meta.contrib._
+import scalafix.syntax._
+import scala.meta.semantic.v1.Signature
+
+import org.scalameta.logger
+
+case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
+
+  private[this] val singleTuplePattern = "^Lscala\\/Tuple\\d{1,2}$".r.pattern
+
+  private[this] def addWrappingParens(ctx: RewriteCtx,
+                                      args: Seq[Term.Arg]): Patch =
+    Seq(
+      ctx.addLeft(args.head.tokens.last, "("),
+      ctx.addRight(args.last.tokens.last, ")")
+    ).asPatch
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    ctx.tree.collect {
+      case x @ q"$fun(...$argss)"
+          if argss.length > 0 && fun.isInstanceOf[Ref] =>
+        mirror
+          .symbol(fun.asInstanceOf[Ref])
+          .toOption
+          .collect {
+            case Symbol.Global(_, Signature.Method(_, jvmSignature)) =>
+              jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
+          }
+          .map { argListSignatures =>
+            argListSignatures
+              .zip(argss)
+              .foldLeft(Seq.empty[Patch]) {
+                case (patches, (argListSignature, args)) =>
+                  if (singleTuplePattern
+                        .matcher(argListSignature)
+                        .matches && args.length > 1) {
+                    patches :+ addWrappingParens(ctx, args)
+                  } else {
+                    patches
+                  }
+              }
+              .asPatch
+          }
+          .getOrElse(Patch.empty)
+    }.asPatch
+  }
+}
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -12,7 +12,7 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
 
   private[this] def addWrappingParens(ctx: RewriteCtx,
                                       args: Seq[Term.Arg]): Patch =
-    ctx.addLeft(args.head.tokens.last, "(") +
+    ctx.addLeft(args.head.tokens.head, "(") +
       ctx.addRight(args.last.tokens.last, ")")
 
   override def rewrite(ctx: RewriteCtx): Patch = {
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -12,40 +12,35 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
 
   private[this] def addWrappingParens(ctx: RewriteCtx,
                                       args: Seq[Term.Arg]): Patch =
-    Seq(
-      ctx.addLeft(args.head.tokens.last, "("),
+    ctx.addLeft(args.head.tokens.last, "(") +
       ctx.addRight(args.last.tokens.last, ")")
-    ).asPatch
 
   override def rewrite(ctx: RewriteCtx): Patch = {
     // "hack" to avoid fixing an argument list more than once due
     // to recursive matching of multiple parameters lists.
     val fixed = MutableSet.empty[Seq[Term.Arg]]
     ctx.tree.collect {
-      case x @ q"$fun(...$argss)"
-          if argss.length > 0 && fun.isInstanceOf[Ref] =>
+      case q"${fun: Term.Ref}(...$argss)" if argss.nonEmpty =>
         mirror
-          .symbol(fun.asInstanceOf[Ref])
+          .symbol(fun)
           .toOption
           .collect {
             case Symbol.Global(_, Signature.Method(_, jvmSignature)) =>
-              jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
-          }
-          .map { argListSignatures =>
-            argListSignatures
-              .zip(argss)
-              .foldLeft(Seq.empty[Patch]) {
-                case (patches, (argListSignature, args)) =>
-                  if (!fixed(args) && singleTuplePattern
-                        .matcher(argListSignature)
-                        .matches && args.length > 1) {
-                    fixed += args // dirty hack, see explanation above
-                    patches :+ addWrappingParens(ctx, args)
-                  } else {
-                    patches
-                  }
-              }
-              .asPatch
+              val argListSignatures =
+                jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
+              argListSignatures
+                .zip(argss)
+                .foldLeft(Patch.empty) {
+                  case (patches, (argListSignature, args)) =>
+                    if (!fixed(args) && singleTuplePattern
+                          .matcher(argListSignature)
+                          .matches && args.length > 1) {
+                      fixed += args // dirty hack, see explanation above
+                      patches + addWrappingParens(ctx, args)
+                    } else {
+                      patches
+                    }
+                }
           }
           .getOrElse(Patch.empty)
     }.asPatch
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -4,6 +4,7 @@ package rewrite
 import scala.meta._
 import scalafix.syntax._
 import scala.meta.semantic.Signature
+import scala.collection.mutable.{Set => MutableSet}
 
 case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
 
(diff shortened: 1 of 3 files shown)


gabro:
Ouch, I also found another failing test case.

```scala
val foo = (a: (Int, Boolean)) => a
foo(42, true)
```

When querying the semantic db, I get this `Symbol` for `foo`

```scala
_root_.scala.`Function1`#apply(Ljava/lang/Object;)Ljava/lang/Object;
```

🤔

olafurpg:
Wow, the progress on this rewrite is impressive! Great work @gabro 

> When querying the semantic db, I get this Symbol for foo

That is a good question. Maybe the symbol signature is not enough to handle that case 🤔 However, I have a scalafix WIP branch to update for https://github.com/scalameta/scalameta/pull/808 that adds Denotations. It seems denotations will include the information we need

```scala

Names:
[8..19): autotupling => _root_.autotupling.
[29..32): tup => _root_.autotupling.tup.
[41..44): foo => _root_.autotupling.tup.foo.
[48..49): a => _root_.autotupling.tup.foo.$anonfun.(a)
[52..55): Int => _root_.scala.Int#
[57..64): Boolean => _root_.scala.Boolean#
[70..71): a => _root_.autotupling.tup.foo.$anonfun.(a)
[74..77): foo => _root_.scala.Function1#apply(Ljava/lang/Object;)Ljava/lang/Object;. // this could be problematic, should probably be a Symbol.Multi and include _root_.autotupling.tup.foo.

Denotations:
_root_.autotupling. => package autotupling
_root_.autotupling.tup. => final object tup
_root_.autotupling.tup.foo. => val foo: ((Int, Boolean)) => (Int, Boolean) // <-- How cool is this?!
_root_.autotupling.tup.foo.$anonfun.(a) => param a: (Int, Boolean)

Sugars:
[87..87) [Int, Boolean]
```

I propose we put this rewrite on hold until the scalameta dependency has been updated. Denotations are still a bit rough around the edges, in https://github.com/scalameta/scalameta/issues/822 we are discussing what data structures we want to use to represent them, currently they're just a String.

gabro:
Hey, thanks for the feedback! I don't really know what denotations are (although they look interesting). Any quick explanation or resources you can point me to? Thanks!

olafurpg:
> it fails with multiple parameter lists since the tree is matched multiple times by q"$fun(..$argss)".

I've hit on similar problems myself. One possible but dirty workaround is to avoid the second match by keeping track of which trees have already been rewritten. 

```scala
val matched = mutable.Set[Term.Arg]
tree.collect {
  case x @ q"fun(..$args)" =>
    args.collect {
      case arg if !matched(arg) =>
      matched += x
    }
    ..
```

I don't have a cleaner solution at the moment, maybe it's possible to bake some idempotency checks into the Patch system.

olafurpg:
> Any quick explanation or resources you can point me to? Thanks!

Nothing official yet, we are still figuring this out ourselves 😄 maybe this discussion helps https://github.com/scalameta/scalameta/pull/808/files/d5596df536dd85f8dbc5c7b88c7e7e2d19d4e819#r113825329

Denotations contain information about a symbol such as 

- modifiers (isFinal, isLazy)
- the shape of the definition (curried or not)

olafurpg:
btw, if you want to poke into denotations you can try checking out this branch here https://github.com/scalacenter/scalafix/compare/master...olafurpg:meta-1.8?expand=1 

The listing above for `foo` is printed at this commit https://github.com/scalacenter/scalafix/commit/1ccaaa4a289e08d035a83fb0b80a0404ab5667dd

gabro:
Thanks for the pointers!

> I don't have a cleaner solution at the moment, maybe it's possible to bake some idempotency checks into the Patch system.

Another solution I'm considering for the multiple paramss problem is using the signature from the semantic db. When matching an `Apply` node with a N parameter lists I can check the signature of `foo` and if it's something in the for of `sig;sig` I skip that iteration until the number of params lists match.

Does it make any sense?

olafurpg:
@gabro You can rebase now on top of master which includes #145 that added denotations. Note that denotation.info is currently a string but in the future it will be a richer data structure.

gabro:
Thanks, I'm playing around with it!
Only, I cannot get the denotation for `foo`, given

```scala
val foo = (a: (Int, Boolean)) => a
foo(42, true)
```

What I'm doing is

```scala
ctx.tree.collect {
   case x @ q"$fun(...$argss)" if argss.length > 0 && fun.isInstanceOf[Ref] =>
     println(fun.asInstanceOf[Ref].symbol.info)
```

and I get

> scala.meta.semantic.SemanticException: error: fatal error: semantic DB doesn't contain _root_.scala.Function1#apply(Ljava/lang/Object;)Ljava/lang/Object;.

gabro:
In the meanwhile, I went for the "dirty hack" solution and multiple parameters lists seem to work now! 🎉

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+      ctx.addLeft(args.head.tokens.last, "("),
+      ctx.addRight(args.last.tokens.last, ")")
+    ).asPatch
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    // "hack" to avoid fixing an argument list more than once due
+    // to recursive matching of multiple parameters lists.
+    val fixed = MutableSet.empty[Seq[Term.Arg]]
+    ctx.tree.collect {
+      case x @ q"$fun(...$argss)"
```
olafurpg: `q"${fun: Term.Ref}(...$argss)"` should avoid the `isInstanceOf`/`asInstanceOf`

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+          .symbol(fun.asInstanceOf[Ref])
+          .toOption
+          .collect {
+            case Symbol.Global(_, Signature.Method(_, jvmSignature)) =>
+              jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
+          }
+          .map { argListSignatures =>
+            argListSignatures
+              .zip(argss)
+              .foldLeft(Seq.empty[Patch]) {
```
olafurpg: You can start with `Patch.empty` and use `Patch.+(Patch)`

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -0,0 +1,89 @@
+rewrites = [NoAutoTupling]
```
olafurpg: I haven't updated the other test suites to take advantage of the fact that you can now skip the `[]` and write `rewrites = NoAutoTupling`. Gonna add support to write the singular `rewrite = NoAutoTupling`. I will update it in a future PR, so don't worry about it, just an FYI :)

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+      case x @ q"$fun(...$argss)"
+          if argss.length > 0 && fun.isInstanceOf[Ref] =>
+        mirror
+          .symbol(fun.asInstanceOf[Ref])
+          .toOption
+          .collect {
+            case Symbol.Global(_, Signature.Method(_, jvmSignature)) =>
+              jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
+          }
+          .map { argListSignatures =>
```
olafurpg: I think we can fuse the `collect` and `map` here.

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+                                      args: Seq[Term.Arg]): Patch =
+    Seq(
+      ctx.addLeft(args.head.tokens.last, "("),
+      ctx.addRight(args.last.tokens.last, ")")
+    ).asPatch
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    // "hack" to avoid fixing an argument list more than once due
+    // to recursive matching of multiple parameters lists.
+    val fixed = MutableSet.empty[Seq[Term.Arg]]
```
olafurpg: This is OK for now. I think we might be able bake some idempotency into the patch system, because I think this will pop up regularly.

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -0,0 +1,89 @@
...
+  def sum(a: Int, b: (Int, String)): Int = ???
+  sum(1, (2, "foo"))
+}
+>>>
+object tup {
+  def sum(a: Int, b: (Int, String)): Int = ???
+  sum(1, (2, "foo"))
+}
+
+<<< auto-tupling with lambdas
```
olafurpg: I OK with adding a SKIP to this test for now. It might be good to add another similar test that does currying

```scala
@ def f: (((Int, String)) => ((String, List[Int])) => Int) = a => b => a._1
defined function f
@ f(1, "")("a", Nil) // original
res17: Int = 1
@ f((1, ""))(("a", Nil)) // expected
res16: Int = 1
```
because the current implementation uses `.zip` which will not work across curried anonymous functions.

However, as long as the rewrite does nothing in these cases and documents this limitation then it's OK for now. I opened an issue in scalameta https://github.com/scalameta/scalameta/issues/837 but I don't see a reason to block this PR until that issue is fixed. This rewrite is already very useful as long as it does not have false positives.

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -0,0 +1,89 @@
...
+  def baar(t: (Boolean, Int, String))(s: (Boolean, String))(k: (Int, String)): String = ???
+  baar(true, 1, "foo")((false, "42"))(42, "foo")
+}
+>>>
+object tup {
+  def fooo(t: (Int, Int))(s: (String, String)): Int = ???
+  fooo((1, 2))(("a", "b"))
+
+  def baar(t: (Boolean, Int, String))(s: (Boolean, String))(k: (Int, String)): String = ???
+  baar((true, 1, "foo"))((false, "42"))((42, "foo"))
```
olafurpg: 😻

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+    ctx.tree.collect {
+      case x @ q"$fun(...$argss)"
+          if argss.length > 0 && fun.isInstanceOf[Ref] =>
```
olafurpg: `.nonEmpty`?

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+import scala.collection.mutable.{Set => MutableSet}
+
+case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
+
+  private[this] val singleTuplePattern = "^Lscala\\/Tuple\\d{1,2}$".r.pattern
+
+  private[this] def addWrappingParens(ctx: RewriteCtx,
+                                      args: Seq[Term.Arg]): Patch =
+    Seq(
+      ctx.addLeft(args.head.tokens.last, "("),
```
olafurpg: Should this be `args.head.tokens.head`?

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -0,0 +1,53 @@
...
+      ctx.addRight(args.last.tokens.last, ")")
+    ).asPatch
```
olafurpg: Patch.+(Patch) would also work, but either way is fine.

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -0,0 +1,89 @@
+rewrites = [NoAutoTupling]
+<<< add explicit tuple
+object tup {
+  def fooo(t: (Int, Int)): Int = ???
+  fooo(1, 2)
```
olafurpg: Try this

```
  fooo(1 + 2, 2)
```

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
olafurpg: See `fooo(1 + 2, 2)` comment below.

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
gabro: oh, that's how you do it! I was wondering... Thanks

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
gabro: Ah right, I've seen that issue. Ok, I'll leave it like this, so that it can be fixed together with the rest.

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
gabro: Woops, nice catch, it's definitely `tokens.head`, thanks!

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
gabro: Definitely. Currying was on my list too. I'll add a test to check it doesn't raise a false positive.

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
gabro: nice to know! I totally missed it, thanks

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
gabro: Yes, I think it would make things much easier to work with.

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
gabro: Done, it doesn't seem like a false positive (no rewrite is applied).

The "curried" case falls together with the lambda case, since we're stuck with a `Ljava/lang/Object` signature and we have no way to access its denotation at call site.

I'll open an issue here on scalafix to track the two skipped tests, pointing to the scalameta issue.

PR #145, opened 2017-05-05
Upgrade to scalameta 1.8 pre-release and cleanup

olafurpg:


project/Dependencies.scala
```diff
@@ -2,12 +2,10 @@ import sbt._
 /* scalafmt: { maxColumn = 120 }*/
 
 object Dependencies {
-  val scalametaV = "1.7.0"
+  val scalametaV = "1.8.0-594-3890ed93"
```
allanrenucci: This version does not include the changes to the xml literal tokenizer

olafurpg: Good catch, I will update it to 1.8.0-600-ae298f26

olafurpg:
It seems I need to update paradise and metaconfig first before proceeding with this PR. I will be unable to complete it today, I will try to get it in monday @allanrenucci

PR #146, opened 2017-05-05
Rewrite Xml Literals to Xml Interpolator

allanrenucci:
Based on #145 and [Olaf's work](https://github.com/olafurpg/scalafix/commits/xml-lit-to-xml-interpolator)


scalafix-tests/src/test/resources/checkSyntax/NoXml.source
```diff
@@ -0,0 +1,99 @@
...
+}
+
+<<< splice in attribute position
+object A {
+  <foo bar={"Hello"}/>
+}
+>>>
+import scala.xml.quote._
+object A {
+  xml"""<foo bar=${"Hello"}/>"""
```
allanrenucci: This fails to typecheck. I think it is a bug in scalameta

olafurpg: What's the compilation error? Scalameta does not typecheck

allanrenucci: [Here](https://gist.github.com/allanrenucci/83db9369e47fd3eb3d1a599c598dc7d5)

allanrenucci:
@olafurpg I published on bintray. Shouldn't it work?

scalafix-tests/src/test/resources/checkSyntax/NoXml.source
olafurpg: I've hit on this error as well when I tried the changes from  https://github.com/scalameta/scalameta/pull/808 it's a bug in the scalahost compiler plugin, I opened https://github.com/scalameta/scalameta/issues/836

olafurpg:
@allanrenucci #145  has now been merged!

scalafix-tests/src/test/resources/checkSyntax/NoXml.source
allanrenucci: Do you want me to skip this test to get this merged or should we wait for the bug to be fixed?

olafurpg:
I propose we skip this test and open an issue to remind us to enable this
test later.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitUnit.scala
```diff
@@ -5,6 +5,7 @@ import scala.meta._
 
 case object ExplicitUnit extends Rewrite {
   override def rewrite(ctx: RewriteCtx): Patch = {
+    org.scalameta.logger.elem(ctx.tree)
```
allanrenucci: @olafurpg This was introduced in your previous commit. Do you want me to get rid of this?

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRewriteSuite.scala
```diff
@@ -97,10 +98,10 @@ abstract class SemanticRewriteSuite(
 
     val fixed = fix(diffTest.wrapped(), diffTest.config)
     val obtained = parse(diffTest.unwrap(fixed))
+    logger.elem(obtained)
```
allanrenucci: Same

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRewriteSuite.scala
```diff
@@ -97,10 +98,10 @@ abstract class SemanticRewriteSuite(
...
     val expected = parse(expectedStr)
     try {
       typeChecks(diffTest.wrapped(fixed))
-      checkMismatchesModuloDesugarings(obtained, expected)
```
allanrenucci: same

allanrenucci:
I'd like to emit a warning on `{{` occurrences in xml part. Is it possible with current scalafix?
I can do it in a separate PR.

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
```diff
@@ -0,0 +1,42 @@
...
+        else Patch.empty
+
+    }.asPatch
+
+    // Emit warning on `{{`
+    ctx.tree.tokens.foreach {
+      case tok @ Token.Xml.Part(part) if part.contains("{{") =>
+        val offset = part.indexOf("{{")
+        val pos =
+          Position.Range(tok.input, tok.start + offset, tok.start + offset + 2)
```
allanrenucci: Is there a better way to construct position from existing ones?

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
```diff
@@ -0,0 +1,42 @@
...
+        ctx.addLeft(tok, "$")
+
+      case tok @ Token.Xml.Part(part) =>
+        if (part.contains('$'))
+          ctx.replaceToken(tok, part.replaceAllLiterally("$", "$$"))
+        else Patch.empty
+
+    }.asPatch
+
+    // Emit warning on `{{`
```
allanrenucci: How can I test this?

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import scala.meta._
+import patch.TreePatch.AddGlobalImport
+
+case object NoXml extends Rewrite {
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    val patch = ctx.tree.tokens.collect {
+      case tok @ Token.Xml.Start() =>
+        ctx.addLeft(tok, "xml\"\"\"")
```
olafurpg: Would it be possible to determine simple cases where single quotes would be enough? For small literals, I think many users would prefer single quotes 

```scala
// before
<br/>
// now
xml"""<br/>"""
// after
xml"<br/>"
```

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
olafurpg: Good question. I've actually considered replacing this mutable reporter (`info(): Unit`) with empty patches like `ctx.info(msg, pos): Patch`. That should make it easier to test.

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRewriteSuite.scala
olafurpg: This is OK, I've been meaning to remove this for a while.

scalafix-tests/src/test/resources/checkSyntax/NoXml.source
```diff
@@ -0,0 +1,99 @@
...
+  val foo =
+    <div>
+      <span>Hello</span>
+    </div>
+}
+>>>
+import scala.xml.quote._
+object A {
+  val foo =
+    xml"""<div>
```
olafurpg: Would it make a semantic difference if this was formatted like this?

```scala
val foo =
  xml"""
  <div>
    <span>...</span>
  </div>
  """
would the whitespace in the start/end get included?

File: scalafix-tests/src/test/resources/checkSyntax/NoXml.source (hunk 1 above)
Comment (olafurpg) [inline scalafix-tests/src/test/resources/checkSyntax/NoXml.source]: Skip for now, this rewrite is actually syntactic so we shouldn't even need to typecheck with scalahost-nsc. It will be possible to run this rewrite from the cli without sbt-scalafix.

File: scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala (hunk 6 above)
Comment (olafurpg) [inline scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala]: Either way, we should probably make it possible to test linter warnings like they do in scala/scala 

```scala
>>>
object a {
  xml"<foo>{{" // scalafix warn: no longer needed
}
```
that comment would then get remove in the diff and turn into check.

olafurpg: For now, manual tests is OK

olafurpg: cc/ #92

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
```diff
@@ -0,0 +1,42 @@
...
+    }.asPatch
+
+    // Emit warning on `{{`
+    ctx.tree.tokens.foreach {
+      case tok @ Token.Xml.Part(part) if part.contains("{{") =>
+        val offset = part.indexOf("{{")
+        val pos =
+          Position.Range(tok.input, tok.start + offset, tok.start + offset + 2)
+        ctx.reporter.warn(
+          "Single opening braces within XML text don't need to be doubled",
```
allanrenucci: The issue is that there are many places in xml literals where braces don't need to be escaped:
```scala
scala> <a b="{{"/>
res0: scala.xml.Elem = <a b="{{"/>

scala> <![CDATA[{{]]> 
res1: scala.xml.PCData = <![CDATA[{{]]>

scala> <xml:unparsed>{{</xml:unparsed>
res2: scala.xml.Unparsed = {{

scala> <?foo {{?>
res3: scala.xml.ProcInstr = <?foo {{?>
```
We would need to parse the xml literal to implement this rewrite

scalafix-tests/src/test/resources/checkSyntax/NoXml.source
allanrenucci: The whitespaces in the start/end don't get included. Bu otherwise all the whitespace are preserved. So we should not shift right/left the xml literal.

build.sbt
```diff
@@ -85,6 +85,7 @@ lazy val allSettings = List(
   resolvers += Resolver.bintrayRepo("scalameta", "maven"),
   resolvers += Resolver.sonatypeRepo("releases"),
   triggeredMessage in ThisBuild := Watched.clearWhenTriggered,
+  incOptions := incOptions.value.withRecompileOnMacroDef(false),
```
allanrenucci: Yes you did. I'll remove it

build.sbt
```diff
@@ -228,9 +229,11 @@ lazy val tests = project
         "; tests/it:testQuick" // hack to workaround cyclic dependencies in test.
     )(state.value),
     parallelExecution in Test := true,
+    resolvers += Resolver.bintrayRepo("allanrenucci", "maven"),
```
allanrenucci: The plan is to have an official release when it gets merged into scala-xml. cc/ @densh

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
allanrenucci: If we work on the parsed tree, then I guess we can do something like:
```scala
ctx.tree.collect { case xml @ Xml(parts, args) =>
  val quote =
    if (isMutliLine(xml)) "\"\"\""
    else if (parts.exists(containsEscapeSequence)) "\"\"\""
    else "\""

  val patch = ...
}
```

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
allanrenucci: I guess I'd like to be able to do something like:
```scala
tok.start.toPos(size = 2)
Position.Range(tok.start, size = 2)
```
I think it is weird to have to pass the input to the Position's constructor when `start: Point` and `end: Point` also carry the input.
WDYT?

scalafix-core/src/main/scala/scalafix/rewrite/NoXml.scala
```diff
@@ -0,0 +1,42 @@
+package scalafix
+package rewrite
+
+import scala.meta._
+import patch.TreePatch.AddGlobalImport
+
+case object NoXml extends Rewrite {
```
allanrenucci: +1 for RemoveXmlLiterals

scalafix-tests/src/test/resources/checkSyntax/NoXml.source
allanrenucci: Shouldn't it be solved by a formater?

allanrenucci: This seems tricky to do

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
```diff
@@ -0,0 +1,105 @@
...
+
+import scala.meta._
+import patch.TreePatch.AddGlobalImport
+
+/** Rewrite Xml Literal to Xml Interpolator
+  *
+  * e.g.
+  * {{{
+  *   // before:
+  *   <div>{ "Hello }</div>
```
olafurpg: Is there missing a closing quote?

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
```diff
@@ -0,0 +1,112 @@
...
+object A {}
+
+<<< single line, no splice
+object A {
+  <div></div>
+}
+>>>
+import scala.xml.quote._
+object A {
+  xml"<div></div>"
```
olafurpg: Nice! 😍  I think users will appreciate having single quotes for simple xml literals.

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
```diff
@@ -0,0 +1,112 @@
...
+}
+
+<<< SKIP protect curly brace
+object A {
+  <div>{{</div>
+}
+>>>
+import scala.xml.quote._
+object A {
+  xml"<div>{</div>"
```
olafurpg: Awesome!

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /** Emit warning on `{{` */
+    def warnOnDoubledBrace(xml: Tree) = xml.tokens.foreach {
+      case tok @ Token.Xml.Part(part) if part.contains("{{") =>
+        val offset = part.indexOf("{{")
+        val pos =
+          Position.Range(tok.input, tok.start + offset, tok.start + offset + 2)
+        ctx.reporter.warn(
+          "Single opening braces within XML text don't need to be doubled",
```
olafurpg: Is this still necessary? The test case rewrites the `{{` to `{`. 

If this is still necessary, can we rephrase the warning to include more details. Something like

> Single opening braces don't need be escaped with `{{` inside the xml interpolator, unlike xml literals. For example `<x>{{</x>` is identical to `xml"<x>{</x>"`. (explain caveats if rewrite tries to fix this automatically but potential limitations/issues).

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
```diff
@@ -0,0 +1,112 @@
...
+<<< single line, triple quoted
+object A {
+  val a = <div b="Hello"/>
+  val b = <a>\</a>
+}
+>>>
+import scala.xml.quote._
+object A {
+  val a = xml"""<div b="Hello"/>"""
+  val b = xml"""<a>\</a>"""
```
olafurpg: How come this is triple quote while the test above is single quote? Is it the escape `\`?

build.sbt
olafurpg: Sounds good, can you please add a comment to explain this?

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
```diff
@@ -0,0 +1,112 @@
...
+}
+
+<<< nested xml literals
+object A {
+  <a>{<a>{"Hello"}</a>}</a>
+}
+>>>
+import scala.xml.quote._
+object A {
+  xml"<a>${xml"<a>${"Hello"}</a>"}</a>"
```
olafurpg: Impressive!

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    val patch = ctx.tree.collect {
+      case xml @ Xml(parts) =>
+        warnOnDoubledBrace(xml)
+        val tripleQuote = isMultiLine(xml) || parts.exists(
+          containsEscapeSequence)
+        patchXml(xml, tripleQuote)
+    }.asPatch
+
+    if (patch.nonEmpty) patch + AddGlobalImport(importer"scala.xml.quote._")
```
olafurpg: The way AddGlobalImport currently works is that it's a semantic rewrite, `AddGlobalImport` is `private[scalafix]` and it is unsafe to  use it directly like this. The correct way to build patches is with `ctx.addGlobalImport()`.

Your options are

-  make RemoveXmlLiterals semantic `case class RemoveXmlLit(mirror: Mirror) extends SemanticRewrite(mirror)`, this means
  - the rewrite needs build integration like sbt-scalafix
  - will be much slower
  - can't run directly from the cli 
- write a `ctx.addLeft(token, "import scala.xml.quote._\n")` patch that inserts the import somewhere at the top of the file, I'm open to bake this syntactic patch into the scalafix api since I believe it will be useful for many
- don't insert any import, I think this would be annoying for many users.

I propose option 2. WDYT?

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    def isMultiLine(xml: Tree) =
+      xml.pos.start.line != xml.pos.end.line
+
+    /** Contains '"' or '\' */
+    def containsEscapeSequence(xmlPart: Lit) = {
+      val Lit(value: String) = xmlPart
+      value.exists(c => c == '\"' || c == '\\')
```
allanrenucci: @olafurpg Am I missing characters that need to be triple quoted?

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
allanrenucci: yes

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
allanrenucci: Indeed, it is triple quoted in order to escape `\`

scalafix-tests/src/test/resources/checkSyntax/RemoveXmlLiterals.source
allanrenucci: This test is skipped. As I said previously:

The issue is that there are many places in xml literals where braces don't need to be escaped:
```scala
scala> <a b="{{"/>
res0: scala.xml.Elem = <a b="{{"/>

scala> <![CDATA[{{]]> 
res1: scala.xml.PCData = <![CDATA[{{]]>

scala> <xml:unparsed>{{</xml:unparsed>
res2: scala.xml.Unparsed = {{

scala> <?foo {{?>
res3: scala.xml.ProcInstr = <?foo {{?>
```
We would need to parse the xml literal to implement this rewrite

olafurpg: Lol, I totally missed that. However, does this mean that the rewrite is not semantically preserving? A previously rendered `{` will now render as `{{`? Which case do you think is more common, that

- `{{` is a legitimate escape that should be rendered as `{` OR
- `{{` should be rendered as `{{` in places like attribute quotes or CDATA?

It's possible to do the rewrite and emit a warning that user should double check if `{{` appeared in a weird location and rendered as `{{`.

allanrenucci: The first case is probably the most common. Doing the rewrite and emitting a warning seems to be the best solution.

scalafix-core/src/main/scala/scalafix/rewrite/RemoveXmlLiterals.scala
allanrenucci: If I understand, you are suggesting to keep it as a Syntactic Rewrite and use `ctx.addLeft(token, "import scala.xml.quote._\n")` instead of `ctx.addGlobalImport(...)`.
I tried 
```scala
def importXmlQuote = ctx.addLeft(ctx.tree.tokens.head, "import scala.xml.quote._\n")
```
But it doesn't work because it has to go after package declaration. Any suggestion?

olafurpg: Maybe we can extract this bit into something reusable https://github.com/scalacenter/scalafix/blob/b68ea04d29731f60c731da89316321fcbc883adc/scalafix-core/src/main/scala/scalafix/patch/OrganizeImports.scala#L68

allanrenucci: This will not help me if there is no import initially

olafurpg:
There is some code there that handles that case too, I can't recall exactly
where. We don't need a Seq[Import] for this, only a Token.

allanrenucci:
> Did you try to run this with the cli on https://github.com/lift/framework ?

I haven't tried yet. Will do

allanrenucci:
@olafurpg Ran the rewrite on https://github.com/lift/framework/pull/1849
I was able to compile and pass the tests.

olafurpg:
Amazing news @allanrenucci! So are we good to merge this? 😄

allanrenucci:
Once CI is green, yes!

olafurpg:
One build timed out :/ I restarted it


cc/ @jvican This is happening quite often, I'm not sure what we can do about it. Should we maybe try to upgrade  the drone installation?

olafurpg:
Congrats with this awesome rewrite @allanrenucci. 

I plan to release scalafix v0.4 next week Monday. I'm currently busy cleaning up the scalameta v1.8 milestone here https://github.com/scalameta/scalameta/milestone/2

PR #149, opened 2017-05-06
Add NoAutoTupling test cases involving classes and case classes

gabro:
Just being paranoid here.

### DIFF
--- a/scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
+++ b/scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
@@ -98,3 +98,25 @@ object tup {
   def foo: (((Int, String)) => ((String, List[Int])) => Int) = a => b => a._1
   foo((1 + 2, "foo"))(("bar", 1 :: 2 :: Nil))
 }
+
+<<< auto-tupling with class constructors
+object tup {
+  class Foo(t: (Int, String))(s: (Boolean, List[Int]))
+  new Foo(1, "foo")(true, Nil)
+}
+>>>
+object tup {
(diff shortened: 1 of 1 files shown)


gabro:
Hum, the build has failed for a timeout

olafurpg:
Hmm, weird, I restarted the build.

cc/ @jvican the build timed out while syncing the cache, any idea what could have caused it? Any idea if we could setup auto-restart in these cases?

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -98,3 +98,25 @@ object tup {
...
+
+<<< auto-tupling with class constructors
+object tup {
+  class Foo(t: (Int, String))(s: (Boolean, List[Int]))
+  new Foo(1, "foo")(true, Nil)
+}
+>>>
+object tup {
+  class Foo(t: (Int, String))(s: (Boolean, List[Int]))
+  new Foo((1, "foo"))((true, Nil))
```
olafurpg: Awesome! I'm impressed this works actually 😄

olafurpg:
cc/ @jvican the build timed out while syncing the cache, any idea what could have caused it? Any idea if it's possible to setup auto-restart in these cases?

jvican:
The drone documentation doesn't say anything about it, so I would bet it's not possible to auto restart it. I don't know what it may have caused it. Maybe flaky connection?

scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
gabro: So was I 😂

Issue #126, opened 2017-04-18
Add rewrite to remove unused imports


jvican:
I would like a Scalafix rewrite that removes unused imports. This rewrite is useful for unclutter the name scope of a Scala file where names are unused.
  
In an informal discussion with @olafurpg, he suggested me to parse compiler error messages to implement this functionality since the current scalahost compiler plugin is being deprecated. Olaf suggests that the API entrypoint is `.messages: Seq[CompilerMessage]`.
  
I will provide an implementation for this some time in the future, if someone doesn't beat me to it!

olafurpg:
This got fixed in https://github.com/scalacenter/scalafix/pull/166

PR #152, opened 2017-05-07
Add NoUnusedImports semantic rewrite

gabro:
This a stab at #126, following the idea of parsing compiler messages.

However, this strategy falls short when trying to fix selective imports.

Here's a failing test case

```scala
import scala.collection.mutable.{ Set, ListBuffer }
val l = ListBuffer.empty[Int]
```
The expected result would be

```scala
import scala.collection.mutable.{ ListBuffer }
val l = ListBuffer.empty[Int]
```

but it's actually

```scala
val l = ListBuffer.empty[Int]
```

which of course doesn't compile.

The reason is that the positions in the compiler message are not precise in case of unused members of a selective import and just point to the whole import statement, making it impossible to discriminate which member is used and which isn't.

/cc @jvican @olafurpg

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
@@ -1,0 +1,54 @@
+package scalafix
+package rewrite
+
+import scala.meta._
+import scala.annotation.tailrec
+
+case class NoUnusedImports(mirror: Mirror) extends SemanticRewrite(mirror) {
+
+  @tailrec
+  private[this] def removeTrailingWhitespacesOrSemi(
+      ctx: RewriteCtx,
+      t: Token,
+      patch: Patch = Patch.empty): Patch = {
+    val nextToken = ctx.tokenList.next(t)
+    if (nextToken.is[Token.LF] || nextToken.is[Token.Semicolon]) {
+      patch + ctx.removeToken(nextToken)
+    } else if (nextToken.is[Token.Space] || nextToken.is[Token.Tab]) {
+      removeTrailingWhitespacesOrSemi(ctx,
+                                      nextToken,
+                                      patch + ctx.removeToken(nextToken))
+    } else {
+      patch
+    }
+  }
+
+  @tailrec
+  private[this] def removeLeadingWhitespaces(
+      ctx: RewriteCtx,
+      t: Token,
+      patch: Patch = Patch.empty): Patch = {
+    val prevToken = ctx.tokenList.prev(t)
+    if (prevToken.is[Token.Space] || prevToken.is[Token.Tab]) {
+      removeLeadingWhitespaces(ctx,
+                               prevToken,
+                               patch + ctx.removeToken(prevToken))
+    } else {
+      patch
+    }
+  }
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    val patches = mirror.database.messages.collect {
+      case Message(Anchor(_, start, end), _, "Unused import") =>
+        val toRemove =
+          ctx.tokens.dropWhile(_.start < start).takeWhile(_.end <= end)
+
+        val removeImport = toRemove.map(ctx.removeToken).asPatch
+        removeImport +
+          removeLeadingWhitespaces(ctx, toRemove.head) +
+          removeTrailingWhitespacesOrSemi(ctx, toRemove.last)
+    }
+    patches.asPatch
+  }
+}
--- a/scalafix-tests/src/test/resources/checkSyntax/NoUnusedImports.source
+++ b/scalafix-tests/src/test/resources/checkSyntax/NoUnusedImports.source
@@ -1,0 +1,44 @@
+rewrites = [NoUnusedImports]
+<<< remove unused imports
+object imp {
+  import scala.collection.mutable.Set
+  import scala.collection.mutable.ListBuffer
+  import java.util._
+
+  val l = ListBuffer.empty[Int]
+}
+>>>
+object imp {
+  import scala.collection.mutable.ListBuffer
+
+  val l = ListBuffer.empty[Int]
+}
+
+<<< funky formatting
+object imp {
+  import scala.collection.mutable.Set    ; import scala.collection.mutable.ListBuffer
+  import scala.util.Success; import java.util.Date
+  import scala.util.Try
+
+  val d = new Date
+}
+>>>
+object imp {
+ import java.util.Date
+
+  val d = new Date
+}
+
+<<< selective import
+object imp {
+  import scala.collection.mutable.{ Set, ListBuffer }
+
+  val l = ListBuffer.empty[Int]
+}
+>>>
+object imp {
+  import scala.collection.mutable.{ ListBuffer }
+
+  val l = ListBuffer.empty[Int]
+}
+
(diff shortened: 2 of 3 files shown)


jvican:
Thanks for taking a shot at this, I'm really looking forward to having this rewrite!
  
If we want to do this in a principled way, we will probably need to modify scalac to report better positions for these cases. I can imagine several workarounds to get this precise information by hijacking global, but they are less than ideal and I would prefer to stay away from them.
  
Another thing that we should be aware of is that unused imports can be different accross scala versions (because of implicit conversions or classes). A perfect approach would remove the unused imports by compiling first with all the cross scala versions and taking the intersection of all the unused names. (To the best of my knowledge, scalafix doesn't accommodate this use case for now.)
  
However, this means that even if we get better position information in 2.12.x, we can never reliably report on unused imports and remove them without the risk of making the project not compile because we lack that precise information in the 2.11.x series.
  
There is a tradeoff here and I don't know what the general goals of the official rewrites are. But it should be well documented that the use of this feature can cause compilation failures and, therefore, should be applied in isolation of the other rewrites.

gabro:
> If we want to do this in a principled way, we will probably need to modify scalac to report better positions for these cases

Apparently (at least in Scala 2.12) this is already happening. Notice the positions of the `^` in the warning messages below. Could it be that the reporter in scalameta is not picking those up?

```
❯ scala -Ywarn-unused-import
Welcome to Scala 2.12.0 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_112).
Type in expressions for evaluation. Or try :help.

scala> import scala.util.{ Try, Success, Failure }
<console>:11: warning: Unused import
       import scala.util.{ Try, Success, Failure }
                           ^
<console>:11: warning: Unused import
       import scala.util.{ Try, Success, Failure }
                                ^
<console>:11: warning: Unused import
       import scala.util.{ Try, Success, Failure }
                                         ^
import scala.util.{Try, Success, Failure}
```

As per the cross-version validation it's a very good point. I don't know where it fits within the scope of scalafix's goals.

gabro:
I tried on scala 2.11.8 and the REPL is capable of printing those `^` at the right position as well (same output), so there may be something missing in the scalameta message.

xeno-by:
@gabro Looks like a bug in meta. Can you submit a repro and a faulty semantic db to https://github.com/scalameta/scalameta/issues/new?

gabro:
sure, I'll do it as soon as I get back to a computer

scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
```diff
@@ -0,0 +1,54 @@
...
+                               patch + ctx.removeToken(prevToken))
+    } else {
+      patch
+    }
+  }
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    val patches = mirror.database.messages.collect {
+      case Message(Anchor(_, start, end), _, "Unused import") =>
+        val toRemove =
```
olafurpg: I think we will need a more robust way to remove importees, for example if `c` is unused in

```scala
import a.{b, c}
```

then we might want to emit

```scala
import a.b
```

scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
```diff
@@ -0,0 +1,54 @@
+package scalafix
+package rewrite
+
+import scala.meta._
+import scala.annotation.tailrec
+
+case class NoUnusedImports(mirror: Mirror) extends SemanticRewrite(mirror) {
```
olafurpg: One challenge with imports is that I suspect many rewrites will want to modify them so there's a risk of patch conflicts. My eventual plan is to take a holistic approach towards imports management and bake it into the Patch system. Instead of eagerly removing/adding import tokens inside rewrites, we can collect high-level patches (e.g., add/remove import, rename) inside rewrites and evaluate them to token patches as late as possible. This might make it easier to compose different rewrites.

My dream is that rewrite authors won't have to worry too much about low-level details of imports (because they're kinda complex in Scala) and scalafix users can configure how they like to organize their imports https://github.com/scalacenter/scalafix/blob/996f30ff87ad9442c53a06ccf505ac6e845b9672/scalafix-core/src/main/scala/scalafix/config/ImportsConfig.scala

However, to begin with, I think there is value in having a custom rewrite only to remove unused imports. My experience writing an OrganizeImports step was that it's quite tricky to get right, so it may take a few iterations until we get there and in the meantime a RemoveUnusedImports rewrite might turn out to be useful enough.

scalafix-tests/src/test/resources/checkSyntax/NoUnusedImports.source
```diff
@@ -0,0 +1,44 @@
+rewrites = [NoUnusedImports]
+<<< remove unused imports
```
olafurpg: There is a test suite here https://github.com/scalacenter/scalafix/blob/996f30ff87ad9442c53a06ccf505ac6e845b9672/scalafix-tests/src/test/resources/checkSyntax/OrganizeImportsRemoveUnused.source which might be relevant to take a look at. It contains a few tricky cases like macro/language imports which the compiler message approach should handle.

scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
```diff
@@ -0,0 +1,54 @@
...
+                               prevToken,
+                               patch + ctx.removeToken(prevToken))
+    } else {
+      patch
+    }
+  }
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    val patches = mirror.database.messages.collect {
+      case Message(Anchor(_, start, end), _, "Unused import") =>
```
olafurpg: Heads up, Anchor is going away in favor of Position.Range https://github.com/scalameta/scalameta/pull/828

scalafix-core/src/main/scala/scalafix/rewrite/NoUnusedImports.scala
```diff
@@ -0,0 +1,54 @@
...
+      removeLeadingWhitespaces(ctx,
+                               prevToken,
+                               patch + ctx.removeToken(prevToken))
+    } else {
+      patch
+    }
+  }
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    val patches = mirror.database.messages.collect {
```
olafurpg: A Mirror can span across multiple source files, so iterating through them here and filtering tokens by start/end offset would trigger false positives. Once Anchor is replaced with Position.Range, it should be easier to filter the by the position's `.input`.

gabro:
Thanks for the review @olafurpg! I'll get back to this as soon as the relevant issues are closed on scalameta.

Speaking of which, @xeno-by, I've just opened https://github.com/scalameta/scalameta/issues/839 ;)

olafurpg:
I just merged https://github.com/scalacenter/scalafix/pull/157 that upgrades to the latest scalameta which includes accurate positions for unused import warnings. The positions should now match `Importee.Rename/Name.pos`. I think the position of `Importee.Wildcard` may still point to the enclosing `Import.pos`.

olafurpg:
Heads up, I'm gonna try to take a swing at this implementing this rewrite tomorrow. I'm eager to use it for my own projects 😄

gabro:
🎉 great news!

olafurpg:
Superseded by #166

PR #155, opened 2017-05-10
Correctly patch empty token

allanrenucci:


scalafix-core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -71,6 +71,14 @@ package object syntax {
 
   implicit class XtensionToken(token: Token) {
     def posTuple: (Int, Int) = token.start -> token.end
+    // copy pasted from:
+    // github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
+    def hash: Long = {
```
allanrenucci: Maybe you don't want to expose this. I can inline it at the use site.

scalafix-core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -71,6 +71,14 @@ package object syntax {
 
   implicit class XtensionToken(token: Token) {
     def posTuple: (Int, Int) = token.start -> token.end
```
allanrenucci: This is no longer used in the project

allanrenucci:
Let me know if I put the tests at the right place and if the naming is OK.

scalafix-core/src/main/scala/scalafix/syntax/package.scala
olafurpg: We can remove this.

scalafix-core/src/main/scala/scalafix/syntax/package.scala
olafurpg: Can we make this `private [scalafix]`?

Which reminds me I should setup MiMa to keep track of the public api.

scalafix-tests/src/main/scala/scalafix/test/FqnRewrite.scala
```diff
@@ -26,3 +26,12 @@ object LambdaRewrites {
   }
 
 }
+
+case object FqnRewrite3 extends Rewrite {
```
olafurpg: Can we also give this rewrite a more descriptive name? For example, `PatchTokenWithEmptyRange`. The `FqnRewrite` names are for testing the classloading feature.

scalafix-core/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -70,7 +70,14 @@ package object syntax {
   }
 
   implicit class XtensionToken(token: Token) {
-    def posTuple: (Int, Int) = token.start -> token.end
+    // copy pasted from:
+    // github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```
olafurpg: I just noticed this links to master branch, can you please canonicalize it?

allanrenucci: What do you mean?
scalameta/scalafmt/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
Like this?

olafurpg: If you press "y" while browsing a file on github.com, the url will expand into something like this 

https://github.com/scalameta/scalafmt/blob/ba319e0aaee6d4e1b30098ad57940c67910c33e9/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala

this url will still work even if the file is edited/moved/removed

allanrenucci: Done

olafurpg:
Thank you for fixing this!

PR #156, opened 2017-05-14
Fix obsolete reference to `fatcli` in the docs

mlangc:


readme/Installation.scalatex
```diff
@@ -81,7 +81,7 @@
       ./scalafix --help
 
     @p
```
olafurpg: We can remove this paragraph altogether, the --scalahost-nsc-plugin-path flag is no longer necessary thanks to the fact that the scalameta semantic api no longer depends on scala-compiler or scalahost-nsc.

mlangc: I'll push that shortly

mlangc:
I find no more references of `fatcli` now...

PR #157, opened 2017-05-14
Upgrade to latest scalameta pre-release.

olafurpg:
It seems https://github.com/scalameta/scalameta/pull/847 broke a few
rewrites. However, I still prefer the new behavior since I belive
synthetic .apply and should belong to aliases and that the rewrites that
broke with this change were only conincidentally working.
Once https://github.com/scalameta/scalameta/issues/846 is fixed then we
should be able to go back to the old behavior, if we so chose.


scalafix-tests/src/test/resources/checkSyntax/NoAutoTupling.source
```diff
@@ -101,22 +101,15 @@ object tup {
...
-  class Foo(t: (Int, String))(s: (Boolean, List[Int]))
-  new Foo((1, "foo"))((true, Nil))
-}
-
-<<< auto-tupling with case class constructors
-object tup {
   case class Foo(t: (Int, String))(s: (Boolean, List[Int]))
-  Foo(1, "foo")(true, Nil)
+  new Foo(1, "foo")(true, Nil)
+  Foo(1, "foo")(true, Nil) // blocked by https://github.com/scalameta/scalameta/issues/846
```
gabro: good thing we added those extra tests then, I'll try to fix them asap!
Thanks for the heads up :)

olafurpg: I'm not sure if it's possible to fix that test atm because of https://github.com/scalameta/scalameta/issues/846

gabro: aw, right. It's probably better to come back to it as soon as the missing sugars are there.

olafurpg: It may take awhile since we may be tempted to address https://github.com/scalameta/scalameta/issues/822 first. Currently, sugars are plain strings with no fany info like symbols. It's totally doable, just need to figure a way to position the sugar symbols.

gabro: Yep, I've see that issue. Structured info would be lovely and definitely worth waiting.

PR #159, opened 2017-05-16
Add NoValInForComprehension syntactic rewrite

### LINKED ISSUE #153: Remove val in for comprehension
```scala

    for {
      srcFile <- javaSources
      val contents = Source.fromFile(srcFile).mkString
      if isAnnotation(contents)
    } yield {



[warn] /Users/ollie/dev/scalatest/project/JavaTagDocumenter.scala:142: val keyword in for comprehension is deprecated
[warn]       val contents = Source.fromFile(srcFile).mkString
[warn]                    ^
```

It would be nice to have a scalafix rewrite that does this diff

```diff
-      val contents = Source.fromFile(srcFile).mkString
+      contents = Source.fromFile(srcFile).mkString
```

gabro:
Closes #153 

I couldn't come up with many creative tests for this, but I'd be glad to add more edge cases if anyone can think of any.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoValInForComprehension.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoValInForComprehension.scala
@@ -1,0 +1,18 @@
+package scalafix
+package rewrite
+
+import scala.meta._
+import scala.meta.contrib._
+
+case object NoValInForComprehension extends Rewrite {
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    ctx.tree.collect {
+      case v: Enumerator.Val =>
+        val valTokens =
+          v.tokens.takeWhile(t => t.syntax == "val" || t.is[Whitespace])
+        valTokens.map(ctx.removeToken).asPatch
+    }.asPatch
+  }
+
+}
--- a/scalafix-tests/src/test/resources/checkSyntax/NoValInForComprehension.source
+++ b/scalafix-tests/src/test/resources/checkSyntax/NoValInForComprehension.source
@@ -1,0 +1,42 @@
+rewrites = [NoValInForComprehension]
+<<< no val in for comprehension
+object noval {
+  for {
+    n <- List(1, 2, 3)
+    val inc = n + 1
+    inc2 = inc + 1
+    val inc3 = inc2 + 1
+    if inc > 1
+  } yield inc2
+}
+>>>
+object noval {
+  for {
+    n <- List(1, 2, 3)
+    inc = n + 1
+    inc2 = inc + 1
+    inc3 = inc2 + 1
+    if inc > 1
+  } yield inc2
+}
+
+<<< no val in for comprehension without yield
+object noval {
+  for {
+    n <- List(1, 2, 3)
+    val inc = n + 1
+    inc2 = inc + 1
+    val inc3 = inc2 + 1
+    if inc > 1
+  } ()
+}
+>>>
+object noval {
+  for {
+    n <- List(1, 2, 3)
+    inc = n + 1
+    inc2 = inc + 1
+    inc3 = inc2 + 1
+    if inc > 1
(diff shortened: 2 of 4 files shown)


olafurpg:
Thank you for this contribution @gabro! You are on fire :) 

I will be away from the laptop for a few days. I will try to review as soon as I get back.

scalafix-core/src/main/scala/scalafix/rewrite/NoValInForComprehension.scala
```diff
@@ -0,0 +1,21 @@
...
+package rewrite
+
+import scala.meta._
+import scala.meta.contrib._
+
+case object NoValInForComprehension extends Rewrite {
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    ctx.tree.collect {
+      case f: Term.ForYield =>
```
olafurpg: I believe this rewrite should also work for `Term.For`. We can make an extractor

```scala
  object TermFor {
    def unapply(tree: Tree): Option[Seq[Enumerator]] = tree match {
      case Term.For(enums, _) => Some(enums)
      case Term.ForYield(enums, _) => Some(enums)
      case _ => None
    }
  }
```

olafurpg: Wait, I don't think we need to match against `Term.For` in the first place. I think it's enough to match directly against `Enumerator.Val`

scalafix-tests/src/test/resources/checkSyntax/NoValInForComprehension.source
```diff
@@ -0,0 +1,21 @@
...
+}
+>>>
+object noval {
+  for {
+    n <- List(1, 2, 3)
+    inc = n + 1
+    inc2 = inc + 1
+    inc3 = inc2 + 1
+    if inc > 1
+  } yield inc2
```
olafurpg: We should probably add another test for Term.For without `yield`.

scalafix-core/src/main/scala/scalafix/rewrite/NoValInForComprehension.scala
gabro: Right, I actually though of it, but I couldn't remember whether a `for` without `yield` were valid scala syntax.
Matching `Enumerator.Val` could actually do the trick!

gabro: Oh, ok, I'm ufficially confused by the grammar.

I tried

```scala
(for { x <- List(1, 2, 3) })
```

and the parser complains. By looking at the grammar:


```
`for' (`(' Enumerators `)' | `{' Enumerators `}') {nl} [`yield'] Expr
```

`Expr` is mandatory, so something like


```scala
(for { x <- List(1, 2, 3) } 42)
```

works instead (the type of the expression is `Unit`). I'm wondering why `Expr` is needed without a `yield`...

gabro:
Ok @olafurpg I've simplified the `collect` logic (matching only against `Enumerator.Val` is enough, indeed, nice catch!) and added a test against a `for` without `yield`.

scalafix-core/src/main/scala/scalafix/rewrite/NoValInForComprehension.scala
```diff
@@ -0,0 +1,18 @@
...
+package rewrite
+
+import scala.meta._
+import scala.meta.contrib._
+
+case object NoValInForComprehension extends Rewrite {
+
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    ctx.tree.collect {
+      case v: Enumerator.Val =>
```
olafurpg: Nice! So small 😄

gabro: small is beautiful 💃

PR #163, opened 2017-05-24
Upgrade

olafurpg:
- significantly refactors scalafix-cli to work with latest fat semanticdb files and greater error accumulation
- adds ExplicitImplicit back 👏 Tested on slick and it works like a charm, besides the fact that inserted types are fully qualified. I believe we can improve on that in future PRs.
- Cleans up sbt-scalahost to depend on sbt-scalahost instead of copy-pasting ScalahostPlugin.


olafurpg:
It seems bintray is still experiencing problems despite their status page saying all is OK. I had to restart the build to get it green due to resolution errors on sbt artifacts. We've had similar problems in scalameta https://platform-ci.scala-lang.org/scalameta/scalameta/323/3 causing the docs to skip publish. Maybe our IP address is just getting rate limited, I'll have to investigate.

Either way, the tests pass and this is the last important bit before the v0.4 release 🎉

PR #165, opened 2017-05-25
Clean up tests

olafurpg:
The scalafix repo has accumulated a lot of successful and not-so-successful experiments over the past year. This PR cleans up what I found to be not-so-successful around our testing infrastructure an imports management:

- unit tests in .source files. When I started scalafix is was only syntactic like scalafmt so I ported the test infrastructure from scalafmt. Now that scalafix does semantic rewrites, I feel that the .source files have a lot of limitation: no IntelliJ support, no syntax highlighting, bad compiler error message, dependency on scala-compiler, fragile interface with ScalahostPipeline and other compiler APIs. Basically, the test suite has become a half-baked built tool. This PR removes all of that in favor of using sbt for compiling/building the Semantic API, see 9beff13
- OrganizeImports that was trying to do too much. OrganizeImports was implemented before Symbols appeared and therefore had a lot of inherent design issues. I ported the implementation to something less ambitious: no sorting, no expanding relative imports, no grouping, no ImportsConfig. This current implementation is far from perfect, but it won't mess up with people's imports as much.


olafurpg:
cc/ @ShaneDelmore scalafix-testkit just got a whole lot simpler, since it no longer uses `.source` files. Files are written in regular .scala files instead with full IntelliJ support. This change also buys us sbt-zinc incremental compilation for free 😄 The only downside is that code for original/expected pairs are no longer next to each other in the same file. However,  the diffs for failed tests are still just as pretty. I believe the benefits outweigh the downsides.

scalafix-core/src/main/scala/scalafix/package.scala
```diff
@@ -26,8 +26,8 @@ package object scalafix {
       extends SyntacticPatchOps(ctx) {
     def semanticOps: SyntacticPatchOps = this
   }
-  implicit class XtensionSeqPatch(patches: Seq[Patch]) {
-    def asPatch: Patch = Patch.fromSeq(patches)
+  implicit class XtensionSeqPatch(patches: Iterable[Patch]) {
+    def asPatch: Patch = Patch.fromIterable(patches)
```
xeno-by: What's the motivation behind `Iterable`? I wonder what use cases you found for this more general type.

scalafix-tests/input/src/main/scala/test/ExplicitReturnTypes.scala
```diff
@@ -0,0 +1,69 @@
+/*
+rewrites = ExplicitReturnTypes
+ */
```
xeno-by: 👍

scalafix-core/src/main/scala/scalafix/package.scala
olafurpg: iirc, I'm using views on `Tokens` to build patches. Views are not Seq[]

PR #178, opened 2017-05-31
fix missing double quotations in scalafix-testkit doc

taisukeoe:
Currently the doc says:

```
addSbtPlugin("org.scalameta" % "sbt-scalahost" % 1.8.0)
```

This misses double quotations in version.

At open source spree at Copenhagen, I find this :)


olafurpg:
Docs don't update automatically at the moment, I will update later!

Issue #177, opened 2017-05-31
All files indexed in the semantic db are implicitly fixed


stuhood:
It's not completely clear that this is a bug, but currently all files in the classpath provided to the CLI tool are fixed. This means that if you don't intend to fix the entire graph of some module (including its dependencies), you have to pass exactly one classpath entry: the one corresponding to the sources.

Again, it's possible that this is intended. But the fact that all files in all introspected databases are fixed sortof conflicts with the fact that you can pass explicit file lists to fix. IMO, only fixing explicitly passed files might make more sense.

PR #183, opened 2017-06-01
Fix #176 and #177

### LINKED ISSUE #176: sourceroot not used by cli tool
The CLI tool does not use the `sourceroot` value that is passed in, and the result is that the files that are operated on are always the ones relative to `CWD`.

The `Input` values that come from the scalameta database will be `Input.LabeledString` (afaict), with relative paths names as the labels. scalafix will need to convert these to absolute paths using the supplied `sourceroot`.

olafurpg:


scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
```diff
@@ -77,9 +82,11 @@ case class CliRunner(
     display.stop()
     exitCode.get()
   }
+  private val sourceroot: AbsolutePath =
+    database.map(_.sourceroot).getOrElse(common.workingPath)
 
   def safeHandleInput(input: Input): ExitStatus = {
-    def path = input.path(common.workingPath)
+    def path = input.path(sourceroot)
```
olafurpg: Review @stuhood This line makes the relative filename in the Mirror absolute according to --sourceroot.

scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
```diff
@@ -220,11 +237,25 @@ object CliRunner {
...
+    }
+    val resolvedInputs: Configured[Seq[Input]] = for {
+      fromMirror <- mirrorPaths
+      pathMatcher <- resolvedPathMatcher
+      sourceroot <- resolvedSourceroot
+    } yield {
+      def inputOK(input: Input) = pathMatcher.matches(input.path(sourceroot))
+      val builder = Seq.newBuilder[Input]
+      fromMirror.withFilter(inputOK).foreach(builder += _)
+      explicitPaths.withFilter(inputOK).foreach(builder += _)
```
stuhood: It's still a bit awkward that there is both a list of explicit files, and a list of filters... but probably ok, because they'll be used in different contexts.

scalafix-core/src/main/scala/scalafix/config/FilterMatcher.scala
```diff
@@ -11,6 +11,7 @@ case class FilterMatcher(
     @ExtraName("include") includeFilters: Regex,
     @ExtraName("exclude") excludeFilters: Regex
 ) {
+  def matches(path: AbsolutePath): Boolean = matches(path.toString())
```
stuhood: Should this match absolute paths, or only the portion relative to the sourceroot?

scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
olafurpg: It should be possible to support a flag to not implicitly add files from the mirror. That might be preferable for build tools that programmatically invoke the cli. We can move that discussion to https://github.com/scalacenter/scalafix/issues/187

scalafix-core/src/main/scala/scalafix/config/FilterMatcher.scala
olafurpg: Good question. This matches against the absolute path but I think --include/--exclude should match against the relative path. I updated the PR.

PR #188, opened 2017-06-02
Add NoExtendsApp rewrite

### LINKED ISSUE #174: Add main function using DelayedInit (extends App)
See https://github.com/lampepfl/dotty/blob/836ec499de1d8f52de6786f6090aeb52dd128218/docs/docs/reference/dropped/delayedInit.md and https://github.com/lampepfl/dotty/issues/559

```scala
// before
object DelayedInitToMain extends App {
  println(args)
}
// after
object DelayedInitToMain {
  def main(args: Array[String]): Unit = {
    println(args)
  }
}
```

The rewrite should only trigger if `args` is accessed.

Note https://github.com/scalameta/scalameta/issues/905

gabro:
Closes #174 

image

### DIFF
--- a/readme/Rewrites.scalatex
+++ b/readme/Rewrites.scalatex
@@ -197,5 +197,28 @@
     @p
       The two syntaxes are equivalent and the presence of the @code{val} keyword has been deprecated since Scala 2.10.
 
+  @sect(NoExtendsApp.toString)
+    @p
+      Removes use of the @code{scala.App} trait.
+
+    @hl.scala
+      // before
+      object Main extends App {
+        println(s"Hello ${args(0)}")
+      }
+
+      // after
+      object Main {
+        def main(args: Array[String]) = {
+          println(s"Hello ${args(0)}")
+        }
+      }
+
+    @p
+      The @code{App} trait makes use of the @code{DelayedInit} feature which has been dropped in Dotty. While it will still be possible to use @code{App} for a concise setup of a main program, the semantics have slighly changed from Scala 2.x and it won't be possible to access the command line arguments without an explicit @code{main}.
+    @p For these reasons, this rewrite is conservative and it always drops the use of @code{App}, introducing a @code{main} method.
+
+    @p More details can be found @lnk("here", "https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/delayed-init.md").
+
   @sect{Planned rewrites...}
     See @lnk("here", "https://github.com/scalacenter/scalafix/labels/rewrite").
--- a/readme/Rewrites.scalatex
+++ b/readme/Rewrites.scalatex
@@ -199,7 +199,7 @@
 
   @sect(NoExtendsApp.toString)
     @p
-      Removes use of the @code{scala.App} trait.
+      Replaces usage of the @code{scala.App} trait with an explicit main function.
 
     @hl.scala
       // before
@@ -215,10 +215,8 @@
       }
 
     @p
-      The @code{App} trait makes use of the @code{DelayedInit} feature which has been dropped in Dotty. While it will still be possible to use @code{App} for a concise setup of a main program, the semantics have slighly changed from Scala 2.x and it won't be possible to access the command line arguments without an explicit @code{main}.
-    @p For these reasons, this rewrite is conservative and it always drops the use of @code{App}, introducing a @code{main} method.
-
-    @p More details can be found @lnk("here", "https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/delayed-init.md").
+      The @code{scala.App} trait uses @code{DelayedInit}, which has been dropped in Dotty.
+      More info @lnk("here", "https://github.com/lampepfl/dotty/blob/master/docs/docs/reference/dropped/delayed-init.md").
 
   @sect{Planned rewrites...}
     See @lnk("here", "https://github.com/scalacenter/scalafix/labels/rewrite").
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
@@ -1,7 +1,7 @@
 package scalafix
 package rewrite
 
-import scala.meta.{Symbol => _, _}
+import scala.meta._
 import scalafix.syntax._
 import scalafix.util.{Whitespace => _, _}
 import scala.meta.contrib._
@@ -27,7 +27,7 @@ case class NoExtendsApp(mirror: Mirror) extends SemanticRewrite(mirror) {
 
     ctx.tree.collect {
       case t: Defn.Object =>
-        ctx.removeParentFromTemplate("_root_.scala.App.", t.templ) +
+        ctx.removeParentFromTemplate(Symbol("_root_.scala.App."), t.templ) +
           wrapBodyInMain(t.templ)
     }.asPatch
   }
@@ -77,12 +77,12 @@ object NoExtendsAppSyntax {
         .asPatch
     }
 
-    def removeParentFromTemplate(normalized: String, template: Template)(
+    def removeParentFromTemplate(normalized: Symbol, template: Template)(
         implicit m: Mirror): Patch = {
       val maybePatch = for {
         treeToRemove <- template.parents
           .collect { case c: Ctor.Ref => c }
-          .find(_.symbolOpt.map(_.normalized.syntax) == Some(normalized))
+          .find(_.symbolOpt.map(_.normalized) == Some(normalized))
         nameToRemove <- treeToRemove.tokens.headOption
         leadingExtendsOrWithToken <- ctx.tokenList
           .leading(nameToRemove)
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
@@ -18,8 +18,12 @@ case class NoExtendsApp(mirror: Mirror) extends SemanticRewrite(mirror) {
         .map { body =>
           val open =
             ctx.addLeft(body.head, s"\n  def main(args: Array[String]) = {")
-          val close = ctx.addRight(body.last, s"  }\n")
-          open + ctx.indent(body) + close
+          val indentBody = ctx.indent(body)
+          // this handles bodies on a single line like:
+          //   object Main extends App { println(args(0)) }
+          val closingSpaces = if (indentBody.isDefined) " " * 2 else ""
+          val close = ctx.addRight(body.last, s"$closingSpaces}\n")
+          open + indentBody.asPatch + close
         }
         .asPatch
 
@@ -47,13 +51,14 @@ object NoExtendsAppSyntax {
       maybeTokens.filterNot(_.isEmpty)
     }
 
-    def indent(tokens: Tokens, numberOfSpaces: Int = 2): Patch = {
-      val lastNewLine = tokens.dropRightWhile(t => !t.is[Newline]).last
-      tokens.collect {
-        case nl @ Newline() if nl != lastNewLine =>
-          ctx.addRight(nl, " " * numberOfSpaces)
-      }.asPatch
-    }
+    def indent(tokens: Tokens, numberOfSpaces: Int = 2): Option[Patch] =
+      tokens.dropRightWhile(t => !t.is[Newline]).lastOption.map {
+        lastNewLine =>
+          tokens.collect {
+            case nl @ Newline() if nl != lastNewLine =>
+              ctx.addRight(nl, " " * numberOfSpaces)
+          }.asPatch
+      }
 
     private[scalafix] def removeTokensBetween(
         from: Token,
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
@@ -57,7 +57,7 @@ object NoExtendsAppSyntax {
       }.asPatch
     }
 
-    def removeTokensBetween(from: Token,
+    private[scalafix] def removeTokensBetween(from: Token,
                             to: Token,
                             removeLeadingWhitespace: Boolean = true): Patch = {
       val toRemove = ctx.tokenList
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
@@ -1,0 +1,124 @@
+package scalafix
+package rewrite
+
+import scala.meta.{Symbol => _, _}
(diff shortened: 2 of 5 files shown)


olafurpg:
Thank you for taking on this rewrite! I will try to review after the weekend :)

gabro:
thanks! In the meanwhile, I'll try to abstract over the template parents removal. That should really be a higher lever API in the same spirit of ImportPatchOps

xeno-by:
We're going to change the ASTs around parents very soon: https://github.com/scalameta/scalameta/pull/907. You may want to hold off developing this particular helper :)

gabro:
thanks for the heads up! I'll wait then :)
I must admit Template was in general a bit unwieldy to work with, so thanks for the work in making Trees "friendlier"

gabro:
@olafurpg I've pushed a working version with more test cases. I've extracted most of the utilities and added them to `ctx`. As of now I've left them in the same file, you may want to take a look and see whether it's reasonable to extract them.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,115 @@
...
+            .takeRightWhile(t => !t.is[Newline])
+      }
+    }
+
+    def removeTokensBetween(from: Token, to: Token): Patch =
+      ctx.tokenList
+        .slice(
+          ctx.tokenList.prev(from),
+          ctx.tokenList
+            .next(ctx.tokenList.next(to)) // apply next twice to include the trailing space
```
gabro: probably not the best idea...

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,115 @@
...
+        )
+        .map(ctx.removeToken)
+        .asPatch
+
+    // FIXME(gabro): this is an **extremely** naive way of approaching this
+    def removeParentFromTemplate(normalized: String, template: Template)(
```
gabro: this essentially looks for the name to remove (like `App`) then check the previous keyword (`extends` or `with`). Three cases then arise:

1) `with App` => easy, just remove it
2) `extends App` => if not followed by other (i.e.` parents.length == 1`), just remove it
3) `extends App with Something` => remove `App with ` (naive but it seems to be working)

gabro: not sure how solid this is, if anyone can come up with trickier test cases, it'd be nice.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,124 @@
...
+          ctx.tokenList.leading(from).takeWhile(_.is[Whitespace])
+        else
+          Nil
+
+      (toRemove ++ trailingToRemove)
+        .map(ctx.removeToken)
+        .asPatch
+    }
+
+    def removeParentFromTemplate(normalized: String, template: Template)(
```
gabro: this essentially looks for the name to remove (like `App`) then check the previous keyword (`extends` or `with`). Three cases then arise:

1) `with App` => easy, just remove it
2) `extends App` => if not followed by other (i.e.` parents.length == 1`), just remove it
3) `extends App with Something` => remove `App with ` (naive but it seems to be working)

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,124 @@
...
+    ctx.tree.collect {
+      case t: Defn.Object =>
+        ctx.removeParentFromTemplate("_root_.scala.App.", t.templ) +
+          wrapBodyInMain(t.templ)
+    }.asPatch
+  }
+}
+
+object NoExtendsAppSyntax {
+  implicit class XtensionRewriteCtx(ctx: RewriteCtx) {
```
olafurpg: We need a consistent story for these utilities. I'm actually thinking if we should mix them into RewriteCtx to avoid the whole extension method business. What do you think?

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def templateBodyTokens(template: Template): Option[Tokens] = {
```
gabro: yeah that's I what I thought too after seeing the changes in that PR. It would just make the implementation simpler, but nothing hard to deal with.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def indent(tokens: Tokens, indentation: Tokens): Patch = {
+      val lastNewLine = tokens.dropRightWhile(t => !t.is[Newline]).last
+      tokens.collect {
+        case nl @ Newline() if nl != lastNewLine =>
+          ctx.addRight(nl, indentation.syntax)
+      }.asPatch
+    }
+
+    def indentationTokens(tokens: Tokens): Tokens = {
```
gabro: Assuming indentation of 2 spaces seems a bit optimistic, but you've definitely seen more codebases than me ;)

scalafix-tests/input/src/main/scala/test/NoExtendsApp.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object Main7
+  extends Something
+  with App
+
+object Main8
+  extends Something with
+  App
+
+object Main9 extends
```
gabro: good point, I'll add a test.

scalafix-tests/input/src/main/scala/test/NoExtendsApp.scala
```diff
@@ -0,0 +1,46 @@
+/*
+rewrites = [NoExtendsApp]
```
gabro: right, I forgot about that change

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
gabro: I was thinking the same. The method extension looks unnecessarily complex and get just get away with traits.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
olafurpg: Those who are not using 2 space indent should at least be using a formatter. Scalafmt does not have an option to configure indent spaces because 2 space indent is pretty much universal.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
olafurpg: Great. Let's keep everything in this file then since moving things around will be a binary breaking change, I can do the migration for 0.5 https://github.com/scalacenter/scalafix/issues/193

scalafix-tests/input/src/main/scala/test/NoExtendsApp.scala
gabro: it fails miserably of course 😈 

I'll be working on it

olafurpg: Hehe, token level rewrites are hard. I wish I could come up with a more intuitive or higher level API for common tasks

gabro: done! Definitely easier than expected!

gabro:
@olafurpg all set, ready for another review :)

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
```diff
@@ -0,0 +1,51 @@
...
+object Main7
+  extends Something
+
+object Main8
+  extends Something
+
+object Main9 extends
+  Something
+
+import scala.{ App => Ppa }
```
gabro: minor thing, in these (hopefully rare) cases, this import becomes unused. I deemed fixing this in a single pass too daunting. Users may have to manually handle this or run the `NoUnusedImports` rewrite right after.

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
```diff
@@ -0,0 +1,51 @@
...
+}
+
+object Main3 extends Something{
+  def main(args: Array[String]) = {
+    println(s"Hello, ${args(0)}")
+    println(s"Hello 2, ${args(0)}")
+  }
+}
+
+object Main4 extends Something
```
olafurpg: Is this expected? In these cases, I'm not sure what we should do honestly. Removing App won't fix the problem however, what do you think about emitting a warning instead?

> Missing body for object that extends `scala.App`. See http://dotty.epfl.ch/docs/reference/dropped/delayed-init.html for possible workarounds.

It's possible to report warnings with `ctx.reporter.warn(pos, msg)`. There's currently no way to test reported messages.

readme/Rewrites.scalatex
```diff
@@ -197,5 +197,28 @@
...
+
+      // after
+      object Main {
+        def main(args: Array[String]) = {
+          println(s"Hello ${args(0)}")
+        }
+      }
+
+    @p
+      The @code{App} trait makes use of the @code{DelayedInit} feature which has been dropped in Dotty. While it will still be possible to use @code{App} for a concise setup of a main program, the semantics have slighly changed from Scala 2.x and it won't be possible to access the command line arguments without an explicit @code{main}.
```
gabro: I agree. The repetition is just not worth it.

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
gabro: Right, I think adding a warning is our only option.
Honestly I don't know how common it is, but yes, let's warn the users.

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
```diff
@@ -0,0 +1,51 @@
...
+import scala.{ App => Ppa }
+object Main10 extends Something {
+  def main(args: Array[String]) = {
+    println(s"Hello, ${args(0)}")
+    println(s"Hello 2, ${args(0)}")
+  }
+}
+
+object Main11 {
+  def main(args: Array[String]) = { println(args(0)) }
```
gabro: My expected output was way worse :D (I'm starting to like tokens...)

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -0,0 +1,114 @@
...
+import NoExtendsAppSyntax._
+
+case class NoExtendsApp(mirror: Mirror) extends SemanticRewrite(mirror) {
+  override def rewrite(ctx: RewriteCtx): Patch = {
+    def wrapBodyInMain(template: Template) =
+      ctx
+        .templateBodyTokens(template)
+        .map { body =>
+          val open =
+            ctx.addLeft(body.head, s"\n  def main(args: Array[String]) = {")
```
gabro: I thought about this, but I took the example verbatim from the dotty notes. Does `main` always return `Unit`?

olafurpg: I see, let's still add `: Unit =`. Although that honestly just a personal preference of mine. Unit returning methods should scream "I'm doing stuff here!"

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
olafurpg: Tokens make some things really easy and other things really hard

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
gabro: Same preference. I was just concerned that `main` may return a termination value, but apparently can only be done with `sys.exit`. Moreover adding `Unit` would match the example here: https://www.scala-lang.org/documentation/getting-started.html

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
gabro: Should we remove the `App` trait anyway?

gabro:
Ready for the third pass 💃

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
```diff
@@ -14,18 +14,24 @@ case class NoExtendsApp(mirror: Mirror) extends SemanticRewrite(mirror) {
...
       ctx
-        .templateBodyTokens(template)
-        .map { body =>
+        .templateBodyTokens(template) match {
+        case None =>
+          ctx.reporter.warn(
+            "Missing or empty body for object that extends scala.App. See http://dotty.epfl.ch/docs/reference/dropped/delayed-init.html for possible workarounds.",
+            template.pos
+          )
+          Patch.empty
```
olafurpg: I've considered adding a `Message(Patch, msg: String, Severity, Position)` patch that can be used to attach a message to a patch. We could then expose an api like `ctx.warn(pos, msg) = Message(Patch.empty, msg, warn, pos)`. We would then make `ctx.reporter` private and issue the warnings at the end of the world, sorted by position location.

I think that would make more sense than the current `warn(msg): Unit` signature. What do you think?

scalafix-tests/output/src/main/scala/test/NoExtendsApp.scala
olafurpg: I wouldn't worry about that, RemoveUnusedImport also creates unused imports in cases like

```scala
import foo.bar
import bar._ // unused import

// after rewrite
import foo.bar // unused import
```

I'm afraid users will need to run scalafix at least twice to be sure all unused imports got removed.

scalafix-core/src/main/scala/scalafix/rewrite/NoExtendsApp.scala
gabro: I agree it would make sense to accumulate messages, but why would they need to be attached to a `Patch`?

olafurpg: You might want to attach a message to a non-empty patch to explain why a particular diff happened. This is not so useful right now since rewrites run in batch mode but it will make more sense if you want to integrate scalafix into IDEs, and create a yellow inspection message saying something like  "Hey you should not use auto-tupling, it's bad!"

gabro: so we're essentially writing a linter. I'd love that, and I think we could merge the discussion into #92

PR #194, opened 2017-06-06
Handle auto-tupling on functions

### LINKED ISSUE #147: Make NoAutoTupling work with Function1
See the discussion [here](https://github.com/scalacenter/scalafix/pull/143#issuecomment-299592113) and [here](https://github.com/scalacenter/scalafix/pull/143#discussion_r115117022).

The failing cases (rewrite not applied) are

```scala
def f: (((Int, String)) => ((String, List[Int])) => Int) = a => b => a._1
f(1, "")("a", Nil)
```

and 

```scala
val f = (a: (Int, Boolean)) => a
f(2, true)
```

~This is currently blocked by https://github.com/scalameta/scalameta/issues/837~

We can tackle this in two ways:
- by reading denotations #194 (currently blocked by https://github.com/scalameta/scalameta/issues/930)
- by reading compiler messages (blocked by imprecise positions in errors emitted by Scalameta)

gabro:
Closes #147 

Still WIP, but the previously failing test cases are now green ✅ 

By the way, I just realized that maybe this whole rewrite can be a lot simpler if you use compiler messages.

Here's what scalac emits if the auto-tupling warning is on

```sg
[warn] /Users/gabro/buildo/scalafix/scalafix-tests/input/src/main/scala/test/NoAutoTupling.scala:15: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: tup1.fooo(t: (Int, Int))(s: (String, String)): Int
[warn]   given arguments: 1, 2
[warn]  after adaptation: tup1.fooo((1, 2): (Int, Int))
[warn]     fooo(1, 2)("a", "b")
[warn]         ^
[warn] /Users/gabro/buildo/scalafix/scalafix-tests/input/src/main/scala/test/NoAutoTupling.scala:15: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: tup1.fooo(t: (Int, Int))(s: (String, String)): Int
[warn]   given arguments: "a", "b"
[warn]  after adaptation: tup1.fooo(("a", "b"): (String, String))
[warn]     fooo(1, 2)("a", "b")
[warn]               ^             
```

The carets are conveniently pointing to the positions we need.

However, I think scalameta outputs inconsistent positions similarly to what it used to do for importees.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
@@ -15,32 +15,68 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
     ctx.addLeft(args.head.tokens.head, "(") +
       ctx.addRight(args.last.tokens.last, ")")
 
+  private[this] def fixMethods(ctx: RewriteCtx,
+                               jvmSignature: String,
+                               argss: Seq[Seq[Term.Arg]],
+                               fixed: mutable.Set[Seq[Term.Arg]]): Patch = {
+    val argListSignatures =
+      jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
+    argListSignatures
+      .zip(argss)
+      .foldLeft(Patch.empty) {
+        case (patches, (argListSignature, args)) =>
+          if (!fixed(args) && singleTuplePattern
+                .matcher(argListSignature)
+                .matches && args.length > 1) {
+            fixed += args // dirty hack, see explanation above
+            patches + addWrappingParens(ctx, args)
+          } else {
+            patches
+          }
+      }
+  }
+
+  private[this] def fixFunctions(ctx: RewriteCtx,
+                                 symbol: Symbol,
+                                 argss: Seq[Seq[Term.Arg]],
+                                 fixed: mutable.Set[Seq[Term.Arg]],
+                                 mirror: Mirror): Patch =
+    mirror.database.denotations
+      .get(symbol)
+      .map { denot =>
+        val argListSignatures =
+          denot.info.split("=>").map(_.trim).toList.dropRight(1)
+        argListSignatures
+          .zip(argss)
+          .foldLeft(Patch.empty) {
+            case (patches, (argListSignature, args)) =>
+              if (!fixed(args) && argListSignature.startsWith("((") && argListSignature
+                    .endsWith("))")
+                  && args.length > 1) {
+                fixed += args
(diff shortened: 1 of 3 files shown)


scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -15,32 +15,68 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
...
+  private[this] def fixFunctions(ctx: RewriteCtx,
+                                 symbol: Symbol,
+                                 argss: Seq[Seq[Term.Arg]],
+                                 fixed: mutable.Set[Seq[Term.Arg]],
+                                 mirror: Mirror): Patch =
+    mirror.database.denotations
+      .get(symbol)
+      .map { denot =>
+        val argListSignatures =
+          denot.info.split("=>").map(_.trim).toList.dropRight(1)
```
olafurpg: Exposing richer data structures for denotations is definitely a TODO :)

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -15,32 +15,68 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
...
+        argListSignatures
+          .zip(argss)
+          .foldLeft(Patch.empty) {
+            case (patches, (argListSignature, args)) =>
+              if (!fixed(args) && argListSignature.startsWith("((") && argListSignature
```
olafurpg: Pro tip, scalafmt does not force line breaks on infix operators so you have to it yourself ;) http://scalameta.org/scalafmt/#Infixapplications

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -15,32 +15,68 @@ case class NoAutoTupling(mirror: Mirror) extends SemanticRewrite(mirror) {
...
+          }
+      }
+  }
+
+  private[this] def fixFunctions(ctx: RewriteCtx,
+                                 symbol: Symbol,
+                                 argss: Seq[Seq[Term.Arg]],
+                                 fixed: mutable.Set[Seq[Term.Arg]],
+                                 mirror: Mirror): Patch =
+    mirror.database.denotations
```
olafurpg: FYI, this list currently only contains denotations from this compilation run. It won't have denotations for 3rd party libraries, but that is planned to change in the future.

PR #197, opened 2017-06-08
Remove note from NoAutotupling docs

gabro:
That is now fixed.


olafurpg:
I just realized this caveat is still relevant, see https://github.com/scalacenter/scalafix/issues/147#issuecomment-307062340 I propose we keep this section in the docs until #147 is completely fixed.

PR #208, opened 2017-06-09
Overhaul of NoAutoTupling, using compiler warnings

gabro:
The new implementation is a huge win, being much simpler and robust at the same time.

The only limitation is that it doesn't work anymore for constructor application using `new`.

I'm not sure how to approach that. This is the compiler message for `new Foo(1, "foo")(true, Nil)`

```
[warn] /Users/gabro/buildo/scalafix/scalafix-tests/input/src/main/scala/test/NoAutoTupling2.scala:26: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: Foo(t: (Int, String))(s: (Boolean, List[Int])): NoAutoTupling2.this.Foo
[warn]   given arguments: 1, "foo"
[warn]  after adaptation: new Foo((1, "foo"): (Int, String))
[warn]   new Foo(1, "foo")(true, Nil)
[warn]   ^
```

as opposed to the one for `Foo(1, "foo")(true, Nil)`

```
[warn] /Users/gabro/buildo/scalafix/scalafix-tests/input/src/main/scala/test/NoAutoTupling.scala:30: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: Foo.apply(t: (Int, String))(s: (Boolean, List[Int])): NoAutoTupling2.this.Foo
[warn]   given arguments: 1, "foo"
[warn]  after adaptation: Foo((1, "foo"): (Int, String))
[warn]   Foo(1, "foo")(true, Nil)
[warn]      ^
[warn] /Users/gabro/buildo/scalafix/scalafix-tests/input/src/main/scala/test/NoAutoTupling.scala:30: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: Foo.apply(t: (Int, String))(s: (Boolean, List[Int])): NoAutoTupling2.this.Foo
[warn]   given arguments: true, Nil
[warn]  after adaptation: Foo((true, Nil): (Boolean, scala.collection.immutable.Nil.type))
[warn]   Foo(1, "foo")(true, Nil)
[warn]                ^
```

Unfortunately the compiler emits much more precise positions for function applications than for constructor application.

The upside is that this approach should not generate false positives.


gabro:
Ouch, apparently positions are much better in Scala 2.12 so this approach may only work there...

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -2,81 +2,24 @@ package scalafix
...
+        pos
+    }.toSet
 
-          case s @ Symbol.Global(_, Signature.Term(x))
-              if mirror.database.denotations.get(s).isDefined =>
-            fixFunctions(ctx, s, argss, fixed, mirror)
-        }.asPatch
+    ctx.tree.collect {
+      case t: Term.Apply if adaptations.contains(t.pos) =>
+        addWrappingParens(ctx, t.args)
```
olafurpg: WOW 😍  25 LOC

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -2,81 +2,24 @@ package scalafix
...
                                       args: Seq[Term.Arg]): Patch =
     ctx.addLeft(args.head.tokens.head, "(") +
       ctx.addRight(args.last.tokens.last, ")")
 
-  private[this] def fixMethods(ctx: RewriteCtx,
-                               jvmSignature: String,
-                               argss: Seq[Seq[Term.Arg]],
-                               fixed: mutable.Set[Seq[Term.Arg]]): Patch = {
-    val argListSignatures =
-      jvmSignature.stripPrefix("(").takeWhile(_ != ')').split(';')
```
olafurpg: I won't miss these hacks.

scalafix-core/src/main/scala/scalafix/rewrite/NoAutoTupling.scala
```diff
@@ -2,81 +2,24 @@ package scalafix
...
-    val fixed = mutable.Set.empty[Seq[Term.Arg]]
-    ctx.tree.collect {
-      case q"${fun: Term.Ref}(...$argss)" if argss.nonEmpty =>
-        fun.symbolOpt.collect {
-          case s @ Symbol.Global(_, Signature.Method(_, jvmSignature)) =>
-            fixMethods(ctx, jvmSignature, argss, fixed) +
-              fixFunctions(ctx, s, argss, fixed, mirror)
+    val adaptations = mirror.database.messages.toIterator.collect {
+      case Message(pos, _, msg)
+          if msg.startsWith("Adapting argument list by creating a") =>
```
olafurpg: Further motivation for https://github.com/scalameta/scalameta/issues/924 Given that we use message for two rewrites now then I think we should prioritize that issue.

olafurpg:
Speaking of names, what do you think about renaming the rewrite to `NoAdaptedArgs`? To match with the scalac option.

I've thought about adding support to provide alternative names for rewrites, since I think the current naming scheme is a bit inconsistent. We can rename the rewrite in a separate PR.

gabro:
Thanks for the review and for the commits!
I'm OK with changing the name, provided that we handle automatic unit insertion too (caught by the same flag)

gabro:
That said, I would do it on a separate issue.

olafurpg:
Awesome 👍 Let's go ahead and merge this then ! 😄

PR #210, opened 2017-06-09
Fine-tune github: protocol.

olafurpg:
I hadn't actually tried it yet, whoops. This expansion should match the
boilerplate created by g8 scalacenter/scalafix.g8

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -163,6 +163,8 @@ lazy val cli = project
     allSettings,
     publishSettings,
     isFullCrossVersion,
+    mainClass in assembly := Some("scalafix.cli.Cli"),
+    assemblyJarName in assembly := "scalafix.jar",
     libraryDependencies ++= Seq(
       "com.github.alexarchambault" %% "case-app"      % "1.1.3",
       "com.martiansoftware"        % "nailgun-server" % "0.9.1"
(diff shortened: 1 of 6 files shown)


gabro:
mm, how can I test this?

I'm trying:

```sh
sbt new scalacenter/scalafix.g8 --rewrite "reponame" --version=v1.0
```

but I get 

```
Template not found for:
```

olafurpg:
Good catch @gabro !  The assignment `=` character is apparently necessary

```
sbt new scalacenter/scalafix.g8 --rewrite="reponame" --version=v1.0
```

I updated the docs and also improved the error message when the user mistypes the `github:` protocol

gabro:
image


mm what am I getting wrong? 🤔

olafurpg:
Are you using the latest sbt version? I'm on 0.13.15

gabro:
Same...
Do I need to do anything specific to pull the template in?

The odd thing is that it doesn't even say which template it can't find.

gabro:
Oh, apparently it worked in interactive mode. Something fishy going on with zsh...

gabro:
For the record 

```sh
sbt "new scalacenter/scalafix.g8 --rewrite=fix --version=v1.0"
```

works as expected instead. Apparently without the quotes the arguments are not passed down to `sbt new`.

PR #211, opened 2017-06-11
Fix #158, fix ExplicitReturnType for non-implicits

### LINKED ISSUE #158: Respect configuration in ExplicitReturnTypes
Following an offline chat with @dotta at Scala Italy, adding an explicit return type should be within the realm of the possible and it could greatly improve compile times.

```diff
-def foo(x: Int) = if (x > 3) Some(x) else None
+def foo(x: Int): Option[Int] = if (x > 3) Some(x) else None
```

This kind of mimics the idea of ExplicitImplicit.

@olafurpg do you see any blocker for this one?

taisukeoe:
This fixes #158 to add ExplicitReturnType with respecting configuration.

ATM memberKind and memberVisibility configuration supports are implemented in "AND" basis.

For example if you specify `[Val, Def]` for memberKind and `[Protected]` for memberVisibility, only `protected val` and `protected def`s are rewritten.
I believe it should be so, but please feel free to let me know if it's different from what you expect.

And I'm so sorry for making issue reference a bit dirty.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
@@ -51,8 +51,8 @@ case class ExplicitReturnTypes(mirror: Mirror)
       base.parse[Type].toOption
     else
       /*
-      Currently a symbol of Var points to its setter function.
-      That's why its argument type should be extracted via pattern-match.
+    Currently a symbol of Var points to its setter function.
+    That's why its argument type should be extracted via pattern-match.
        */
       for {
         stat <- base.parse[Stat].toOption
@@ -86,38 +86,40 @@ case class ExplicitReturnTypes(mirror: Mirror)
       } yield ctx.addRight(replace, s": ${typ.treeSyntax}")
     }.to[Seq]
 
-    import ctx.config.explicitReturnTypes._
+    def isRewriteCandidate[D <: Defn](
+        defn: D,
+        mods: Traversable[Mod],
+        body: Term)(implicit ev: Extract[D, Mod]): Boolean = {
+      import ctx.config.explicitReturnTypes._
 
-    def matchesMemberVisibility(mods: Seq[Mod]): Boolean =
-      memberVisibility.contains(visibility(mods))
+      def matchesMemberVisibility(): Boolean =
+        memberVisibility.contains(visibility(mods))
 
-    def matchesMemberKind(defn: Defn): Boolean =
-      kind(defn).exists(memberKind.contains)
+      def matchesMemberKind(): Boolean =
+        kind(defn).exists(memberKind.contains)
 
-    def matchesSimpleDefinition(rhs: Term): Boolean =
-      rhs.is[Lit] && skipSimpleDefinitions
+      def matchesSimpleDefinition(): Boolean =
+        body.is[Lit] && skipSimpleDefinitions
+
+      defn.hasMod(mod"implicit") && !isImplicitly(body) ||
+      !matchesSimpleDefinition() &&
+      !defn.hasMod(mod"implicit") &&
+      matchesMemberKind() &&
+      matchesMemberVisibility()
+    }
 
     tree
       .collect {
         case t @ Defn.Val(mods, _, None, body)
-            if t.hasMod(mod"implicit") && !isImplicitly(body)
-              || !matchesSimpleDefinition(body)
-                && !t.hasMod(mod"implicit")
-                && matchesMemberKind(t)
-                && matchesMemberVisibility(mods) =>
+            if isRewriteCandidate(t, mods, body) =>
           fix(t, body)
 
         case t @ Defn.Var(mods, _, None, Some(body))
-            if !matchesSimpleDefinition(body)
-              && (matchesMemberKind(t)
-                && matchesMemberVisibility(mods)) =>
+            if isRewriteCandidate(t, mods, body) =>
           fix(t, body)
 
         case t @ Defn.Def(mods, _, _, _, None, body)
-            if (t.hasMod(mod"implicit")
-              || !matchesSimpleDefinition(body)
-                && matchesMemberKind(t)
-                && matchesMemberVisibility(mods)) =>
+            if isRewriteCandidate(t, mods, body) =>
           fix(t, body)
       }
       .flatten
--- a/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
@@ -27,17 +27,17 @@ case class ExplicitReturnTypes(mirror: Mirror)
     case Defn.Def(_, name, _, _, _, _) => name
   }
 
-  def visibility(mod:Mod):Option[MemberVisibility] = Option(mod).flatMap{
-    case _:Mod.Private => Some(MemberVisibility.Private)
-    case _:Mod.Protected => Some(MemberVisibility.Protected)
+  def visibility(mod: Mod): Option[MemberVisibility] = Option(mod).flatMap {
+    case _: Mod.Private => Some(MemberVisibility.Private)
+    case _: Mod.Protected => Some(MemberVisibility.Protected)
     case _ => None
   }
 
-  def kind(defn:Defn):Option[MemberKind] = Option(defn).flatMap{
-    case _:Defn.Val => Some(MemberKind.Val)
-    case _:Defn.Def => Some(MemberKind.Def)
-    case _:Defn.Var => Some(MemberKind.Var)
-    case _          => None
+  def kind(defn: Defn): Option[MemberKind] = Option(defn).flatMap {
+    case _: Defn.Val => Some(MemberKind.Val)
+    case _: Defn.Def => Some(MemberKind.Def)
+    case _: Defn.Var => Some(MemberKind.Var)
+    case _ => None
   }
 
   def parseDenotationInfo(denot: Denotation): Option[Type] = {
@@ -48,10 +48,12 @@ case class ExplicitReturnTypes(mirror: Mirror)
         throw new UnsupportedOperationException(
           s"Can't parse type for denotation $denot, denot.info=${denot.info}")
       }
-    if(denot.isVal || denot.isDef)
+    if (denot.isVal || denot.isDef)
       base.parse[Type].toOption
     else
-      base.parse[Stat].toOption.flatMap{_.collect{case Term.Ascribe(_,typ) => typ}.headOption}
+      base.parse[Stat].toOption.flatMap {
+        _.collect { case Term.Ascribe(_, typ) => typ }.headOption
+      }
   }
 
   def defnType(defn: Defn): Option[Type] =
@@ -76,27 +78,28 @@ case class ExplicitReturnTypes(mirror: Mirror)
         replace <- lhsTokens.reverseIterator.find(x =>
           !x.is[Token.Equals] && !x.is[Whitespace])
         typ <- defnType(defn)
-      } yield {
-        ctx.addRight(replace, s": ${typ.treeSyntax}")
-      }
+      } yield ctx.addRight(replace, s": ${typ.treeSyntax}")
     }.to[Seq]
 
-    def checkModsScope(mods:Seq[Mod]):Boolean =
-      ctx.config.explicitReturnTypes.memberVisibility.contains(mods.flatMap(visibility).headOption.getOrElse(Public))
-    def checkDefnScope(defn:Defn):Boolean =
+    def checkModsScope(mods: Seq[Mod]): Boolean =
+      ctx.config.explicitReturnTypes.memberVisibility
+        .contains(mods.flatMap(visibility).headOption.getOrElse(Public))
+    def checkDefnScope(defn: Defn): Boolean =
       kind(defn).exists(ctx.config.explicitReturnTypes.memberKind.contains)
 
     tree
       .collect {
         case t @ Defn.Val(mods, _, None, body)
-          if  t.hasMod(mod"implicit") && !isImplicitly(body)
-            || !t.hasMod(mod"implicit") && checkDefnScope(t) && checkModsScope(mods) =>
+            if t.hasMod(mod"implicit") && !isImplicitly(body)
+              || !t.hasMod(mod"implicit") && checkDefnScope(t) && checkModsScope(
+                mods) =>
           fix(t, body)
         case t @ Defn.Var(mods, _, None, Some(body))
-          if checkDefnScope(t) && checkModsScope(mods) =>
+            if checkDefnScope(t) && checkModsScope(mods) =>
           fix(t, body)
         case t @ Defn.Def(mods, _, _, _, None, body)
-          if t.hasMod(mod"implicit") || checkDefnScope(t) && checkModsScope(mods) =>
+            if t.hasMod(mod"implicit") || checkDefnScope(t) && checkModsScope(
+              mods) =>
           fix(t, body)
       }
       .flatten
--- a/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
+++ b/scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
@@ -50,6 +50,10 @@ case class ExplicitReturnTypes(mirror: Mirror)
     if (denot.isVal || denot.isDef)
       base.parse[Type].toOption
     else
+    /*
+      Currently a symbol of Var points to its setter function.
+      That's why its argument type should be extracted via pattern-match.
+     */
       for {
         stat <- base.parse[Stat].toOption
         typ <- stat.collectFirst { case Term.Ascribe(_, typ) => typ }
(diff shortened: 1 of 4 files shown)


scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,20 +23,40 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
       else if (denot.isDef) denot.info.replaceFirst(".*\\)", "")
       else {
         throw new UnsupportedOperationException(
           s"Can't parse type for denotation $denot, denot.info=${denot.info}")
       }
-    base.parse[Type].toOption
+    if(denot.isVal || denot.isDef)
+      base.parse[Type].toOption
+    else
+      base.parse[Stat].toOption.flatMap{_.collect{case Term.Ascribe(_,typ) => typ}.headOption}
```
taisukeoe: This is added to extract a "argument" type from dennotation of Var like `var c_= (x$1 : Int):Unit`
I would appreciate it if you can suggest a smarter way!

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +23,37 @@ case class ExplicitReturnTypes(mirror: Mirror)
 
   def defnName(defn: Defn): Option[Name] = Option(defn).collect {
     case Defn.Val(_, Seq(Pat.Var.Term(name)), _, _) => name
+    case Defn.Var(_, Seq(Pat.Var.Term(name)), _, _) => name
     case Defn.Def(_, name, _, _, _, _) => name
   }
 
+  def visibility(mod: Mod): Option[MemberVisibility] = Option(mod).flatMap {
```
gabro: You could use `collect` here and avoid the `wrapping` inside the `flatMap` as well as the fallback case.

```scala
Option(mod).collect {
  case _: Mod.Private => MemberVisibility.Private
  case _: ModProtected => MemberVisibility.Protected
}
```

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +23,37 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
+    case _: Mod.Private => Some(MemberVisibility.Private)
+    case _: Mod.Protected => Some(MemberVisibility.Protected)
+    case _ => None
+  }
+
+  def kind(defn: Defn): Option[MemberKind] = Option(defn).flatMap {
+    case _: Defn.Val => Some(MemberKind.Val)
```
gabro: same as above

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Fixed.
And changed `visibility` signature to be able to return `MemberVisibility.Public`.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Fixed.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +22,39 @@ case class ExplicitReturnTypes(mirror: Mirror)
 
   def defnName(defn: Defn): Option[Name] = Option(defn).collect {
     case Defn.Val(_, Seq(Pat.Var.Term(name)), _, _) => name
+    case Defn.Var(_, Seq(Pat.Var.Term(name)), _, _) => name
     case Defn.Def(_, name, _, _, _, _) => name
   }
 
+  def visibility(mods: Traversable[Mod]): MemberVisibility =
+    mods
+      .collect {
```
olafurpg: `.collectFirst` == `.collect.headOption`

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +22,39 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
+        case _: Mod.Private => MemberVisibility.Private
+        case _: Mod.Protected => MemberVisibility.Protected
+      }
+      .headOption
+      .getOrElse(MemberVisibility.Public)
+
+  def kind(defn: Defn): Option[MemberKind] = Option(defn).collect {
+    case _: Defn.Val => MemberKind.Val
+    case _: Defn.Def => MemberKind.Def
+    case _: Defn.Var => MemberKind.Var
```
olafurpg: It seems we should make the `implicit` part configurable. I propose we include it under MemberKind, so that 
`explicitReturnTypes.memberKind = [Implicit]` means "insert explicit types for any implicit member, regardless if it's def/var/val. WDYT/

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -1,10 +1,12 @@
...
+import scala.collection.immutable.Seq
+import scala.meta._
+import scala.meta.contrib._
+import scalafix.config.{MemberKind, MemberVisibility}
 import scalafix.syntax._
-import scala.meta._, contrib._
 import scalafix.util.Whitespace
-import scala.collection.immutable.Seq
 
 // TODO: implement ExplicitReturnTypesConfig
```
olafurpg: No longer TODO :)

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +22,39 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
       else if (denot.isDef) denot.info.replaceFirst(".*\\)", "")
       else {
         throw new UnsupportedOperationException(
           s"Can't parse type for denotation $denot, denot.info=${denot.info}")
       }
-    base.parse[Type].toOption
+    if (denot.isVal || denot.isDef)
+      base.parse[Type].toOption
+    else
+      base.parse[Stat].toOption.flatMap {
```
olafurpg: Interesting, it seems the symbol for vars point to the setter. This may change in the future, FYI, but this solution is OK for now! We have tests :)

Can we use a for comprehension instead?

```scala
// NOTE: The symbol for vars points to the getter, this may change in the future
// but for now we parse the structure of the getter's denotation.  
for {
  stat <- base.parse[Stat].toOption
  typ <- stat.collectFirst { case Term.Ascribe(_, typ) => typ }
} yield typ
```

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -58,14 +82,27 @@ case class ExplicitReturnTypes(mirror: Mirror)
         typ <- defnType(defn)
       } yield ctx.addRight(replace, s": ${typ.treeSyntax}")
     }.to[Seq]
+
+    def checkModsScope(mods: Seq[Mod]): Boolean =
```
olafurpg: s/checkModsScope/matchesMemberVisibility/

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -58,14 +82,27 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
+      ctx.config.explicitReturnTypes.memberVisibility
+        .contains(visibility(mods))
+
+    def checkDefnScope(defn: Defn): Boolean =
```
olafurpg: s/checkDefnScope/matchesMemberKind/

scalafix-tests/input/src/main/scala/test/ExplicitReturnTypes.scala
```diff
@@ -1,9 +1,21 @@
...
+explicitReturnTypes.memberKind = [Def, Var]
+explicitReturnTypes.memberVisibility = [Public, Protected]
  */
 package test
 
 object ExplicitReturnTypes {
+  val a = 1
+  def b() = List
+  var c = true
+  protected val d = 1.0f
```
olafurpg: I think intellij-scala excludes cases when the right-hand side is a plain literal, see https://blog.jetbrains.com/scala/2016/10/05/beyond-code-style/

We could do something similar, have an option to skip the rewrite if the right-hand side body is a `Lit`.

```
explicitReturnTypes.skipSimpleDefinitions: Boolean // true by default
```

We can later expand on the heuristic to determine if a definition is "simple" or not.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Thanks. Fixed.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Fixed.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: I would think `Implicit` should not be an option and be always applied, since `Implicit` should always have explicit return type / type annotated even in Scala 2.x , like:
https://issues.scala-lang.org/browse/SI-9130

While I believe there is no use case that implicit should NOT have explicit return type, please let me know if there is.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: As you thought, `Var` part is the most tricky one in this PR.
While I didn't expect that `Var` denotation would be the setter function, do you think it's a bug in scala.meta? (If so, I'd love to contribute some to scala.meta :) )

ATM, I'll add some comments on this.

As for `for comprehension`, I fixed.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Fixed.

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Fixed.

scalafix-tests/input/src/main/scala/test/ExplicitReturnTypes.scala
taisukeoe: It's good idea to add `skipSimpleDefinitions` into ExplicitReturnType option. I added.

As you mentioned, we can revisit a definition of `simple` later on :)

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -20,18 +21,43 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
         throw new UnsupportedOperationException(
           s"Can't parse type for denotation $denot, denot.info=${denot.info}")
       }
-    base.parse[Type].toOption
+    if (denot.isVal || denot.isDef)
+      base.parse[Type].toOption
+    else
+      /*
+      Currently a symbol of Var points to its setter function.
+      That's why its argument type should be extracted via pattern-match.
```
xeno-by: Let's point this to a getter. This would be a simple matter of changing https://github.com/scalameta/scalameta/blob/48eff97ac6951e5b58eb2ec7251f69d53a38a73a/scalahost/nsc/src/main/scala/scala/meta/internal/semantic/DenotationOps.scala#L15.

Would you consider contributing this to scalameta/scalameta?

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -58,14 +85,39 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
+    import ctx.config.explicitReturnTypes._
+
+    def matchesMemberVisibility(mods: Seq[Mod]): Boolean =
+      memberVisibility.contains(visibility(mods))
+
+    def matchesMemberKind(defn: Defn): Boolean =
+      kind(defn).exists(memberKind.contains)
+
+    def matchesSimpleDefinition(rhs: Term): Boolean =
+      rhs.is[Lit] && skipSimpleDefinitions
```
xeno-by: What's a "simple definition"?

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: Thanks for your confirmation and suggestion!
Yes, I'd love to. So can I file this issue, and try to solve it?

xeno-by: Yes. Please do :)

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: With borrowing @olafurpg 's comment, I would define a "simple definition" as "its right-hand-side is a plain literal".

https://github.com/scalacenter/scalafix/pull/211/files/d49d74bbbb7eeece82e70af0df6f8b3ecf76d3f4#r121351776

I would also think a return type of a plain literal looks obvious, and needn't to be rewritten. What do you think about it?

xeno-by: At ScalaDays Copenhagen, I remember chatting with @olafurpg about binpacking in scalafmt, and he said that he had binpacking behaving differently for literals and non-literals. Afterwards, the special behavior for literals ended up being so useful that Olaf upgraded it from just literals to literals and short identifiers.

When I saw a notion of a "simple definition" in this pull request, I thought that maybe that scalafmt story and this pull request relate in any way. Could you comment on that? /cc @olafurpg

olafurpg: I'd say in this case we want to treat identifiers as "complex definitions" since inferring their type may require inferring complex definitions

```scala
def foo = Generic[Bar] // infers to complex type `String :+: Int :+ ... :+: HNil`
implicit val x  = foo // not simple, we want explicit type here.
```

I think literals are a good enough for "simple definitions", at least to start with. We may later want to handle more sophisticated cases like

```
val foo = Option.empty[Int]
// after rewrite
val foo: Option[Int] = None
```

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
```diff
@@ -58,14 +85,39 @@ case class ExplicitReturnTypes(mirror: Mirror)
...
+
     tree
       .collect {
         case t @ Defn.Val(mods, _, None, body)
-            if t.hasMod(mod"implicit") &&
-              !isImplicitly(body) =>
+            if t.hasMod(mod"implicit") && !isImplicitly(body)
+              || !matchesSimpleDefinition(body)
```
olafurpg: Seems like this complex condition is repeated for val/var/def. Can we extract it into a helper? Something like

```scala
def isRewriteCandidate(t: Defn, mods: Seq[Mod], body: Term): Boolean = ???
```

taisukeoe: Indeed.
I have extracted it into a helper function, and happen to find that `implicit var` was not include in rewrite scope. I believe it should be, and have applied the same helper function into `Var` as `Val` or `Def`.

taisukeoe: Like:

```scala
private implicit var j = 1

//after rewrite

private implicit var j: Int = 1
```

scalafix-core/src/main/scala/scalafix/rewrite/ExplicitReturnTypes.scala
taisukeoe: As @xeno-by suggested, it seems to be hard to guess "what is simple?".
So now I think `skipDefinitionsByLiteral` would be a more right word than `skipSimpleDefinitions` ATM.
Of course I'm open to revisit the name after we expand the definition of "simple".

olafurpg: I think I prefer to have the flexibility to redefine simple in future releases. We could make the option an enum, `skipDefinitions = literals/simple`, but I think that's overkill for now. Let's keep it `skipSimpleDefinitions`.

taisukeoe:
Thanks!

Issue #224, opened 2017-06-22
Automatically detect --syntactic


olafurpg:
Currently, with #220, it's necessary to pass --syntactic for syntactic rewrites, `scalafix --syntactic --rewrites RemoveXmlLiterals`. It would be nice if --syntactic was inferred if the passed rewrites are only syntactic.

olafurpg:
Fixed in #228

PR #227, opened 2017-06-26
Cross-compile core to Scala.js

gabro:
Very WIP.

This PR serves a base for a discussion :)

Blocking things I've noticed so far:
- `metaconfig-typesafe-config` seems to be missing from maven
- `googleDiff` is the only JVM-only dependency. We can probably extract the diffing part into a separate module that uses `googleDiff` for the JVM and some other diffing library for JS.

Minor:
- The strategy of putting deps in `Dependencies.scala` falls short for cross deps (I've just inlined them for now)
- I couldn't find an easy way of making `.configure(setId)` work for `CrossProject`. I've just manually configured the cross project for now

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
+++ b/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
@@ -14,37 +14,33 @@ import java.util.concurrent.atomic.AtomicReference
 import java.util.function.UnaryOperator
 import java.util.regex.Pattern
 import java.util.regex.PatternSyntaxException
-import java.util.stream.Collectors
 import scala.meta._
 import scala.meta.inputs.Input
 import scala.meta.internal.inputs._
 import scala.meta.internal.io.PlatformFileIO
 import scala.meta.internal.semantic.vfs
 import scala.meta.io.AbsolutePath
-import scala.util.Success
 import scala.util.Try
 import scala.util.control.NonFatal
 import scalafix.cli.termdisplay.TermDisplay
 import scalafix.config.Class2Hocon
 import scalafix.config.FilterMatcher
+import scalafix.config.LazyMirror
+import scalafix.config.MetaconfigPendingUpstream
 import scalafix.config.PrintStreamReporter
+import scalafix.config.RewriteKind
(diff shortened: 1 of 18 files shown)


gabro:
This looks like a valid candidate for the JS implementation of diff.

https://github.com/kpdecker/jsdiff

gabro:
I just realized I've no idea of how to implement the same interface differently for JS and JVM in a cross build. Is this even possible?

Ideally for the JVM project we'd use `googleDiff` and for the JS one `jsdiff`.

gabro:
Ok, I figured that one out (people were nice on gitter :D). I've just pushed a tentative `scalafix-diff` module that abstracts over the diff generation cross platform.

If you're ok with this approach @olafurpg I'll add some tests to ensure consistent outputs across the two platforms.

olafurpg:
@gabro This is awesome! 

To quickly answer your questions
- `metaconfig-hocon` can be used on JS, http://search.maven.org/#artifactdetails%7Ccom.geirsson%7Cmetaconfig-hocon_sjs0.6_2.12%7C0.4.0%7Cjar That's what scalafmt does https://github.com/scalameta/scalafmt/blob/d204ae1437247aae0f0b8a9901e433368057e678/build.sbt#L41-L46 metaconfig-hocon is not as robust as the typesafe config variant, but we can always fix that if it becomes an issue.
- The `Patch.diff` method can be moved to `scalafix.util.Diff(Input, Input)` with the current implementation the `jvm/` directory with a dummy implementation in core. No need for a separate project.

I would like to get the fix for #224 in before we do this change, since I have a fairly large pending diff. Will review closer after that fix is ready ;)

olafurpg:
As for the tests, we should be able to run most of the SemanticTests under JS. It's fine to move them all to jvm only to begin with, but only the ones that do classloading or runtime compilation need to be on the jvm. The io module works on Node.js and semantic api work in JS too.

olafurpg:
#228 is the PR I'd like to merge first to avoid rebase issues.

olafurpg:
Heads up, I'm rebasing this PR on top of master and fixing the resolution errors.

PR #229, opened 2017-06-27
Add Scala.js support, fixes #226

### LINKED ISSUE #226: Cross compile core to Scala.js
As discussed on gitter, it shouldn't be that hard and it could enable some interesting usage from browser-based applications.

olafurpg:
Supersedes #227


scalafix-core/shared/src/main/scala/scalafix/config/ScalafixConfig.scala
```diff
@@ -6,7 +6,9 @@ import scala.meta._
 import scala.meta.dialects.Scala211
 import scala.meta.parsers.Parse
 import metaconfig._
-import metaconfig.typesafeconfig.typesafeConfigMetaconfigParser
+//import metaconfig.typesafeconfig.typesafeConfigMetaconfigParser
```
olafurpg: This is removed in #230

PR #231, opened 2017-06-28
Add tests for scalafix-diff

gabro:
This adds a basic test for `DiffUtils.unifiedDiff` to ensure the output is consistent between JS and JVM implementations.

I found some discrepancies, namely:

- [x] the JS library also accepts a String to append to the header after the file name. Despite passing an empty String a `\t` was still appended, so I manually trimmed it.

- [x] the JS library starts the output with a line made of `===`. I dropped it.

- [ ] the JS library appends a `\ No newline at end of file` to the diff, whereas the Java library doesn't. I'm not sure of who's right and how to work around this. @olafurpg any idea in merit?

### DIFF
--- a/scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
+++ b/scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
@@ -33,12 +33,11 @@ object DiffUtils {
     def trimHeader(line: String) =
       if (line.startsWith("+++") || line.startsWith("---")) line.trim else line
 
-    def removeNewlineDiff(diffLines: Array[String]) =
+    def removeNewlineDiff(diffLines: Iterator[String]) =
       diffLines.filterNot(_ == "\\ No newline at end of file")
 
     removeNewlineDiff(
-      diff
-        .split("\n")
+      diff.lines
         .drop(1) // remove ==== separator
         .map(trimHeader) // remove whitespaces at the end of headers
     ).mkString("\n")
--- a/scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
+++ b/scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
@@ -29,12 +29,18 @@ object DiffUtils {
       "",
       "",
       js.Dynamic.literal("context" -> contextSize))
+
     def trimHeader(line: String) =
       if (line.startsWith("+++") || line.startsWith("---")) line.trim else line
-    diff
-      .split("\n")
-      .drop(1) // remove ==== separator
-      .map(trimHeader) // remove whitespaces at the end of headers
-      .mkString("\n")
+
+    def removeNewlineDiff(diffLines: Array[String]) =
+      diffLines.filterNot(_ == "\\ No newline at end of file")
+
+    removeNewlineDiff(
+      diff
+        .split("\n")
+        .drop(1) // remove ==== separator
+        .map(trimHeader) // remove whitespaces at the end of headers
+    ).mkString("\n")
   }
 }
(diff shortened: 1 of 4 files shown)


scalafix-diff/shared/src/test/scala/scalafix/DiffUtilsSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  val contextSize = 4
+
+  val expectedDiff =
+    s"""|--- SomeFile.scala
+        |+++ SomeFileRenamed.scala
+        |@@ -1,7 +1,9 @@
+        | object Foo {
+        |-  const x = Map(
+        |+  const y = Map(
```
olafurpg: Are we testing the context size here?

scalafix-diff/shared/src/test/scala/scalafix/DiffUtilsSpec.scala
```diff
@@ -0,0 +1,55 @@
+package scalafix.diff
+
+import org.scalatest.FunSuite
+
+class DiffUtilsSpec extends FunSuite {
```
olafurpg: s/Spec/Test/

scalafix-diff/shared/src/test/scala/scalafix/DiffUtilsSpec.scala
gabro: ah right, it's probably not enough, I'll add more statements in between.

Also, lol for using `const` instead of `val`. Living on the edge...

gabro:
> What happens when you apply a diff with "no newline at eof" with git?

I don't think ti makes a difference. From what I understand reading [the implementation](https://github.com/kpdecker/jsdiff/blob/64eb9206b7bede737ca2032b8bd04ed958c8dc89/src/patch/create.js#L74-L78), it is only used for context.

In fact, my last commit (to test context in a more meaningful way) incidentally made the EOF not to be included in the context and the difference disappeared.

The Java implementation never adds that line (there's no trace in the implementation), so (for consistency with what we use now) I propose that we simply filter it out from the output.

gabro:
@olafurpg I pushed an implementation that does it. If you have a different proposal, I can remove the last commit.

scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
```diff
@@ -29,12 +29,18 @@ object DiffUtils {
...
-      .drop(1) // remove ==== separator
-      .map(trimHeader) // remove whitespaces at the end of headers
-      .mkString("\n")
+
+    def removeNewlineDiff(diffLines: Array[String]) =
+      diffLines.filterNot(_ == "\\ No newline at end of file")
+
+    removeNewlineDiff(
+      diff
+        .split("\n")
```
olafurpg: s/.split(..)/.lines/ will return an iterator, which is lazy and avoids the creation of an intermediary array in all steps.

scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
```diff
@@ -26,9 +26,21 @@ object DiffUtils {
...
+      "",
+      "",
       js.Dynamic.literal("context" -> contextSize))
-    diff.split("\n").drop(1).mkString("\n")
+
+    def trimHeader(line: String) =
+      if (line.startsWith("+++") || line.startsWith("---")) line.trim else line
+
+    def removeNewlineDiff(diffLines: Array[String]) =
+      diffLines.filterNot(_ == "\\ No newline at end of file")
```
olafurpg: This is reasonable.

scalafix-diff/js/src/main/scala/scalafix/diff/DiffUtils.scala
gabro: nice, fixed it

gabro:
@olafurpg done, check it out

Issue #192, opened 2017-06-05
Move internal API to an 'internal' package


gabro:
This is an idea discussed offline with @olafurpg.

Currently private (low-level) and public (high-level) APIs are mixed, with private ones being marked as `private[scalafix]`.
While this works from a technical point of view, there's no obvious separation between the two.

(By "public"/"private" here I mean from rewrites authors' perspective)

A cleaner approach could be moving all the low-level internal stuff to an `internal` package (similarly to what Scalameta does) which would clarify the intention.

Thoughts anyone?

xeno-by:
Can recommend to a friend. I think this has been working extremely well for us in meta.

olafurpg:
I agree this is a good idea. We can make this refactoring as part of v0.5

PR #235, opened 2017-06-30
Move patch operations into RewriteCtx, fixes #193.

### LINKED ISSUE #193: Refactor patch extension methods into plain methods on RewriteCtx
PatchOps are currently extension methods on RewriteCtx defined here https://github.com/scalacenter/scalafix/blob/8641796376b924d2d8719cfdc9c7c4975719569e/scalafix-core/src/main/scala/scalafix/package.scala#L20-L28

I think it would make more sense to mix those methods directly into RewriteCtx and make semantic operations take an `implicit Mirror` parameter.

Related: https://github.com/scalacenter/scalafix/pull/188

olafurpg:
This avoid the need for extension methods. The implementations of the
patch operations are now abstract to make it easier to read.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala
@@ -5,12 +5,8 @@ package patch
 import scala.meta._
 import org.scalameta.FileLine
 
-object PatchOps {
-  def removeTokens(tokens: Tokens): Patch =
-    tokens.foldLeft(Patch.empty)(_ + TokenPatch.Remove(_))
-}
-
-trait SyntacticPatchOps {
+trait PatchOps {
+  // Syntactic patch ops.
   def removeImportee(importee: Importee): Patch
   def replaceToken(token: Token, toReplace: String): Patch
   def removeTokens(tokens: Tokens): Patch
@@ -18,9 +14,8 @@ trait SyntacticPatchOps {
   def rename(from: Name, to: Name)(implicit fileLine: FileLine): Patch
   def addRight(tok: Token, toAdd: String): Patch
   def addLeft(tok: Token, toAdd: String): Patch
-}
 
-trait SemanticPatchOps {
+  // Semantic patch ops.
   def removeGlobalImport(symbol: Symbol)(implicit mirror: Mirror): Patch
   def addGlobalImport(importer: Importer)(implicit mirror: Mirror): Patch
   def replace(from: Symbol,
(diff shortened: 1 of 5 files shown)


olafurpg:
Quite nice to see no changes were necessary on the rewrites. I was tempted to mix the operations into `Rewrite` but then I remembered that we may need the input/ctx in order to implement some particular patches. One idea that I have flirted with is to add extension methods on Tree/Token so it's possible to write

```scala
tree.replaceWith(newTree)
```

The RewriteCtx would need to be implicit in that case. However, I personally prefer the `ctx.*` approach since it makes it easy to discover what operations are available.

gabro:
Nice! I'll review it soon, in the meantime the CI is failing because there's a faulty reference in the readme.

olafurpg:
I'm about to finish #192 soon, which follows up on this PR.

scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
```diff
@@ -1,20 +1,27 @@
...
 import scalafix.util.MatchingParens
 import scalafix.util.TokenList
+import org.scalameta.FileLine
 import org.scalameta.logger
 
 /** Bundle of useful things when implementing [[Rewrite]]. */
-case class RewriteCtx(tree: Tree, config: ScalafixConfig) {
-  def syntax =
+case class RewriteCtx(tree: Tree, config: ScalafixConfig) extends PatchOps {
+  ctx =>
```
gabro: this is just an alias for `this` right?

scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala
```diff
@@ -3,35 +3,25 @@ package scalafix
...
+  def rename(from: Name, to: Name)(implicit fileLine: FileLine): Patch
+  def addRight(tok: Token, toAdd: String): Patch
+  def addLeft(tok: Token, toAdd: String): Patch
 
-class SemanticPatchOps(ctx: RewriteCtx, mirror: Mirror) {
-  def removeGlobalImport(symbol: Symbol): Patch = RemoveGlobalImport(symbol)
-  def addGlobalImport(importer: Importer): Patch = AddGlobalImport(importer)
+  // Semantic patch ops.
+  def removeGlobalImport(symbol: Symbol)(implicit mirror: Mirror): Patch
+  def addGlobalImport(importer: Importer)(implicit mirror: Mirror): Patch
```
gabro: I think this makes a lot of sense 👍

scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
olafurpg: Yes, but I like the convention of calling it `ctx`. The ctx is not much used to implement the patches at the moment but I suspect that later we may need it to support configuration etc.

olafurpg:
Thank you for the review!

PR #236, opened 2017-06-30
Move internal code to internal package, fixes #192.

### LINKED ISSUE #192: Move internal API to an 'internal' package

olafurpg:
This PR is a followup of #193.


project/Mima.scala
```diff
@@ -0,0 +1,10 @@
+import com.typesafe.tools.mima.core._
+import com.typesafe.tools.mima.core.ProblemFilters._
+
+object Mima {
+  val ignoredABIProblems: Seq[ProblemFilter] = {
+    Seq(
+      exclude[MissingClassProblem]("scalafix.package$XtensionRewriteCtx")
```
gabro: I'm not super familiar with MiMa, can you please explain what is this for?

olafurpg: It's a way to document binary breaking changes. `sbt mimaReportBinaryIssues` compares the classfiles of the current HEAD with the classfiles of the latest published release to Maven and reports issues such as "missing class/method". It's possible to ignore those issues by acknowledging them like this, to indicate you the breaking change is OK.

olafurpg: More info here https://github.com/typesafehub/migration-manager/wiki/sbt-plugin#basic-usage

olafurpg: After the v0.5.0 release, we can run `mimaReportBinaryIssues` in CI so that PRs will need to document the breaking changes in the public API.

gabro: Thanks. And what about this specific line?
Why are we ignoring that case?

olafurpg: It really shouldn't be there, I kept it as an example of how to filter out errors. I'll remove it.

gabro: Ah ok, got it!

PR #239, opened 2017-06-30
Remove unused imports.

olafurpg:
I've been delaying this commit for too long hoping to fix it with
scalafix. However, it turns out that we can't yet use scalahost in the
scalafix repo because of https://github.com/scalameta/scalameta/issues/979.

I could not get myself to manually remove the imports so I implemented
instead an sbt plugin called sbt-messagehost that collects reported messages
from the compilerReporter in sbt and writes .semanticdb files. Using
messagehost, I was able to run RemoveUnusedImports on the scalafix
codebase. I suspect messagehost may become popular as a lightweight
alternative to scalahost.

I plan to publish messagehost later, it will not be part of the scalafix
repo.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -387,6 +387,7 @@ lazy val readme = scalatex
         .dependsOn(run.in(Compile).toTask(" --validate-links"))
         .value
     },
+    scalacOptions ~= (_.filterNot(_ == "-Xlint")),
     test := run.in(Compile).toTask(" --validate-links").value,
     libraryDependencies ++= Seq(
       "com.twitter" %% "util-eval" % "6.42.0"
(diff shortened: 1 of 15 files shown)


ShaneDelmore:
This is a great idea.  I have been wanting to automate removing unused imports as they are a little annoying to have to keep removing manually but not wanting the full weight of semantic db for a case where the compiler message clearly has the needed information.  Love this ❤️

olafurpg:
Glad you like the idea @ShaneDelmore I needed to make minor accommodations in RemoveUnusedImport to work with sbt-messagehost's offset positions instead of range positions, but it was not too tricky.

PR #240, opened 2017-06-30
Add utilities to help debug mirror for custom ctx, fixes #232

### LINKED ISSUE #232: Utility to print database for single RewriteCtx
Currently, `logger.elem(mirror)` will print out the database for all files in the mirror. It would be nice if there was some `ctx.debug()` that prints out the database for that file only.

olafurpg:


scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
```diff
@@ -30,8 +30,21 @@ case class RewriteCtx(tree: Tree, config: ScalafixConfig) extends PatchOps {
   lazy val tokenList: TokenList = new TokenList(tokens)
   lazy val matching: MatchingParens = MatchingParens(tokens)
   lazy val comments: AssociatedComments = AssociatedComments(tokens)
+  lazy val input: Input = tokens.head.input
   val reporter: ScalafixReporter = config.reporter
 
+  // Debug utilities
+  def mirror(implicit mirror: Mirror): Database =
```
gabro: It's a bit strange that mirror returns a Database.

gabro: How about 'database' instead?

olafurpg: Good catch, I'll let it return a Mirror instead. I find the database vs. mirror distinction confusing, ideally we should consolidate them. cc/ @xeno-by

olafurpg: I opted for the return type `Mirror` instead of calling the method `database`

gabro:
Code looks 👍  but the CI is unhappy, although it looks unrelated...

olafurpg:
It seems some merge got broken with all the parallel PRs, master became red even if I only merged on green :O I've rebase on master, which should fix the issue.

PR #248, opened 2017-07-13
Implement crude MoveSymbol patch.

olafurpg:
This patch can be used to move or rename call-sites of a global
definition like a class, trait, object or static def. This
implementation is crude, it is likely produce invalid code in many
cases. The patch tries to preserve the prefix when possible. For
example:

```
// config
from = a.b.c
to = d.b.x
// code before
import a.b.c
import a.b
b.c
c
// code after
import a.b
import d.b
import d.b.x
b.x
x
```

Room for improvement:

- Observe that `import a.b` is not removed as it may be used to prefix
  other identifiers from the `a.b.` package. It should be possible to
  remove the imports if its not used elsewhere.
- It's possible that duplicate imports are added, for example if
  the source contains `import a.b_` then scalafix may still add
  redundant `import a.b.c`.
- Imports are always inserted fully qualified at the bottom of the
  global imports list. It might be nice if it handled
  - imports grouped by curly braces
  - import sorting
  - local imports

### DIFF
--- a/bin/new-rewrite.sh
+++ b/bin/new-rewrite.sh
@@ -2,6 +2,6 @@
 set -eux
 rewrite=$1
 
-touch scalafix-core/src/main/scala/scalafix/rewrite/${rewrite}.scala
+touch scalafix-core/shared/src/main/scala/scalafix/rewrite/${rewrite}.scala
 touch scalafix-tests/input/src/main/scala/test/${rewrite}.scala
 touch scalafix-tests/output/src/main/scala/test/${rewrite}.scala
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,8 +1,7 @@
 addSbtPlugin("com.eed3si9n" % "sbt-doge" % "0.1.5")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
-resolvers += Resolver.bintrayIvyRepo("scalameta", "sbt-plugins") // only needed for scalatex 0.3.8-pre
-addSbtPlugin("com.lihaoyi" % "scalatex-sbt-plugin" % "0.3.8-pre")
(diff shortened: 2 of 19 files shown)


olafurpg:
This implementation needs more battle testing, but I'm gonna go ahead and merge it to unblock v0.5.

ShaneDelmore:
@olafurpg This only changes the callers to use a different implementation, it doesn’t actually move the implementation does it?  If so, how about calling it something like retarget or switchImplementation.  I’m actually not coming up with many ideas for names right now and will update the issue if I think of something better but if I am understanding this functionality correct it would be nice if it communicated that it is going to use an implementation of the method called from another location, it is not actually moving or altering the implementation.

olafurpg:
Yes, that's been nagging me too. I agree "Move" should correspond to the IDE "Move class" which also handles the definition site. Coming up with a good name is challenging, suggestions are welcome. 

I used "MoveSymbol" in the context of a library, where users bump up a version and the deprecation messages say "this method has been moved there" or something. In this scenario you don't have to move the definition.

olafurpg:
MoveSymbolUsage?

ShaneDelmore:
The move work also makes me think I am moving an implementation.

If I wasn’t trying to come up with a specific word I think I would describe this as “use different implementation”.  

Actually, isn’t this just the old replaceSymbol?  I like replace, retarget, use, because they don’t have the connotation of modifying the implementation itself, just choosing to use a different one.  

I’m curious if anyone else will weigh in but so far I like replaceSymbol the most, and if that is not a good fit it would be helpful to know where my misunderstanding is on what the difference is.

olafurpg:
`replaceSymbol` is good, I will update the name. Thanks for pitching in @ShaneDelmore, I often get too lost in the details 😅

olafurpg:
https://github.com/scalacenter/scalafix/pull/262

Issue #215, opened 2017-06-13
Handle overridden member in NoExtendsApp


olafurpg:
See https://github.com/lampepfl/dotty/issues/559#issuecomment-308057980

NoExtendsApp currently wraps all statements inside the main function, include overridden members.

```scala
// input
trait Foo {
  def bar = 23
  def baz
}
object Main12 extends Foo with App {
  override def bar: Int = 42
  def baz = 14
  println(bar)
  println(baz)
}

// current output
trait Foo {
  def bar = 23
  def baz
}
object Main12 extends Foo {
  def main(args: Array[String]): Unit = {
    override def bar: Int = 42
    def baz = 14
    println(bar)
    println(baz)
  }
}

// expected output
trait Foo {
  def bar = 23
}
object Main12 extends Foo {
  override def bar: Int = 42
  def baz = 14
  def main(args: Array[String]): Unit = {
    println(bar)
  }
}
```

gabro:
Oh, this sounds tricky.
Is there some denotation that can help in dealing with this?

(also, are you missing `baz` in the expected output?)

olafurpg:
It's gets even hairier with code that relies on eager evaluation of vals/vars

```scala
trait Foo{
  def x: Int
}
object Main12 extends App with Foo {
  println("root")
  val x = {
    println("x")
    1
  }
}
```
In that case I think it's necessary to do something like
```scala
trait Foo{
  def x: Int
}
object Main12 extends App with Foo {
  val x = {
    println("x")
    1
  }
  def main = {
    println("root")
    x // <- synthesize this one here
  }
}
```
to trigger the evaluation of `x`. With that workaround, we wouldn't have to check if the member overrides a parent member or not, just keep all def/val/var in the top scope.

gabro:
So, to recap:

- [ ] non-lazy vals need to be synthesized in the main function, to preserve the eager evaluation semantics
- [ ] members that override/implement members of a parent should be hoisted (left in the outer scope)

Both sound doable, assuming the semanticdb provides enough information for the second one.

gabro:
In general, we may think of leaving all the member definitions in the outer scope and synthesize the non-lazy vals. @olafurpg do you see any counter-argument to this approach?

olafurpg:
I think we can look at it this way instead

- keep all members in the top-scope unchanged
- move non-member statements into the `main` body along with synthetic calls non-lazy val/var

[...]

PR #250, opened 2017-07-14
Upgrade to Scalameta v2

olafurpg:


readme/Rewrites.scalatex
```diff
@@ -219,26 +219,5 @@
     @p
       The two syntaxes are equivalent and the presence of the @code{val} keyword has been deprecated since Scala 2.10.
 
-  @sect(NoExtendsApp.toString)
```
olafurpg: cc/ @gabro 

I'm going on vacation next week so I didn't have time to closely debug the failures. I suspect it's not too hard to fix, but given #215 I felt it would be best to remove the rewrite since it already had enough problems. I will try to add the rewrite back again once I'm back from vacation in August.

gabro: sure, and if I find some time I'll take a look myself!
Enjoy your vacation! 🏖

PR #251, opened 2017-07-15
Upgrade to meta v2 and metaconfig v0.5

olafurpg:
- Remove sbt-bintray.
- Configure CI to publish to sonatype on tag push.
- Turn of npm warnings about licence, we don't publish packages to npm, yet.


gabro:
🙌 nice!

olafurpg:
First run

> ./bin/ci-publish.sh: line 18: [: missing `]'

:sob:

I got it working for the scalameta and metaconfig repo. Release on tag push is so promising :smile: and surprisingly simple to setup once you figure out the right gpg commands

gabro:
bash is hard, you can probably dry run the script locally with the right env variables

gabro:
maybe running it inside a docker Linux image, osx bash is a bit different

olafurpg:
I did dry run it locally, but OSX allows single brackets. I'm planning to move this into an sbt-plugin to avoid this bash business

olafurpg:
NB. Release on tag push will make it much simpler to stick to a release schedule. I'm very excited about this. Releasing is incredibly time consuming esp. for larger projects like scalameta that contain >60 modules when you take into account the full cross-build

olafurpg:
https://travis-ci.org/scalacenter/scalafix/jobs/253903395#L2274
> [info] repositoryClosed: id:chepflscala-1056

It works 🎉

gabro:
Cool! That definitely will allow releasing more often!

PR #259, opened 2017-07-17
Fix #253, add option to group inserted imports by ref.

### LINKED ISSUE #253: addGlobalImport should group imports by ref
Currently it produces output like this
```scala
import scala.Predef.{ augmentString => _ }
import scala.Predef.{ intArrayOps => _ }
import scala.`package`.{ List => _ }
import scala.`package`.{ Vector => _ }
import scala.Predef.{ Map => _ }
import scala.`package`.{ :: => _ }
import scala.`package`.{ Seq => _ }
import scala.`package`.{ #:: => _ }
import strawman.collection.immutable.HashMap
import strawman.collection.immutable.Vector
import strawman.collection.immutable.Map
import strawman.collection.immutable.Nl
import strawman.collection.immutable.List
import strawman.collection.immutable.LazyList
import strawman.collection.mutable.ArrayBuffer
```
it would be nice if it grouped by shared prefix.

olafurpg:


ShaneDelmore:
Love this change, thanks!

olafurpg:
Glad you like it! This may seem identical to the imports configuration that existed before but the difference now is that

- scalafix does not fiddle with existing imports
- the configuration only applies to the imports scalafix inserts

PR #263, opened 2017-07-18
Add DottyEnumKeyword rewrite, fixes #247.

### LINKED ISSUE #247: Backtick enum identifiers for Dotty
`enum` is now a keyword in Dotty. It would be nice to have scalafix rewrite that backticks `enum` identifiers to avoid compilation errors like this

```scala
[error] -- [E039] Syntax Error: /home/geirsson/scalapb/compiler-plugin/src/main/scala/com/trueaccord/scalapb/compiler/ProtobufGenerator.scala:1288:44
[error] 1288 |      b.setName(file.scalaDirectory + "/" + enum.getName + ".scala")
[error]      |                                            ^^^^
[error]      |                                      ')' expected, but 'enum' found
```

olafurpg:
Could someone from the Dotty team maybe review, @smarter?


smarter:
Looks fine to me, should this be generalized to other keywords, like `inline` ?

gabro:
We could also have a compound "DottyKeyword" composing this and other ones like inline

olafurpg:
Thanks for the feedback! I like the idea of a `DottyKeywords` rewrite that also handles inline. Inline is a bit trickier since Dotty allows it in annotation position under -language:Scala2.

PR #268, opened 2017-08-02
Run the CLI using a specialized sbt scope

jonas:
This changes the sbt plugin to use the normal Ivy resolution mechanism by "tagging" the Scalafix CLI dependencies with a custom scope.

The output is changed because it now uses sbt's JVM run mechanism, e.g.:

    Running scalafix.cli.Cli --config .../.scalafix.conf ...

---

This is mostly a PoC based on a solution developed for the [sangria-codegen sbt plugin](https://github.com/mediative/sangria-codegen). The magic part is this line

    ivyConfigurations += Scalafix

which allows to resolve the CLI dependencies and run it using the provided sbt API.

It's likely creative(?) sbt hackery and I don't know if this would be considered an acceptable approach by the high priests and priestesses of sbt.


olafurpg:
Thank you for this PR! This is in fact how the original sbt-scalafmt plugin worked. This approach has problems when scalaVersion conflicts with scalafixScalaVersion (see https://github.com/scalameta/scalafmt/issues/485 and https://github.com/sbt/sbt/issues/2786). neo-scalafmt works around that issue with `ivyScala := ivyScala.value.map(_.copy(overrideScalaVersion = false))` (see https://github.com/scalameta/scalafmt/issues/923#issuecomment-299934986) but that seems to require special handling for alternative resolvers like coursier (http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22sbt-scalafmt-coursier%22).

The root of this issue seems to be that the combination of

1.  sbt 0.13 being stuck on 2.10 and
2. the sbt 0.13 library management api is lacking, see https://github.com/sbt/sbt/issues/2879

I've given up on the synthetic project approach used in sbt-scalafix 0.4 because it seems to only work in simple builds. 

Have you tried sbt-scalafix from the current master? sbt-ensime uses coursier in the same way.

jonas:
Sorry, I didn't know this was how it originally worked (modulo forking a new JVM). Apart from the sbt regression I didn't find any problems with projects using Scala 2.10 so far.

olafurpg:
There's a lot of great stuff in this PR. In particular, I think CliWrapperPlugin can be replaced with the `runner` parts. One thing I've wanted is the ability to invoke scalafix once for all modules, that will allow more global dead code analysis. Currently, it only runs per-project.

I will keep this PR in mind while iterating further on sbt-scalafix. The only part I'm unsure about is using hidden custom configurations to fetch artifacts. I prefer to use coursier for that, even if that approach has its problems too.

PR #275, opened 2017-08-09
Add PatchOps.renameSymbol.

olafurpg:
This is a helpful patch to rename a method. The implementation piggy
backs on the ReplaceSymbol since the two are very similar.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
@@ -6,6 +6,7 @@ import scala.meta.tokens.Tokens
 import scalafix.syntax._
 import scalafix.config.ScalafixConfig
 import scalafix.config.ScalafixReporter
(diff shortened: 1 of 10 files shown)


scalafix-core/shared/src/main/scala/scalafix/rewrite/RewriteCtx.scala
```diff
@@ -77,5 +78,10 @@ case class RewriteCtx(tree: Tree, config: ScalafixConfig) extends PatchOps {
   def replaceSymbol(fromSymbol: Symbol.Global, toSymbol: Symbol.Global)(
       implicit mirror: Mirror): Patch =
     TreePatch.ReplaceSymbol(fromSymbol, toSymbol)
+  def renameSymbol(fromSymbol: Symbol.Global, toName: String)(
+      implicit mirror: Mirror): Patch =
+    TreePatch.ReplaceSymbol(
```
olafurpg: We may later want to introduce proper node for renames instead of re-using ReplaceSymbol. This won't affect the public API however since it's an implementation detail.

PR #287, opened 2017-08-14
Integrate sbtfix into scalafix

olafurpg:
This commit integrates the sbt 1.0 migration rewrites into scalafix. This means that users can use sbt-scalafix for both regular project sources as well as sbt build file sources. To run semantic rewrites on build sources, users will need to install the sbthost compiler plugin globally. Here is what I have setup on my laptop

```scala
// ~/.sbt/0.13/build.sbt
libraryDependencies ++= {
  val sbthost = "org.scalameta" % "sbthost-nsc" % "0.3.1" cross CrossVersion.full
  val isMetabuild = {
    val p = thisProject.value
    p.id.endsWith("-build") && p.base.getName == "project"
  }
  if (isMetabuild) compilerPlugin(sbthost) :: Nil
  else Nil
}
```

One neat effect of this change is that it's possible to write custom rewrites for sbt build sources. `> scalafixBuild github:foo/bar/v2` will work just like for `> scalafix` :D

### DIFF
--- a/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -10,6 +10,7 @@ import sbt.plugins.JvmPlugin
 import scalafix.internal.sbt.ScalafixCompletions
 import scalafix.internal.sbt.ScalafixJarFetcher
 import sbt.Def
+import sbt.Def
 
 object ScalafixPlugin extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
@@ -55,6 +56,19 @@ object ScalafixPlugin extends AutoPlugin {
     cliWrapperMainClass := "scalafix.cli.Cli$",
     scalafixEnabled := true,
     scalafixVerbose := false,
+    scalafixBuild := Def.inputTaskDyn {
+      val baseDir = (baseDirectory in ThisBuild).value
+      val sbtDir: File = baseDir./("project")
+      val sbtFiles = baseDir.*("*.sbt").get
+      scalafixTaskImpl(
+        scalafixParser.parsed,
+        Seq.empty[String],
+        sbtDir +: sbtFiles,
+        "sbt-build",
+        streams.value
+      )
+    }.evaluated,
+    aggregate.in(scalafixBuild) := false,
     scalafixSourceroot := baseDirectory.in(ThisBuild).value,
     scalafixVersion := Versions.version,
     scalafixSemanticdbVersion := Versions.scalameta,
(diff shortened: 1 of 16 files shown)


olafurpg:
This PR does not have test for `scalafixBuild`. I tested it manually locally on twitter/scalding  and it seems to do the right thing. I will to add some tests tomorrow, just wanted to open the PR to share progress.

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -124,69 +127,109 @@ object ScalafixPlugin extends AutoPlugin {
     }
   )
   lazy val scalafixTaskSettings = Seq(
+    scalafixBuild := Def.inputTaskDyn {
+      val extracted = Project.extract(state.value)
+      val root = extracted.rootProject(extracted.structure.root)
+      val current = thisProject.value.id
+      if (root != current) {
+        // TODO(olafur) come up with less hacky approach to avoid aggregating
```
jvican: Have you tried `aggregate in scalafixBuild := false`?

olafurpg:
This PR is ready for review :)

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -50,6 +56,19 @@ object ScalafixPlugin extends AutoPlugin {
     cliWrapperMainClass := "scalafix.cli.Cli$",
     scalafixEnabled := true,
     scalafixVerbose := false,
+    scalafixBuild := Def.inputTaskDyn {
```
Duhemm: Would it be worth it to check the `scalacOptions` to make sure that `sbthost` is present and fail early if not? What happens if someone runs `scalafixBuild` without the compiler plugin at the moment?

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -10,12 +10,16 @@ import sbt.plugins.JvmPlugin
 import scalafix.internal.sbt.ScalafixCompletions
 import scalafix.internal.sbt.ScalafixJarFetcher
 import sbt.Def
+import sbt.Def
```
olafurpg: Removed.

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
olafurpg: Currently scalafix will run and semantic rewrites will get no symbols/denotations, which may be confusing for users. Ideally, in the future, we can trigger a reload with the compiler plugin enabled to guarantee that the .semanticdb files are in sync with the sources.

I'm not sure if we can query the libraryDependencies of metabuilds from here. When invoked, the task is typically scoped to the root project. I'll add a note to explain.

PR #289, opened 2017-08-15
Fail fast on missing semanticdb.

### LINKED ISSUE #264: Silent failures during semanticdb loading
I was running a semantic scalafix rewrite on the stdlib and got this cryptic error message

```scala
java -jar ~/scalafix/scalafix-cli/target/scala-2.12/scalafix.jar -r NoAutoApply --verbose
Automatic classpath=/home/geirsson/scala/src/library/target
error: Type mismatch;
  found    : String (value: "NoAutoApply")
  expected : scalafix.rewrite.Rewrite

Unknown input 'NoAutoApply'. Expected one of: NoValInForComprehension, RemoveXmlLiterals, VolatileLazyVal, ProcedureSyntax, ExplicitUnit, DottyVarArgPattern.

Type mismatch;
  found    : String (value: "NoAutoApply")
  expected : scalafix.rewrite.Rewrite
```

The semanticdb files are there. I suspect an error happened while loading them that disappeared.

olafurpg:
Previously, scalafix-cli would fail silently if running a semantic
rewrite with no semanticdb files. Now it fails fast.
Fixes #264

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
+++ b/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
@@ -21,17 +21,17 @@ import scala.meta.io.AbsolutePath
 import scala.meta.sbthost.Sbthost
 import scala.util.Try
 import scala.util.control.NonFatal
-import scalafix.internal.config.Class2Hocon
-import scalafix.internal.config.FilterMatcher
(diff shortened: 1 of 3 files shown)


olafurpg:
FYI @Duhemm This PR makes scalafixBuild fail fast when sbthost is not enabled.

Issue #241, opened 2017-07-03
Escape hatch / disable rewrites over regions


olafurpg:
Currently, scalafix runs all rewrites with no option to opt-out of rewrites over particular regions. It would be nice if scalafix offered an "escape hatch" to disable rewrites for particular regions. It could support a syntax like described by @ShaneDelmore in #238 (inlined below)

```
//scalafix: off. –disables scalafix for the rest of the file until
//scalafix: on is found
//scalafix: ignore –don’t scalafix this line (or method?) at all
//scalafix: ignore.RemoveUnused –Only ignore the RemoveUnused rewrite but apply any other matching rewrites.
//scalafix: ignore.{ RemoveUnused, ExplicitReturnTypes } –handle multiple ignores with a syntax similar to imports maybe?
```

olafurpg:
Related discussion in https://github.com/scalacenter/scalafix/issues/238#issuecomment-312446114

olafurpg:
Escape hatches for rewrites are blocked by #368. This should be quite simple to implement for linters, since each linter message is attached to a single position.

olafurpg:
Fixed by @MasseGuillaume in #415

PR #291, opened 2017-08-16
Add support for linters

olafurpg:
Just opening this PR to show progress
- [x] clean up commit history
- [x] move scalafix.internal.rewrite to separate PR (see #294 )
- [x] improve RewriteName, use List[Name] and Name has fields like `.deprecated`.
- [x] more linter tests
- [x] LintCategory.Info
- [x] remove akka from build import?
- [x] scalafix.lint docstrings

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -292,7 +292,6 @@ lazy val testsInput = project
     resolvers += Resolver.bintrayRepo("allanrenucci", "maven"),
     libraryDependencies ++= testsDeps
   )
-  .disablePlugins(ScalafixPlugin)
   .dependsOn(testsShared)
 
 lazy val testsOutput = project
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,6 +15,5 @@ addSbtPlugin(
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.18")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.7.0")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.14")
-addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.0-M4")
 
 libraryDependencies += "org.scala-sbt" % "scripted-plugin" % sbtVersion.value


ShaneDelmore:
This looks great so far 👍   New lint rules look very straightforward to implement.

olafurpg:
Do you think it's confusing that a linter `extends Rewrite`? I've considered if we a better name.

ShaneDelmore:
That’s tricky as sometimes we will want a lint or be a rewrite (or have a rewrite associated) but not always.  I’m heading off to bed but if I think of a better name I will let you know.  In some ways this reminds me of a more powerful version of a classic regex. You have your capture groups, and optionally your replacement but that still doesn’t immediately suggest a better name to me.

It’s possible it could be generalized to some sort of Matcher with an optional Lint and an optional Rewrite to convey the idea that we always have to match a pattern of code, then we can either warn, rewrite, or both.

olafurpg:
Borrowing terminology from pylint, we could use `Checker` (https://pylint.readthedocs.io/en/latest/how_tos/custom_checkers.html). I personally prefer sticking to `Rewrite` however.

gabro:
I agree that `Rewrite` is confusing for rules that check only.

The confusion arises by the strange path we're taking: usually projects start as linters and then add some fixing capabilities, but scalafix started from the fixing part and we're retrofitting the linting.

Instead of having a unique concept, I would rather see two different entities: a `Checker`, `Rule`, or whatever and a `Rewrite` associated to it, but I'm not sure we have a clear way of drawing the line between the two.

WDYT?

olafurpg:
I am not sure about separating rewrites from linters since that will leave us without a good name for rewrites that emit linter messages (like RemoveXmlLiterals in this PR). The API is identical for both linters and rewrites, so I would edge towards using a unified name. Maybe `scalafix.Rule`?

gabro:
`Rule` sounds good. It may be useful (for future integrations or even for the readme) to include a flag that specifies whether this is a pure linting rule or whether it includes fixes as well.

olafurpg:
Agreed, I'm leaning towards adding more metadata methods on `Rewrite` (or `Rule` if we go that route) such as

```
Rewrite.description: String
Rewrite.lintMessages: List[LintMessage]
...
```

gabro:
OT: Preparing a talk on Scalafix is very entertaining these days. I need a Slidefix tool :P

olafurpg:
I opened https://github.com/scalacenter/scalafix/issues/293 to continue our discussion on renaming `Rewrite`

olafurpg:
> OT: Preparing a talk on Scalafix is very entertaining these days. I need a Slidefix tool :P

:rofl: 

I apologize for the endless changes, I'm not good at sitting down and designing thoroughly before writing code.

gabro:
> I apologize for the endless changes, I'm not good at sitting down and designing thoroughly before writing code.

No apologies needed. I would be more worried about an exciting project slowing down in its early stages because of minor compatibility issues (just don't rename stuff 10 minutes before my talk!  I'll keep an eye on you 👀 )

olafurpg:
All right, I think this PR is ready now :) 

I ended up doing a lot of testing infrastructure changes in this PR. This may pollute the actual diff but it gives me a lot more confidence the changes work as intended. I am super happy with the CliTest rewrite and the lint message assertions via comments in scalafix-testkit are quite nice too.

The main data structures went through some renaming in the final commits while I was writing docs

- LintCategory: a certain kind/type of lint message
- LintMessage: a (message, position, LintCategory) triplet.

Escape hatches #241 are still a TODO, but those will have to wait for another PR.

Issue #254, opened 2017-07-17
ExplicitReturnTypes doesn't fully qualify enough


olafurpg:
After running ExplicitReturnTypes in the Akka repo I got the errors below. The full diff is available here https://gist.github.com/olafurpg/82b6fa8261ba2f1dc147dccd595098ab

```scala
[error] [1] /akka-actor/src/main/scala/akka/actor/ActorPath.scala:346:
[error]     type mismatch;
[error]      found   : java.lang.StringBuilder
[error]      required: scala.collection.mutable.StringBuilder
[error]         val sb: StringBuilder = buildToString(new JStringBuilder(length + 12), length, 0, _.toString)
[error]                                              ^
[error] [2] /akka-actor/src/main/scala/akka/actor/ActorPath.scala:347:
[error]     type mismatch;
[error]      found   : scala.collection.mutable.StringBuilder
[error]      required: java.lang.StringBuilder
[error]         appendUidFragment(sb).toString
[error]                           ^
[error] [3] /akka-actor/src/main/scala/akka/actor/ActorPath.scala:366:
[error]     type mismatch;
[error]      found   : java.lang.StringBuilder
[error]      required: scala.collection.mutable.StringBuilder
[error]         val sb: StringBuilder = buildToString(new JStringBuilder(length + 12), length, diff, _.toStringWithAddress(addr))
[error]                                              ^
[error] [4] /akka-actor/src/main/scala/akka/actor/ActorPath.scala:367:
[error]     type mismatch;
[error]      found   : scala.collection.mutable.StringBuilder
[error]      required: java.lang.StringBuilder
[error]         appendUidFragment(sb).toString
[error]                           ^
[error] [5] /akka-actor/src/main/scala/akka/actor/ActorSelection.scala:101:
[error]     type mismatch;
[error]      found   : java.lang.StringBuilder
[error]      required: scala.collection.mutable.StringBuilder
[error]         val builder: StringBuilder = new java.lang.StringBuilder()
[error]                                      ^
[error] [6] /akka-actor/src/main/scala/akka/actor/ActorSelection.scala:131:
[error]     type mismatch;
[error]      found   : java.lang.StringBuilder
[error]      required: scala.collection.mutable.StringBuilder
[error]         val builder: StringBuilder = new java.lang.StringBuilder()
[error]                                      ^
[error] [7] /akka-actor/src/main/scala/akka/actor/Address.scala:53:
[error]     type mismatch;
[...]

PR #310, opened 2017-08-22
Fix #309, handle trickier symbolic identifiers.

### LINKED ISSUE #309: ExplicitReturnTypes rewrite invalid pattern in type ascription
Before

```scala
implicit val materializer_ = ActorMaterializer()
```

After

```scala
implicit val materializer_: akka.stream.ActorMaterializer = ActorMaterializer()
```

This is a syntax error.

```scala
 val foo_: Int = 1
<console>:1: error: illegal start of simple pattern
val foo_: Int = 1
              ^

scala> val foo_ : Int = 1
foo_: Int = 1
```

the `_` and the `:` must be separated by a space

olafurpg:


dragos:
LGTM! Thanks for the very quick reaction!

dragos:
@olafurpg how can I try the fixes? do you publish SNAPSHOTS?

olafurpg:
I just pushed the tag to publish M7 with this fix, it usually takes ~30 minutes to run via CI and sync. publishLocal until then ;)

We don't publish SNAPSHOT, we used to publish to bintray on every merge but no longer. It should probably be quite easy to configure the CI to publish a sonatype SNAPSHOT on every merge into master, it already know how to publish non-SNAPSHOTs on tag push 😄

olafurpg:
FYI, https://github.com/scalameta/scalameta/pull/1076 will soon make ExplicitReturnTypes more robust. Currently it randomly shortens magic names like List/StringBuilder/Long resulting in bugs like #254

dragos:
Cool, I'll try M7. BTW, what's the status on trailing commas? It seems like it should be supported, but I see errors.

olafurpg:
> what's the status on trailing commas?

trailing commas have been supported in the scalameta parser since february, it could be that scalafix is using the wrong dialect. Can you report an issue with reproduction?

Issue #318, opened 2017-08-31
Remove cyclic source dependencies in codebase


olafurpg:
I ran https://github.com/lihaoyi/acyclic on the scalafix codebase and it reported ~25 cyclic dependencies 😞 I tried to see if I could fix the cycles easily but it seems it will require a bit of a redesign. I'm opening up this issue to keep track of the major refactorings necessry to remove cyclic source dependencies

- [ ] Rewrite => Config => Rewrite
- [ ] Patch => RewriteCtx => Patch

olafurpg:
Closing due to inactivity.

PR #320, opened 2017-08-31
Harden ExplicitReturnTypes

olafurpg:
Prefixing with `_root_.` is a necessary evil. with `unsafeShortenName`, users have an opt-out solution so I think it's fine to go all-in on safety by default. I think it's most important that the rewrite produces compiling code.

### DIFF
--- a/readme/Rewrites.scalatex
+++ b/readme/Rewrites.scalatex
@@ -34,17 +34,11 @@
         Inserts type annotations only for implicit definitions by default. To configure the
         rewrite to run on non-implicit definitions
         @config
-          explicitReturnTypes {
-             memberKind = [
-               Def
-               Var
-               Val
-             ]
-             memberVisibility = [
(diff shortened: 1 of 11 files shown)


olafurpg:
Here is a diff from running ExplicitReturnTypes on akka https://gist.github.com/olafurpg/1a6ce127350930393adc43b1a4d3342b with `unsafeShortenNames=true`. We are blocked by https://github.com/scalameta/scalameta/issues/1099 to support short names on defs.

xeno-by:
I'll take care of 1099 today.

olafurpg:
Awesome, thank you @xeno-by !

olafurpg:
Publishing v0.5.0-M8 with this fix, hopefully the latest milestone release. There are parts of the public API that I would like to refactor, see #318, but I think those will have to wait for v0.6

PR #325, opened 2017-09-06
The great rename

olafurpg:


olafurpg:
The great rename PR got moved to #325

PR #332, opened 2017-09-07
Align the readme to the website

gabro:
I think it's nicer to have this consistent with the website. This is not a Dotty-only tool anymore :)

Also, I don't know why the scaladex thinks `0.3.4+44-1091b539` is the latest version :/


olafurpg:
As for the weird version number, I suspect scaladex is using the version from bintray. We should really remove those artifacts from bintray.

PR #333, opened 2017-09-07
Implement NoInfer linter.

olafurpg:
This is the first scalafix linter rule. There are still a few rough corners around how to display carets around synthetics but the api is already usable. It was a great exercise to implement this linter since I found a few necessary parts missing in testkit to make the experience nice.

### DIFF
--- a/scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
+++ b/scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
@@ -9,7 +9,6 @@ import org.scalameta.logger
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.FunSuite
 import org.scalatest.exceptions.TestFailedException
-import scala.meta.internal.inputs.XtensionPositionFormatMessage
 import scala.util.matching.Regex
 import org.langmeta.internal.ScalafixLangmetaHacks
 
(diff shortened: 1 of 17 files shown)


scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
```diff
@@ -95,7 +96,7 @@ abstract class SemanticRuleSuite(
       val key = lintMessages.head.category.key(rule.name)
       val explanation =
         s"""|To fix this problem, suffix the culprit lines with
-            |   // scalafix: $key
+            |   // assert: $key
```
gabro: how does the `// assert` comment work?

olafurpg: You can assert which linter message is reported at which specfic lines https://github.com/scalacenter/scalafix/blob/master/scalafix-tests/input/src/main/scala/test/NoInfer.scala#L7

The test fails if
- the linter category doesn't match,
- a message is not reported on that line
- a message is reported at a line without an assert

gabro: Nice! This is only for testkit right?

olafurpg: Yes, this is only testkit. Escape hatches for users are yet unimplemented https://github.com/scalacenter/scalafix/issues/241

Issue #349, opened 2017-09-12
scalafixTest and sbtfixTest are missing in sbt-scalafix


olafurpg:
See https://contributors.scala-lang.org/t/scalafix-v0-5-released/1101/2

The cli already supports --test, so this should be easy to support.

olafurpg:
Fixed in #350

PR #350, opened 2017-09-13
Introduce scalafixTest and sbtfixTest sbt tasks

Krever:
I have implemented two tasks mentioned in #349 

I'm not 100% happy with the implementation itself but it fits into my "good enough" category. Let me know what to improve :)

I have modified the only sbt test that was present but I can put this logic into a separate one. Let me know whats your preference.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,9 @@ $ sbt scalafix-tests/it:test # Slow integration tests running rules on
                              # open source projects.
 
 # For SBT plugin
-$ sbt scalafix-sbt/test     # publishes modules locally and runs scripted (slow)
+$ sbt scalafix-sbt/it:test  # publishes modules locally and runs scripted (slow)
+                            # for hassle-free execution run with version specified
+                            # e.g. -Dscalafix.version=0.5-SNAPSHOT
 $ sbt scalafix-sbt/scripted # only run scripted tests (still slow, but skips
                             # publishLocal for core/cli/nsc scalafix modules)
 ```
--- a/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -17,9 +17,14 @@ object ScalafixPlugin extends AutoPlugin {
   object autoImport {
     val scalafix: InputKey[Unit] =
       inputKey[Unit]("Run scalafix rule.")
+    val scalafixTest: InputKey[Unit] =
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -32,7 +32,9 @@ $ sbt scalafix-tests/it:test # Slow integration tests running rules on
                              # open source projects.
 
 # For SBT plugin
-$ sbt scalafix-sbt/test     # publishes modules locally and runs scripted (slow)
+$ sbt scalafix-sbt/it:test  # publishes modules locally and runs scripted (slow)
+                            # for hassle-free execution run with version specified
+                            # e.g. -Dscalafix.version=0.5-SNAPSHOT
```
olafurpg: 👍 Thank you!

scalafix-sbt/src/sbt-test/sbt-scalafix/cross-build/test
```diff
@@ -1,8 +1,12 @@
 # compile should not affect scalafix
 > compile
+-> scalafixTest ProcedureSyntax
 > scalafix ProcedureSyntax
+> scalafixTest ProcedureSyntax
 # Should fail because 2.10 has no semanticdb
 # Other >2.10 projects should succeed
 -> scalafix
+-> sbtfixTest Sbt1
```
olafurpg: Nice!

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -54,7 +60,29 @@ object ScalafixPlugin extends AutoPlugin {
...
+        scalafixVersion.value
+      )
+      if (jars.isEmpty) {
+        throw new MessageOnlyException("Unable to download scalafix-cli jars!")
+      }
+      jars
+    }
+  )
+
+  private def sbtfixImpl(runAsTest: Boolean = false) = {
```
olafurpg: Instead of `runAsTest: Boolean`, can  we propagate `extraOptions: Seq[String]`? That would make it easier to extend for other flags in the future.

Krever: Sure! Would you like the same implementation for `scalafixTest` ?

olafurpg: Yes, that would be great

Krever: Fixed

Issue #214, opened 2017-06-12
Rewrite automatic unit insertion


gabro:
```scala
// before
val x: Option[Unit] = Option()

// after
val x: Option[Unit] = Option(())
```

As discussed in #208, we should include this into `NoAutotupling`, and rename it to `NoAdaptedArgs` in order to match the scalac flag that detects both cases.

The implementation should be simple enough, since the compiler provides detailed messages for this case too.

olafurpg:
This would be a nice rewrite. Question is whether we should bundle this rewrite with NoAutoTupling or keep it separate. I think it might make more sense to keep it separate as `NoUnitInsertion`. We can compose the two rewrites 

```scala
val NoAdaptedArgs = NoUnitInsertion.andThen(NoAutoTupling)
```

olafurpg:
For rewrites like this it might be useful to have some kind of `PositionPatch` that works on positions, that would be an even lower-level patch than token patches. What do you think?

gabro:
> We can compose the two rewrites
> 
> ```scala
> val NoAdaptedArgs = NoUnitInsertion.andThen(NoAutoTupling)
> ```

I like that! Composition FTW 👍 

> For rewrites like this it might be useful to have some kind of PositionPatch that works on positions

Yes, especially for rewrites working with compiler messages, some position-based operations would be extremely useful!

gabro:
I also think this would be a good first contribution, taking `NoAutotupling` as reference implementation.

olafurpg:
Fixed in #352

PR #352, opened 2017-09-15
No unit insertion as an addition of no auto tupling

insdami:
This PR is for #214.
I'm not quite sure how to compose it with `NoAutoTupling` to run it using `NoAdaptedArgs` as you propose on the comments.

Have a look at it and let me know, please :)


gabro:
I'll take a better look from a computer, but it looks 👌

I'm thinking we can probably merge this into NoAutoTupling and just rename the whole thing NoArgumentListAdaptation or something along those lines.

WDYT @olafurpg?

gabro:
Also many thanks @insdami for picking this up, it's a really nice addition to Scalafix!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoUnitInsertion.scala
```diff
@@ -0,0 +1,25 @@
...
+import scalafix.rule.{RuleCtx, SemanticRule}
+
+case class NoUnitInsertion(index: SemanticdbIndex)
+    extends SemanticRule(index, "NoUnitInsertion") {
+
+  private[this] def insertUnit(ctx: RuleCtx, t: Term.Apply): Patch =
+    ctx.addRight(t.tokens.init.last, "()")
+
+  override def fix(ctx: RuleCtx): Patch = {
+    val adaptations = index.messages.toIterator.collect {
```
olafurpg: This will iterate through all reported messages in the workspace/project on every source file, we should move it outside of the `fix` method.

insdami: Sure, I'll do the same in NoAutoTupling

Issue #335, opened 2017-09-07
Rename configuration explicitReturnTypes to explicitResultTypes


MasseGuillaume:
The rule ExplicitReturnTypes is now deprecated, but we still use it as a configuration key:

```
explicitReturnTypes.memberKind = [Val, Def, Var]
explicitReturnTypes.memberVisibility = [Public, Protected]
```

olafurpg:
Fixed in #354

PR #354, opened 2017-09-15
Renamed configuration explicitReturnTypes to explicitResultTypes

insdami:
Small contribution from Dublin Scala Spree on #335


scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -33,7 +33,7 @@ case class ExplicitResultTypes(
     this(index, ExplicitResultTypesConfig.default)
   override def init(config: Conf): Configured[Rule] =
     config
-      .getOrElse("explicitReturnTypes")(ExplicitResultTypesConfig.default)
+      .getOrElse("explicitResultTypes")(ExplicitResultTypesConfig.default)
```
olafurpg: This will break people's existing configuration, I think we will have to support both names throughout 0.5.x. We currently have no means to emit a deprecation warning here, so no worries about that.

There's no high-level api on `Conf` to fallback on a different key, but a simple workaround is to check if the resulting setting is still default and then try the fallback name.

olafurpg: Oh, and one more comment, see https://github.com/scalacenter/scalafix/pull/355/files#r139296465 what do you think about making it "ExplicitResultTypes" with capital E in the beginning?

insdami: @olafurpg cool! I just pushed a new commit. To support the deprecated config I've used `extraNames` in `Config#getOrElse`. Let me know if there is something else that could be improved.

PR #355, opened 2017-09-15
Fix #266: Extended NoInfer adding extraSymbols field

### LINKED ISSUE #266: ShowSynthetics
It would be nice to have a rewrite that expands all sugars so that inferred type parameters/implicits are made explicit.

```scala
// before
  val x = List(1).map(_ + 1)
  x + "foobar"
// after
  val x = List.apply[Int](1).map[Int, List[Int]](_ + 1)(scala.collection.immutable.List.canBuildFrom[Int])
  scala.Predef.any2stringadd[List[Int]](x) + "foobar"
```

bmatthews68:
Done in Scala Center's spree in Dublin.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
@@ -1,0 +1,22 @@
+package scalafix.internal.config
+
+import metaconfig._
+import org.langmeta._
+
+case class NoInferConfig(
+    extraSymbols: List[Symbol] = Nil
+) {
+  implicit val reader: ConfDecoder[NoInferConfig] =
+    ConfDecoder.instanceF[NoInferConfig] { c =>
+        c.getOrElse("extraSymbols")(extraSymbols).map(xs => NoInferConfig(xs))
+    }
+  implicit val symbolReader: ConfDecoder[Symbol] =
+    ConfDecoder.instanceF[Symbol] { c =>
+      c.as[String].map(Symbol.apply)
+    }
+}
+
+object NoInferConfig {
+  val default = NoInferConfig()
+  implicit val reader: ConfDecoder[NoInferConfig] = default.reader
+}
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
@@ -1,25 +1,39 @@
 package scalafix.internal.rule
 
+import metaconfig.{Conf, Configured}
+
 import scala.meta._
+import scalafix.internal.config.NoInferConfig
 import scalafix.lint.LintCategory
 import scalafix.lint.LintMessage
-import scalafix.rule.RuleCtx
-import scalafix.rule.SemanticRule
+import scalafix.rule.{Rule, RuleCtx, SemanticRule}
 import scalafix.util.SemanticdbIndex
 import scalafix.util.SymbolMatcher
 
-case class NoInfer(index: SemanticdbIndex)
+case class NoInfer(index: SemanticdbIndex,
+                   config: NoInferConfig = NoInferConfig.default)
     extends SemanticRule(index, "NoInfer")
     with Product {
-  private val badSymbol = SymbolMatcher.exact(NoInfer.badSymbols: _*)
+  private val badSymbol = SymbolMatcher.exact(NoInfer.badSymbols ++ config.extraSymbols: _*)
+
+  def this(index: SemanticdbIndex) =
+    this(index, NoInferConfig.default)
+
+  override def init(config: Conf): Configured[Rule] =
+    config
+      .getOrElse("noInfer")(NoInferConfig.default)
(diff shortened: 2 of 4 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.langmeta._
+
+case class NoInferConfig(
+    extraSymbols: List[Symbol] = Nil
+) {
+  implicit val reader: ConfDecoder[NoInferConfig] =
+    ConfDecoder.instanceF[NoInferConfig] { c =>
+        c.getOrElse("extraSymbols")(extraSymbols).map(xs => NoInferConfig(xs))
+    }
+  implicit val symbolReader: ConfDecoder[Symbol] =
```
olafurpg: There is a Symbol.Global reader in `scalafix.internal.config` which we should use instead, it handles a few corner cases.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
```diff
@@ -1,25 +1,39 @@
...
     with Product {
-  private val badSymbol = SymbolMatcher.exact(NoInfer.badSymbols: _*)
+  private val badSymbol = SymbolMatcher.exact(NoInfer.badSymbols ++ config.extraSymbols: _*)
+
+  def this(index: SemanticdbIndex) =
+    this(index, NoInferConfig.default)
+
+  override def init(config: Conf): Configured[Rule] =
+    config
+      .getOrElse("noInfer")(NoInferConfig.default)
```
olafurpg: I'm starting to think a good convention is to make the configuration key match exactly with the rule name (instead of lowercasing first character like I've done until now). What do you think?

scalafix-sbt/src/sbt-test/sbt-scalafix/cross-build/.scalafix.conf
```diff
@@ -1,4 +1,8 @@
 rules = [
   ProcedureSyntax
   NoAutoTupling
-]
\ No newline at end of file
+]
+
+NoInfer.extraSymbols = [
```
olafurpg: Is this change necessary?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
```diff
@@ -1,25 +1,39 @@
...
     LintCategory.error(
       """The Scala compiler sometimes infers a too generic type such as Any.
         |If this is intended behavior, then the type should be explicitly type
         |annotated in the source.""".stripMargin
     )
-  override def check(ctx: RuleCtx): Seq[LintMessage] =
-    ctx.index.synthetics.flatMap {
+
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+
```
olafurpg: Is this change necessary?

scalafix-tests/input/src/main/scala/test/NoInfer.scala
```diff
@@ -14,4 +17,6 @@ case object NoInfer {
   }
   case class A()
   List(NoInfer, A())// assert: NoInfer.product
+  case class B()
+  List(B())// assert: NoInfer.b
```
olafurpg: Nice! :+1:

scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
```diff
@@ -0,0 +1,22 @@
+package scalafix.internal.config
+
+import metaconfig._
+import org.langmeta._
+
+case class NoInferConfig(
```
olafurpg: God there is an incredible amount of ceremony involved around adding configuration, we should try to come up with something easier 🤔

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
gabro: 👍 to this

scalafix-sbt/src/sbt-test/sbt-scalafix/cross-build/.scalafix.conf
bmatthews68: No. I will take it out.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
bmatthews68: No. I didn't undo enough when backtracking a change.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoInfer.scala
bmatthews68: I agree. It would feel more intuitive without the lower case.

olafurpg:
Superseded by #398

olafurpg:
Thank you for this PR @bmatthews68 !

PR #356, opened 2017-09-15
depend only on the part of cats actually used

SethTisue:
this is better in general... but also, dbuild doesn't understand the
combo-style dependency, so this is needed for Scala community build
compatibility


olafurpg:
Thank you @SethTisue ! LGTM 👍

PR #365, opened 2017-09-19
Fix #345 Add replaceSymbols(Seq[(String, String)])

### LINKED ISSUE #345: replaceSymbols that accepts a Seq
It's fairly common to have a `Seq` to feed `replaceSymbols` and `: _*` is cumbersome to use.

rberenguel:
Started writing a version with Seq[(String, String)] and passing to the current (String, String)*, but they are the same after type erasure. Added a couple of fake implicits to differentiate them (that works, sure) but actually just dropping the variadic arg is equivalent (or seems to, since all tests in `test` are green)

Ran scalafmt before the PR (should probably be added to the contributing guidelines?)


gabro:
Thank you for the contribution @rberenguel!
I'm not sure we want to drop the vararg variant, but I also wouldn't miss it so much. I agree that the `DummyImplicit` alternative doesn't seem like a wonderful solution to maintain. WDYT @olafurpg?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -96,14 +96,15 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
   def replaceSymbol(fromSymbol: Symbol.Global, toSymbol: Symbol.Global)(
       implicit index: SemanticdbIndex): Patch =
     TreePatch.ReplaceSymbol(fromSymbol, toSymbol)
-  def replaceSymbols(toReplace: (String, String)*)(
```
olafurpg: We will have to keep this option for a while to preserve compatibility. I am fine with using the dummy implicit trick as a workaround throughout v0.5.

olafurpg:
Can you also maybe add `./scalafmt` to CONTRIBUTING? :)

scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala
```diff
@@ -114,6 +114,16 @@ trait PatchOps {
...
       implicit index: SemanticdbIndex): Patch
+
+  /** Helper for calling replaceSymbols without needing _*
+    *
+    * Defers work to replaceSymbols((String, String)*). Needs a dummy implicit
+    * to differentiate from replaceSymbols((String, String)*) after type erasure
+    */
+  implicit val noopSeq: Seq[(String, String)] = Seq(("", ""))
+  def replaceSymbols(toReplace: Seq[(String, String)])(
+      implicit noop: Seq[(String, String)],
```
gabro: You can probably use scala.Predef.DummyImplicit

rberenguel:
Thanks @gabro I totally did not know about this (but makes total sense for it to exist)

CONTRIBUTING.md
```diff
@@ -39,6 +39,10 @@ $ sbt scalafix-sbt/scripted # only run scripted tests (still slow, but skips
                             # publishLocal for core/cli/nsc scalafix modules)
 ```
 
+## Formatting
+
+Be sure to run `scalafmt` (available in the root folder) to ensure code formatting. You can read more about it at http://scalafmt.org
```
olafurpg: 👍

olafurpg:
It seems like adding the new method is binary breaking :/

We will need to come up with another technique to add methods without breaking compatibility on every PR. I will try to come back to this sometime next week.

gabro:
mm, in general overloading are 👍 

> The Java programming language is designed to prevent additions to contracts and accidental name collisions from breaking binary compatibility. Specifically, addition of more methods overloading a particular method name does not break compatibility with pre-existing binaries. The method signature that the pre-existing binary will use for method lookup is chosen by the method overload resolution algorithm at compile time (§15.12.2).

but I think we're hitting some edge case due to how varargs are represented by scalac.

rberenguel:
@gabro @olafurpg that's what I had assumed from MiMa's documentation too:

> A key aspect of MiMa to be aware of is that it only looks for syntactic binary incompatibilities. The semantic binary incompatibilities (such as adding or removing a method invocation) are not considered.

([source](https://github.com/typesafehub/migration-manager/wiki#what-is-the-migration-manager))

It's weird. I wonder if there is any way to hard-test it (as in, is the new binary actually usable). Otherwise this will need to wait until 1.0?

olafurpg:
This change is breaking in the cases when an external module has implemented the PatchOps trait. Providing a default implementation would make it binary compatible in 2.12 but not in 2.11. 

I've had doubts about making PatchOps a trait, this situation makes it more tempting to refactor it into a class or object.

olafurpg:
I think we are safe to add the mima filter to ignore this breaking change, since it only affects clients that extend `RuleCtx`. Can you please update `project/Mima.scala` using the recommended error from mima @rberenguel ? We will probably change the `RuleCtx` trait for 0.6 to regain more control over its evolvement.

rberenguel:
@olafurpg yup, I’ll do ASAP, sorry for the delay

Issue #238, opened 2017-06-30
Remove unused terms


olafurpg:
scalac can report warnings on unused terms like this

```scala
[warn] /Users/ollie/dev/scalafix/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala:211: local val originalLines in method unifiedDiff is never used
[warn]     val originalLines = original.asString.lines.toSeq.asJava
[warn]         ^
[warn] /Users/ollie/dev/scalafix/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala:211: local val originalLines in method unifiedDiff is never used
[warn]     val originalLines = original.asString.lines.toSeq.asJava
[warn]         ^
```

it would be nice if scalafix could automatically remove those definitions. Note that the right-hand side of the definition may contain side-effects so we can't remove those unless they are 100% side-effect free (for example literals)

gabro:
I like this! We could also have an "aggressive" option that removes them indiscriminately, for users that are confident no side-effects happen on the rhs.

olafurpg:
That should be doable.

Note that this rewrite will need to provide some escape hatch for cases like this in the scalafix codebase
```scala
[warn] <macro>:3: private val FilterMatcherReader in class PrintStreamReporter is never used
[warn]   private val FilterMatcherReader = null
[warn]               ^
[warn] one warning found
```
That statement is necessary to shadow another implicit, removing the line will cause a compilation error. What would be a sane escape hatch?

```scala
private val FilterMatcherReader = null // scalafix: ignore
private val FilterMatcherReader = null // This line is necessary because ...
```

gabro:
Ugh, I didn't know you could do something like that :D In that specific case, why isn't `FilterMatcherReader` implicit?

Anyway, a general `scalafix: ignore` would probably be useful in some other cases.

ShaneDelmore:
+1 to scalafix: ignore.  It follows the style of wartRemover suppressWarnings, scalafmt ignore, scalastyle ignore, etc.  All of these tools end up needing an escape hatch sooner or later, ignore seems like the popular way to do it.  Just two weeks ago I tried to run scalafix on sbt codebase and ended up wanting to ignore a bunch of generated files that were in source with a //Do not modify by hand comment on them.

olafurpg:
[...]

PR #367, opened 2017-09-20
#238 Rule to remove unused local val and var

DanielaSfregola:
This PR implements the rule described in #238.

In the specific, it removed unused local vals and vars. If litteral expressions are involved, the whole expression is removed, otherwise only the val definition is removed.

I'll try to send another PR for unused def/trait/classes in the near future.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -306,7 +306,7 @@ lazy val testsInput = project
     scalacOptions += s"-P:semanticdb:sourceroot:${sourceDirectory.in(Compile).value}",
     scalacOptions ~= (_.filterNot(_ == "-Yno-adapted-args")),
     scalacOptions += "-Ywarn-adapted-args", // For NoAutoTupling
-    scalacOptions += "-Ywarn-unused-import", // For RemoveUnusedImports
+    scalacOptions += "-Ywarn-unused", // For RemoveUnusedImports and RemoveUnusedTerms
     logLevel := Level.Error, // avoid flood of compiler warnings
     // TODO: Remove once scala-xml-quote is merged into scala-xml
     resolvers += Resolver.bintrayRepo("allanrenucci", "maven"),
--- a/build.sbt
+++ b/build.sbt
@@ -306,7 +306,8 @@ lazy val testsInput = project
     scalacOptions += s"-P:semanticdb:sourceroot:${sourceDirectory.in(Compile).value}",
     scalacOptions ~= (_.filterNot(_ == "-Yno-adapted-args")),
     scalacOptions += "-Ywarn-adapted-args", // For NoAutoTupling
-    scalacOptions += "-Ywarn-unused", // For RemoveUnusedImports and RemoveUnusedTerms
+    scalacOptions += "-Ywarn-unused-import", // For RemoveUnusedImports
+    scalacOptions += "-Ywarn-unused", // For RemoveUnusedTerms
     logLevel := Level.Error, // avoid flood of compiler warnings
     // TODO: Remove once scala-xml-quote is merged into scala-xml
     resolvers += Resolver.bintrayRepo("allanrenucci", "maven"),
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
@@ -1,0 +1,36 @@
+package scalafix.internal.rule
+
+import scala.meta._
+import scalafix.Patch
+import scalafix.SemanticdbIndex
+import scalafix.rule.RuleCtx
+import scalafix.rule.SemanticRule
+
+case class RemoveUnusedTerms(index: SemanticdbIndex)
+    extends SemanticRule(index, "RemoveUnusedTerms") {
+
+  private val unusedTerms = {
+    val UnusedLocalVal = """local (.*) is never used""".r
+    index.messages.toIterator.collect {
+      case Message(pos, _, UnusedLocalVal(_*))  =>
+        pos
+    }.toSet
+  }
+
+  private def isUnused(defn: Defn) =
+    unusedTerms.contains(defn.pos)
+
+  private def removeLiterals(rhs: Term): String =
+    rhs match {
+      case Lit(_) => ""
+      case r => r.syntax
+    }
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.debugIndex()
+    ctx.tree.collect {
+      case i: Defn.Val if isUnused(i) => ctx.replaceTree(i, removeLiterals(i.rhs))
+      case i: Defn.Var if isUnused(i) => ctx.replaceTree(i, i.rhs.fold("")(removeLiterals))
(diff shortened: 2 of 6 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,36 @@
...
+    unusedTerms.contains(defn.pos)
+
+  private def removeLiterals(rhs: Term): String =
+    rhs match {
+      case Lit(_) => ""
+      case r => r.syntax
+    }
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.debugIndex()
```
gabro: Forgot this in?

scalafix-tests/output/src/main/scala/test/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,13 @@
+package test
+
+object RemoveUnusedTerms {
+
+  def foo {
+
```
gabro: You may want to remove the leading spaces including the newline (like I did in my presentation from yesterday)

gabro: We definitely need a ctx.removeLeading or something

scalafix-tests/output/src/main/scala/test/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,13 @@
...
+object RemoveUnusedTerms {
+
+  def foo {
+    
+    println(5)
+    println(0)
+    println(1)
+  }
+
+  val unused = 0
```
gabro: Why does this stay?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,36 @@
...
+    }.toSet
+  }
+
+  private def isUnused(defn: Defn) =
+    unusedTerms.contains(defn.pos)
+
+  private def removeLiterals(rhs: Term): String =
+    rhs match {
+      case Lit(_) => ""
+      case r => r.syntax
```
gabro: You're potentially modifying the syntax of the original rhs here, I would probably change approach and produce an Option[Patch]

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
DanielaSfregola: Yes, my bad! Removing it

scalafix-tests/output/src/main/scala/test/RemoveUnusedTerms.scala
DanielaSfregola: As far as I could see, this wasn't flagged by the compiler as unused. But I have to admit this looks a bit weird. // cc @olafurpg ?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
olafurpg: I propose that we only remove the necessary tokens (val/identifier/=) and leave the rhs unchanged, there may be other rules touching the tokens inside the rhs.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.debugIndex()
+    ctx.tree.collect {
+      case i: Defn.Val if isUnused(i) => ctx.replaceTree(i, removeLiterals(i.rhs))
```
olafurpg: It requires a bit more work, but as I mention above, I think it makes sense to `ctx.removeToken` only for the tokens that should be removed.

scalafix-tests/output/src/main/scala/test/RemoveUnusedTerms.scala
olafurpg: Yes, I suggested at the spree that we leave the leading space unchanged until we add better utilities to handle this.

scalafix-tests/output/src/main/scala/test/RemoveUnusedTerms.scala
olafurpg: This is a globally accessible val, it could be used by another project.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
olafurpg: Maybe something like `i.tokens.filterNot(i.rhs.tokens.contains).map(ctx.removeToken)` might do the trick.

build.sbt
```diff
@@ -306,7 +306,7 @@ lazy val testsInput = project
     scalacOptions += s"-P:semanticdb:sourceroot:${sourceDirectory.in(Compile).value}",
     scalacOptions ~= (_.filterNot(_ == "-Yno-adapted-args")),
     scalacOptions += "-Ywarn-adapted-args", // For NoAutoTupling
-    scalacOptions += "-Ywarn-unused-import", // For RemoveUnusedImports
+    scalacOptions += "-Ywarn-unused", // For RemoveUnusedImports and RemoveUnusedTerms
```
olafurpg: Seems we need to keep `-Ywarn-unused-import` for 2.11 https://travis-ci.org/scalacenter/scalafix/jobs/278586505

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -0,0 +1,42 @@
...
+      case Message(pos, _, UnusedLocalVal(_*)) =>
+        pos
+    }.toSet
+  }
+
+  private def isUnused(defn: Defn) =
+    unusedTerms.contains(defn.pos)
+
+  private def removeDeclarationTokens(i: Defn, rhs: Term): Tokens = {
+    val startDef = i.tokens.start
```
olafurpg: TIL I have not noticed Tokens.start 😄

DanielaSfregola: Yeah, I struggled a bit with this function as `Tokens.filter` returns a `IndexSeq[Token]` rather than a `Tokens`. Also, the constructor of `Tokens` is private to the package `meta`, so writing something like 
```
i.tokens.filterNot(rhs.tokens.contains)
```
 would have been nicer but unfortunately it didn't quite work :(

olafurpg:
Thank you! 🎆

PR #375, opened 2017-09-23
Cross-build sbt plugin for 0.13 and 1.0

### LINKED ISSUE #374: Crossbuilding with SBT 1.0
Hi,

That would be cool to crossbuild with SBT 1.0.

We currently have a dependency on sbt-dotty as far as I understand in https://github.com/scalacenter/scalafix/issues/281, but if we upgrade to 0.13.16 we should be able to cross build?

Cheers

gabro:
Closes #374

Not sure what I should change to make this automatically published by the CI

### DIFF
--- a/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -146,8 +146,9 @@ object ScalafixPlugin extends AutoPlugin {
 
   lazy val scalafixScalacSettings: Seq[Def.Setting[_]] = Def.settings(
     scalacOptions ++= {
+      val options = scalafixScalacOptions.value
       if (isSupportedScalaVersion.value) {
-        scalafixScalacOptions.value
+        options
       } else {
         Nil
       }
--- a/bin/ci-publish.sh
+++ b/bin/ci-publish.sh
@@ -21,7 +21,7 @@ if [[ "$TRAVIS_SECURE_ENV_VARS" == true && "$CI_PUBLISH" == true ]]; then
   if [ -n "$TRAVIS_TAG" ]; then
     echo "$PGP_SECRET" | base64 --decode | gpg --import
     echo "Tag push, publishing release to Sonatype."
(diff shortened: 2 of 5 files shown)


scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -146,8 +146,9 @@ object ScalafixPlugin extends AutoPlugin {
 
   lazy val scalafixScalacSettings: Seq[Def.Setting[_]] = Def.settings(
     scalacOptions ++= {
+      val options = scalafixScalacOptions.value
       if (isSupportedScalaVersion.value) {
-        scalafixScalacOptions.value
+        options
```
gabro: This is to fix

```
[error] Problem: `scalafixScalacOptions` is inside the if expression of a regular task.
[error]   Regular tasks always evaluate task inside the bodies of if expressions.
[error]
[error] Solution:
[error]   1. If you only want to evaluate it when the if predicate is true or false, use a dynamic task.
[error]   2. Otherwise, make the static evaluation explicit by evaluating `scalafixScalacOptions` outside the if expression.
[error]
[error]         scalafixScalacOptions.value
[error]         ^
[error] one error found
```

gabro: By the way, ❤️ @jvican for that beautiful error message 👆

jvican: Thanks @gabro, lovely comment.

gabro:
I've moved sbt versions to `Dependencies.scala` and tried to upgrade the `ci-publish` script.

The last one is admittedly a shot in the dark, but I don't have a decent way of testing it.

gabro:
mm, I had high hopes but nope, didn't work:
https://travis-ci.org/scalacenter/scalafix/jobs/279044336

Any sbt expert around?

gabro:
@olafurpg I've removed the tag, but left the release notes in draft (https://github.com/scalacenter/scalafix/releases/tag/untagged-0b502c49a33fb0c5374e)

olafurpg:
I'll take a look at this Monday.

PR #378, opened 2017-09-27
Fix #373: addLeft's implementation doesn't match its specification

### LINKED ISSUE #373: addLeft's implementation doesn't match its specification
PatchOps.addLeft claims to add a string to the *first* token of the tree: https://github.com/scalacenter/scalafix/blob/f095438f25a050bebf635415f92f5f7a54cdbd26/scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala#L52

However, the implementation in `RuleCtxImpl` adds the string to the *last* token:

```
def addLeft(tree: Tree, toAdd: String): Patch =
  toks(tree).lastOption.fold(Patch.empty)(addLeft(_, toAdd))
```
https://github.com/scalacenter/scalafix/blob/f095438f25a050bebf635415f92f5f7a54cdbd26/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala#L88

Should that be `headOption` instead of `lastOption`? If so, I'd be happy to send a PR.
If not, IMO we should update the comment to reflect what the code is doing.

abeln:
Return `headOption` instead of `lastOption` in addLeft's implementation.
Additionally, add a unit test, and refactor the syntax tests.

TESTED=unit tests pass

### DIFF
--- a/scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
+++ b/scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
@@ -8,8 +8,18 @@ import scalafix.syntax._
 
 import org.scalatest.FunSuiteLike
 
-class SyntacticRuleSuite(rule: Rule) extends FunSuiteLike with DiffAssertions {
+class SyntacticRuleSuite(defaultRule: Rule = Rule.empty)
+    extends FunSuiteLike
+    with DiffAssertions {
   def check(name: String, original: String, expected: String): Unit = {
+    check(defaultRule, name, original, expected)
+  }
+
+  def check(
+      rule: Rule,
+      name: String,
+      original: String,
+      expected: String): Unit = {
     test(name) {
       import scala.meta._
       val obtained = rule.apply(Input.String(original))
@@ -18,6 +28,10 @@ class SyntacticRuleSuite(rule: Rule) extends FunSuiteLike with DiffAssertions {
   }
 
   def checkDiff(original: Input, expected: String): Unit = {
+    checkDiff(defaultRule, original, expected)
+  }
+
+  def checkDiff(rule: Rule, original: Input, expected: String): Unit = {
     test(original.label) {
       val ctx = RuleCtx(original.parse[Source].get, ScalafixConfig.default)
       val obtained = rule.diff(ctx)
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/SyntacticTests.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/SyntacticTests.scala
@@ -9,24 +9,27 @@ import java.nio.file.Files
 import java.nio.file.Paths
 import scalafix.internal.rule.ProcedureSyntax
 
-class ErrorSuite extends SyntacticRuleSuite(ProcedureSyntax) {
+class ErrorSuite extends SyntacticRuleSuite {
   test("on parse error") {
     intercept[ParseException] {
       ProcedureSyntax.apply(Input.String("object A {"))
     }
   }
 }
-class PatchSuite
-    extends SyntacticRuleSuite(Rule.syntactic("PatchSuite")(ctx =>
-      ctx.addRight(ctx.tree.tokens.find(_.is[Ident]).get, "bba"))) {
+class PatchSuite extends SyntacticRuleSuite {
 
-  val original =
+  val original: String =
     """// Foobar
       |object a {
       |  val x = 2
       |}""".stripMargin
 
+  val addRightRule: Rule = Rule.syntactic("addRight") { (ctx) =>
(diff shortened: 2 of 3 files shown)


abeln:
The binary compatibility test ("mima"?) fails (because I renamed the test suite). Is that a blocker?

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
```diff
@@ -8,16 +8,20 @@ import scalafix.syntax._
 
 import org.scalatest.FunSuiteLike
 
-class SyntacticRuleSuite(rule: Rule) extends FunSuiteLike with DiffAssertions {
```
olafurpg: The migration manager (mima) is complaining because of this change, scalafix-testkit is a published artifact to help scalafix users develop custom rules in repositories outside of scalacenter/scalafix.

I think we can make this change without breaking compatibility by changing this parameter to `defaultRule: Rule = Rule.empty`.

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
```diff
@@ -8,16 +8,20 @@ import scalafix.syntax._
...
-  def check(name: String, original: String, expected: String): Unit = {
```
olafurpg: Forward this to new `check` overload with `defaultRule`

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
```diff
@@ -8,16 +8,20 @@ import scalafix.syntax._
...
+      original: String,
+      expected: String): Unit = {
     test(name) {
       import scala.meta._
       val obtained = rule.apply(Input.String(original))
       assertNoDiff(obtained, expected)
     }
   }
 
-  def checkDiff(original: Input, expected: String): Unit = {
```
olafurpg: Forward this to new `checkDiff` overload with `defaultRule`

scalafix-tests/unit/src/test/scala/scalafix/tests/SyntacticTests.scala
```diff
@@ -9,24 +9,31 @@ import java.nio.file.Files
...
-class ErrorSuite extends SyntacticRuleSuite(ProcedureSyntax) {
+class ErrorSuite extends SyntacticRuleSuite {
   test("on parse error") {
     intercept[ParseException] {
       ProcedureSyntax.apply(Input.String("object A {"))
     }
   }
 }
-class PatchSuite
-    extends SyntacticRuleSuite(Rule.syntactic("PatchSuite")(ctx =>
```
olafurpg: This is a nice cleanup, removing the rule definition in the constructor 👍

scalafix-tests/unit/src/test/scala/scalafix/tests/SyntacticTests.scala
```diff
@@ -9,24 +9,31 @@ import java.nio.file.Files
...
-      ctx.addRight(ctx.tree.tokens.find(_.is[Ident]).get, "bba"))) {
+class PatchSuite extends SyntacticRuleSuite {
 
-  val original =
+  val addRightRule: Rule = Rule.syntactic("addRight") { (ctx) =>
+    ctx.addRight(ctx.tree.tokens.find(_.is[Ident]).get, "bba")
+  }
+
+  val addLeftRule: Rule = Rule.syntactic("addLeft") { (ctx) =>
```
olafurpg: Can we move this definition right next to the next where it's used?

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
abeln: Done.

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
abeln: Done.

scalafix-testkit/src/main/scala/scalafix/testkit/SyntacticRuleSuite.scala
abeln: Done.

scalafix-tests/unit/src/test/scala/scalafix/tests/SyntacticTests.scala
abeln: Done.

abeln:
Thanks for the review. PTAL.

abeln:
The binary compatibility test now passes.

PR #381, opened 2017-09-28
Add resultType extension method on Symbol

gabro:
This seemed like a useful utility so I extracted it from `ExplicitResultTypes`

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/util/DenotationOps.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/util/DenotationOps.scala
@@ -1,0 +1,32 @@
+package scalafix.internal.util
+
+import scala.meta._
+
+object DenotationOps {
+  val defaultDialect =
+    dialects.Scala212.copy(allowMethodTypes = true, allowTypeLambdas = true)
+
+  def resultType(
+      symbol: Symbol,
+      denot: Denotation,
+      dialect: Dialect): Option[Type] = {
+    def getDeclType(tpe: Type): Type = tpe match {
+      case Type.Method(_, tpe) if denot.isDef => tpe
+      case Type.Lambda(_, tpe) if denot.isDef => getDeclType(tpe)
+      case Type.Method((Term.Param(_, _, Some(tpe), _) :: Nil) :: Nil, _)
+          if denot.isVar =>
+        // Workaround for https://github.com/scalameta/scalameta/issues/1100
+        tpe
+      case x =>
+        x
+    }
+    val signature =
+      if (denot.isVal || denot.isDef | denot.isVar) denot.signature
+      else {
+        throw new UnsupportedOperationException(
+          s"Can't parse type for denotation $denot, denot.info=${denot.signature}")
+      }
+    val input = Input.Denotation(signature, symbol)
+    (dialect, input).parse[Type].toOption.map(getDeclType)
+  }
+}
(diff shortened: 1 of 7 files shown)


scalafix-tests/unit/src/test/scala/scalafix/tests/DenotationOpsTest.scala
```diff
@@ -0,0 +1,32 @@
...
+class DenotationOpsTest extends BaseSemanticTest("DenotationOpsTest") {
+
+  test("resultType") {
+    val source = docs.input.parse[Source].get
+    source.collect {
+      case t @ Pat.Var(Name("x")) =>
+        for {
+          symbol <- t.symbol
+          resultType <- symbol.resultType
+        } yield assert(resultType.structure == Type.Name("Boolean").structure)
```
gabro: is there a smarter way to compare trees?

scalafix-core/shared/src/main/scala/scalafix/internal/util/DenotationOps.scala
```diff
@@ -0,0 +1,32 @@
+package scalafix.internal.util
+
+import scala.meta._
+
+object DenotationOps {
+  val defaultDialect =
+    dialects.Scala212.copy(allowMethodTypes = true, allowTypeLambdas = true)
+
+  def resultType(
```
gabro: the name is up to debate. `resultType` is general enough, but it's a bit of a stretch when referring to simple `val`/`var`

olafurpg: Result type is the correct nomenclature from scalac 👍

scalafix-core/shared/src/main/scala/scalafix/syntax/package.scala
```diff
@@ -30,6 +31,10 @@ package object syntax {
     @deprecated("Renamed to denotation", "0.5.0")
     def denotOpt: Option[Denotation] = denotation
     def denotation: Option[Denotation] = index.denotation(symbol)
+    def resultType(dialect: Dialect): Option[Type] =
+      denotation.flatMap(denot =>
+        DenotationOps.resultType(symbol, denot, dialect))
+    def resultType: Option[Type] = resultType(DenotationOps.defaultDialect)
```
olafurpg: Can we only keep this overload and let advanced users with custom dialects call `DenotationOps.resultType(customDialect)`?

scalafix-tests/unit/src/test/scala/scalafix/tests/DenotationOpsTest.scala
olafurpg: There is! `import scala.meta.contrib._; q"a.b".isEqual(q"a.b.")`

scalafix-core/shared/src/main/scala/scalafix/syntax/package.scala
gabro: fine by me

scalafix-tests/unit/src/test/scala/scalafix/tests/DenotationOpsTest.scala
gabro: thanks! I'll fix it

olafurpg:
We should filter out `scalafix.internal.*` from mima like here https://github.com/scalameta/scalafmt/blob/143bff64f24c4db6b53998a927d317ee0d874bcc/project/Mima.scala#L11

We need to upgrade Mima to 0.1.15 https://github.com/scalacenter/scalafix/blob/6b8cccd9438d8786742f02479865e16e434343fe/project/plugins.sbt#L17 which added support for regexes in filters

gabro:
@olafurpg I've added the mima change in this PR, so that we can test it on the spot. Hope you don't mind

PR #382, opened 2017-09-29
Fix #372: addGlobalImport(Importer) is syntactic

### LINKED ISSUE #372: Created semantic patch without extending SemanticRule
I have a rule that seems (to me) like it's syntactic and not semantic: [code](https://pastebin.com/eMPfbbNc).

However, when I test it, I get an exception saying that one of the patches (the added import statement) is semantic. Should it be possible to create a semantic patch without extending SemanticRule?

```
[info] Nullify_Tests:
[info] - scala/ca/uwaterloo/cs/nullify/NullifyTest.scala *** FAILED ***
[info]   scalafix.internal.util.Failure$Unsupported: Semantic patches are not supported without a SemanticdbIndex: List(AddGlobalImport(DenullifyWrapper3571.denullify))
[info]   at scalafix.patch.Patch$.apply(Patch.scala:158)
[info]   at scalafix.testkit.SemanticRuleSuite.$anonfun$runOn$1(SemanticRuleSuite.scala:131)
[info]   at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:12)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FunSuiteLike$$anon$1.apply(FunSuiteLike.scala:186)
[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
[info]   ...
[info]   Cause: java.lang.UnsupportedOperationException: Semantic patches are not supported without a SemanticdbIndex: List(AddGlobalImport(DenullifyWrapper3571.denullify))
[info]   at scalafix.internal.util.Failure$Unsupported.<init>(Failure.scala:29)
[info]   at scalafix.patch.Patch$.apply(Patch.scala:158)
[info]   at scalafix.testkit.SemanticRuleSuite.$anonfun$runOn$1(SemanticRuleSuite.scala:131)
[info]   at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:12)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.FunSuiteLike$$anon$1.apply(FunSuiteLike.scala:1
```

olafurpg:
Review @abeln

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/patch/ImportPatchOps.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/patch/ImportPatchOps.scala
@@ -68,13 +68,13 @@ object ImportPatchOps {
     }
 
   // NOTE(olafur): This method is the simplest/dummest thing I can think of
-  // to convert
   private[scalafix] def superNaiveImportPatchToTokenPatchConverter(
       ctx: RuleCtx,
       importPatches: Seq[ImportPatch])(
       implicit index: SemanticdbIndex): Iterable[Patch] = {
     val allImports = ctx.tree.collect { case i: Import => i }
     val allImporters = allImports.flatMap(_.importers)
+    lazy val allImportersSyntax = allImporters.map(_.syntax)
     val allImportees = allImporters.flatMap(_.importees)
     val allNamedImports = allImportees.collect {
       case Importee.Name(n) if index.names.contains(n.pos) =>
@@ -116,7 +116,8 @@ object ImportPatchOps {
           isAlreadyImported += symbol
           SymbolOps.toImporter(symbol).toList
         case TreePatch.AddGlobalImport(importer)
-            if !allImporters.exists(_.syntax == importer.syntax) =>
+            // best effort deduplication for syntactic addGlobalImport(Importer)
+            if !allImportersSyntax.contains(importer.syntax) =>
           importer :: Nil
         case _ => Nil
       }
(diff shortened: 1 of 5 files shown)


abeln:
LGTM. Thanks for fixing this!
Is there a reason why we can't use the type system to prevent this type of error (e.g. remove the `semanticIndex` option from `Ctx`)? Is it related to the fact that we want to be able to combine syntactic and semantic patches?

olafurpg:
> Is there a reason why we can't use the type system to prevent this type of error

The implicit SemanticdbIndex parameter on ctx.addGlobalImport is used to statically guard against using semantic from syntactic rules. The index is not used in the method implementation however, it's not included as a field in the underlying Patch adt for several reasons
- then we need to validate that all of the Patches have the same index
- we only need the index to "interpret" the Patches into diffs, not when building them

I admit I'm still not very happy with the current design. I feel like there should be a much cleaner way to model things.

> remove the semanticIndex option from Ctx

Can you elaborate?

>  Is it related to the fact that we want to be able to combine syntactic and semantic patches?

A lot of the complications in the scalafix api design are caused by the desire to abstract over syntactic/semantic rules. The fact that the semantic API has gone through several revisions has not helped either, but I suspect it will be stable onwards.

PR #384, opened 2017-09-29
handle coursier's --standalone classloader

xeno-by:
When coursier created a standalone executable for Scalafix, then the app
is run in an unusual classloader. The thing is a URLClassLoader, but
urls of the individual jars are  weird - their getProtocol is not file,
but bootstrapsomethingsomething, and their getPath doesn't point to
anything on the file system.

This is not at all a problem for running Scalafix, but it becomes an
issue when Scalafix is doing runtime compilation to load rules using
the file: protocol. The Scala compiler instance that's spun up for
runtime compilation doesn't understand those weird urls, so compilation
fails.

The proposed patch fixes the compilation failure by extracting the jars
into temporary files on disk, and then passing paths to those files to
the Scala compiler instance.

### DIFF
--- a/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
+++ b/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
@@ -63,9 +63,22 @@ class Compiler() {
   settings.outputDirs.setSingleOutput(target)
   getClass.getClassLoader match {
     case u: URLClassLoader =>
-      settings.classpath.value = u.getURLs
-        .map(x => URLDecoder.decode(x.getPath, "UTF-8"))
-        .mkString(File.pathSeparator)
+      val paths = u.getURLs.toList.map(u => {
+        if (u.getProtocol.startsWith("bootstrap")) {
+          import java.io._
+          import java.nio.file._
+          val stream = u.openStream
+          val tmp = File.createTempFile("bootstrap-" + u.getPath, ".jar")
+          Files.copy(
+            stream,
+            Paths.get(tmp.getAbsolutePath),
+            StandardCopyOption.REPLACE_EXISTING)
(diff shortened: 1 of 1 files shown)


scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
```diff
@@ -63,9 +63,19 @@ class Compiler() {
...
+          import java.nio.file._
+          val stream = u.openStream
+          val tmp = File.createTempFile("bootstrap-" + u.getPath, ".jar")
+          Files.copy(stream, Paths.get(tmp.getAbsolutePath), StandardCopyOption.REPLACE_EXISTING)
+          tmp.getAbsolutePath
+        } else {
+          URLDecoder.decode(u.getPath, "UTF-8")
+        }
+      })
+      settings.classpath.value = paths.mkString(File.pathSeparator)
```
olafurpg: I was thinking, is it possible to pass in a `Classloader` instead of `classpath: String`?

xeno-by: Nope.

The Scala compiler needs to traverse the entire classpath in order to correctly load symbols. A generic `ClassLoader` doesn't provide this functionality.

We have tried this in `ReflectGlobal` that takes a `ClassLoader` (it's used inside `Toolbox` which is part of our runtime reflection framework) and ran into hard- or impossible-to-fix problems on pretty common Scala code. After spending some time trying to fix those problems, we decided to abandon that direction of research.

olafurpg: I take your word for it! It's a shame.

xeno-by:
Formatting fixed.

PR #385, opened 2017-10-01
Fix broken links

DavidDudson:
Clearly broke when crossbuilding was introduced.


olafurpg:
Thank you @DavidDudson !

PR #388, opened 2017-10-06
Fix #387, emit linter warnings from cli.

### LINKED ISSUE #387: Scalafix-CLI does not show linter messages
I have been tinkering with scalafix-cli, and although when I developed the rule in `scalafix-testkit` all the linter messages were showing, when I run the tool the following happens:

- `fix` shows the `println` statements, but not the linter messages.
- `check` does not show any output.

I used the following code:
https://gist.github.com/blorente/7c22bda345ee158250dbaf034414f4b5

olafurpg:
This commit changes the cli to
- use the correct OutputStream for emitting lint messages
- call both fix and check instead of only fix

Review @blorente

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
+++ b/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
@@ -80,11 +80,15 @@ sealed abstract case class CliRunner(
       display.taskProgress(msg, progress)
       code
     }
+    val exit: ExitStatus = {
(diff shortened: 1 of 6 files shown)


blorente:
Sory it took me so long, not quite used to sbt yet. 

It works for me, now both check and fix display the println, and check displays the linter messages.

olafurpg:
Thank you for the review @blorente Please considering updating CONTRIBUTING.md if you think it could benefit from more information

PR #386, opened 2017-10-01
Switch website to microsite

gabro:
Still WIP, but looking good :)

Here's a deploy of the current version (which I'll keep updated as we go) https://gabro.github.io/scalafix/

A few notes:

Todo/Wishlist:
- [x] Replace the call-for-action "View on GitHub" with "View Documentation"
- [x] Have a custom background (or no background at all)
- [x] Customize the palette (maybe it could match the scalacenter website)
- [x] Include a _Edit on GitHub_ link
- [x] Add "Creating your own rules" section
- [x] Address TODOs in build.sbt
- [x] Make links look like links (they're currently black)
- [x] Add hash links to subsection headers (like GitHub does)
- [x] Convert all the `@sect.ref`
- [x] Add Configuration section
- [x] Update the contribution guide
- [x] Remove social media buttons and consider putting a Gitter link instead
- [x] Increase text contrast (gray on white is hard to read)
- [x] Make sure the website looks good on very large screens
- [x] Handle Edit on GitHub at smaller widths
- [x] Add a proper intro to Installation section
- [x] Decide about CLI `--help`
- [x] Automatically deploy with Travis
- [x] Include API docs? (this is a longer shot, but it should be doable)
- [x] Remove `readme` subproject
- [x] Add info about `menu.yml` to contribution guide

PR #389, opened 2017-10-09
Remove dependency on scalamacros/paradise

olafurpg:
Should unblock unidoc setup in #386

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/util/TokenClasses.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/util/TokenClasses.scala
@@ -1,29 +1,32 @@
 package scalafix.util
 
-import scala.meta.internal.classifiers.classifier
+import scala.meta.classifiers.Classifier
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 
-@classifier
 trait Whitespace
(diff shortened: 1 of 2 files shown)


scalafix-core/shared/src/main/scala/scalafix/util/TokenClasses.scala
```diff
@@ -1,29 +1,32 @@
...
 import scala.meta.tokens.Token._
 
-@classifier
 trait Whitespace
 object Whitespace {
-  def unapply(token: Token): Boolean = {
-    token.is[Space] || token.is[Tab] || token.is[Newline] || token.is[FF]
-  }
+  implicit val WhitespaceClassifier: Classifier[Token, Whitespace] =
+    new Classifier[Token, Whitespace] {
```
gabro: sigh, this would be so much nicer with a SAM

```scala
implicit val WhitespaceClassifier: Classifier[Token, Whitespace] =
  (token: Token) => token.is[Space] || token.is[Tab] || token.is[Newline] || token.is[FF]
```

but we cross-compile for 2.11, so 👍

gabro:
mima is not happy though :( Can you check @olafurpg

PR #395, opened 2017-10-12
Release/website fixes

### LINKED ISSUE #392: Automatically cross-publish sbt plugin
`sbt-scalafix` cross-builds correctly but the CI only publishes the 0.13 version.

olafurpg:
- Fixes #392

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,12 +17,12 @@ name := {
 
 lazy val crossVersions = Seq(scala211, scala212)
 
-// Custom scalafix release command. Tried sbt-release but didn't play well with sbt-doge.
-commands += Command.command("release") { s =>
+commands += Command.command("ci-release") { s =>
   "clean" ::
     "very publishSigned" ::
-    "sonatypeRelease" ::
-    "gitPushTag" ::
+    "^^ 1.0.2 " ::
+    "scalafix-sbt/publishSigned" ::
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -17,12 +17,12 @@ name := {
...
 lazy val crossVersions = Seq(scala211, scala212)
 
-// Custom scalafix release command. Tried sbt-release but didn't play well with sbt-doge.
-commands += Command.command("release") { s =>
+commands += Command.command("ci-release") { s =>
   "clean" ::
     "very publishSigned" ::
-    "sonatypeRelease" ::
-    "gitPushTag" ::
+    "^^ 1.0.2 " ::
```
gabro: doesn't this publish only 1.0.2? Shouldn't it be `^ scalafix-sbt/publishSigned :: sonatypeReleaseAll`?

olafurpg: `very publishSigned` publishes the 0.13 version

gabro: Ah ok, I'm still fairly surprised that ^ publishSigned doesn't do what I thought :(

olafurpg: The way I understand it, `^` will aggregate `sbtCrossVersions` like `very` does for `crossScalaVersions`. I haven't found a way to combine the two with `^ very publishSigned`, this was the easiest solution I could think of.

Issue #393, opened 2017-10-11
Website chores


gabro:
- [x] add `noPublish`
- [x] use 2.12 for building the website (so that apidocs are for 2.12, which is much nicer)
- [ ] test against 404 links in CI (maybe using something like http://digitaldrummerj.me/jekyll-validating-links-and-images/ or https://www.npmjs.com/package/broken-link-checker)
- [ ] get prettier thumbnail 

<img width="647" alt="screen shot 2017-10-21 at 12 06 17" src="https://user-images.githubusercontent.com/1408093/31850749-44c2446c-b658-11e7-951c-2ba7eea5e156.png">

olafurpg:
The website has now been ported to docusaurus! We still haven't integrated 404 link checking, but mdoc does validate that local hyperlinks point to existing sections :)

PR #404, opened 2017-10-27
Publish using 2.12.3 on the CI (this fixes the Scaladoc version issue)

gabro:
I just switched to 2.12.3 the entire CI job. I don't see any obvious drawback, since we're running the slow version on scala 2.11.11 anyway.

This fixes the issue with the version of Scaladoc used (reported in #393)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ env:
   matrix:
   - CI_TEST: ci-fast
     CI_SCALA_VERSION: 2.11.11
-    CI_PUBLISH: true
   - CI_TEST: ci-fast
     CI_SCALA_VERSION: 2.12.3
+    CI_PUBLISH: true
   - CI_TEST: ci-slow
     CI_SCALA_VERSION: 2.11.11
   - CI_TEST: scalafmt


.travis.yml
```diff
@@ -6,7 +6,7 @@ jdk:
 env:
   matrix:
   - CI_TEST: ci-fast
-    CI_SCALA_VERSION: 2.11.11
+    CI_SCALA_VERSION: 2.12.3
```
olafurpg: This disables ci for 2.11, can we move ci_publish to 2.12 entry instead?

gabro: doesn't the CI run anyway (although in "slow" mode?)

gabro: anyway, yes, your solution makes much more sense. Done.

.travis.yml
```diff
@@ -6,7 +6,7 @@ jdk:
...
     CI_PUBLISH: true
   - CI_TEST: ci-fast
     CI_SCALA_VERSION: 2.12.3
```
olafurpg: This entry here runs tests in 2.12 but no publish, we should run ci_publish here

PR #409, opened 2017-11-01
Add --tool-classpath to cli for file: and class: protocols.

### LINKED ISSUE #383: --tool-classpath to extend classpath when classloading custom rules
Currently, the cli uses the cli's it's own classpath when classloading custom rules. It would be nice if the cli supported --tool-classpath flag (name up for debate) to pass in a custom classpath with for example additional library dependencies.

olafurpg:
Previously, you needed to include the classpath as part of the scalafix
cli binary in order to use 3rd party dependencies. This commit makes it
possible to use the same scalafix binary with a custom classpath for
loading rules.

Fixes #383

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/LazySemanticdbIndex.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/LazySemanticdbIndex.scala
@@ -12,20 +12,20 @@ import org.langmeta.io.AbsolutePath
 // the moment we instantiate the rule.
 //type LazySemanticdbIndex = RuleKind => Option[SemanticdbIndex]
 class LazySemanticdbIndex(
-    f: RuleKind => Option[SemanticdbIndex],
-    val reporter: ScalafixReporter,
-    val workingDirectory: AbsolutePath
+    f: RuleKind => Option[SemanticdbIndex] = _ => None,
+    val reporter: ScalafixReporter = ScalafixReporter.default,
+    // The working directory when compiling file:relativepath/
+    val workingDirectory: AbsolutePath = AbsolutePath.workingDirectory,
+    // Additional classpath entries to use when compiling/classloading rules.
+    val toolClasspath: List[AbsolutePath] = Nil
 ) extends Function[RuleKind, Option[SemanticdbIndex]] {
   override def apply(v1: RuleKind): Option[SemanticdbIndex] = f(v1)
 }
 
 object LazySemanticdbIndex {
-  lazy val empty = new LazySemanticdbIndex(
-    _ => None,
-    ScalafixReporter.default,
-    AbsolutePath.workingDirectory)
+  lazy val empty = new LazySemanticdbIndex()
   def apply(
       f: RuleKind => Option[SemanticdbIndex],
       cwd: AbsolutePath = AbsolutePath.workingDirectory): LazySemanticdbIndex =
-    new LazySemanticdbIndex(f, ScalafixReporter.default, cwd)
+    new LazySemanticdbIndex(f, ScalafixReporter.default, cwd, Nil)
 }
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
@@ -1,6 +1,7 @@
 package scalafix
 package internal.config
 
+import java.io.File
 import scala.meta.Ref
 import scala.meta._
 import scala.meta.parsers.Parse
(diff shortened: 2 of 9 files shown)


scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
```diff
@@ -39,6 +47,14 @@ class ScalafixToolbox {
       code: Input,
       index: LazySemanticdbIndex): Configured[Rule] =
     synchronized {
```
xeno-by: Unrelated, but I'm wondering whether getRule itself should be synchronized.

scalafix-core/shared/src/main/scala/scalafix/internal/config/LazySemanticdbIndex.scala
```diff
@@ -12,20 +12,20 @@ import org.langmeta.io.AbsolutePath
 // the moment we instantiate the rule.
 //type LazySemanticdbIndex = RuleKind => Option[SemanticdbIndex]
 class LazySemanticdbIndex(
-    f: RuleKind => Option[SemanticdbIndex],
-    val reporter: ScalafixReporter,
-    val workingDirectory: AbsolutePath
+    f: RuleKind => Option[SemanticdbIndex] = _ => None,
+    val reporter: ScalafixReporter = ScalafixReporter.default,
+    // The working directory when compiling file:relativepath/
```
xeno-by: Oh, looks like the fix to the bug that I mentioned the other day! 👍

scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
```diff
@@ -25,6 +28,7 @@ import metaconfig.Configured.Ok
 import scalafix.internal.config.MetaconfigParser.{parser => hoconParser}
 import scalafix.internal.rule.ConfigRule
 import scalafix.patch.TreePatch
+import org.scalameta.logger
```
xeno-by: Ah the good old extraneous `org.scalameta.logger` :)

scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
```diff
@@ -132,7 +136,14 @@ trait ScalafixMetaconfigReaders {
   def classloadRuleDecoder(index: LazySemanticdbIndex): ConfDecoder[Rule] =
     ConfDecoder.instance[Rule] {
       case UriRuleString("scala" | "class", fqn) =>
-        ClassloadRule(fqn, classloadRule(index))
+        val classloader =
+          if (index.toolClasspath.isEmpty) ClassloadRule.defaultClassloader
+          else {
+            val urls =
+              index.toolClasspath.iterator.map(_.toURI.toURL).toArray
+            new URLClassLoader(urls, ClassloadRule.defaultClassloader)
```
xeno-by: Yep, that's exactly what we need!

scalafix-core/shared/src/main/scala/scalafix/internal/util/ClassloadRule.scala
```diff
@@ -115,7 +115,7 @@ class ClassloadRule[T](classLoader: ClassLoader)(implicit ev: ClassTag[T]) {
 }
 
 object ClassloadRule {
-  lazy val defaultClassloader = getClass.getClassLoader
+  def defaultClassloader: ClassLoader = getClass.getClassLoader
```
xeno-by: Why is this no longer lazy?

scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
olafurpg: Yes, I should make it a top-level package like `pprint` so I don't have to import anything 😅

scalafix-core/shared/src/main/scala/scalafix/internal/util/ClassloadRule.scala
olafurpg: not necessary change, just remove lazy because lazy should only be used when necessary.

scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
olafurpg: Hmm, I'm not sure. I recall this was enough to handle my manual tests running compilation for 20 project concurrently.

scalafix-core/shared/src/main/scala/scalafix/internal/config/LazySemanticdbIndex.scala
olafurpg: You will also be happy to know sbt-scalafix `file:` tab completes absolute paths and `..` paths! #408

PR #411, opened 2017-11-01
Migrate to utest

### LINKED ISSUE #339: Port scalafix-testkit to utest
If I remove all tokens from an input file to testkit, then I get the following error

```scala
[info] - scala/app/domain/entity.scala *** FAILED ***
[info]   java.util.NoSuchElementException: None.get
[info]   at scala.None$.get(Option.scala:349)
[info]   at scala.None$.get(Option.scala:347)
[info]   at scalafix.testkit.SemanticRuleSuite$.stripTestkitComments(SemanticRuleSuite.scala:20)
[info]   at scalafix.testkit.SemanticRuleSuite.$anonfun$runOn$1(SemanticRuleSuite.scala:123)
[info]   at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.java:12)
```

Here 
https://github.com/scalacenter/scalafix/blob/c0a26983514599172b30e515c9b2029cd92e1f64/scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala#L23
we can return a bogus token `.getOrElse("".tokenize.get(0))` instead of calling `.get`

olafurpg:
- remove dependency on scalatest
- remove ONLY hack from scalafix-testkit, thanks to ability to target
  individual test with testOnly in utest.
- cleaner test names for simpler testOnly
- highlight stacktraces from scalafix namespace

Fixes #339


olafurpg:
The main motivation for this change was to fix #339. Scalafix-testkit had features that can be replaced with utest, those features had bugs so instead of fixing the bugs I ported the tests to utest instead. This brings lots of nice improvements, for example prettier reports and stack traces

screen shot 2017-11-01 at 17 55 38

screen shot 2017-11-01 at 17 29 50

wdyt?

project/Dependencies.scala
```diff
@@ -23,6 +23,7 @@ object Dependencies {
 
   def scalameta = Def.setting("org.scalameta" %%% "contrib" % scalametaV)
   def scalatest = Def.setting("org.scalatest" %%% "scalatest" % "3.0.0")
```
gabro: why is scalatest still a dependency? Is it used elsewhere?

olafurpg:
I updated the PR to keep source-compatibility with the g8 template. Although I'm personally excited about this change, I think it's best to let it wait until 0.6. We should focus instead on non-breaking improvements.

PR #415, opened 2017-11-02
Disable rules with comments

MasseGuillaume:


scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -48,18 +51,28 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
     logger.elem(values: _*)
   }
 
-  def printLintMessage(msg: LintMessage, owner: RuleName): Unit = {
+  def reportLintMessage(msg: LintMessage, owner: RuleName): Boolean = {
```
MasseGuillaume: Maybe we should just collect the message to be reported here (eg returning (Message, Positon, Severity, Escaped | Not Escaped) )

scalafix-tests/unit/src/test/scala/scalafix/tests/core/DisableRangeTest.scala
```diff
@@ -0,0 +1,65 @@
+package scalafix.tests.core
```
MasseGuillaume: Should I leave unit tests or scalafix-testkits are enough?

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+      anchors.sortBy(_.position.offset).grouped(2).toList.collect {
+        case List(FilterAnchor.Start(start, rules), FilterAnchor.End(end)) => {
+          CommentFilter(start, Some(end), rules)
+        }
+
+        case List(FilterAnchor.Start(start, rules)) => {
+          CommentFilter(start, None, rules)
+        }
+
+        case _ => sys.error("invalid anchors")
```
MasseGuillaume: This would fail with

```
Start a
Start b
```

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+}
+
+object FilterAnchor {
+  case class Start(position: PointPosition, rules: Set[String])
+      extends FilterAnchor
+  case class End(position: PointPosition) extends FilterAnchor
+}
+
+object EscapeHatch {
+  private val FilterAnchorStart = "scalafix:off (.*)".r
```
olafurpg: Can we can allow a space before and after: `" ?scalafix ?`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+  private val FilterAnchorEnd = "scalafix:on".r
```
olafurpg: Is is possible to selectively enable a rule? `// scalafix:on Name`?

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
olafurpg: Can we include the anchor in the message to help debug what went wrong?

olafurpg: We are inside a collect, why do we fail here?

olafurpg: What happens in this case?
```scala
// scalafix:off A
// scalafix:off B
// scalafix:on B
// scalafix:on A
```

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        case _ => sys.error("invalid anchors")
+      }
+
+    new EscapeHatch(filters)
+  }
+}
+
+class EscapeHatch(filters: List[CommentFilter]) {
```
olafurpg: I feel like we can simplify the logic a bit if we change the semantics to be

> a rule `r` is disabled in position `p` if there is a comment disabling `r` at position `p1 < p` and there is no comment enabling `r` in position `p2` where `p1 < p2 < p`.

This does not handle push/pop semantics like
```
// scalafix:off A
// scalafix:off A
// scalafix:on A
object ReportA // A is reported because there is a single `on:A` between it and `off:A`
```
But I question how useful push/pop is in cases like that, `// scalafix:on A` should IMO always enable rule `A` no matter how many leading `off: A` there are in the file.


```scala
class EscapeHatch(
    starts: TreeMap[Int, FilterMatcher],
    ends: TreeMap[Int, FilterMatcher]) {
  def isEnabled(id: String, pos: Int): Boolean = !isDisabled(id, pos)
  def isDisabled(id: String, pos: Int): Boolean = {
    starts.to(pos).exists {
      case (spos, sname) =>
        sname.matches(id) && {
          !ends.range(spos, pos + 1).exists {
            case (epos, ename) =>
              ename.matches(id)
          }
        }
    }
  }
}
object EscapeHatch {
  case class Comment(pos: Int, id: String, kind: Kind)
  sealed abstract class Kind
  case object On extends Kind
  case object Off extends Kind
  def apply(comments: List[Comment]): EscapeHatch = {
    val starts = TreeMap.newBuilder[Int, FilterMatcher]
    val ends = TreeMap.newBuilder[Int, FilterMatcher]
    comments.foreach {
      case Comment(pos, id, Off) => starts += (pos -> FilterMatcher(id))
      case Comment(pos, id, On) => ends += (pos -> FilterMatcher(id))
    }
    new EscapeHatch(starts.result(), ends.result())
  }
}
```
`FilterMatcher` comes from internal.config. One benefit of this approach is that we avoid having to match on/off comments.

What do you think?

scalafix-core/shared/src/main/scala/scalafix/rule/RuleCtx.scala
```diff
@@ -51,9 +51,9 @@ trait RuleCtx extends PatchOps {
   // Private scalafix methods, subject for removal without notice.
   private[scalafix] def toks(t: Tree): Tokens
   private[scalafix] def config: ScalafixConfig
-  private[scalafix] def printLintMessage(
+  private[scalafix] def reportLintMessage(
       msg: LintMessage,
-      owner: RuleName): Unit
+      owner: RuleName): Boolean
```
olafurpg: Mima may complain over this change, we can safely add a filter for it in project/Mima.scala since it's a private api.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/DisableRangeTest.scala
olafurpg: testkit is enough I think, that avoids us testing against the scalafix internal API.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+
+import scala.meta._
+
+object RangePosition {
+  def apply(position: Position): RangePosition =
+    RangePosition(position.start, position.end)
+}
+
+case class RangePosition(start: Int, end: Int) {
+  def withIn(other: RangePosition): Boolean = {
```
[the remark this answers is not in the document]
MasseGuillaume: Done.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,97 @@
...
+    start <= other.start && other.end <= end
+  }
+}
+case class PointPosition(offset: Int)
```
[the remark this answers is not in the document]
MasseGuillaume: Done.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/DisableRangeTest.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import org.scalatest.FunSuite
+
+import scala.meta._
+
+import scalafix.internal.patch._
+
+class DisableRangeTest extends FunSuite {
```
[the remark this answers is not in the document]
MasseGuillaume: Done.

scalafix-core/shared/src/main/scala/scalafix/rule/RuleCtx.scala
MasseGuillaume: If it's a private api we should mark this `private[scalafix]`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
MasseGuillaume: Done.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
MasseGuillaume: Done.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
MasseGuillaume: Done. Added test case.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
MasseGuillaume: I think it's easier te reason about scalafix:on|off as brackets `{` and `}` so if I convert the thing above in bracket notation it look like:

```
{
{
}
```

MasseGuillaume: I asked for a third opinion (Martin D.) and he said the way you propose makes more sense. I agree with you guys now ;-).

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -48,18 +55,28 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
     val key = msg.category.key(owner)
-    if (config.lint.ignore.matches(key)) ()
```
olafurpg: Can we maybe implement this filtering in `Patch.lintMessages` and this method reports everything?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -48,18 +55,28 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
-        msg.position,
-        category.toSeverity
-      )
+
+      val ruleName = msg.id(owner)
+      val position = msg.position
+
+      val isEscaped = escapeHatch.isEnabled(ruleName, position)
+
+      if (isEscaped) false
```
olafurpg: Same here, `Patch.lintMessages`

scalafix-core/shared/src/main/scala/scalafix/lint/LintMessage.scala
```diff
@@ -16,6 +16,12 @@ final case class LintMessage(
     position: Position,
     category: LintCategory
 ) {
+
+  def id(owner: RuleName): String = {
```
olafurpg: I just realized this is implemented in LintCategory.key https://github.com/scalacenter/scalafix/blob/master/scalafix-core/shared/src/main/scala/scalafix/lint/LintCategory.scala#L20

scalafix-tests/input/src/main/scala/test/escapeHatch/EscapeHatchRange.scala
```diff
@@ -0,0 +1,16 @@
...
+  "class:scalafix.test.EscapeHatchDummyLinterB"
+]
+ */
+package test
+
+object EscapeHatchRange {
+  //scalafix:off EscapeHatchDummyLinterA
+  //scalafix:off EscapeHatchDummyLinterB
+  object A
+  // scalafix:on EscapeHatchDummyLinterA
```
olafurpg: Either remove or test that A is reported following this line.

scalafix-tests/input/src/main/scala/test/escapeHatch/EscapeHatchEOF.scala
```diff
@@ -0,0 +1,12 @@
+/*
+rules = [
+  "class:scalafix.test.EscapeHatchDummyLinterA",
+]
+*/
+
+//scalafix:off EscapeHatchDummyLinterA
```
olafurpg: let it "breathe" 😜  `// scalafix:off`

MasseGuillaume:
I have to [make Mima happy](https://travis-ci.org/scalacenter/scalafix/jobs/298519540#L805-L814)

NB: There is a broken scaladoc comment somewhere.

since `website/makeMicrosite` crashes with `[error] (website/scalaunidoc:unidoc::scalacOptions) java.util.NoSuchElementException: None.get`

MasseGuillaume:
I still need to support this case:

```scala
object abc { // scalafix:ok LowerCaseName
  val foo = 1
}
```

This ok escape is not attached to `object abc`

website/src/main/tut/docs/users/configuration.md
```diff
@@ -109,4 +111,38 @@ debug.printSymbols = true
 
 To build custom rules see {% doc_ref Rule Authors %}.
 
+## Per source file
+
+It's possible to supress false positives in the source code itself to have a more fine grain on what to report and what not to report. There is two techniques you could use to disable linting: disable/enable until end of file or disable per expression.
```
olafurpg: > It is possible to suppress false positive linter messages over source file regions using comments.
> This provides fine grained control over what to report ...

...

>  There are two alternative ways to disable linter messages: a single expression using `scalafix:ok` or until the until the end of the source file with `scalafix:off/on`

website/src/main/tut/docs/users/configuration.md
```diff
@@ -109,4 +111,38 @@ debug.printSymbols = true
...
+
+### Enable or Disable until the EOF
```
olafurpg: s/EOF/end of file/

website/src/main/tut/docs/users/configuration.md
```diff
@@ -109,4 +111,38 @@ debug.printSymbols = true
...
+
+```scala
+
+// scalafix:off
+foo(null)
+1.asInstanceOf[String]
+// scalafix:on
```
olafurpg: I would first show an example without `scalafix:on`. Then show another example how to re-enable.

website/src/main/tut/docs/users/configuration.md
```diff
@@ -109,4 +111,38 @@ debug.printSymbols = true
...
 
+```
+
+### Disable for an expression
+
+```scala
+List(1, "") // scalafix:ok
```
olafurpg: Show more examples, leading comment + trailing comment + per-line with `object a { // ok`

website/src/main/tut/docs/users/configuration.md
```diff
@@ -109,4 +111,38 @@ debug.printSymbols = true
...
+```scala
+// scalafix:off Disable.null, Disable.asInstanceOf
+foo(null)
+1.asInstanceOf[String]
+// scalafix:on
+```
+
+```scala
+List(1, "") // scalafix:ok NoInfer.any
+```
```
olafurpg: Should we also document that scalafix will report warning for unused `off/on/ok` comments? I think it's a pretty great selling point

CONTRIBUTING.md
```diff
@@ -115,6 +115,14 @@ If you need to link to a page in the apidocs you can use the `apidocs_ref` tag,
 // Expands to: /scalafix/docs/api/scalafix/patch/PatchOps.scala
 ```
 
+## Binary Compatibility
+
+To avoid breaking binary compatiblity we use the Migration Manage for Scala or [Mima](https://github.com/lightbend/migration-manager) for short.
```
olafurpg: 👍

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+import scala.meta._
+import scala.meta.tokens.Token
+import scala.meta.contrib._
+
+import scala.collection.mutable
+import scala.collection.immutable.TreeMap
+
+/* EscapeHatch is an algorithm to selectively disable rules
+ *
+ * Rules are disable via comments with a specific syntax
```
olafurpg: disabled

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+                isEnabled
+              }
+          }
+        }
+      }
+
+    (!isDisabled, culprit)
+  }
+
+  def filter(lints: List[LintMessage]): List[LintMessage] = {
```
olafurpg: Add a brief explanation

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+    ) extends Escape
+  }
+
+  private sealed abstract class Toogle
+  private object Toogle {
+    case object Disable extends Toogle
+    case object Enable extends Toogle
+  }
+
+  private val unusedScalafixSupressionId = "UnusedScalafixSupression"
```
olafurpg: This should be a `RuleName`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+
+  private[scalafix] def unusedScalafixSupression(
+      idSuffix: String,
+      explain: String) =
+    LintCategory.warning(unusedScalafixSupressionId + idSuffix, explain)
+
+  private val unusedEnableWarning =
+    unusedScalafixSupression(
```
olafurpg: `LintCategory...`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+
+  private sealed abstract class Toogle
+  private object Toogle {
+    case object Disable extends Toogle
+    case object Enable extends Toogle
+  }
+
+  private val unusedScalafixSupressionId = "UnusedScalafixSupression"
+
+  private[scalafix] def unusedScalafixSupression(
```
olafurpg: We can remove this helper in favor of `LintCategory.withOwner(RuleName)`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+      }
+    }
+
+    tree.tokens.foreach {
+      // matches off anchors
+      //
+      // // scalafix:off RuleA
+      // ...
+      //
+      case comment @ Token.Comment(FilterDisable(rules)) =>
```
olafurpg: Can we match onces on `Token.Comment` and move the `FilterDisable` match inside the case body?

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+        val position = AnchorPosition(comment.pos)
+        escapes += Escape.UntilEOF(position, rules, Toogle.Disable)
+        trackUnusedRules(rules, Toogle.Disable, position)
+
+      // matches on anchors
+      //
+      // ...
+      // // scalafix:on RuleA
+      //
+      case comment @ Token.Comment(FilterEnable(rules)) =>
```
olafurpg: Merge with case above

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+        val position = AnchorPosition(comment.pos)
+        escapes += Escape.UntilEOF(position, rules, Toogle.Enable)
+        trackUnusedRules(rules, Toogle.Enable, position)
+
+      // matches expressions not handled by AssociatedComments
+      //
+      // object Dummy { // scalafix:ok EscapeHatchDummyLinter
+      //   1
+      // }
+      case comment @ Token.Comment(FilterExpression(rules)) =>
```
olafurpg: ditto

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -0,0 +1,324 @@
...
+    EscapeHatch(escapes.result(), unusedEnable.result())
+  }
+
+  private def apply(
+      comments: List[Escape],
+      unusedEnable: List[LintMessage]): EscapeHatch = {
+    val enableRules = TreeMap.newBuilder[EscapeOffset, EscapeFilter]
+    val disableRules = TreeMap.newBuilder[EscapeOffset, EscapeFilter]
+
+    def matcher(rules: String): FilterMatcher = {
```
olafurpg: Can we move this out of the `apply` method?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -48,19 +50,18 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
-        .getConfiguredSeverity(key)
-        .getOrElse(msg.category.severity)
-      config.lint.reporter.handleMessage(
-        msg.format(owner, config.lint.explain),
-        msg.position,
-        category.toSeverity
-      )
-    }
+  def printLintMessage(msg: LintMessage): Unit = {
+    val key = msg.category.id
```
olafurpg: Remove?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -48,19 +50,18 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
+
+    val category = config.lint
```
olafurpg: We may want to move this into `LintCategory.withConfig(LintConfig): LintCategory`

olafurpg: We can open an issue and fix it in a separate PR.

scalafix-core/shared/src/main/scala/scalafix/lint/LintCategory.scala
```diff
@@ -29,6 +25,18 @@ final case class LintCategory(
     LintMessage(message, Position.None, this)
   def at(position: Position): LintMessage =
     LintMessage(explanation, position, noExplanation)
+
+  @deprecated("use withOwner(owner: RuleName): LintCategory", "0.5.4")
+  def key(owner: RuleName): String =
+    deprecatedKey(owner)
+
+  private[scalafix] def withOwner(owner: RuleName): LintCategory =
```
olafurpg: This can be public

scalafix-tests/input/src/main/scala/test/escapeHatch/EscapeHatchDoubleOn.scala
```diff
@@ -0,0 +1,18 @@
...
+// of how many time it was turned off
+
+/* scalafix:off */ // assert: UnusedScalafixSupressionDisable
+/* scalafix:off EscapeHatchDummyLinter */ // assert: UnusedScalafixSupressionDisable
+/* scalafix:off EscapeHatchDummyLinter */ // assert: UnusedScalafixSupressionDisable
+// scalafix:on EscapeHatchDummyLinter
+
+object EscapeHatchDoubleOn {
+  object Dummy // assert: EscapeHatchDummyLinter
+}
```
olafurpg: note to self, https://github.com/scalameta/scalafmt/issues/1073

scalafix-tests/input/src/main/scala/test/escapeHatch/EscapeHatchExpression.scala
```diff
@@ -0,0 +1,51 @@
...
+NoInfer.symbols = [
+  "scala.Predef.any2stringadd"
+]
+*/
+
+// `scalafix:ok` can be used to disable expressions
+// it can be attached to an expressions that spans multiple lines
+// or inside the first line of the template of an object/class/trait
+
+package test
```
olafurpg: test.escapeHatch

scalafix-tests/input/src/main/scala/test/escapeHatch/EscapeHatchUnused.scala
```diff
@@ -0,0 +1,53 @@
...
+    1,
+    2
+  ) /* scalafix:ok EscapeHatchDummyLinter */ // assert: UnusedScalafixSupressionDisable
+
+  object Ok { /* scalafix:ok EscapeHatchDummyLinter */ // assert: UnusedScalafixSupressionDisable
+    val a = 1
+  }
+}
+
+// UnusedScalafixSupression
```
olafurpg: Is this needed?

website/src/main/tut/docs/users/configuration.md
olafurpg: suppress

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -144,18 +150,20 @@ object EscapeHatch {
...
   }
 
-  private val unusedScalafixSupressionId = "UnusedScalafixSupression"
+  private val unusedScalafixSupression = RuleName("UnusedScalafixSupression")
 
-  private[scalafix] def unusedScalafixSupression(
-      idSuffix: String,
-      explain: String) =
-    LintCategory.warning(unusedScalafixSupressionId + idSuffix, explain)
+  // private[scalafix] def unusedScalafixSupression(
```
olafurpg: Remove comments?

olafurpg:
And please don't force push changes that address review comments, it's easier to review if there is a commit with message "Address review feedback".

PR #428, opened 2017-11-13
Disable Keywords: Test case & Configuration

MasseGuillaume:


MasseGuillaume:
Closing this in favor of `DisableSyntax`. The `Disable` is a SemanticRule and requires a compilation of the code where `DisableSyntax` can simply be a Rule (purely syntactic).

PR #431, opened 2017-11-16
Disable syntax

MasseGuillaume:


project/Dependencies.scala
```diff
@@ -5,7 +5,7 @@ import org.scalajs.sbtplugin.ScalaJSPlugin.autoImport._
 object Dependencies {
 
   val scalametaV = "2.0.1"
-  val metaconfigV = "0.5.2"
+  val metaconfigV = "0.5.2+1-9d714689"
```
olafurpg: 0.5.3

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+
+import metaconfig.{Conf, ConfError, ConfDecoder, Configured}
+import org.langmeta._
+import MetaconfigPendingUpstream.XtensionConfScalafix
+
+import scala.meta.tokens.Token
+
+case class DisableSyntaxConfig(
+    keywords: Set[DisabledKeyword] = Set(),
+    tab: Boolean = false,
```
olafurpg: tabs?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+    semicolon: Boolean = false,
```
olafurpg: semicolons

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+  implicit val reader: ConfDecoder[DisableSyntaxConfig] = default.reader
+}
+
+case class DisabledKeyword(keyword: String)
+
+object DisabledKeyword {
+  def unapply(token: Token): Option[String] = {
+    import Token._
+
+    token match {
```
olafurpg: Can we avoid this match and use product prefix instead? This approach is unsafer since there's a risk we accidentally add a typo in the match. We should also avoid doing a large match like this on every token.

olafurpg: The config reader already validates that the configuration contains only valid keywords.

MasseGuillaume:
@olafurpg For some reason `DisableSyntax` is not getting pick up by `scalafix.tests.rule.SemanticTests`. You told me yesterday to enable it somewhere. Can you point me to the relevant part of the code? Thanks

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+          case Conf.Bool(value) =>
+            if (value) readKeyword("true")
+            else readKeyword("false")
+
+          case _ => Configured.typeMismatch("String", conf)
+        }
+      }
+    }
+
+  // XXX: This is from metaconfig.ConfError
```
olafurpg: Why XXX?

scalafix-tests/unit/src/test/scala/scalafix/tests/config/DisableSyntaxConfigSuite.scala
```diff
@@ -0,0 +1,81 @@
...
+import scala.meta.inputs.Input
+
+import metaconfig.Configured
+import metaconfig.Configured.NotOk
+import metaconfig.ConfError
+
+import scalafix.internal.config.MetaconfigParser
+import scalafix.internal.config.{DisableSyntaxConfig, DisabledKeyword}
+
+class DisableSyntaxConfigSuite extends FunSuite {
```
olafurpg: Nice 👍

scalafix-tests/unit/src/test/scala/scalafix/tests/config/DisableSyntaxConfigSuite.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  test("Provide suggestions when typos are present in keywords") {
+    val rawConfig =
+      """|keywords = [
+         |  overide
+         |]
+         |""".stripMargin
+    val errorMessage =
+      "overide is not in our supported keywords. (Did you mean: override?)"
```
olafurpg: 👏

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
MasseGuillaume: https://softwareengineering.stackexchange.com/a/65470/38296

it is used as a marker for code that requires attention.

MasseGuillaume:
@olafurpg all comments addressed!

olafurpg:
The CI error should be fixed if you rebase on top of master.

MasseGuillaume:
@olafurpg done.

MasseGuillaume:
@olafurpg the minor comment got lost in the rebase. What was it?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,149 @@
...
+
+object DisableSyntaxConfig {
+  val default = DisableSyntaxConfig()
+  implicit val reader: ConfDecoder[DisableSyntaxConfig] = default.reader
+}
+
+object Keyword {
+  def unapply(token: Token): Option[String] = {
+    val prefix = token.productPrefix
+    if (prefix.take(2) == "Kw") {
```
olafurpg: We can avoid this string manipulation on every token by precomputing `val keywordProductPrefixes: Set[String] = keywords.toIterator.map(kw => s"Kw${kw.capitalize}").toSet`

olafurpg: PS, you can use `String.startsWith`

olafurpg:
GitHub review loosing comments is absolutely terrifying, thank god it was only one comment in this case! 🙈

MasseGuillaume:
@olafurpg let's use simple comment and create an empty review to approve/disapprove.

MasseGuillaume:
Done, I added the kw chache.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,148 @@
...
+  def unapply(token: Token): Option[String] = {
+    val prefix = token.productPrefix
+    if (kwSet.contains(prefix)) {
+      Some(prefix.drop(2).toLowerCase)
+    } else {
+      None
+    }
+  }
+  val all = List(
+    "abstract",
```
olafurpg: FWIW scalafmt actually has a setting to binpack argument lists consisting only of literals, it does it for numeric literals by default but for some reason I excluded strings in the default setting.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -0,0 +1,148 @@
...
+
+object DisableSyntaxConfig {
+  val default = DisableSyntaxConfig()
+  implicit val reader: ConfDecoder[DisableSyntaxConfig] = default.reader
+}
+
+object Keyword {
+  def unapply(token: Token): Option[String] = {
+    val prefix = token.productPrefix
+    if (kwSet.contains(prefix)) {
```
olafurpg: Thanks!

olafurpg:
I am investigating what is causing the build failures, it's unrelated to this PR. I started once we upgraded to 2.12.4, but I'm unable to reproduce it locally or in some cases PRs like #430 Very weird

PR #433, opened 2017-11-16
Downgrade to Scala 2.11.11 and 2.12.3.

olafurpg:
After upgrading to 2.12.4 and 2.11.12, we started getting irregular
compile errors in CI (example https://travis-ci.org/scalacenter/scalafix/jobs/303142842#L4658), only when the cache is not clean. I don't think
we are doing anything weird with the cache, this commit is to verify if
this regression is caused by misconfiguration in the scalafix builds or
the 2.12.4/2.11.12 upgrade.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,12 @@ jdk:
 env:
   matrix:
   - CI_TEST: ci-fast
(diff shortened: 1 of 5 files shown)


olafurpg:
Closing in favor of non-master branch pr.

PR #435, opened 2017-11-16
Document DisableSyntax. fix #434

### LINKED ISSUE #434: Document: DisableSyntax
I think it felt in the cracks, there is no documentation for the DisableSyntax Rule.

MasseGuillaume:


website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,84 @@
+---
+layout: docs
+title: Disable
+---
+
+# DisableSyntax
+
+_Since 0.5.4_
+
+This rule reports errors when a "disallowed" syntax is used.
```
olafurpg: Explain what we mean by syntax? Essentially, just make it clear this does not require types or compilation.

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,84 @@
...
+
+```scala
+MyCode.scala:7: error: [DisableSyntax.xml] xml is disabled.
+  <a>xml</a>
+  ^
+```
+
+## Configuration
+
+By default, this rule does allows all syntax.
```
olafurpg: Does not disable any syntax.

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,84 @@
...
+* semicolons (`;`)
+* tabs
+* xml literals
+
+To disallow a syntax:
+
+```
+DisableSyntax {
+  keywords = [
+    abstract
```
olafurpg: Let's keep this list small, only the usual suspects like null, throw, return

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,84 @@
...
+    type
+    val
+    var
+    while
+    with
+    yield
+  ]
+  carriageReturn = true
+  semicolons = true
+  tabs = true
```
olafurpg: Seeing this here makes me wonder if we should name the field "noTabs", since "tabs = true" is easy to misinterpret, wdyt? Same for other fields like xml

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,84 @@
...
+  xml = true
```
olafurpg: How hard would it be to support arbitrary regexp? 

Also, make it possible to add a custom message/explanation for each kind? 

We can leave it for another pr or release, just wondering

website/src/main/tut/docs/rules/DisableSyntax.md
MasseGuillaume: yup noTabs, noFoo, etc make sense.

website/src/main/tut/docs/rules/DisableSyntax.md
MasseGuillaume: It would be nice indeed: 
https://github.com/apache/spark/blob/7f2e62ee6b9d1f32772a18d626fb9fd907aa7733/scalastyle-config.xml#L206-L215

MasseGuillaume: https://github.com/scalacenter/scalafix/issues/436

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -7,21 +7,21 @@ import MetaconfigPendingUpstream.XtensionConfScalafix
 import scala.meta.tokens.Token
 
 case class DisableSyntaxConfig(
-    carriageReturn: Boolean = false,
+    noCarriageReturn: Boolean = false,
```
olafurpg: noCarriageReturn**s**?

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,48 @@
...
+* carriage return aka <CR> (windows uses <CR><LF>)
+* semicolons (`;`)
+* tabs
+* xml literals
+
+To disallow a syntax:
+
+```
+DisableSyntax {
+  keywords = [
```
olafurpg: What do you think about using 

```
Disable.keywords = [ ... ]
Disable.noTabs = true
...
```
instead of `DisableSyntax { ... }`. I think it makes it easier to copy-paste a single setting.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -7,21 +7,21 @@ import MetaconfigPendingUpstream.XtensionConfScalafix
...
+    noSemicolons: Boolean = false,
+    noTabs: Boolean = false,
+    noXml: Boolean = false
 ) {
   implicit val reader: ConfDecoder[DisableSyntaxConfig] =
     ConfDecoder.instanceF[DisableSyntaxConfig](
       c =>
         (
-          c.getField(carriageReturn) |@|
+          c.getField(noCarriageReturn) |@|
```
olafurpg: Did no tests fail from this change?

MasseGuillaume: Adding a test for this.

MasseGuillaume: I think we should remove this one. I modified the file with an hex editor to add a <CR> but it was not picked up by scalameta.

MasseGuillaume:
@olafurpg All comments addressed

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
```diff
@@ -7,19 +7,20 @@ DisableSyntax {
...
 }
 */
 package test
 
 case object DisableSyntax {
-  var a = 1                 // assert: DisableSyntax.var
-  null                      // assert: DisableSyntax.null
-  def foo: Unit = return    // assert: DisableSyntax.return
-  throw new Exception("ok") // assert: DisableSyntax.throw
+  var a = 1                 // assert: DisableSyntax.keywords.var
```
olafurpg: I think I preferred  `DisableSyntax.var`, what was the motivation here?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -7,7 +7,7 @@ import MetaconfigPendingUpstream.XtensionConfScalafix
 import scala.meta.tokens.Token
 
 case class DisableSyntaxConfig(
-    noCarriageReturn: Boolean = false,
+    noCarriageReturns: Boolean = false,
```
olafurpg: Remove this field?

olafurpg:
Can you please rebase on top of master where I've switched back to 2.12.3

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
MasseGuillaume: to be consistent with the configuration. I think this way it's easier to see where it came from.

MasseGuillaume:
Rebased.

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
olafurpg: Fair enough. I've wondered if `// scalafix:ok var` should be enough to disable `DisableSyntax.keywords.var`, since the latter is quite verbose. However, I just checked and `// scalafix:ok var` actually is enough to disable this message, which is good.

olafurpg: I have wondered whether the full category id should be printed to the console since it does add a lot of bloat

```scala
scala/test/Disable.scala:17: error: [Disable.asInstanceOf] asInstanceOf is disabled
  val w = List(1, 2, 3).asInstanceOf[Seq[String]] // assert: Disable.asInstanceOf
```

but that's unrelated to this PR :)

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,44 @@
...
+# DisableSyntax
+
+_Since 0.5.4_
+
+This rule reports errors when a "disallowed" syntax is used. We take advantage of the ScalaMeta parser to quickly
+scan the source code. This rules like the other syntaxtic rules does not require any compilation.
+
+Example:
+
+```scala
```
olafurpg: Can we also include an example for `return` or `throw`?

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
MasseGuillaume: This is why we should add the error message to the testkit :). You can still use assert:foo but you can also output the full message to tune the message UX.

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -0,0 +1,44 @@
+---
+layout: docs
+title: Disable
+---
+
+# DisableSyntax
+
+_Since 0.5.4_
+
+This rule reports errors when a "disallowed" syntax is used. We take advantage of the ScalaMeta parser to quickly
```
olafurpg: s/ScalaMeta/Scalameta/

We don't have to mention this however, this is enough

> This is a syntactic rule, which means it does not require compilation to run unlike the `Disable` rule.

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
olafurpg: you can already see the error messages in testkit ;) asserting against the full message contents is overtesting IMO, testkit should make it as easy as possible to add tests, copy-pasting the message or tweaking the caret adds an overhead which might demotivate people to add more tests.

MasseGuillaume: It would be in addition to what we already have.

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -18,6 +17,12 @@ MyCode.scala:7: error: [DisableSyntax.xml] xml is disabled.
   ^
 ```
 
+```scala
+MyCode.scala:10: error: [DisableSyntax.keywords.return] return is disabled.
+  return
```
olafurpg: You could spice up this example you know 😜 

```scala
if (isBad) return error
```

Issue #440, opened 2017-11-17
Testkit: Alternative way to assert rules


MasseGuillaume:
Related: https://github.com/scalacenter/scalafix/issues/436#issuecomment-345193468

```scala
Await.result(longComputation)// assert: Disable.Await.result
```

to

```scala
Await.result(longComputation)
/*
^ DisableSyntax.Await.result

Are you sure that you want to use Await.ready? In most cases, you should use ThreadUtils.awaitReady instead.
If you must use Await.ready, wrap the code block with
// scalafix:off DisableSyntax.Await.result
Await.ready(...)
// scalafix:on DisableSyntax.Await.result
*/
```

I think we should add an alternative way to test lint message by asserting the caret position and the message content. This is also really helpful to test custom messages: #439.

@olafurpg is concern that the test suite becomes too verbose.

@gabro WDYT ?

olafurpg:
My personal view is that testing the actual message content is not necessary since messages are typically only string literals with minimal splicing. You can manually check how the message looks by skipping an `// assert` and testkit prints the message out to the console just like the cli does.

MasseGuillaume:
Just to clarify I don't want to change the current method of asserting in testkit. I want to provide an additional way to assert.

olafurpg:
I may have misunderstood you @MasseGuillaume, I think adding a `/* assertMessage */` could be a nice addition.

```scala
object Disable {
  Await.result(...) /* assertMessage: Await.result is disabled
  ^
Explanation: Are you sure that you want to use Await.ready? In most cases, you should use ThreadUtils.awaitReady instead.
If you must use Await.ready, wrap the code block with
// scalafix:off DisableSyntax.Await.result
Await.ready(...)
// scalafix:on DisableSyntax.Await.result
*/

```

MasseGuillaume:
Fixed in: https://github.com/scalacenter/scalafix/pull/448

PR #443, opened 2017-11-17
Disable Rule: add a way to customize the lint message.

MasseGuillaume:
I would need https://github.com/scalacenter/scalafix/issues/440 to test this.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/CustomMessage.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/CustomMessage.scala
@@ -1,0 +1,29 @@
+package scalafix.internal.config
+
+import metaconfig.{Conf, ConfError, ConfDecoder, Configured}
+import org.langmeta._
+import MetaconfigPendingUpstream.XtensionConfScalafix
+
+import scala.language.implicitConversions
+
+case class CustomMessage[T](value: T, message: Option[String])
+object CustomMessage {
+  def decoder[T](field: String)(
+      implicit ev: ConfDecoder[T]): ConfDecoder[CustomMessage[T]] =
+    ConfDecoder.instance[CustomMessage[T]] {
+      case obj: Conf.Obj =>
+        (obj.get[T](field) |@| obj.get[String]("message")).map {
+          case (value, message0) => {
+            val message =
+              if (message0.split(System.lineSeparator).size > 1) {
+                System.lineSeparator + message0.stripMargin
+              } else {
+                message0
+              }
+            CustomMessage(value, Some(message))
+          }
+        }
+      case els =>
+        ev.read(els).map(value => CustomMessage(value, None))
+    }
+}
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
@@ -1,11 +1,26 @@
 package scalafix.internal.config
 
 import metaconfig.ConfDecoder
-import org.langmeta.Symbol
-
 import MetaconfigPendingUpstream.XtensionConfScalafix
+import org.langmeta.Symbol
+import scalafix.internal.util.SymbolOps
 
-case class DisableConfig(symbols: List[Symbol.Global] = Nil) {
+case class DisableConfig(symbols: List[CustomMessage[Symbol.Global]] = Nil) {
+  def allSymbols = symbols.map(_.value)
+
+  private val messageFromSymbol: Map[String, String] =
+    symbols
+      .flatMap(custom =>
+        custom.message.map(message => (custom.value.syntax, message)))
+      .toMap
+
+  def customMessage(symbol: Symbol.Global): Option[String] = {
+    messageFromSymbol.get(SymbolOps.normalize(symbol).syntax)
+  }
+
+  implicit val customMessageReader: ConfDecoder[CustomMessage[Symbol.Global]] =
+    CustomMessage.decoder(field = "symbol")
+
   implicit val reader: ConfDecoder[DisableConfig] =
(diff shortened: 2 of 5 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/config/CustomMessage.scala
```diff
@@ -0,0 +1,29 @@
...
+object CustomMessage {
+  def decoder[T](field: String)(
+      implicit ev: ConfDecoder[T]): ConfDecoder[CustomMessage[T]] =
+    ConfDecoder.instance[CustomMessage[T]] {
+      case obj: Conf.Obj =>
+        (obj.get[T](field) |@| obj.get[String]("message")).map {
+          case (value, message0) => {
+            val message =
+              if (message0.split(System.lineSeparator).size > 1) {
+                System.lineSeparator + message0.stripMargin
```
olafurpg: Is stripMargin necesary? We I think a `trim` should be enough, so you can write configuration like this
```
Disable {
  message = """
Are you sure?

blah
blah
"""
}
```

scalafix-core/shared/src/main/scala/scalafix/internal/config/CustomMessage.scala
```diff
@@ -0,0 +1,29 @@
...
+case class CustomMessage[T](value: T, message: Option[String])
+object CustomMessage {
+  def decoder[T](field: String)(
+      implicit ev: ConfDecoder[T]): ConfDecoder[CustomMessage[T]] =
+    ConfDecoder.instance[CustomMessage[T]] {
+      case obj: Conf.Obj =>
+        (obj.get[T](field) |@| obj.get[String]("message")).map {
+          case (value, message0) => {
+            val message =
+              if (message0.split(System.lineSeparator).size > 1) {
```
olafurpg: System.lineSeparator will produce inconsistent behavior across OS since `.scalafix.conf` contains the same bytes, we should use "\n" instead. `.split().size` can also be replaced with `"".contains('\n')`

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -27,3 +27,40 @@ Disable.symbols = [
...
+## Custom error messages
+
+It's possible to provide a custom error message.
+
+```
+scala/test/Disable.scala:47: error: [Disable.get]
+Option.get is the root of all evils
+
+If you really want to do this do:
+// scalafix:off Option.get
```
olafurpg: Advertise `// scalafix:ok get` instead of verbose off/on?

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -27,3 +27,40 @@ Disable.symbols = [
...
+
+```
+rules = Disable
+Disable.symbols = [
+  "scala.Any.asInstanceOf"
+  "test.Disable.D.disabledFunction"
+  {
+    symbol = "scala.Option.get"
+    message = 
+"""|Option.get is the root of all evils
```
olafurpg: Ditto

scalafix-tests/input/src/main/scala/test/Disable.scala
```diff
@@ -1,9 +1,19 @@
-/*
+/* ONLY
```
olafurpg: CI fails with "ONLY" ;)

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -27,3 +27,40 @@ Disable.symbols = [
   "scala.Any.asInstanceOf"
 ]
 ```
+
+_Since 0.5.3_
```
olafurpg: 0.5.4? Should this be below "Custom error messages"?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,11 +1,26 @@
...
+import scalafix.internal.util.SymbolOps
 
-import MetaconfigPendingUpstream.XtensionConfScalafix
+case class DisableConfig(symbols: List[CustomMessage[Symbol.Global]] = Nil) {
+  def allSymbols = symbols.map(_.value)
+
+  private val messageFromSymbol: Map[String, String] =
+    symbols
+      .flatMap(custom =>
+        custom.message.map(message => (custom.value.syntax, message)))
```
olafurpg: `custom.normalize.syntax`?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,11 +1,26 @@
...
 import metaconfig.ConfDecoder
+import MetaconfigPendingUpstream.XtensionConfScalafix
 import org.langmeta.Symbol
+import scalafix.internal.util.SymbolOps
 
-import MetaconfigPendingUpstream.XtensionConfScalafix
+case class DisableConfig(symbols: List[CustomMessage[Symbol.Global]] = Nil) {
+  def allSymbols = symbols.map(_.value)
+
+  private val messageFromSymbol: Map[String, String] =
```
olafurpg: `messageBySymbol`

scalafix-core/shared/src/main/scala/scalafix/internal/config/CustomMessage.scala
MasseGuillaume: both would work with stripMargin

olafurpg: Aha, of course 😅  sure we can keep stripMargin, but I think we should advertise the non-stripMargin syntax in the docs since it's less annoying IMO.

MasseGuillaume:
superseeded: https://github.com/scalacenter/scalafix/pull/448

PR #445, opened 2017-11-17
ExplicitResultTypes: skip local implicits 

### LINKED ISSUE #34: ExplicitImplicit: allow to omit types for local implicit vals
Type annotations for local implicit vals are optional for local implicit vals, see https://github.com/lampepfl/dotty/pull/1785

```scala
def foo = {
  // this is OK
  implicit val localImplicit = 2
}
```
ExplicitImplicit should have a flag to not insert type annotations for local implicit vals.

erikwj:
fixes #34

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
@@ -38,9 +38,6 @@ case class ExplicitResultTypes(
         ExplicitResultTypesConfig.default)
       .map(c => ExplicitResultTypes(index, c))
 
-  //FIXME
-  implicit val i = index
-
   // Don't explicitly annotate vals when the right-hand body is a single call
   // to `implicitly`. Prevents ambiguous implicit. Not annotating in such cases,
   // this a common trick employed implicit-heavy code to workaround SI-2712.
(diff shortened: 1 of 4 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -96,8 +96,12 @@ case class ExplicitResultTypes(
...
       ScalafixScalametaHacks.resetOrigin(tree).syntax
 
-    def isRuleCandidate[D <: Defn](defn: D, mods: Traversable[Mod], body: Term)(
-        implicit ev: Extract[D, Mod]): Boolean = {
+    def isRuleCandidate[D <: Defn](
+        defn: D,
+        nm: Name,
+        mods: Traversable[Mod],
+        body: Term,
+        printLocalTypes: Boolean)(implicit ev: Extract[D, Mod]): Boolean = {
```
olafurpg: is `printLocalTypes` required? Seems `config.skipLocalImplicits` is only used as the argument

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -112,11 +116,16 @@ case class ExplicitResultTypes(
...
         defn.parent.exists(_.is[Template])
 
-      isImplicit || {
-        isLocal &&
+      def isLocal = nm.symbol match {
+        case Some(value) => value.isInstanceOf[scala.meta.Symbol.Local]
+        case None => false
+      }
+
+      isImplicit && (if (printLocalTypes) !isLocal else true) || {
```
olafurpg: I would prefer to keep this boolean condition as simple as possible and delegate logic to local defs. We can move this if/else into `isLocal`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -125,16 +134,16 @@ case class ExplicitResultTypes(
     }
 
     ctx.tree.collect {
-      case t @ Defn.Val(mods, _, None, body)
-          if isRuleCandidate(t, mods, body) =>
+      case t @ Defn.Val(mods, Seq(Pat.Var(name)), None, body)
```
olafurpg: nitpick, scalameta trees only use `List`, conventionally we would pattern match this like `Pat.Var(name) :: Nil`

website/src/main/tut/docs/rules/ExplicitResultTypes.md
```diff
@@ -19,11 +19,23 @@ implicit val tt = liftedType
...
+  implicit val localImplicit = 2
+}
+```
+
 ### Configuration
 
 ```scala
-ExplicitResultTypes.memberKind = [Val, Def, Var]
-ExplicitResultTypes.memberVisibility = [Public, Protected]
+ExplicitResultTypes.memberKind = [Val, Def, Var] // empty by default
```
olafurpg: Nice 👍

PR #453, opened 2017-11-24
Fix #452, use flexible scala dialect.

### LINKED ISSUE #452: Use Scala212 dialect by default
Currently we default  to 2.11 which doesn't support trailing commas.

olafurpg:
Scalafmt uses the same dialect with the only difference that we don't
treat `inline` as a keyword here (I think that was a premature choice in
scalafmt)


MasseGuillaume:
LGTM. Why we have a default dialect?

olafurpg:
The default dialect was primarily used for syntactic rules but we also use them for semantic rules since typically you just care about parsing as mucha s possible.

PR #454, opened 2017-11-24
Refactor build to staticly define scala versions.

MasseGuillaume:


build.sbt
```diff
@@ -113,32 +125,46 @@ lazy val `scalafix-sbt` = project
...
-  )
+  ))
 
-lazy val testsInput = project
-  .in(file("scalafix-tests/input"))
-  .settings(
+lazy val testsShared211 = testsShared(scala211)
+lazy val testsShared212 = testsShared(scala212)
+
+lazy val inputOutputSetting = Seq(
```
olafurpg: Custom settings should be in ScalafixBuild.scala, build.sbt should only contain project definitions.

build.sbt
```diff
@@ -9,59 +9,71 @@ name := "scalafixRoot"
 onLoadMessage := s"Welcome to Scalafix ${version.value}"
 noPublish
 
-lazy val diff = crossProject
-  .in(file("scalafix-diff"))
-  .settings(
+val diff = MultiScalaCrossProject(
```
olafurpg: Can we use something more ergonomic like `val diff = scalafixProject("diff")`?

MasseGuillaume:
Ready for review.

build.sbt
```diff
@@ -2,65 +2,143 @@ import sbt.ScriptedPlugin
 import sbt.ScriptedPlugin._
 import Dependencies._
 
-version.in(ThisBuild) ~= { old: String =>
```
olafurpg: This line should stay, it's useful to testing the sbt plugin and patching broken releases

build.sbt
```diff
@@ -2,65 +2,143 @@ import sbt.ScriptedPlugin
...
+lazy val allScala211Projects: List[ProjectReference] =
+  scala211Projects ++ scalaFixedProjects
+
+lazy val scala211ProjectsDependencies: List[ClasspathDep[ProjectReference]] =
+  scala211Projects.map(ClasspathDependency(_, None))
+
+lazy val scalafix = project
+  .in(file("."))
+  .settings(
+    moduleName := "root",
```
olafurpg: I avoid "root" because it makes difficult to find the project in the "Open recent" tab in IJ 
screen shot 2017-11-25 at 09 14 52

build.sbt
```diff
@@ -2,65 +2,143 @@ import sbt.ScriptedPlugin
...
+      allJSSettings,
+      npmDependencies in Compile += "diff" -> "3.2.0"
+    )
+    .jsConfigure(_.enablePlugins(ScalaJSBundlerPlugin))
+)
+
+val diff211 = diff(scala211)
+val diff212 = diff(scala212)
+
+lazy val diff211JVM = diff211.jvm
```
olafurpg: Remind me we should move `diff` to `scalacenter/diff` since it's barely related to scalafix.

build.sbt
```diff
@@ -2,65 +2,143 @@ import sbt.ScriptedPlugin
...
   .settings(
-    moduleName := "scalafix-core",
+    moduleName := "scalafix211",
+    noPublish,
+    scalaVersion := scala211
+  )
+  .aggregate(allScala211Projects: _*)
+  .dependsOn(scala211ProjectsDependencies: _*)
+
+val diff = MultiScalaCrossProject(
```
olafurpg: Running the test suite from IntelliIJ fails on something related to diff

screen shot 2017-11-25 at 09 24 55

This is not a blocker, I personally don't even run tests from intellij but I know many contributors do.

olafurpg: it could be that the tests didn't run in IJ before this PR either, but I know they ran at some point 😅

project/plugins.sbt
```diff
@@ -1,4 +1,3 @@
-addSbtPlugin("com.eed3si9n" % "sbt-doge" % "0.1.5")
```
olafurpg: 🎉

PR #458, opened 2017-11-25
Upgrade binaries

olafurpg:
- scalafix 0.5.6
- scalafmt 1.3.0


gabro:
do we have a way to make sure this doesn't happen again? Maybe the CI can commit the new binary upon releases (it should be easy with a deploy key)

olafurpg:
Automating this would be very nice indeed, the problem is that this can't be done until ~30 minutes after a release. I feel like a release checklist could do enough here, wdyt?

olafurpg:
I've opened https://github.com/scalacenter/scalafix/issues/459 to discuss how to improve the release process.

PR #465, opened 2017-11-27
Drop-in

MasseGuillaume:
fix #465


scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
```diff
@@ -18,7 +20,8 @@ case class ScalafixConfig(
     reporter: ScalafixReporter = ScalafixReporter.default,
     patches: ConfigRulePatches = ConfigRulePatches.default,
     dialect: Dialect = ScalafixConfig.DefaultDialect,
-    lint: LintConfig = LintConfig.default
+    lint: LintConfig = LintConfig.default,
+    diffs: Option[List[GitDiff]] = None
```
olafurpg: Is this configurable via .scalafix.conf? Is there a way we can keep GitDiff out RuleCtx/ScalafixConfig?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -37,6 +37,7 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
   lazy val comments: AssociatedComments = AssociatedComments(tokens)
   lazy val input: Input = tokens.head.input
   lazy val escapeHatch = EscapeHatch(tree, comments)
+  lazy val diffDisable = config.diffs.map(diffs => new DiffDisable(diffs))
```
olafurpg: If possible I think we should try to keep this out of RuleCtxImpl

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
```diff
@@ -144,7 +144,23 @@ case class ScalafixOptions(
...
+    @HelpMessage(
+      "If set, only apply scalafix to added and edited files in git diff against master.")
+    diff: Boolean = false,
+    @HelpMessage(
+      "If set, only apply scalafix to added and edited files in git diff against a provided branch.")
+    diffBranch: Option[String] = None,
+    @HelpMessage(
+      """|If set, only scalafix to added and edited files in a git diff provided in the stdin
+         |
+         |scalafix --diff
```
olafurpg: Can we move these examples to the list of other examples in Cli.helpMessage?

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,173 @@
...
+
+    def commit(): Unit = {
+      git(s"commit -m 'r$revision'")
+      revision += 1
+    }
+
+    private def git(argsRaw: String): Unit = {
+      val args = argsRaw.split(' ').toList
+      import scala.collection.JavaConverters._
+      val builder = new ProcessBuilder(("git" :: args).asJava)
```
olafurpg: Duplication with scalafix-cli? Would be good to reuse the impl.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DiffDisable.scala
```diff
@@ -0,0 +1,21 @@
...
+import scalafix.LintMessage
+
+import org.langmeta.inputs.Input
+
+class DiffDisable(diffs: List[GitDiff]) {
+  val newFiles: Set[Input] = diffs.collect {
+    case NewFile(path) => Input.File(path)
+  }.toSet
+
+  // todo RANGE modified
```
olafurpg: Very excited about this!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
olafurpg: If the cli uses a different Rule.apply here https://github.com/scalacenter/scalafix/blob/bb0edd04fa59c291e4174f3900314b5f56e37e91/scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala#L125 then we can remove GitDiff out of ScalafixConfig and RuleCtx. What do you think?

scalafix-diff/shared/src/main/scala/scalafix/diff/GitDiffParser.scala
```diff
@@ -0,0 +1,193 @@
...
+  private val OriginalFile = "--- (.*)$".r
+  private val RevisedFile = "^\\+\\+\\+ (.*)$".r
+  private val RenameFrom = "^rename from (.*)$".r
+  private val RenameTo = "^rename to (.*)$".r
+  private val NoNewLine = "\\ No newline at end of file"
+
+  def path(relative: String): Path =
+    workingDir.resolve(relative)
+
+  private def accept(regex: Regex): Unit = {
```
olafurpg: Fancy, I like it 👍

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,173 @@
...
+
+    val obtained = noColor(cli.run("--diff"))
+
+    val expected =
+      s"""|$newCodeAbsPath:3: error: [DisableSyntax.keywords.var] keywords.var is disabled
+          |  var newVar = 1
+          |  ^
+          |""".stripMargin
+
+    assertNoDiff(obtained, expected)
```
olafurpg: Thanks for using this!

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,173 @@
...
+         |  // New code, no vars
+         |  var newVar = 1
+         |}""".stripMargin)
+    git.add(newCode)
+    git.commit()
+
+    val obtained = noColor(cli.run("--diff"))
+
+    val expected =
+      s"""|$newCodeAbsPath:3: error: [DisableSyntax.keywords.var] keywords.var is disabled
```
olafurpg: 😍 👏 Nice!

One trick to avoid the `$newCodeAbsPath` splice is to `replace(tmpDirectory.toString, "<...>")`. I find it makes the test readable, but this is fine too

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
MasseGuillaume: Sure, in user doc?

scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
MasseGuillaume: Nope only via cli, I guess I should add it to the config file as well.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -53,6 +54,16 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
     logger.elem(values: _*)
   }
 
+  def filterLintMessage(lints: List[LintMessage]): List[LintMessage] = {
+    val lints0 = escapeHatch.filter(lints)
+
+    diffDisable
+      .map(
```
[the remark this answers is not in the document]
MasseGuillaume: or maybe just a pattern match, I always forget the order in fold.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
MasseGuillaume: I'm not too sure what you mean here.

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
olafurpg: See `Cli.helpMessage` "examples" section

scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
olafurpg: I think we can keep this out of ScalafixConfig with the suggestion I made to use a different Rule.apply in CliRunner.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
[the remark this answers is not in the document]
olafurpg: pattern match is fine, `.map().getOrElse` is harder to follow.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
olafurpg: In CliRunner we are currently using the convenience method `Rule.apply(RuleCtx)` that runs ctx.printLintMessage`. However, if we do that by ourselves here then we can filter the messages in the cli without adding new private methods to `RuleCtx`. Does that make sense? `printLintMessage` is already a hack that should ideally be removed from `RuleCtx`, we should avoid adding new methods like that.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
MasseGuillaume: woot

we should also diff for color. ^

olafurpg: That is actually a problem with scalatest adding coloring, let's not worry about this now, we can open a ticket to track it.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
MasseGuillaume: The duplication is really small, not worth it.

MasseGuillaume:
@olafurpg ready for round 2

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
olafurpg: Can we remove this message here now that's it's explained with other examples.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DiffDisable.scala
```diff
@@ -0,0 +1,34 @@
...
+    case ModifiedFile(path, changes) => {
+      val ranges = changes.map {
+        case Range(start, length) => (start, start + length)
+      }
+      Input.File(path) -> IntervalSet(ranges)
+    }
+  }.toMap
+
+  def filter(lints: List[LintMessage]): List[LintMessage] = {
+    lints.filter(
```
olafurpg: Can we please break down this expression into a few steps?

scalafix-core/shared/src/main/scala/scalafix/internal/util/IntervalSet.scala
```diff
@@ -0,0 +1,27 @@
+package scalafix.internal.util
+
+import scala.collection.BitSet
+
+class IntervalSet(range: BitSet) {
+  def contains(elem: Int): Boolean =
+    range.contains(elem)
+
+  def intersect(start: Int, end: Int): Boolean = {
```
olafurpg: intersect**s**

scalafix-core/shared/src/main/scala/scalafix/rule/Rule.scala
```diff
@@ -69,24 +69,24 @@ abstract class Rule(ruleName: RuleName) { self =>
...
-  final def apply(ctx: RuleCtx, patch: Patch): String = {
+  final def apply(ctx: RuleCtx, patch: Patch): String =
     apply(ctx, Map(name -> patch))
-  }
-  final def apply(ctx: RuleCtx, patches: Map[RuleName, Patch]): String = {
-    val result = Patch(patches.values.asPatch, ctx, semanticOption)
-    // This overload of apply if purely for convenience
-    // Use fixWithName to iterate over LintMessage
-    // without printing to the console
-    Patch.lintMessages(patches, ctx).foreach(ctx.printLintMessage)
```
olafurpg: The old `apply` should have unchanged behavior, otherwise it would be a breaking change.

scalafix-core/shared/src/main/scala/scalafix/rule/RuleCtx.scala
```diff
@@ -51,7 +51,9 @@ trait RuleCtx extends PatchOps {
   // Private scalafix methods, subject for removal without notice.
   private[scalafix] def toks(t: Tree): Tokens
   private[scalafix] def config: ScalafixConfig
-  private[scalafix] def printLintMessage(msg: LintMessage): Unit
```
olafurpg: This should remain unchanged, but we should consider removing it in 0.6

scalafix-core/shared/src/main/scala/scalafix/rule/RuleCtx.scala
```diff
@@ -51,7 +51,9 @@ trait RuleCtx extends PatchOps {
...
+  private[scalafix] def filterLintMessage(
```
olafurpg: Was this change necessary for this PR?

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -122,9 +122,6 @@ object Patch {
       patches: Map[RuleName, Patch],
       ctx: RuleCtx): List[LintMessage] = {
 
-    // TODO(olaf): Remove this cast
-    val ctxImpl = ctx.asInstanceOf[RuleCtxImpl]
```
olafurpg: I'm eager to get rid of this cast but this change seems unrelated to the PR.

scalafix-diff/shared/src/main/scala/scalafix/diff/GitDiffParser.scala
```diff
@@ -0,0 +1,193 @@
...
+case class NewFile(path: Path) extends GitDiff
+case class ModifiedFile(path: Path, changes: List[Range]) extends GitDiff
+
+private[diff] case class Hunk(
+    originalLine: Int,
+    originalCount: Int,
+    revisedLine: Int,
+    revisedCount: Int)
+
+private[diff] object HunkExtractor {
```
olafurpg: Can this be `Hunk` without "Extractor"?

olafurpg: Can we move this to `internal.diff`? The main motivation is simplified mima maintenance, you will need to write a lot of filters for every new/removed field in `private[diff]` classes.

olafurpg: Can we also split Hunk/GitDiffParser/PeekableSource into different files.

scalafix-diff/shared/src/main/scala/scalafix/diff/GitDiffParser.scala
```diff
@@ -0,0 +1,193 @@
+package scalafix.diff
+
+import scala.util.matching.Regex
+
+import java.nio.file.Path
+
+case class Range(start: Int, length: Int)
+
+sealed trait GitDiff
```
olafurpg: Can we introduce a `case class GitDiffs(diffs: List[GitDiff])` instead of passing around `List[GitDiff]` everywhere?

I think we can then move `GitDiffParser` to `scalafix.internal.diff` and add a constructor via
```scala
object GitDiffs {
  def apply(input: Iterator[String], workingDir: Path): GitDiffs =
    new scalafix.internal.diff.GitDiffParser(input, workingDir).parse()
}
```
That way we minimize the surface of our public API.

scalafix-diff/shared/src/main/scala/scalafix/diff/GitDiffParser.scala
```diff
@@ -0,0 +1,193 @@
+package scalafix.diff
+
+import scala.util.matching.Regex
+
+import java.nio.file.Path
+
+case class Range(start: Int, length: Int)
```
olafurpg: `s/Range/GitChange/`? Range is already in Predef and it has different fields.

scalafix-diff/shared/src/main/scala/scalafix/diff/GitDiffParser.scala
```diff
@@ -0,0 +1,193 @@
...
+  }
+  def peek: Option[T] = peeker
+  def drop(n: Int): Unit = {
+    it.drop(n - 1)
+    setPeeker()
+  }
+}
+
+object GitDiffParser {
+  def show(diffs: List[GitDiff]): Unit = {
```
olafurpg: We could move this into `GitDiffs`.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,258 @@
...
+    test(name) {
+      val fs = new Fs()
+      val git = new Git(fs.workingDirectory)
+      val cli = new Cli(fs.workingDirectory)
+
+      body(fs, git, cli)
+    }
+  }
+
+  private class Fs() {
```
olafurpg: This looks super useful, can we move it out of the test suite?

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,258 @@
...
+        content: String,
+        op: StandardOpenOption): Unit = {
+      Files.write(path(filename), content.getBytes, op)
+    }
+
+    private def path(filename: String): Path =
+      workingDirectory.resolve(filename)
+  }
+
+  private class Git(workingDirectory: Path) {
```
olafurpg: Same here, make it reusable in other test suites.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    private def git(argsRaw: String): Unit = {
+      val args = argsRaw.split(' ').toList
+      import scala.collection.JavaConverters._
+      val builder = new ProcessBuilder(("git" :: args).asJava)
+      builder.directory(workingDirectory.toFile)
+      builder.redirectErrorStream(true)
+      val process = builder.start()
+      // val input = process.getInputStream()
```
olafurpg: remove

scalafix-tests/unit/src/test/scala/scalafix/tests/util/IntervalSetSpec.scala
```diff
@@ -0,0 +1,46 @@
+package scalafix.tests.util
+
+import org.scalatest.FunSuite
+import org.scalactic.source.Position
+
+import scalafix.internal.util.IntervalSet
+
+class IntervalSetSpec() extends FunSuite {
```
olafurpg: IntervalSetTests to follow convention. I may soon rename `s/Tests/Suite/` but for now let's try to keep it consistent.

scalafix-diff/shared/src/test/scala/scalafix/GitDiffParser.scala
```diff
@@ -0,0 +1,43 @@
+package scalafix.diff
+
+import java.nio.file.Paths
+import org.scalatest.FunSuite
+import scala.io.Source
+
+class GitDiffParserSpec extends FunSuite {
```
olafurpg: s/Spec/Tests/

Can we also try to have the filename match the test suite.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliGitDiff.scala
```diff
@@ -0,0 +1,258 @@
...
+import org.scalatest._
+import org.scalatest.FunSuite
+
+import scala.collection.immutable.Seq
+import scala.util._
+import scalafix.cli
+import scalafix.internal.cli.CommonOptions
+import scalafix.testkit.DiffAssertions
+
+class CliGitDiff() extends FunSuite with DiffAssertions {
```
olafurpg: CliGitDiffTests

scalafix-tests/unit/src/test/scala/scalafix/tests/util/IntervalSetSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  // format: off
```
olafurpg: Not convinced the minor vertical alignment is paying off here, can we please skip this? You can make the comments align by adding more spaces. I personally once bailed out on a PR to a project I was contributing to just because there was a lot of manual vertical alignment and I didn't know how to keep things pretty. It's a distraction.

MasseGuillaume:
@olafurpg ready for round 3.

scalafix-core/shared/src/main/scala/scalafix/rule/RuleCtx.scala
olafurpg: ?

scalafix-diff/shared/src/main/scala/scalafix/internal/diff/Hunk.scala
```diff
@@ -0,0 +1,33 @@
+package scalafix.internal.diff
+
+private[scalafix] case class Hunk(
```
olafurpg: `private [scalafix]` is implicit from `scalafix.internal`, but we can keep this

scalafix-tests/unit/src/test/scala/scalafix/tests/util/IntervalSetTest.scala
```diff
@@ -5,24 +5,27 @@ import org.scalactic.source.Position
...
+  test("intersects") {
     val set = IntervalSet((1, 2), (5, 8))
-    def in(start: Int, end: Int)(implicit pos: Position): Unit  = assert( set.intersect(start, end))
-    def out(start: Int, end: Int)(implicit pos: Position): Unit = assert(!set.intersect(start, end))
+    def in(start: Int, end: Int)(implicit pos: Position): Unit =
+      assert(set.intersects(start, end))
+    def out(start: Int, end: Int)(implicit pos: Position): Unit =
+      assert(!set.intersects(start, end))
 
+    // format: off
```
olafurpg: You can avoid this by padding more spaces before the comment
```
//                  //    0 1 2 3 4 5 6 7 8 9 
```

But it's not necessary.

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
olafurpg: Don't get me wrong, this change is nice and we can keep it here. I feel it could have have been a separate PR to make this PR easier to review. I admit I'm personally bad at keeping PRs focused on one change, so I promise to try and do better myself.

MasseGuillaume:
> Seems to be missing column numbers in the new tests https://travis-ci.org/scalacenter/scalafix/jobs/310001150#L1287

I don't know why, but locally it's correct.

> 'similarity index 100%'

corrected the regex, thanks

olafurpg:
Now there is a new error

```
$ scalafix-nightly -r DisableSyntax --config-str="DisableSyntax.keywords=[def]" --diff scalafix-cli     
error: java.lang.Exception: Unexpected: diff --git a/scalafix-diff/shared/src/test/scala/scalafix/GitDiffParserTest.scala b/scalafix-diff/shared/src/test/scala/scalafix/GitDiffParserTest.scala
scalafix.internal.diff.GitDiffParser.accept(GitDiffParser.scala:32)
```

It would be good to add tests for the failing cases.

olafurpg:
We should also try to use a `GitDiffException` instead of using `java.lang.Exception` (I'm guilty of not doing this myself, but I'm not proud of it)

MasseGuillaume:
Ready for round 4!

MasseGuillaume:
Travis just gives random results. :thinking:

olafurpg:
Got a new error testing out on a larger diff in scalameta/scalameta
```
$ git diff -U0 v1.0.0 | scalafix-nightly -r DisableSyntax --config-str="DisableSyntax.keywords=[def]" --diff-stdin

scalafix.internal.diff.GitDiffParser.parse(GitDiffParser.scala:75)
scalafix.internal.rule.DiffDisable$.apply(DiffDisable.scala:20)
scalafix.cli.CliRunner$Builder.<init>(CliRunner.scala:571)
```

scalafix-diff/shared/src/test/resources/git3.diff
```diff
@@ -0,0 +1,3181 @@
+diff --git a/project/Mima.scala b/project/Mima.scala
```
olafurpg: I'm not sure this is a sustainable approach, can try to make these tests more targeted to the discovered bugs?

MasseGuillaume:
We are going to use JGit, parsing git output is super fragile. Also, the Hunk header is not the finest granularity. We can get the edit list with the most accurate diff.

PR #466, opened 2017-11-28
Revive sbt 1.0 cross-build of sbt-scalafix

olafurpg:


MasseGuillaume:
I think you should add user documentation and say what is supported in sbt 1.0 (syntax) and what is supported in sbt 0.13 (syntax + semantic). If the semantic in 0.13 is not working properly we should delete it.

olafurpg:
Good idea @MasseGuillaume, updated docs.

project/ScalafixBuild.scala
```diff
@@ -174,10 +174,17 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         "scalafix/publishSigned" ::
         "scalafix211/publishSigned" ::
         "scalafix-sbt/publishSigned" ::
+        s"^^ $sbt1 scalafix-sbt/publishSigned" ::
         "sonatypeReleaseAll" ::
         s
     },
-    commands += Command.command("ci-slow") { s =>
+    commands += Command.command("ci-sbt") { s =>
+      // Don't run tests on sbt 1.0, blocked by
```
MasseGuillaume: It's not really blocked. You don't really want to support sbt 1.0. I think we should close https://github.com/scalameta/scalameta/issues/1157

olafurpg: It is blocked, if we fix that issue then we can run semantic rules for sbt 1.0. That issue is totally legitimate, it's just not high priority enough for me to work on it right now.

MasseGuillaume: I think we should close the issue if we are not going to work on it. There is not a lot of value to run scalafix on sbt build and it's really hard to do it right. Syntax rewrite can do the job.

olafurpg: There are more potential users of that issue than scalafix, building semanticdbs for *.sbt can be useful in other applications like code search. The actual work to complete that issue is not that much, but someone needs to find motivation and time to work on it.

PR #467, opened 2017-11-28
Include column number in linter messages

### LINKED ISSUE #460: moar output on lint failures
Some thoughts:

- the format should match what scalac outputs so that regex based file/line detection still works in a multitude of tools, c.f. https://github.com/ensime/emacs-sbt-mode/issues/130
- including the column information would be :100:  `pos.startColumn` is missing here https://github.com/scalacenter/scalafix/blob/0566b480109c449f84e0733f2d200806fd89752c/scalafix-core/shared/src/main/scala/org/langmeta/internal/ScalafixLangmetaHacks.scala#L15

olafurpg:
This makes it easy to extract the exact position from the cli to
display in the editor.

Fixes #460 

Review @fommil

### DIFF
--- a/scalafix-core/shared/src/main/scala/org/langmeta/internal/ScalafixLangmetaHacks.scala
+++ b/scalafix-core/shared/src/main/scala/org/langmeta/internal/ScalafixLangmetaHacks.scala
@@ -11,8 +11,10 @@ object ScalafixLangmetaHacks {
       message: String): String = {
     if (pos != Position.None) {
       val input = pos.input
+      val startLine = pos.startLine + 1
+      val startColumn = pos.startColumn + 1
       val header =
-        s"${input.syntax}:${pos.startLine + 1}: $severity: $message"
+        s"${input.syntax}:$startLine:$startColumn $severity: $message"
       val line = {
         val start = input.lineToOffset(pos.startLine)
         val notEof = start < input.chars.length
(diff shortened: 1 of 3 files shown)


scalafix-core/shared/src/main/scala/org/langmeta/internal/ScalafixLangmetaHacks.scala
```diff
@@ -11,8 +11,10 @@ object ScalafixLangmetaHacks {
       message: String): String = {
     if (pos != Position.None) {
       val input = pos.input
+      val startLine = pos.startLine + 1
+      val startColumn = pos.startColumn + 1
       val header =
-        s"${input.syntax}:${pos.startLine + 1}: $severity: $message"
+        s"${input.syntax}:$startLine:$startColumn $severity: $message"
```
fommil: you still need another `:` after the `startColumn`.

```
s"${input.syntax}:$startLine:$startColumn: $severity: $message"
```

fommil: and AFAICT scalac doesn't use `severity`

olafurpg: Fixed.

 scalac includes the severity before the message

```
foo.scala:2: error: type mismatch;
 found   : Int(1)
 required: String
  val x: String = 1
                  ^
```

PR #469, opened 2017-11-29
Sbt 1.0

olafurpg:


MasseGuillaume:
test failure in scripted test:
https://travis-ci.org/scalacenter/scalafix/jobs/308923874#L1766
https://github.com/scalacenter/scalafix/blob/master/scalafix-sbt/src/sbt-test/sbt-scalafix/scalafixEnable/build.sbt#L21

PR #470, opened 2017-12-01
debug

MasseGuillaume:


build.sbt
```diff
@@ -410,6 +410,11 @@ lazy val website = project
   )
   .dependsOn(testkit212, core212JVM, cli212)
 
+TaskKey[Unit]("debug-cred") := {
+  println(sys.env.get("SONATYPE_USERNAME"))
```
olafurpg: This won't give you results in PRs ;)

MasseGuillaume: why not ?

olafurpg: If these leak in PRs then that's a security vulnerability, it would mean any contributor that opens a PR can see our credentials. These secrets should only be available after code review and merge into master

Issue #380, opened 2017-09-28
Consider moving PatchOps out of RuleCtx


gabro:
Following a discussion in https://github.com/scalacenter/scalafix/issues/379#issuecomment-332743569 it would probably make sense to move the path operations into a static object, rather than having them on the `ctx` instance.

If there are ops that require configuration, we can pass it as a parameter.

olafurpg:
This was fixed in #737, the `PatchOps` methods now live a simple defs on the `Patch` object.
https://github.com/scalacenter/scalafix/blob/264a8ffc15e74c5ce994d913786bb9ffd4ea5929/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala#L114-L169

Issue #370, opened 2017-09-21
Provide some API for dealing with leading whitespaces


gabro:
When removing all the tokens of a line it's very common to be willing to remove also the leading whitespaces including the newline.

e.g.

removing `bar` from

```
foo
  bar
baz
```

should result in

```
 foo
 baz
```

as opposed to

```
foo
  
baz
```

Currently doing this is fairly tedious. We should provide a better API for this use case.

For example

```scala
ctx.removeLeadingWhitespace(t, includingNewline = true)
```

or even a modifier on existing methods (binary compat not going to be happy, though):

```scala
ctx.removeTokens(t.tokens, removeLeadingWhitespaces = true)
```

ShaneDelmore:
I’m bikeshedding here but given how common this is, what about shorter more common names like:
- trim
- trimLeft
- trimRight

olafurpg:
Fixed in #474

PR #474, opened 2017-12-05
Add methods for dealing with whitespaces

marcelocenerine:
When working on a scalafix rule for `scala/collection-strawman` I implemented my own logic to get rid of whitespaces. @olafurpg [mentioned in the review](https://github.com/scala/collection-strawman/pull/291/files#r149894285) that it would be nice to have such functionality supported by Scalafix. Later I realized that there was even an issue filed by @gabro to look into that: #370.

I explored a few alternatives to introduce the functionality. There might be better ways to achieve this, so I'd appreciate your feedback and will be willing to accept suggestions.

The code changes in this PR implement the approach that I personally prefer. There are other 2 alternatives available on separate branches (those need some polishing).

1 - `trimLeft` / `trimRight` / `trim` (https://github.com/scalacenter/scalafix/compare/master...marcelocenerine:trim):
These methods behave similarly to Java's `String.trim` in the sense that they eliminate not only spaces but newline tokens as well. This may not be flexible enough for users who just want to get rid of spaces/tabs and preserve blank lines.
The naming is based on what I originally suggested in the comment above as well as by @ShaneDelmore [here](https://github.com/scalacenter/scalafix/issues/370#issuecomment-331550417)

2 - `trimLeft` / `trimRight` / `trim` with a `includingNewline` parameter (code in the PR):
This is similar to the alternative above, except that the trimming methods take an optional parameter (set to `true` by default) that can be used to indicate when newlines should be preserved. This is based on @gabro's suggestion on #370

3 - `ctx.whitespaces.leftTo` / `rightTo` / `around` (https://github.com/scalacenter/scalafix/compare/master...marcelocenerine:trim_whitespaces):
`RuleCtx` already has ~30 methods and adding 3 more would make it more bloated (#380). This approach basically introduces an utility class that groups functions to find/select whitespaces (similar to `MatchingParens` and `AssociatedComments`). The output of its methods can be passed to `ctx.removeTokens` to perform the actual removal. Example:
```
ctx.removeTokens(ctx.whitespaces.leftTo(tk))
```

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
@@ -32,7 +32,7 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
   override def toString: String = syntax
   def toks(t: Tree): Tokens = t.tokens(config.dialect)
   lazy val tokens: Tokens = tree.tokens(config.dialect)
-  lazy val tokenList: TokenList = new TokenList(tokens)
+  lazy val tokenList: TokenList = TokenList(tokens)
   lazy val matchingParens: MatchingParens = MatchingParens(tokens)
   lazy val comments: AssociatedComments = AssociatedComments(tokens)
   lazy val input: Input = tokens.head.input
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/core/util/TokenListTest.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/core/util/TokenListTest.scala
@@ -6,21 +6,90 @@ import scalafix.util.TokenList
 
 class TokenListTest extends FunSuite {
 
-  val tokens = "".tokenize.get // this contains two tokens: beginningOfFile and endOfFile
-  val firstToken = tokens.head
-  val lastToken = tokens.last
-  val tokenList = new TokenList(tokens)
+  val emptyFileTokens = "".tokenize.get // this contains two tokens: BOF and EOF
+  val nonEmptyFileTokens =
+    """package foo
+      |
+      |object Bar {
+      | val baz   =   10
+      |}
+    """.stripMargin.tokenize.get
 
-  test("Prev returns the firstToken when is given the firstToken") {
-    assert(tokenList.prev(firstToken) == firstToken)
-    assert(tokenList.prev(tokens.last) == firstToken)
+  test("prev returns the preceding token") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.prev(emptyFileTokens.last) == emptyFileTokens.head)
   }
 
-  test("Slice gives an empty list with same token as inputs") {
-    assert(tokenList.slice(firstToken, firstToken) == Seq())
+  test("prev returns self if there is no preceding token") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.prev(emptyFileTokens.head) == emptyFileTokens.head)
   }
 
-  test("Next breaks gives the lastToken if it is given lastToken") {
-    assert(tokenList.next(lastToken) == lastToken)
+  test("next returns the following token") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.next(emptyFileTokens.head) == emptyFileTokens.last)
+  }
+
+  test("next returns self if there is no following token") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.next(emptyFileTokens.last) == emptyFileTokens.last)
+  }
+
+  test("slice returns an empty seq if there is no token in between") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.slice(emptyFileTokens.head, emptyFileTokens.head) == Seq())
+  }
+
+  test("slice returns all tokens between from/to tokens") {
+    val Some(kwObject) = nonEmptyFileTokens.find(_.is[Token.KwObject])
+    val Some(leftBrace) = nonEmptyFileTokens.find(_.is[Token.LeftBrace])
+    val tokenList = TokenList(nonEmptyFileTokens)
+
+    val slice = tokenList.slice(kwObject, leftBrace)
+    assert(slice.size == 3)
+    val Seq(space1, bar, space2) = slice
+    assert(space1.is[Token.Space])
+    assert(bar.syntax.equals("Bar"))
+    assert(space2.is[Token.Space])
+  }
+
+  test("leadingSpaces returns all spaces preceding a token") {
+    val Some(equals) = nonEmptyFileTokens.find(_.is[Token.Equals])
+    val tokenList = TokenList(nonEmptyFileTokens)
+
+    val spaces = tokenList.leadingSpaces(equals)
+    assert(spaces.size == 3)
+    val Seq(s1, s2, s3) = spaces
+    assert(s1 == tokenList.prev(equals))
+    assert(s2 == tokenList.prev(s1))
+    assert(s3 == tokenList.prev(s2))
+  }
+
+  test(
+    "leadingSpaces returns an empty seq if there are no spaces preceding a token") {
+    val Some(kwPackage) = nonEmptyFileTokens.find(_.is[Token.KwPackage])
+    val tokenList = TokenList(nonEmptyFileTokens)
+
+    assert(tokenList.leadingSpaces(kwPackage) == Seq())
+  }
+
+  test("trailingSpaces returns all spaces following a token") {
+    val Some(equals) = nonEmptyFileTokens.find(_.is[Token.Equals])
+    val tokenList = TokenList(nonEmptyFileTokens)
+
+    val spaces = tokenList.trailingSpaces(equals)
+    assert(spaces.size == 3)
+    val Seq(s1, s2, s3) = spaces
+    assert(s1 == tokenList.next(equals))
+    assert(s2 == tokenList.next(s1))
+    assert(s3 == tokenList.next(s2))
+  }
+
+  test(
+    "trailingSpaces returns an empty seq if there are no spaces following a token") {
+    val Some(rightBrace) = nonEmptyFileTokens.find(_.is[Token.RightBrace])
+    val tokenList = TokenList(nonEmptyFileTokens)
+
+    assert(tokenList.trailingSpaces(rightBrace) == Seq())
   }
 }
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -24,11 +24,14 @@ object Mima {
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "scalafix.rule.RuleCtx.printLintMessage"),
       ProblemFilters.exclude[ReversedMissingMethodProblem](
-        "scalafix.rule.RuleCtx.filterLintMessage"
-      ),
+        "scalafix.rule.RuleCtx.filterLintMessage"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "scalafix.cli.CliRunner.this"
-      )
+        "scalafix.cli.CliRunner.this"),
+      ProblemFilters.exclude[FinalClassProblem]("scalafix.util.TokenList"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "scalafix.util.TokenList.leadingSpaces"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "scalafix.util.TokenList.trailingSpaces")
     )
   }
 }
--- a/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
@@ -6,7 +6,7 @@ import scala.meta.tokens.Token
 import scala.meta.tokens.Tokens
 
 /** Helper to traverse tokens as a doubly linked list.  */
-class TokenList(tokens: Tokens) {
+final class TokenList private (tokens: Tokens) {
   def trailing(token: Token): SeqView[Token, IndexedSeq[Token]] =
     tokens.view(tok2idx(token), tokens.length).drop(1)
   def leading(token: Token): SeqView[Token, IndexedSeq[Token]] =
@@ -59,4 +59,13 @@ class TokenList(tokens: Tokens) {
     }
   }
 
+  def leadingSpaces(token: Token): SeqView[Token, IndexedSeq[Token]] =
+    leading(token).takeWhile(_.is[Token.Space])
+
+  def trailingSpaces(token: Token): SeqView[Token, IndexedSeq[Token]] =
+    trailing(token).takeWhile(_.is[Token.Space])
+}
+
+object TokenList {
+  def apply(tokens: Tokens): TokenList = new TokenList(tokens)
 }


MasseGuillaume:
Hi Marcelo, this is a really useful contribution, thanks a lot!

The CI is red because of mima
https://travis-ci.org/scalacenter/scalafix/jobs/311647980#L4256-L4269

the trait `PatchOps` has 3 new methods (`trimLeft`, `trimRight`, `trim`) and this is a public api. This is not super useful because the only implementation is via RuleCtix/RuleCtxImpl. Therefore, you can exclude them in the Mima check: 

```scala
ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.patch.PatchOps.trim")
ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.patch.PatchOps.trimRight")
ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.patch.PatchOps.trimLeft")
```

in https://github.com/scalacenter/scalafix/blob/master/project/Mima.scala#L7

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -93,4 +93,85 @@ class PatchSuite extends SyntacticRuleSuite {
...
+  val trimLeftIncludingNewLineRule: Rule = Rule.syntactic("trimLeft") { ctx =>
+    ctx.tokens.find(_.is[Token.KwVal]).map(ctx.trimLeft(_)).asPatch
+  }
+
+  check(
+    trimLeftIncludingNewLineRule,
+    "trimLeft removes spaces and newline before `val`",
+    original,
+    """// Foobar
+      |object a {val x = 2
```
olafurpg: What happens if there is a // comment after the open curly brace? We should not remove newlines trailing such comments.

olafurpg: We could test this by trimming left of "object"

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -87,6 +88,20 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
+      tokenList.leading(token).takeWhile(canBeTrimmed(_, includeNewLine)))
+  def trimRight(token: Token, includeNewLine: Boolean = true): Patch =
+    doRemoveTokens(
+      tokenList.trailing(token).takeWhile(canBeTrimmed(_, includeNewLine)))
+  def trim(token: Token, includeNewLine: Boolean = true): Patch =
+    trimLeft(token, includeNewLine) + trimRight(token, includeNewLine)
+  private def canBeTrimmed(tk: Token, includeNewLine: Boolean): Boolean =
+    tk match {
+      case Space() | Tab() | FF() => true
+      case Newline() if includeNewLine => true
```
olafurpg: I'm wondering if we should limit this to a single newline. A typical use case if you want to remove a statement then you usually want to keep the leading / trailing blank lines. What do you think?

scalafix-core/shared/src/main/scala/scalafix/patch/PatchOps.scala
```diff
@@ -126,4 +126,20 @@ trait PatchOps {
   def replaceSymbols(toReplace: Seq[(String, String)])(
       implicit noop: DummyImplicit,
       index: SemanticdbIndex): Patch
+
+  /**
+    * Remove whitespaces on the left side of the token. Use `includeNewLine` to
+    * indicate when newline tokens should be preserved.
+    */
+  def trimLeft(token: Token, includeNewLine: Boolean = true): Patch
```
olafurpg: What do you think about having a separate method to trim only spaces? trimLeftSpaces, trimSpaces, trimRightSpaces (name up for debate). Tabs can be ignored, I'm not aware of any Scala project using tabs for indent. 

I think it could result in a cleaner user API, boolean parameters are quite awkward to both write and read.

marcelocenerine: you are right, @olafurpg. Boolean parameters often indicate the need for two separate methods. My only concern is the explosion of methods (potentially 6 new ones) in the `RuleCtx` api (those would be inherited from `PatchOps`). If you think that's not a concern I will go for it.

olafurpg: I think the eventual home for these methods will be simple defs in `object Patch`. I personally think it's OK to have more methods if it buys us a cleaner api.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
gabro: Just to understand, given this example:

```scala
val x = 42

  "foo"

val y = "bar"
```

would the current implementation be able to produce this diff?

```diff
val x = 42
-
-  "foo"

val y = "bar"
```

so that the resulting code is

```scala
val x = 42

val y = "bar"
```

olafurpg: I think the implementation should produce this diff
```diff
val x = 42

-  "foo"

val y = "bar"
```

That would leave two blank lines, but it would also preserve a blank line in the following case instead of remove it
```diff
val x = 42

-  "foo"
val y = "bar"
```

gabro: Right. Probably the job of taking care of multiple blank lines belongs to a formatter

olafurpg: Yes, scalafmt would at least combine the two blank lines into a single blank line.

MasseGuillaume:
@marcelocenerine we are thinking of cutting a release today. Do you think you can address the comments above? This is a really valuable contribution and we would be happy to add it to the upcoming version.

marcelocenerine:
@MasseGuillaume, unfortunately I will only be able to address the comments in the evening. Please feel free to continue the work if you guys  want it to make the release cut today. Otherwise I will update the PR this evening

MasseGuillaume:
@marcelocenerine cool, I'm going to address the PR comments and include it in this release.

MasseGuillaume:
@marcelocenerine  don't know enough about token manipulation to address this PR for the release. I'm putting the ball back on your side for the next release. :)

marcelocenerine:
@MasseGuillaume, ok. I will update the PR tonight :)

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
marcelocenerine: In the example above, the result would be:
```
// Foobar object a {
```
I put some conditions to prevent this issue (comments in the format `/* ... */` are OK). The same would occur when trimming on the right side of a comment though....

Comments are not the only case where trimming would cause an invalid output. Example:

```
object Foo {
   val a = 0
   val b = 0
}
```

`ctx.trimLeft(bValToken)`

Output:

```
object Foo {
   val a = 0 val b = 0
}
```

I feel that the current behavior for newlines is not very intuitive and useful in practice. I can't think of a valid scenario where I would like to get rid of the left/right newline and move the remaining code on a particular line (or line below) up. Is there any use case that comes to your mind? From the example given above and the use case @gabro described in #370, maybe the right API would be sth like `ctx.removeLine(token)` (`trim spaces` is still useful). What do you think?

marcelocenerine:
@olafurpg thanks for your suggestions!!! I updated the PR addressing your comments, but I think we should discuss more on the behavior for newlines. Please see my comment: https://github.com/scalacenter/scalafix/pull/474#discussion_r155696870

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
```diff
@@ -87,6 +89,51 @@ case class RuleCtxImpl(tree: Tree, config: ScalafixConfig) extends RuleCtx {
...
+        if (!isSingleLineComment(prev3)) builder += (prev1, prev2)
+      } else if (prev1.is[Newline] && !isSingleLineComment(prev2))
+        builder += prev1
+    }
+    removeTokens(builder.result())
+  }
+  private def isSingleLineComment(tk: Token): Boolean =
+    tk.is[Comment] && tk.text.startsWith("//")
+  private def isWindowsNewline(tk1: Token, tk2: Token): Boolean =
+    tk1.is[CR] && tk2.is[LF]
```
marcelocenerine: I ran tests on Windows with a file containing CR+LF characters and it seems that scalameta ignore the CR ones, keeping only the LF tokens

marcelocenerine:
Hey @olafurpg, did you get a chance to take a look at my latest changes as well as my comment https://github.com/scalacenter/scalafix/pull/474#discussion_r155696870 ? 

I created this branch https://github.com/scalacenter/scalafix/compare/master...marcelocenerine:trim_normalize with an alternative implementation (`removeLine` instead of trimming newlines) as well as methods to normalize spaces. Please see what you think ;)

olafurpg:
Hey @marcelocenerine so sorry, I have maybe 2-3 different pending reviews for this PR that I never got around to submit. Every time I look at it I start writing tons of comments, then I change my mind and 20 minutes I decide to "look at it again later" 😅  

I took a look at your trim_normalize branch and I'm not convinced about `normalize*` and `removeLine`. Remove line seems particularly dangerous, since it doesn't handle multiline expressions. Normalize seems quite specific so I'm not sure it needs to be in the core API, unless you have a concrete use-case?

I'm wondering if we can simplify things by moving these methods to `TokenList`

```scala
class TokenList {
  def leading
  def leadingSpaces
  def leadingLine // includes spaces + first newline
  def trailing
  def trailingSpaces
  def trailingLine
```
I'm personally most excited about `leadingSpace`, since that one is most annoying to implement and it's quite common for me to remove statements like in https://github.com/scalacenter/scalafix/blob/8bd026de8e6610934ab2cad70f811592f764fc98/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala#L44
Still, that means users will need to feed the output of those methods to `removeTokens`, would that defeat the purpose?

OK, I'm submitting this comment before I change my mind again 😂

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RuleCtxImpl.scala
olafurpg: We have some windows line break related issues that I have never properly looked into https://github.com/scalameta/scalameta/issues/443

marcelocenerine:
Hey @olafurpg, thanks for your feedback. Your hesitation highlights the fact that this PR needs more work (especially when it comes to handling newlines) 😄

As a rule author I'd be more interested in the `leadingSpaces`/`trailingSpaces` for use cases where I know what tokens can precede/follow a given token that I want to remove/replace (silly example: input: `1 + 2` ; output: `1.+(2)`). If I have that awareness and don't care about whitespaces in between, then the leading/trailing spaces methods would work just fine. For the example above, I wouldn't even care whether the whitespaces are spaces, newlines, tabs or a mix of everything.

I think that removing newlines is dangerous unless the rule is aware of what tokens exist in the prev/next line as well as in the current line. As you pointed out, `removeLine` is indeed dangerous if a rule doesn't have that awareness. I think the same would be true for `leadingLine`/`trailingLine`. They would only be useful if users knew that either the current line could be entirely removed or whatever tokens remained in the current line could be safely moved up/down (e.g.: multiline expressions).

I'd be more interested in methods to find the prev/next newline character instead and then write my own logic to slice(newline, token) and figure out what is in between or what precedes/follows the newline token.

```
def nextNewline: Option[LF] // or Option[Token] or else...
def prevNewline: Option[LF]
```

or maybe a method that returned all leading/trailing tokens in the same line.

What do you think? Should I update the PR following your suggestion? Or do the above instead of `leadingLine`/`trailingLine`? Or just focus on spaces for now? 

>Still, that means users will need to feed the output of those methods to removeTokens, would that defeat the purpose?

To be honest, I'd prefer to have a rich API focused in finding/selecting tokens and a cleaner API with basic operations for adding/removing/replacing things (`PatchOps`).

Alternatively, given that additional whitespaces between tokens are irrelevant to determine the correctness of a program (just an aesthetic concern), should rules just ignore them and let a code formatting tool finish the job?

olafurpg:
> I'd be more interested in methods to find the prev/next newline character instead and then write my own logic to slice(newline, token) and figure out what is in between or what precedes/follows the newline token so I could determine if they could be removed.

Would `prevNewline` skip over non-whitespace tokens? Example, would it return None or Some here?
```scala
{
  val x = <<prevNewline>>
```
If it returns the newline after `{` (implemented as `.find(_.is[Newline])`) then it would be less useful IMO.

> To be honest, I'd prefer to have a rich API focused in finding/selecting tokens and a cleaner API with basic operations for adding/removing/replacing things (PatchOps).

I'm glad we agree there!

> Alternatively, given that additional whitespaces between tokens are irrelevant to determine the correctness of a program (just an aesthetic concern), should rules just ignore them and let a code formatting tool finish the job?

Trickier cases I think we should delegate to a formatter, but for some simple rewrites I think it would be nice to offer a best-effort by for example trimming enclosing spaces, for example in #513 

OK, here is my concrete proposal:

-  `TokenList.leadingSpaces/trailingSpaces`
- Make `TokenList` final, adding relevant filters in Mima.

It seems we are still unsure what to do about lines so we can give that a bit more time to think and leave it for another PR. I think `leadingLine(Token/Tree)` (without trailing option) would be a nice complement to `ctx.removeTokens(leadingLine(tree) ++ tree.tokens)`.

marcelocenerine:
>Would prevNewline skip over non-whitespace tokens?

yeah, I was thinking something like that but, like you said, it's already easy to achieve that goal though the current api. Except if we want to save users from having to deal with the intricacies of Windows/Unix line breaks (as long as scalameta cared about that difference).

Your proposal sounds good to me. In my short experience with scalafix I have came across scenarios where I missed utilities to get rid of whitespaces. I wanted to cover the other use case in #370 (newlines) but I think that it needs some more thought so that whatever comes out of it is generic enough for people to use.

I will update the PR shortly 😅

scalafix-tests/unit/src/test/scala/scalafix/tests/core/util/TokenListTest.scala
```diff
@@ -6,21 +6,90 @@ import scalafix.util.TokenList
...
+  test("next returns self if there is no following token") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.next(emptyFileTokens.last) == emptyFileTokens.last)
+  }
+
+  test("slice returns an empty seq if there is no token in between") {
+    val tokenList = TokenList(emptyFileTokens)
+    assert(tokenList.slice(emptyFileTokens.head, emptyFileTokens.head) == Seq())
+  }
+
```
marcelocenerine: it's kinda off topic but I think it's worth mentioning: when I was working on the previous implementation I found 2 things in the API that didn't behave as I expected (http://wiki.c2.com/?PrincipleOfLeastAstonishment):

- `prev`/`next` returns the input token if it is the head/last of the tokenList. Would `Option[Token]` make more sense?
- `slice(from, to)` doesn't behave as in Scala's std collection given that `from` is not inclusive. For instance, while removing tokens what could take a single step (`ctx.removeTokens(tokenList.slice(tk, last)`) actually requires 2 (`ctx.removeToken(tk) + ctx.removeTokens(tokenList.slice(tk, last)`) and is really easy to introduce bugs if users don't dive into the implementation of `slice`

marcelocenerine:
The PR is updated now.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/util/TokenListTest.scala
olafurpg: These are great points, I agree both of these can be improved. Can you open tickets? I am open to fix `slice` immediately (we can even implement is using slice on views from the stdlib) while prev/next will have to wait until v0.6

marcelocenerine: Awesome! Here are the issues I created:
https://github.com/scalacenter/scalafix/issues/516
https://github.com/scalacenter/scalafix/issues/517

PR #475, opened 2017-12-05
Move globalSettings to projectSettings (fix #468)

### LINKED ISSUE #468: publishSigned no longer works
```
sbt diffJVM/publishSigned
... 
[error] (diffJVM/*:publishSigned) java.io.IOException: PUT operation to URL https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-diff_2.12/0.5.6/scalafix-diff_2.12-0.5.6-sources.jar failed with status code 400: Bad Request
```

Can you reproduce @MasseGuillaume ? Any chance this is related to the build refactoring?

MasseGuillaume:
for some reason sbt is not picking up `globalSettings`

### DIFF
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -160,62 +160,12 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   lazy val adhocRepoCredentials = sys.props("scalafix.repository.credentials")
   lazy val isCustomRepository = adhocRepoUri != null && adhocRepoCredentials != null
 
-  override def globalSettings: Seq[Def.Setting[_]] = List(
-    stableVersion := version.in(ThisBuild).value.replaceAll("\\-.*", ""),
-    scalacOptions ++= compilerOptions,
-    scalacOptions in (Compile, console) := compilerOptions :+ "-Yrepl-class-based",
-    libraryDependencies += scalatest.value % Test,
-    testOptions in Test += Tests.Argument("-oD"),
-    updateOptions := updateOptions.value.withCachedResolution(true),
-    resolvers += Resolver.sonatypeRepo("releases"),
-    triggeredMessage in ThisBuild := Watched.clearWhenTriggered,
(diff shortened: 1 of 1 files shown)


project/ScalafixBuild.scala
```diff
@@ -247,15 +238,22 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
-  )
-
-  private val PreviousScalaVersion = Map(
-    "2.11.12" -> "2.11.11",
-    "2.12.4" -> "2.12.3"
-  )
-
-  override def projectSettings: Seq[Def.Setting[_]] = List(
+    ),
+    publishTo := {
```
olafurpg: What happens if you move these to `buildSettings`?

MasseGuillaume: What is the difference?

olafurpg: reload performance, `projectSettings` are evaluated for every project while `buildSettings` are only evaluated once for the build. scalafix has 64 projects, and adding 2.13/native support would increase that number even more.

MasseGuillaume: Done.

project/ScalafixBuild.scala
```diff
@@ -160,62 +160,12 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   lazy val adhocRepoCredentials = sys.props("scalafix.repository.credentials")
   lazy val isCustomRepository = adhocRepoUri != null && adhocRepoCredentials != null
 
-  override def globalSettings: Seq[Def.Setting[_]] = List(
```
olafurpg: Are all of these changes necessary to fix publishing? I thought it was only publishTo that was not picked up. I don't have a strong opinion, this change is fine, just trying to learn

MasseGuillaume: yeah publishTo, scmInfo all of this has an impact on publishSigned.

PR #477, opened 2017-12-05
Second stab at 2.12.4.

olafurpg:
Last time we upgraded to 2.12.4 the CI failed with a missing shapeless
implicit error from case-app. I can reproduce this error locally only on
the first `sbt compile`. Running the command the seconds time succeeds
compilation. I want to see if CI fails again, and investigate if
`-YdisableFlatCpCaching` can fix the error.

EDIT, `-YdisableFlatCpCaching` did not seem to fix the error. Let me try and remove ArgParserImplicits.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,5 @@
 -Dfile.encoding=UTF8
+-Xss8m
 -Xms1G
 -Xmx6G
 -XX:ReservedCodeCacheSize=250M
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
   // https://travis-ci.org/scalacenter/scalafix/jobs/303142842#L4658
   // as well as https://github.com/scala/bug/issues/10609
   def scala211 = "2.11.11"
-  def scala212 = "2.12.3"
+  def scala212 = "2.12.4"
   val currentScalaVersion = scala212
 
   def sbt013 = "0.13.6"


olafurpg:
cc/ @alexarchambault have you experienced issues resolving case-app/shapeless implicits when upgrading to 2.12.4? The error is
```
[error] /home/travis/build/scalacenter/scalafix/scalafix-cli/src/main/scala/scalafix/internal/cli/ArgParserImplicits.scala:14: could not find implicit value for parameter parser: caseapp.core.Parser[scalafix.internal.cli.ScalafixOptions]
[error]   val OptionsParser: Parser[ScalafixOptions] = Parser.apply[ScalafixOptions]
[error]                    
```
Full log: https://travis-ci.org/scalacenter/scalafix/jobs/312965607#L660

Most notably, the `cli` module compiled fine right before https://travis-ci.org/scalacenter/scalafix/jobs/312965607#L605 but the implicit resolution fails because of custom settings for scripted
```scala
    publishLocal := publishLocal
      .dependsOn(
        publishLocal in diff212JVM,
        publishLocal in core212JVM,
        publishLocal in reflect212,
        publishLocal in cli212)
      .value,
```
It's quite tricky to reproduce locally, but I think I've validated that the error is not caused by scaladoc. My best guess is that it's related to  `shapeless.Lazy` and running concurrent aggregated tasks via sbt. I have never experienced any problems with this before, however, it only started in 2.12.4.

alexarchambault:
@olafurpg Nothing specific to 2.12.4. But I remember having manually written some `Parser` in the past, because their derivation was failing for no obvious reason. I haven't investigated more than that…

olafurpg:
Thanks @alexarchambault 

I debugged the issue with -Xlog-implicits and that revealed a stack overflow error

```
[info]  [E518] scalafix-cli/src/main/scala/scalafix/internal/cli/ArgParserImplicits.scala
[info]         shapeless.this.Strict.mkStrict is not a valid implicit value for shapeless.Strict[caseapp.core.HListParser.Aux[Int with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("stackVerbosity")],Int] :: shapeless.HNil,Option[Int] :: shapeless.HNil,scala.collection.immutable.Nil.type :: shapeless.HNil,None.type :: shapeless.HNil,None.type :: shapeless.HNil,Some[caseapp.Hidden] :: shapeless.HNil,None.type :: shapeless.HNil,PT]] because:
[info]         hasMatchingSymbol reported error: exception during macro expansion:
[info]         java.lang.StackOverflowError
[info]         	at scala.tools.nsc.typechecker.StdAttachments.isMacroImplRef(StdAttachments.scala:153)
[info]         	at scala.tools.nsc.typechecker.StdAttachments.isMacroImplRef$(StdAttachments.scala:153)
```
I bumped up the stack size in .jvmopts and it seems to have fixed the issue. Someone is catching fatal errors in the compiler 😭 

Happy we can at least we can upgrade and 2.12.4 is not to blame 😄

PR #480, opened 2017-12-06
Teskit empty assertdiff

MasseGuillaume:
@olafurpg look what I found :p

I should always test edge cases.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
@@ -205,7 +205,7 @@ object Patch {
       case _: LintPatch => hasLintMessage = true
       case _ => onlyLint = false
     }
-    hasLintMessage && onlyLint
+    patch.isEmpty || hasLintMessage && onlyLint
   }
 
   private def foreach(patch: Patch)(f: Patch => Unit): Unit = {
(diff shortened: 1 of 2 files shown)


scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -205,7 +205,7 @@ object Patch {
       case _: LintPatch => hasLintMessage = true
       case _ => onlyLint = false
     }
-    hasLintMessage && onlyLint
+    patch.isEmpty || hasLintMessage && onlyLint
```
MasseGuillaume: I wrote an empty syntactic rule and I ran the SemanticSuite. It complained it was missing an output. This is because it checks if there are lint messages. It's obviously not the best way to fix this.

scalafix-testkit/src/main/scala/scalafix/testkit/CommentAssertion.scala
```diff
@@ -353,24 +353,41 @@ object AssertDiff {
...
+          .flatMap(_.headOption.map(_.assert))
+          .toList
+
+      val unexpected =
+        matrix.columns
+          .filter(_.forall(_.isWrong))
+          .flatMap(_.headOption.map(_.lintMessage))
+          .toList
+
+      val missmatch =
```
olafurpg: mismatch

PR #481, opened 2017-12-06
New Rule: No finalize

MasseGuillaume:


website/src/main/tut/docs/rules/NoFinalize.md
```diff
@@ -0,0 +1,24 @@
+---
+layout: docs
+title: NoFinalize
+---
+
+** syntactic **
```
gabro: Is this a convention we use elsewhere too?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
```diff
@@ -0,0 +1,34 @@
...
+    }.toSeq
+  }
+
+  private val errorCategory: LintCategory =
+    LintCategory.error(
+      "finalizer may be never called and have a severe performance penalty")
+
+  private def error(pos: Position): LintMessage = errorCategory.at(pos)
+
+  private def isFinalized(defn: Defn.Def): Boolean = {
```
gabro: wouldn't a quasi-quote be easier to match against in this case?

Something like

```
ctx.tree.collect {
  case q"..override def finalize(): Unit = $_" => // ...
}
```

MasseGuillaume: yes, but how do you match against all the cases?

```
override def finalize(): Unit = $_
override protected def finalize(): Unit = $_
override protected def finalize() = $_
```

website/src/main/tut/docs/rules/NoFinalize.md
MasseGuillaume: I just introduced it. I think since those rules/rewrite does not require compilation we should tag them.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
gabro: ah, fair point. I guess you can't do fuzzy matching with the quasi-quotes. 

Maybe something like

`case q"..$mods def finalize(): $tpeopt" if mods.exists(_isEqual(Mod.Override)) && ... =>`

but then it's probably not more readable than the existing code. Disregard my comment :D 👍

website/src/main/tut/docs/rules/NoFinalize.md
gabro: Makes sense. Maybe we can refine this to make it look like a tag, but definitely in a future PR.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
```diff
@@ -23,12 +23,11 @@ case object NoFinalize extends Rule("NoFinalize") {
...
-    val unit = Type.Name("Unit")
-    import defn._
-    mods.exists(_.isEqual(Mod.Override())) &&
-    defn.name.value == "finalize" &&
-    tparams == Nil &&
-    paramss == List(Nil) &&
-    decltpe.map(_.isEqual(unit)).getOrElse(true)
+    defn match {
+      case q"override def finalize(): Unit = $_" => true
+      case q"override protected def finalize(): Unit = $_" => true
```
gabro: missing

```scala
q"override def finalize() = $_"
```

?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
olafurpg: This implementation still does not handle cases like
```scala
class A {
    override def finalize() = println(1)
  }
```

There is no need to check `override` since it's not possible to define a method named "finalize"
```scala
@ class A {
    def finalize = println(1)
  }
cmd0.sc:2: overriding method finalize in class Object of type ()Unit;
 method finalize needs `override' modifier
  def finalize = println(1)
      ^
Compilation Failed
```

I propose to use Defn.Def instead
```scala
@ q"def finalize = 2" match {
    case Defn.Def(_, name @ q"finalize", _, Nil | Nil :: Nil, _, _) => name
  }
res5: Term.Name = Term.Name("finalize")

@ q"def finalize(x: Int) = 2" match {
    case Defn.Def(_, name @ q"finalize", _, Nil | Nil :: Nil, _, _) => name
  }
scala.MatchError: def finalize(x: Int) = 2 (of class scala.meta.Defn$Def$DefnDefImpl)
  ammonite.$sess.cmd6$.<init>(cmd6.sc:2)
  ammonite.$sess.cmd6$.<clinit>(cmd6.sc)
res4: Term.Name = Term.Name("finalize")
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
```diff
@@ -0,0 +1,33 @@
...
+case object NoFinalize extends Rule("NoFinalize") {
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+    ctx.tree.collect {
+      case defn: Defn.Def if isFinalized(defn) => error(defn.name.pos)
+    }
+  }
+
+  private val errorCategory: LintCategory =
+    LintCategory.error(
+      "finalizer may be never called and have a severe performance penalty")
```
olafurpg: > finalize may never get called and it incurs a performance penalty

Go any advice on what should the programmer should do instead? Maybe link to blog under `explain = "..."`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  private def error(pos: Position): LintMessage = errorCategory.at(pos)
```
olafurpg: Is this helper necessary? s/errorCategory/error/ and then `error.at(pos)`

website/src/main/tut/docs/rules/NoFinalize.md
```diff
@@ -0,0 +1,24 @@
...
+
+# NoFinalize
+
+Disallow implementing java.lang.Object.finalize.
```
olafurpg: Report an error when `java.lang.Object.finalize` is overridden.

website/src/main/tut/docs/rules/NoFinalize.md
```diff
@@ -0,0 +1,24 @@
...
+More info about finalize on this [blog](https://dzone.com/articles/javas-finalizer-is-still-there)
+
+_Since 0.5.6_
+
+...
+
+Example:
+
+```scala
+MyCode.scala:7: error: [NoFinalize]
```
olafurpg: Missing message?

MasseGuillaume: Maybe we should extract this from the test suite?

MasseGuillaume:
Ready for round 2

website/src/main/tut/docs/rules/NoFinalize.md
olafurpg: We should be able to generate all docs for rules automatically, tables with the configuration options, syntactic/semantic, caveats, example, since version number, etc.

Issue #489, opened 2017-12-07
sbt bugs as observed in ensime-server


fommil:
1. if there are no scala files in the project (e.g. java project)  we get `scalafix.cli.Cli$NonZeroExitCode: Got exit code InvalidCommandLineOption=8`
2. if I use `scalafixEnabled := false` in specific projects that I don't want to touch, the aggregated `scalafix RemoveUnusedImports` gives `scalafix.cli.Cli$NonZeroExitCode: Got exit code InvalidCommandLineOption=8` in the disabled projects instead of skipping them

olafurpg:
Good catch, thanks for reporting!

Pending fix for 1) in #490 

`scalafixEnabled` is a misnomer since what it currently means "do not automatically enable the semanticdb-scalac plugin". I agree `scalafixEnabled := false` should mean "`scalafix`does nothing". Is `disablePlugins(ScalafixPlugin)` a more idiomatic way to accomplish the latter? If so, then I propose we rename `scalafixEnabled` to `scalafixSemanticdbEnable`. I plan to make it false by default in 0.6, since I've found it difficult to make the automatic setup work for all builds.

fommil:
actually you're right, I should be disabling the plugin and the proposed name sounds better.

Issue #331, opened 2017-09-07
Support SBT `IntegrationTest` configuration


avdv:
Hi.

In my project, I have enabled the IntegrationTest config:

```
// build.sbt

Defaults.itSettings
```

I am using `addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.0-RC1")`.

When running `scalafix RemoveUnusedImports` or even `it:scalafix RemoveUnusedImports` there are a lot of warnings about the scala files of the IntegrationTest configuration, a la "No semanticdb associated with" them.

```
[info] Running scalafix -r RemoveUnusedImports --classpath /home/claudio/src/data/target/scala-2.12/classes --sourceroot /home/claudio/src/data /home/claudio/src/data/app /home/claudio/src/data/app
...
error: No semanticdb associated with /home/claudio/src/data/app/persistence/StencilMapperMongo.scala. Is --sourceroot correct?
```

Of course, the classpath is not correctly set, as the classes are written to `targetDirectory / scalaBinaryVersion / "it-classes"`.

I used 
```
scalafix in IntegrationTest := ScalafixPlugin.scalafixTaskImpl(IntegrationTest).evaluated
```

to workaround that problem and it is half-working now - some files are properly processed. Some are not.

Is it possible to set up the scalafix task appropriately for each configuration (`projectConfigurations: Seq[Configuration]`)?

\edit: Note, I also get some exceptions during the compilation stage:

```
[warn] failed to generate semanticdb for /home/claudio/src/data/test/it/helpers/TestApplication.scala:
[warn] org.scalameta.UnreachableError: this code path should've been unreachable
[warn] where tpe = helpers.TestApplication with org.specs2.SpecificationLike
[warn] 	at org.scalameta.UnreachableError$.raise(package.scala:40)
[warn] 	at scala.meta.internal.semanticdb.PrinterOps$SyntheticCodePrinter.printType(PrinterOps.scala:406)
[warn] 	at scala.meta.internal.semanticdb.PrinterOps$SyntheticCodePrinter.processTreePrinting(PrinterOps.scala:801)
[warn] 	at scala.meta.internal.semanticdb.PrinterOps$SyntheticCodePrinter.printTree(PrinterOps.scala:415)
[warn] 	at scala.tools.nsc.ast.Printers$TreePrinter.$anonfun$print$1(Printers.scala:33)
[...]

PR #490, opened 2017-12-08
Use unmanagedSources instead of unmanagedSourceDirectories.

### LINKED ISSUE #340: Use unmanagedSources instead of unmanagedSourceDirectories
See https://github.com/scalacenter/scalafix/issues/331#issuecomment-328434106

A good place to start looking at this is here https://github.com/scalacenter/scalafix/blob/c0a26983514599172b30e515c9b2029cd92e1f64/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L207

The motivation for using unmanagedSourceDirectories was to make the sbt console output pretty when debugging the flags passed to the scalafix-cli. If we do this change, we probably need to minimize the output here
https://github.com/scalacenter/scalafix/blob/c0a26983514599172b30e515c9b2029cd92e1f64/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L263

olafurpg:
- Fixes #340
- Fixes bug for java-only projects, see part 1 of #489

### DIFF
--- a/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
+++ b/scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
@@ -233,7 +233,7 @@ object ScalafixPlugin extends AutoPlugin {
         sources <- unmanagedSources.all(filter).value
         source <- sources
         if source.exists()
-        if canFormat(source)
+        if canFix(source)
       } yield source
       val options: Seq[String] = List("--classpath", classpath) ++ extraOptions
       scalafixTaskImpl(
@@ -294,7 +294,7 @@ object ScalafixPlugin extends AutoPlugin {
     }
   }
 
-  def canFormat(file: File): Boolean = {
+  private def canFix(file: File): Boolean = {
     val path = file.getPath
     path.endsWith(".scala") ||
     path.endsWith(".sbt")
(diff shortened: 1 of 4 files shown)


scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -285,13 +287,19 @@ object ScalafixPlugin extends AutoPlugin {
...
         val finalArgs = args ++ files.map(_.getAbsolutePath)
-        val nonBaseArgs = finalArgs.filterNot(baseArgs).mkString(" ")
+        val nonBaseArgs = args.filterNot(baseArgs).mkString(" ")
         log.info(s"Running scalafix $nonBaseArgs")
         main.main(finalArgs.toArray)
       }
     }
   }
 
+  def canFormat(file: File): Boolean = {
```
MasseGuillaume: `canFix` ?

scalafix-sbt/src/sbt-test/sbt-scalafix/cross-build/javaProject/src/main/java/example/Example.java
```diff
@@ -0,0 +1,7 @@
+package example;
+
+class Foo  {
```
MasseGuillaume: can you add a test like: https://github.com/scalacenter/scalafix/issues/331#issuecomment-327733840

> This may be related to the fact that in my project the integration test .scala files are in a sub-folder of the test/ source directory, e.g.:

olafurpg: Given the implementation uses directly `unmanagedSources` I think it would be an sbt bug if that failed now.

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
olafurpg: Done.

PR #494, opened 2017-12-10
Add regex to the DisableSyntax rule

MasseGuillaume:


scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
```diff
@@ -9,16 +9,40 @@ DisableSyntax.keywords = [
...
   var a = 1                 // assert: DisableSyntax.keywords.var
-  null                      // assert: DisableSyntax.keywords.null
   def foo: Unit = return    // assert: DisableSyntax.keywords.return
   throw new Exception("ok") // assert: DisableSyntax.keywords.throw
 
   "semicolon";              // assert: DisableSyntax.noSemicolons
   <a>xml</a>                // assert: DisableSyntax.noXml
 	                          // assert: DisableSyntax.noTabs
+
+  implicit class StringPimp(value: String) { // assert: DisableSyntax.[P|p]imp
```
olafurpg: Hmm, I wonder if `CustomMessage` should have an optional `id` field to avoid exposing complex messages like this into linter messages. Opened https://github.com/scalacenter/scalafix/issues/495

PR #497, opened 2017-12-11
Fix #493, handle synthetics and symbol signatures in Disable.

### LINKED ISSUE #493: Disable reports errors on symbol signatures
I want to disable usage of some class `Foo`, so I write 
```
rule = Disable
Disable.symbols = [mypackage.Tests.Foo]
```
and corresponding test
```
package mypackage

object Tests {
  class Foo
  class Bar extends Foo // assert: Disable.Foo
}
```

But that test throws a strange error 
```
_root_.mypackage.Tests.Foo#`<init>`()V.:1: error: [Disable.Foo] Foo is disabled
(): Foo
    ^
```
that's impossible to catch.

olafurpg:
Previously, Disable reported an error on the positions of any
ResolvedName to a disabled symbol, including Input.Synthetic and
Input.Denotation which are invisible and don't appear in the source
file. Now Disable ignores symbol signatures and for synthetics it
points a caret at the position where the synthetic got expanded
and includes the pretty-printed representation of the synthetic
to help make sense of the error message.

Review @vovapolu

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
@@ -30,21 +30,34 @@ final case class Disable(index: SemanticdbIndex, config: DisableConfig)
       .getOrElse("disable", "Disable")(DisableConfig.default)
       .map(Disable(index, _))
 
-  override def check(ctx: RuleCtx): Seq[LintMessage] =
-    ctx.index.names.collect {
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+    val buffer = List.newBuilder[LintMessage]
+    def add(name: ResolvedName): Unit = name match {
       case ResolvedName(
           pos,
           disabledSymbol(symbol @ Symbol.Global(_, signature)),
-          false) => {
-
-        val message =
-          config
-            .customMessage(symbol)
-            .getOrElse(s"${signature.name} is disabled")
-
-        errorCategory
+          false) =>
+        val (details, caret) = pos.input match {
+          case synthetic @ Input.Synthetic(_, input, start, end) =>
+            // For synthetics the caret should point to the original position
+            // but display the inferred code.
+            s" and it got inferred as `${synthetic.text}`" ->
+              Position.Range(input, start, end)
+          case _ =>
+            "" -> pos
+        }
+        val message = config
+          .customMessage(symbol)
+          .getOrElse(s"${signature.name} is disabled$details")
+        buffer += errorCategory
           .copy(id = signature.name)
-          .at(message, pos)
-      }
+          .at(message, caret)
+      case _ =>
     }
+    ctx.index.documents.foreach { document =>
+      document.names.foreach(add)
+      document.synthetics.foreach(_.names.foreach(add))
+    }
+    buffer.result()
+  }
 }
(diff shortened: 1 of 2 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -30,21 +30,34 @@ final case class Disable(index: SemanticdbIndex, config: DisableConfig)
       .getOrElse("disable", "Disable")(DisableConfig.default)
       .map(Disable(index, _))
 
-  override def check(ctx: RuleCtx): Seq[LintMessage] =
-    ctx.index.names.collect {
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+    val buffer = List.newBuilder[LintMessage]
```
vovapolu: Maybe make it immutable with `Option[LintMessage]` and `.flatMap`? Or performance is important here?

scalafix-tests/input/src/main/scala/test/Disable.scala
```diff
@@ -52,4 +56,10 @@ If you must Option.get, wrap the code block with
 ...
 // scalafix:on Option.get
 */
-}
\ No newline at end of file
+  val l: ListBuffer[Int] =
+    scala.collection.mutable.ListBuffer.empty[Int] // scalafix:ok Disable.ListBuffer
```
vovapolu: :thinking: why is it ok?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
olafurpg: Refactored to avoid builder.

scalafix-tests/input/src/main/scala/test/Disable.scala
olafurpg: `// scalafix:ok` is about the same as `// assert: `, since scalafix will report an error on `// scalafix:ok` if it's not used. Maybe we should consider removing `// scalafix: assert` 🤔 but assert also has the option to test the exact error message.

scalafix-tests/input/src/main/scala/test/Disable.scala
```diff
@@ -52,4 +56,9 @@ If you must Option.get, wrap the code block with
 ...
 // scalafix:on Option.get
 */
-}
\ No newline at end of file
+  val l: ListBuffer[Int] = scala.collection.mutable.ListBuffer.empty[Int] // assert: Disable.ListBuffer
+  List(1) + "any2stringadd" /* assert: Disable.any2stringadd
+  ^
+any2stringadd is disabled and it got inferred as `scala.Predef.any2stringadd[List[Int]](*)`
+  */
```
MasseGuillaume: so multiline asserts are useful after all ;-).

scalafix-tests/input/src/main/scala/test/Disable.scala
```diff
@@ -52,4 +56,9 @@ If you must Option.get, wrap the code block with
 ...
 // scalafix:on Option.get
 */
-}
\ No newline at end of file
+  val l: ListBuffer[Int] = scala.collection.mutable.ListBuffer.empty[Int] // assert: Disable.ListBuffer
+  List(1) + "any2stringadd" /* assert: Disable.any2stringadd
+  ^
+any2stringadd is disabled and it got inferred as `scala.Predef.any2stringadd[List[Int]](*)`
```
MasseGuillaume: Not blocking, but I don't think this is the most meaningful example.

PR #500, opened 2017-12-12
Add id for CustomMessage (fix #495)

### LINKED ISSUE #495: CustomMessage should include an `id: Option[String]` field
Context https://github.com/scalacenter/scalafix/pull/494/files#diff-439023eeb6466236b27366980c2da737R42

It would be nice if it was possible to provide a custom ID for disabled symbols/patterns/keywords. This would allow Disable to have more accurate ids like `Disable.Option.get` instead of `Option.get`. For `DisableSyntax.regex = [ "[P|p]imp"]` this would allow using `DisableSyntax.pimp` as an id.

MasseGuillaume:


scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,44 @@
...
+import metaconfig.{Conf, ConfError, ConfDecoder, Configured, Metaconfig}
+import org.langmeta._
+import scalafix.internal.config.MetaconfigPendingUpstream.XtensionConfScalafix
+
+import scala.language.implicitConversions
+
+class CustomMessage[T](value: T, message: Option[String], id: Option[String])
+
+object CustomMessage {
+  def getOption[T](obj: Conf.Obj, path: String)(
```
MasseGuillaume: https://github.com/olafurpg/metaconfig/pull/31

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,44 @@
...
+        value => ev.read(value).map(Some(_))
+      )
+      .getOrElse(Configured.Ok(None))
+
+  def decoder[T](field: String)(
+      implicit ev: ConfDecoder[T]): ConfDecoder[CustomMessage[T]] =
+    ConfDecoder.instance[CustomMessage[T]] {
+      case obj: Conf.Obj => {
+        (obj.get[T](field) |@|
+          getOption[String](obj, "message") |@|
```
MasseGuillaume: https://github.com/olafurpg/metaconfig/pull/31

replace with `obj.getOption[String]("message")`

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
```diff
@@ -11,6 +11,7 @@ DisableSyntax.noSemicolons = true
 DisableSyntax.noXml = true
 DisableSyntax.regex = [
   {
+    id = offensive
```
MasseGuillaume: I cannot use pimp here because the assert would trigger the regex again.

olafurpg: Interesting, I wonder if we can avoid this in general by making sure the offensive part does not follow `scalafix:ok/off`. Since we have the start offset, we should be able to handle this. I opened https://github.com/scalacenter/scalafix/issues/507 to track this.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
olafurpg: metaconfig v0.5.4 is out with `getOption` 🔥

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,47 @@
+package scalafix
+package config
+
+import metaconfig.{Conf, ConfError, ConfDecoder, Configured, Metaconfig}
+import org.langmeta._
+import scalafix.internal.config.MetaconfigPendingUpstream.XtensionConfScalafix
+
+import scala.language.implicitConversions
+
+class CustomMessage[T](
```
olafurpg: Now that this is in our public API, can we add 
```scala
type CustomMessage[T] = scalafix.config.CustomMessage[T]
val CustomMessage = scalafix.config.CustomMessage
```
in the scalafix package object? Rule authors should only need to `import scalafix._; import scala.meta._` to get started.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,47 @@
...
+    val value: T,
+    val message: Option[String],
+    val id: Option[String])
```
olafurpg: Can we add `withValue/withMessage/withId` helpers to make it more convenient to copy one field?

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,47 @@
...
+
+object CustomMessage {
```
olafurpg: Should we add a `apply(value: T) = new CustomMessage(value, None, None)` constructor?

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -0,0 +1,47 @@
...
+  def getOption[T](obj: Conf.Obj, path: String)(
```
olafurpg: We should remove this since it's public API now.

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
olafurpg: "offensive" is better anyways.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
MasseGuillaume: You only read from the CustomMessage, I don't think this is necessary.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
MasseGuillaume: ibid. you only read from the conf.

MasseGuillaume:
Ready for round 2.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
olafurpg: Right, I agree.

scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala
olafurpg: You're right, nevermind!

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,4 +1,5 @@
-package scalafix.internal.config
+package scalafix
```
olafurpg: I generally avoid this pattern since it can easily introduce cycles in your codebase. `import scalafix.config.CustomMessage` is fine in our own code. However, I'm not so consistent myself 😸

PR #503, opened 2017-12-13
Escape hatch on Patch (fix #368)

### LINKED ISSUE #368: Rule.apply: List[Patch]
Following a conversation with @xeno-by from his experience with the scalafix api. This is a proposal to replace `Rule.fix/check` with a single method `Rule.apply(ctx: RuleCtx): List[Fix]` where (roughly)

```scala
trait Fix {
  def message: LintMessage // LintMessage.empty by default
  def patch: Patch // Patch.empty by default
}
```

This change addresses several problems with the current design

- escape hatches #241, a single `Patch` can contain a sequence of `ctx.addLeft/addRight/remove` edits that need atomically trigger or not.  See example below, if `Baz` is inserted inside `foo(...)` along with an import at the top of a file, then that import should not be inserted since scalafix is disabled around `foo(...)`.

```scala
// import a.Baz to be inserted by scalafix
object a {
  // scalafix: off
  foo(...) // Baz to be inserted here by scalafix
  // scalafix: on
}
```

- one signature makes it easier to convert a linter into a rewrite (autofix) or vice-versa turn a rewrite into a linter
- avoids the need for sprinkling `.asPatch` in most cases.

I think it's possible to introduce this change while keeping compatibility with the current design, and we deprecate `Rule.fix/check`.

MasseGuillaume:
This is a really raw first pass.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedImports.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedImports.scala
@@ -28,6 +28,6 @@ case class RemoveUnusedImports(index: SemanticdbIndex)
   }
   override def fix(ctx: RuleCtx): Patch =
     ctx.tree.collect {
-      case i: Importee if isUnused(i) => ctx.removeImportee(i)
+      case i: Importee if isUnused(i) => ctx.removeImportee(i).atomic
     }.asPatch
 }
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
@@ -40,6 +40,8 @@ case class RemoveUnusedTerms(index: SemanticdbIndex)
   override def fix(ctx: RuleCtx): Patch =
     ctx.tree.collect {
       case i: Defn if isUnused(i) =>
-        tokensToRemove(i).fold(Patch.empty)(ctx.removeTokens)
+        tokensToRemove(i)
+          .fold(Patch.empty)(token => ctx.removeTokens(token))
+          .atomic
     }.asPatch
 }
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveXmlLiterals.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveXmlLiterals.scala
@@ -14,6 +14,7 @@ import scalafix.rule.RuleCtx
  *   <div>{ "Hello" }</div>
  *
  *   // after:
+ *   import scala.xml.quote._
  *   xml"<div>${ "Hello" }</div>"
  * }}}
  *
@@ -119,9 +120,11 @@ case object RemoveXmlLiterals extends Rule("RemoveXmlLiterals") {
       case xml @ Term.Xml(parts, _) =>
         val tripleQuoted = isMultiLine(xml) || parts.exists(
           containsEscapeSequence)
-        patchXml(xml, tripleQuoted)
+        patchXml(xml, tripleQuoted).atomic
     }.asPatch
 
+    // NB if all patches are not yet escaped here
+    // it can be empty and only have the import
     if (patch.nonEmpty) patch + importXmlQuote
     else patch
   }
--- a/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
@@ -48,12 +48,16 @@ sealed abstract class Patch {
   def ++(other: Iterable[Patch]): Patch = other.foldLeft(this)(_ + _)
   def isEmpty: Boolean = this == EmptyPatch
   def nonEmpty: Boolean = !isEmpty
+
+  /** Skip this entire patch if a part of it is disabled with // scalafix:off */
+  def atomic: Patch = AtomicPatch(this)
 }
 
 //////////////////////////////
 // Low-level patches
 //////////////////////////////
 trait LowLevelPatch
+
 abstract class TokenPatch(val tok: Token, val newTok: String)
     extends Patch
     with LowLevelPatch {
@@ -92,6 +96,7 @@ private[scalafix] object TreePatch {
 }
 
 // implementation detail
+private[scalafix] case class AtomicPatch(underlying: Patch) extends Patch
 private[scalafix] case class LintPatch(message: LintMessage) extends Patch
 private[scalafix] case class Concat(a: Patch, b: Patch) extends Patch
 private[scalafix] case object EmptyPatch extends Patch with LowLevelPatch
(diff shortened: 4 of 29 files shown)


scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -48,12 +48,14 @@ sealed abstract class Patch {
...
   def isEmpty: Boolean = this == EmptyPatch
   def nonEmpty: Boolean = !isEmpty
+  def atomic: Patch = AtomicPatch(this)
 }
 
 //////////////////////////////
 // Low-level patches
 //////////////////////////////
 trait LowLevelPatch
+
```
MasseGuillaume: this fucks up sublime syntax highlighting.

olafurpg: awww, maybe get a better editor 😜

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedImports.scala
```diff
@@ -28,6 +28,6 @@ case class RemoveUnusedImports(index: SemanticdbIndex)
   }
   override def fix(ctx: RuleCtx): Patch =
     ctx.tree.collect {
-      case i: Importee if isUnused(i) => ctx.removeImportee(i)
+      case i: Importee if isUnused(i) => ctx.removeImportee(i).atomic
```
olafurpg: In the case where there is no `.atomic`, should we implicitly assume `.atomic` on the whole returned `Patch`? This would mean one `// scalafix:ok Blah` disables `Blah` for the whole file, which might be counter intuitive. Alternatively, we can also keep the existing behavior where no `.atomic` means the rule does not respect escape hatches and is applied everywhere regardless of `// scalafix:off blah`. I'm personally leaning towards the second option.

scalafix-tests/output/src/main/scala/test/escapeHatch/EscapeHatchPatches.scala
```diff
@@ -0,0 +1,8 @@
+package test.escapeHatch
+
+import scala.collection.immutable // scalafix:ok RemoveUnusedImports
+
+import scala.collection.mutable.{ // scalafix:ok RemoveUnusedImports
```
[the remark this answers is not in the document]
olafurpg: That's likely because removing `Map` or `Set` independently removes the `{` which is escaped. If the comment only applies to `import scala` I suspect it won't work

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedImports.scala
olafurpg: Are there any other scalafix rules that could benefit from `.atomic`?

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -48,12 +48,14 @@ sealed abstract class Patch {
   def ++(other: Iterable[Patch]): Patch = other.foldLeft(this)(_ + _)
   def isEmpty: Boolean = this == EmptyPatch
   def nonEmpty: Boolean = !isEmpty
+  def atomic: Patch = AtomicPatch(this)
```
olafurpg: Add docstring. The other methods have no docstring since I think they're quite self explanatory but atomic is not as obvious

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedImports.scala
MasseGuillaume: Yep, this is what I am doing right now. I'm making sure all the Scalafix rules uses atomic.

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
MasseGuillaume: I will write user documentation with examples and limitation.

scalafix-tests/output/src/main/scala/test/escapeHatch/EscapeHatchPatches.scala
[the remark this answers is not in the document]
MasseGuillaume: This is good then, it makes it really difficult to fall on the non-supported case

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
```diff
@@ -121,7 +105,24 @@ abstract class SemanticRuleSuite(
...
+
+      if (diff.isFailure) {
+        println("###########> Lint       <###########")
+        println(diff.toString)
+      }
+
+      val result = compareContents(obtained, expected)
+      if (result.nonEmpty) {
+        println("###########> Diff       <###########")
+        println(error2message(obtained, expected))
```
olafurpg: Is the motivation for this change to copy-paste the terminal output? If so, we should include the filename since println is detached from the scalatest failure report.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveXmlLiterals.scala
```diff
@@ -123,7 +123,6 @@ case object RemoveXmlLiterals extends Rule("RemoveXmlLiterals") {
         patchXml(xml, tripleQuoted).atomic
     }.asPatch
 
-    if (patch.nonEmpty) (patch + importXmlQuote)
-    else patch
+    importXmlQuote.dependsOn(patch)
```
MasseGuillaume: @olafurpg what do you think about this?

MasseGuillaume:
I just added atomics for other rules like `DottyVolatileLazyVal`. However, I don't think we should allow escaping certain types of fix. For example, `DottyVarArgPattern` since it would result in a compilation error on the target side.

MasseGuillaume:
Ready for round 2.

MasseGuillaume:
Ready for round 3.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -40,6 +40,7 @@ case class RemoveUnusedTerms(index: SemanticdbIndex)
   override def fix(ctx: RuleCtx): Patch =
     ctx.tree.collect {
       case i: Defn if isUnused(i) =>
-        tokensToRemove(i).fold(Patch.empty)(ctx.removeTokens)
```
olafurpg: Could have added `.atomic` on the whole thing 😉

scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveXmlLiterals.scala
```diff
@@ -119,9 +120,11 @@ case object RemoveXmlLiterals extends Rule("RemoveXmlLiterals") {
       case xml @ Term.Xml(parts, _) =>
         val tripleQuoted = isMultiLine(xml) || parts.exists(
           containsEscapeSequence)
-        patchXml(xml, tripleQuoted)
+        patchXml(xml, tripleQuoted).atomic
     }.asPatch
 
+    // NB if all patches are not yet escaped here
```
olafurpg: 👍

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
olafurpg: We should still have a docstring despite docs on the website, a single sentence like

> Skip this entire patch if a part of it is disabled with `// scalafix:off`

will do.

website/src/main/tut/docs/rule-authors/glossary.md
```diff
@@ -42,6 +42,9 @@ utility methods to accomplish common tasks.
 If you experience that it's difficult to implement something that
 seems simple then don't hesitate to ask on {% gitter %}.
 
+It's possible to escape parts of the Patch with `// scalafix:ok`. If you want to treat
```
olafurpg: s/the/a/

website/src/main/tut/docs/rule-authors/glossary.md
```diff
@@ -42,6 +42,9 @@ utility methods to accomplish common tasks.
...
+the Patch as a transaction use `.atomic`.
```
olafurpg: s/the/a/

MasseGuillaume:
Ready for round 4.

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -48,12 +48,16 @@ sealed abstract class Patch {
   def ++(other: Iterable[Patch]): Patch = other.foldLeft(this)(_ + _)
   def isEmpty: Boolean = this == EmptyPatch
   def nonEmpty: Boolean = !isEmpty
+  /** Skip this entire patch if a part of it is disabled with // scalafix:off
```
olafurpg: Nitpick, single line docstrings should be `/** Blah blah */`, I should make scalafmt enforce this 😅

gabro:
This is so cool! 🎉 Thank you @MasseGuillaume, looking forward to use it!

MasseGuillaume:
@gabro hehe, don't forget to set `.atomic` properly.

PR #513, opened 2017-12-19
Add rule to leverage automatic tupling of function parameters in Dotty

marcelocenerine:
http://dotty.epfl.ch/docs/reference/auto-parameter-tupling.html

@olafurpg, is this something you would like to have in scalafix at this point?

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
@@ -1,0 +1,50 @@
+package scalafix.internal.rule
+
+import scala.meta._
+import scala.meta.tokens.Token._
+import scalafix.Patch
+import scalafix.rule.{Rule, RuleCtx}
+
+case object DottyAutoTuplingFunctionArgs
+    extends Rule("DottyAutoTuplingFunctionArgs") {
+  override def description: String =
+    "Rewrite that removes pattern-matching decomposition if function arguments can be automatically tupled in Dotty."
+
+  override def fix(ctx: RuleCtx): Patch =
+    ctx.tree.collect {
+      case Term.Apply((_, List(pf: Term.PartialFunction)))
+          if canBeAutoTupled(pf) =>
+        val pfTokens = pf.tokens
+
+        (for {
+          leftBrace <- pfTokens.headOption if leftBrace.is[LeftBrace]
+          rightBrace <- pfTokens.lastOption if rightBrace.is[RightBrace]
+          kwCase <- pfTokens.find(_.is[KwCase])
+        } yield {
+          val tkList = ctx.tokenList
+          val replacedBraces =
+            if (canReplaceBracesWithParens(pf)) {
+              ctx.replaceToken(leftBrace, "(") +
+                ctx.replaceToken(rightBrace, ")") +
+                ctx.removeTokens(tkList.leadingSpaces(leftBrace)) +
+                ctx.removeTokens(tkList.trailingSpaces(leftBrace)) +
+                ctx.removeTokens(tkList.leadingSpaces(rightBrace))
+            } else Patch.empty
+
+          replacedBraces +
+            ctx.removeToken(kwCase) +
+            ctx.removeTokens(tkList.trailingSpaces(kwCase))
+        }).asPatch
+    }.asPatch
+
+  private def canBeAutoTupled(pf: Term.PartialFunction): Boolean =
+    pf.cases match {
+      case Case(Pat.Tuple(args), None, _) :: Nil
+          if args.forall(a => a.is[Pat.Var] || a.is[Pat.Wildcard]) =>
+        true
+      case _ => false
+    }
+
+  private def canReplaceBracesWithParens(pf: Term.PartialFunction): Boolean =
+    pf.pos.startLine == pf.pos.endLine && pf.cases.head.stats.size == 1
+}
--- a/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
@@ -1,0 +1,36 @@
+/*
+rules = DottyAutoTuplingFunctionArgs
+ */
+package test
+
+object DottyAutoTuplingFunctionArgs {
+  Some((1, 2)).map { case (a, b) => a + b }
+  List((1, 2, 3), (4, 5, 6)).map { case (a, b, c) => println(a); a + b + c }
+  List((1, 2), (3, 4)).foreach { case (a, b) => { val c = a + b; println(c) } }
+  Set((1, 2, 3)).map { case (a, b, c) =>
+    println(a + b + c)
+  }
+  List((1, 1, 1, 1, 1), (2, 2, 2, 2, 2)).flatMap{
+    case (_, _, _, _, n) => List(n)
+  }
+  val xs = List("3", "1", "0").zipWithIndex
+  xs.filter { case (w, i) => w == i.toString }
+
+  // should not rewrite the below
+  xs.map { case (w, i) if i > 1 => s"$w-$i" } // case with guards
+  List((Some(1), Some(1))).map { case (Some(a), Some(b)) => a + b } // extraction
+  xs.map {
+    case (a, b) => a + b
+    case _ => 0
+  } // multiple cases
+  xs.find { case (_, x @ b) => x - b == 0 } // binding
+  xs.map { case ("0", 0) => "zero" } // literals
+  List(("a", 1, false, 1.0), ("b", 2, true, 2.0)).exists{
+    case (_: String, _: Int, c: Boolean, _: Double) => c
+  } // typed parameters
+  class A
+  class B extends A
+  def foo(xs: List[(A, A)]) = xs.map {
+    case (x: B, y: A) => 1
+  } // typed parameters not matching tuple element types
+}
--- a/website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
+++ b/website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
@@ -1,0 +1,17 @@
+---
+layout: docs
+title: DottyAutoTuplingFunctionArgs
+---
+
+# DottyAutoTuplingFunctionArgs
+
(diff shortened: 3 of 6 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,53 @@
...
+      case Case(Pat.Tuple(args), None, _) :: Nil
+          if args.forall(isOptionallyTypedVarOrWildcard) =>
+        true
+      case _ => false
+    }
+
+  private def isOptionallyTypedVarOrWildcard(pat: Pat): Boolean =
+    pat.is[Pat.Var] || pat.is[Pat.Typed] || pat.is[Pat.Wildcard]
+
+  private def initSpaces(tokens: Seq[Token]) = tokens.takeWhile(_.is[Space])
```
marcelocenerine: I just found myself writing this again 😜

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,53 @@
...
+    "Rewrite that removes pattern-matching decomposition if function arguments can be automatically tupled in Dotty."
+
+  override def fix(ctx: RuleCtx): Patch =
+    ctx.tree.collect {
+      case Term.Apply((_, List(pf: Term.PartialFunction)))
+          if canBeAutoTupled(pf) =>
+        val pfTokens = pf.tokens
+
+        (for {
+          leftBrace <- pfTokens.headOption
```
olafurpg: Guard against `.is[LeftBrace/RightBrace]`?

marcelocenerine: I think the curly braces here are mandatory (http://www.lihaoyi.com/post/WartsoftheScalaProgrammingLanguage.html#needing-curliescase-for-destructuring-anonymous-functions). I put the guards anyway to make it explicit  👍

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,53 @@
...
+          rightBrace <- pfTokens.lastOption
+          kwCase <- pfTokens.find(_.is[KwCase])
+          singleLine = pf.pos.startLine == pf.pos.endLine
+        } yield {
+          val bracesReplacedWithParentheses =
+            if (singleLine) {
```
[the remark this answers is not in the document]
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,24 @@
+/*
+rules = DottyAutoTuplingFunctionArgs
+ */
+package test
+
+object DottyAutoTuplingFunctionArgs {
+  List((1, 2), (3, 4)).map { case (a, b) => a + b }
```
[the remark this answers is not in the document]
marcelocenerine: oh, you found a way to break it. Nice catch! Your suggestion https://github.com/scalacenter/scalafix/pull/513/files#r157722339 addresses this.

allanrenucci:
> I also ran it on dotty-compiler-bootstrapped but couldn't test the compilation since it depends on dotty-compiler/compile with scala 2.x

Can you run it on `dotty-compiler` instead and test the compilation by setting the `scalaVersion` value to `0.5.0-RC1`?

smarter:
> I also ran it on dotty-compiler-bootstrapped but couldn't test the compilation since it depends on dotty-compiler/compile with scala 2.x

You can bootstrap from an existing published bootstrap. First publish a working bootstrap:
```scala
> dotty-bootstrapped/publishLocal
```
Then use it to compile the bootstrap:
```scala
> set bootstrapFromPublishedJars in ThisBuild := true
```

allanrenucci:
I have some examples in mind that may be tricky to rewrite:
```scala
scala> class A 
// defined class A
scala> class B extends A 
// defined class B
scala> def foo(xs: List[(A, A)]) = xs.map((x, y) => 1) 
def foo(xs: List[(A, A)]): List[Int]
scala> def foo(xs: List[(A, A)]) = xs.map((x: A, y: A) => 1) 
def foo(xs: List[(A, A)]): List[Int]
scala> def foo(xs: List[(A, A)]) = xs.map((x: B, y: A) => 1) 
1 |def foo(xs: List[(A, A)]) = xs.map((x: B, y: A) => 1)
  |                                                    ^
  |                                                 found:    (B, A) => Int
  |                                                 required: ((A, A)) => Int
  |                                                 
scala> def foo(xs: List[(A, A)]) = xs.map { case (x: B, y: A) => 1 }  // Note this one cannot be rewritten
def foo(xs: List[(A, A)]): List[Int]
scala> def foo(xs: List[(A, A)]) = xs.map((x: Any, y: A) => 1) 
def foo(xs: List[(A, A)]): List[Int]
```

marcelocenerine:
Interesting example... I don't think there is enough type information in scalameta (nor if it is intended to) to allow the rule to evaluate the type relationship between the partial function parameters and the tuple elements so it can be 100% sure that the pattern matching is not needed... 

To be on the safe side, maybe the rule could only rewrite when the partial function parameters are explicitly typed and the types exactly match the types of the corresponding elements in the tuple. This might be doable....

olafurpg:
> Interesting example... I don't think there is enough type information in scalameta (nor if it is intended to in the future) to allow the rule to evaluate the type relationship between the partial function parameters and the tuple elements so it can be 100% sure that the pattern matching is not needed...

We actually have the full signature with attributed types of the invoked symbol. However, there is no `=:=` that handles de-aliasing etc. so I think it's best for now to edge on the safe side and leave `pat.is[Pat.Typed]` out of the cases that get rewritten.

marcelocenerine:
>We actually have the full signature with attributed types of the invoked symbol. 

sorry, I meant that it may not be possible to determine if `B <: A` (I could be wrong)

>I think it's best for now to edge on the safe side and leave pat.is[Pat.Typed] out of the cases that get rewritten.

Done

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
```diff
@@ -0,0 +1,17 @@
...
+Removes pattern-matching decomposition if function arguments can be automatically tupled. See [http://dotty.epfl.ch/docs/reference/auto-parameter-tupling.html](http://dotty.epfl.ch/docs/reference/auto-parameter-tupling.html)
+
+```scala
+// before
+xs.map { case (x, y) => x + y }
+// after
+xs.map((x, y) => x + y)
+```
+
+Auto-tupling is a language feature only supported in Dotty. Therefore, the code produced by this rewrite does not compile with Scala 2.x.
```
olafurpg: 👍

olafurpg:
Here is a diff from running the rewrite on the dotty repo including submodules
https://gist.github.com/olafurpg/d6a393d16487b50a39d8a09a3a825145
@allanrenucci I got
```
[error] (dotty-compiler/compile:compileIncremental) java.lang.NoClassDefFoundError: dotty/tools/dotc/interfaces/ReporterResult
```
when running `++0.5.0-RC1`.  I also tried `set bootstrapFromPublishedJars in ThisBuild := true` after publishLocal but that still compiled dotty-compiler with Scala 2.x. I tested this on scalatest and everything still compiled after.

The diff  in apache/spark is 11k LOC *_* https://gist.github.com/olafurpg/1b52f2b3d070dc8424fd41aa385b9873

allanrenucci:
Actually, I don't think this rewrite can be syntactic:
```scala
scala> def foo(xs: List[Any]) = xs.map { case (x, y) => 1 } 
def foo(xs: List[Any]): List[Int]
scala> def foo(xs: List[Any]) = xs.map((x, y) => 1 ) 
1 |def foo(xs: List[Any]) = xs.map((x, y) => 1 )
  |                                ^^^^^^^^^^^
  |                                wrong number of parameters, expected: 1
```

marcelocenerine:
😞 I'll take a look at #521

olafurpg:
Good catch! @allanrenucci I opened https://github.com/scalacenter/scalafix/issues/521 with details to followup on that. Am I right in thinking that this case only happens for `Any`/`AnyRef` since they are supertypes of `TupleN`?

olafurpg:
@marcelocenerine No worries, getting rewrites 100% correct is really hard! I also think there is value in having a syntactic rewrite that runs fast without compilation, and manually fixing corner cases where you are doing silly stuff like deconstructing `Any` into `(Any, Any)`. Most important bit is to set the expectations correctly in the docs.

smarter:
> I also think there is value in having a syntactic rewrite that runs fast without compilation, and manually fixing corner cases

That sounds exactly like what I don't want in an automatic rewrite :P. It can take as long as it wants as, that's fine as long as I can trust the result.

allanrenucci:
@olafurpg I think it is more complicated than that. You are basically doing exhaustivity checking here. So off the top of my head: `ProductN`, type variables and am sure I am missing many other corner cases. Maybe @liufengyun can comment on this,.

Also, you don't want to rewrite something that expect a partial function for example:
```scala
def foo(f: PartialFunction[(Int, Int), Int]): Nothing
foo { case (x, y) => 1 } // should't be rewritten.
```

olafurpg:
> That sounds exactly like what I don't want in an automatic rewrite :P. It can take as long as it wants, that's fine as long as I can trust the result.

If I have the option between slow correct rewrite and fast but slightly buggy rewrite then I definitely agree with you and pick the slow correct one 😄 But our current choice is unfortunately between no rewrite and fast but slight buggy rewrite (failing in cases where you deconstruct into `(Any, Any)`) so I personally prefer the latter setting expectations correctly in the docs #522 . Migration tools like codemod https://github.com/facebook/codemod are widely used despite producing non-perfect rewrites.

marcelocenerine:
There are a number of missed corner cases pointed out above and in #521. @olafurpg If you agree I'd suggest to revert my commit for now until the rule can handle those scenarios correctly and produce reliable results

olafurpg:
Sure we can revert the commit @marcelocenerine

Issue #512, opened 2017-12-18
Avoid Stale Semanticdb warnings if possible


MasseGuillaume:
If two projects share the same source and they result in the same fix.  We should apply the fix and not emit a "stale semanticdb" warnings

PR #514, opened 2017-12-19
Enable scalafix in the scalafix build.

olafurpg:
I think long overdue to bootstrap and start using scalafix to develop scalafix 😄 We can start with remove unused imports and terms and then add more as we go. I found NoInfer to be too noisy, I will soon have a PR making it possible to tweak the noise by excluding certain symbols.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,4 +1,9 @@
 rules = [
   RemoveUnusedImports
   RemoveUnusedTerms
+  ExplicitResultTypes
 ]
+
+ExplicitResultTypes {
+  unsafeShortenNames = true
+}
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,4 @@
+rules = [
+  RemoveUnusedImports
+  RemoveUnusedTerms
+]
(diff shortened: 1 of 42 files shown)


project/MultiScalaProject.scala
```diff
@@ -126,7 +126,7 @@ object TestProject {
     MultiScalaProject(
       s"tests${sub.capitalize}",
       base(sub),
-      configure
+      configure.andThen(_.disablePlugins(scalafix.sbt.ScalafixPlugin))
```
MasseGuillaume: Because of https://github.com/scalacenter/scalafix/issues/512 ?

.scalafix.conf
```diff
@@ -0,0 +1,4 @@
+rules = [
```
MasseGuillaume: explicit return type for public methods?

MasseGuillaume:
It would be nice to add a --test in the ci matrix.

olafurpg:
We run --test in the ci-fast-212 entry @MasseGuillaume https://travis-ci.org/scalacenter/scalafix/jobs/319146392#L942

PR #515, opened 2017-12-19
DisableUnless rule

vovapolu:
This PR introduces `DisableUnless` rule that forbids given symbols outside of given blocks. It also forbids these symbols in function/method definitions regardless of whether they are in ignored blocks or not. It prevents "escaping" of the symbols.


scalafix-tests/input/src/main/scala/test/fpSpec/NestedBlockingSpec.scala
```diff
@@ -0,0 +1,77 @@
+/*
+rules = [
+  DisableUnless
+]
+
+DisableUnless.symbols = [
+  {
+    mode = "outside"
```
olafurpg: Are there any tests for `inside`?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
```diff
@@ -0,0 +1,113 @@
...
+            context = newContext
+            super.apply(tree)
+            context = oldContext
+        }
+      } else {
+        super.apply(tree)
+      }
+    }
+
+    def result: List[T] = buf.toList
```
olafurpg: can make this call `apply()` so that you only call `new Searcher().result`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
```diff
@@ -0,0 +1,113 @@
...
+    */
+  class SearcherWithContext[T, U](initContext: U)(
+      fn: PartialFunction[(Tree, U), Either[T, U]])
+      extends Traverser {
+    private var context: U = initContext
+    private val buf = scala.collection.mutable.ListBuffer[T]()
+    // it can be immutable...
+
+    override def apply(tree: Tree): Unit = {
+      if (fn.isDefinedAt((tree, context))) {
```
olafurpg: I would use `fn.lift`  or `fn.applyOrElse` to avoid running the pattern twice.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,76 @@
...
+    ConfDecoder.instanceF[UnlessConfig] { c =>
+      (
+        (c.get[String]("mode") match {
+          case Ok("inside") => Ok(UnlessInside)
+          case Ok("outside") => Ok(UnlessOutside)
+          case _ => Ok(UnlessInside)
+        }) |@|
+          c.get[Symbol.Global]("block") |@|
+          c.get[Symbol.Global]("symbol") |@|
+          (c.get[String]("message") match {
```
olafurpg: There is now `getOption[String]` after the upgrade to metaconfig v0.5.4 a couple days ago

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,76 @@
...
+    mode: UnlessMode,
+    block: Symbol.Global,
+    symbol: Symbol.Global,
+    message: Option[String])
+
+object UnlessConfig {
+  implicit val decoder: ConfDecoder[UnlessConfig] =
+    ConfDecoder.instanceF[UnlessConfig] { c =>
+      (
+        (c.get[String]("mode") match {
```
olafurpg: The convention would be to put an `ConfDecoder[UnlessMode]` in the companion object of `UnlessMode` and then `c.get[UnlessMode]("mode")`

vovapolu: I've just removed that setting

olafurpg:
Have you tried running this rule on larger codebase with non-synthetic configuration? I recommend giving it a try to get a feel for the rule and how many false positives/negatives messages it reports.

olafurpg:
What happens in cases like
```scala
def foo() = println(1)
IO(foo())
```
and `foo()` is never called anywhere else?

vovapolu:
>running this rule on larger codebase with non-synthetic configuration
I asked you about this yesterday :D Have you got any examples how to run rules on external repos?

vovapolu:
>What happens in cases like

foo is not in the `IO` block so it's blocked.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
vovapolu: `.collect`, `.transform`, `.traverse` methods also use `.isDefinedAt` and `apply`

olafurpg: Not anymore https://github.com/scalameta/scalameta/pull/1201

olafurpg:
> I asked you about this yesterday :D Have you got any examples how to run rules on external repos?

Manual tests: `publishLocal`, `addSbtPlugin()` and `sbt> scalafix DisableUnless`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
vovapolu: nice :D

olafurpg:
There is https://github.com/scalacenter/scalafix/blob/8bd026de8e6610934ab2cad70f811592f764fc98/scalafix-testkit/src/main/scala/scalafix/testkit/IntegrationPropertyTest.scala to automate manual tests, but I haven't used it in a few months

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,66 @@
...
+    symbol: Symbol.Global,
+    message: Option[String])
+
+object UnlessConfig {
+  implicit val decoder: ConfDecoder[UnlessConfig] =
+    ConfDecoder.instanceF[UnlessConfig] { c =>
+      (
+        c.get[Symbol.Global]("block") |@|
+        c.get[Symbol.Global]("symbol") |@|
+        (c.get[String]("message") match {
```
MasseGuillaume: `obj.getOption[String]("message")`

https://github.com/scalacenter/scalafix/blob/8263bb3ee9aa2f86285a35f08c550982f894b556/scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala#L17

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,66 @@
...
+    symbol: Symbol.Global,
+    message: Option[String])
+
+object UnlessConfig {
+  implicit val decoder: ConfDecoder[UnlessConfig] =
+    ConfDecoder.instanceF[UnlessConfig] { c =>
+      (
+        c.get[Symbol.Global]("block") |@|
+          c.get[Symbol.Global]("symbol") |@|
+          (c.get[String]("message") match {
```
olafurpg: `getOption`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
```diff
@@ -0,0 +1,115 @@
...
+import scalafix.util.SymbolMatcher
+
+object DisableUnless {
+
+  /**
+    * Searches for specific nodes in Tree and adds result of searching to the list.
+    * Also passes context through tree traversal.
+    * fn can either return a value. that stops the searching, or change the context.
+    */
+  class SearcherWithContext[T, U](initContext: U)(
```
olafurpg: I'm not sure the genericity is helping much here since `SearcherWithContext` is only constructed once

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
```diff
@@ -0,0 +1,115 @@
...
+      fn: PartialFunction[(Tree, U), Either[T, U]])
+      extends Traverser {
+    private var context: U = initContext
+    private val buf = scala.collection.mutable.ListBuffer[T]()
+    // it can be immutable...
```
olafurpg: ?

fommil:
@olafurpg I'm unable to run it on anything until it's released.

vovapolu:
I've run it on scalaz, but it's quite hard to find a good code base for such rule. There weren't a lot of IO usages, but on the existing ones it works good. No wrong exceptions with `DisableUnless` rule and expections on selected symbols with `Disable` rule.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
vovapolu: Well, `SearcherWithContext` can be immutable, i.e without `var`s :D

olafurpg:
@fommil Have you tried checking out this branch and `scalafix-sbt/publishLocal`? I can configure the CI to publish SNAPSHOT on merge, but new rules should still be tested on real world projects before merge.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableUnless.scala
olafurpg: without more explanation the comment doesn't help understanding the code better, if you skim through the scalafix codebase you'll find liberal use of mutable collections and vars within small local scopes

fommil:
@olafurpg my wife is about to go into labour so I'm unlikely to be able to try anything for the next week. I :+1: merging early with snapshot feedback, instead of getting caught up in a catch22 situation about releasing vs testing in the wild.

It would be good if there was a snapshot mechanism in place, :100: to that idea.

olafurpg:
> my wife is about to go into labour so I'm unlikely to be able to try anything for the next week.

Woot! ❤️  I opened #538 to configure CI to publish -SNAPSHOT on merge. @vovapolu said he tried the rule on scalaz and that's good enough for me, just wanted to sanity check that this rule was tested outside of unit tests.

Only remaining comment is `getOption` otherwise this is good to go!

olafurpg:
SNAPSHOT on merge is up! Track https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-core_2.12/ for the latest version number, each merge gets a unique version number

fommil:
I'm sure once that's available we can get a lot of scalaz folk to try it out before you cut a release.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,62 @@
...
+import metaconfig.{Conf, ConfDecoder, ConfError, Configured}
+import org.langmeta.Symbol
+
+import scalafix.internal.config.MetaconfigPendingUpstream.XtensionConfScalafix
+import scalafix.internal.util._
+
+case class UnlessConfig(
+    block: Symbol.Global,
+    symbol: Symbol.Global,
+    message: Option[String])
```
olafurpg: FWIW, we recently added a `CustomMessage[T]` helper for exactly this pattern https://github.com/scalacenter/scalafix/blob/f5e0a1c084256e42cf6725208d5c13113f3f4821/scalafix-core/shared/src/main/scala/scalafix/config/CustomMessage.scala I supports multiline message descriptions and also the ability to customize the category id. However, it seems to only supports single field types which won't work here for `block` and `symbol`, I'll think how we can accommodate that :D

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import metaconfig.{Conf, ConfDecoder, ConfError, Configured}
+import org.langmeta.Symbol
+
+import scalafix.internal.config.MetaconfigPendingUpstream.XtensionConfScalafix
+import scalafix.internal.util._
+
+case class UnlessConfig(
+    block: Symbol.Global,
+    symbol: Symbol.Global,
```
olafurpg: Might want to make this a `List[Symbol.Global]` down the road in order to reuse lists of bad symbols

fommil:
Is there an idiot's guide to using the snapshot plugin with fp rules turned to 11?

olafurpg:
I opened #540 to document how to use snapshot releases, @vovapolu or you will need to add docs for DisableUnless

olafurpg:
Snapshot installation docs are up! The version number in the examples seems up-to-date so it can be copy-pasted directly
https://scalacenter.github.io/scalafix/docs/users/installation#snapshot

Issue #516, opened 2017-12-19
Make `from` in TokenList.slice(from, to) inclusive


marcelocenerine:
`TokenList.slice(from, to)` doesn't have the same behavior as `slice` in the Scala's std collection. This is due to the fact that the `from` parameter is exclusive. This is something that rule authors may not expect and can be the cause of bugs for those not familiar with the implementation details of `TokenList`.

The current behavior also affects usability. For instance, while removing tokens what could take a single step (`ctx.removeTokens(tokenList.slice(tk, last)`) actually requires 2 (`ctx.removeToken(tk) + ctx.removeTokens(tokenList.slice(tk, last)`).

See https://github.com/scalacenter/scalafix/pull/474/files#r157779985

PR #534, opened 2017-12-28
Make 'from' parameter in TokenList.slice/find inclusive

marcelocenerine:
- Addresses #516. The original goal was to do this only for `TokenList.slice` but then I realized that `find` was also inconsistent.
  * `find` is not used anywhere in the repo
  * `slice` is used in 2 places: `ProcedureSyntax` & `ExplicitResultTypes`. For both of them this change won't cause any harm.

- Fixes a bug in `TokenList.leading` that was causing the first element in the token list (usually `Token.BOF`) being incorrectly dropped.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
@@ -9,8 +9,10 @@ import scala.meta.tokens.Tokens
 final class TokenList private (tokens: Tokens) {
   def trailing(token: Token): SeqView[Token, IndexedSeq[Token]] =
     tokens.view(tok2idx(token) + 1, tokens.length)
+
   def leading(token: Token): SeqView[Token, IndexedSeq[Token]] =
     tokens.view(0, tok2idx(token)).reverse
+
   private[this] val tok2idx = {
     val map = Map.newBuilder[Token, Int]
     var i = 0
@@ -21,20 +23,11 @@ final class TokenList private (tokens: Tokens) {
     map
       .result()
       .withDefault(t =>
-        throw new NoSuchElementException("token not found: " + t))
+        throw new NoSuchElementException(s"token not found: $t"))
   }
 
-  def find(start: Token)(f: Token => Boolean): Option[Token] = {
-    def loop(curr: Token): Option[Token] = {
-      if (f(curr)) Option(curr)
-      else {
-        val iter = next(curr)
-        if (iter == curr) None // reached EOF
-        else loop(iter)
-      }
-    }
-    loop(next(start))
-  }
+  def find(start: Token)(p: Token => Boolean): Option[Token] =
+    tokens.drop(tok2idx(start)).find(p)
 
   def slice(from: Token, to: Token): SeqView[Token, IndexedSeq[Token]] =
     tokens.view(tok2idx(from), tok2idx(to))
--- a/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
@@ -18,7 +18,10 @@ final class TokenList private (tokens: Tokens) {
       map += (tok -> i)
       i += 1
     }
-    map.result()
+    map
+      .result()
+      .withDefault(t =>
+        throw new NoSuchElementException("token not found: " + t))
   }
 
   def find(start: Token)(f: Token => Boolean): Option[Token] = {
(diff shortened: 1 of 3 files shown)


scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
```diff
@@ -8,40 +8,29 @@ import scala.meta.tokens.Tokens
 /** Helper to traverse tokens as a doubly linked list.  */
 final class TokenList private (tokens: Tokens) {
   def trailing(token: Token): SeqView[Token, IndexedSeq[Token]] =
-    tokens.view(tok2idx(token), tokens.length).drop(1)
+    tokens.view(tok2idx(token) + 1, tokens.length)
+
   def leading(token: Token): SeqView[Token, IndexedSeq[Token]] =
-    tokens.view(0, tok2idx(token)).drop(1).reverse
+    tokens.view(0, tok2idx(token)).reverse
```
marcelocenerine: I believe the goal here was to have the `token` element dropped from the result. However, this is already handled by the `.view` method as the `end` parameter is exclusive

olafurpg: Woah, this was surprising :D 

```scala
@     tokens.view(0, tok2idx(token)).reverse.mkString
res19: String = "(list { a object"

@     tokens.view(0, tok2idx(token)).drop(1).reverse.mkString
res20: String = "(list { a object"
```

Yes, the objective was to drop the token.

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
```diff
@@ -8,40 +8,29 @@ import scala.meta.tokens.Tokens
...
-  def slice(from: Token, to: Token): Seq[Token] = {
-    val builder = Seq.newBuilder[Token]
-    var curr = next(from)
-    while (curr.start < to.start) {
-      builder += curr
-      curr = next(curr)
-    }
-    builder.result()
-  }
+  def slice(from: Token, to: Token): SeqView[Token, IndexedSeq[Token]] =
```
olafurpg: Is this signature change necessary? I think we can keep the signature to avoid a binary breaking change.

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
marcelocenerine:
```scala
val tokens = "foo bar".tokenize.get
tokens: scala.meta.tokens.Tokens = Tokens(, foo,  , bar, )
val barIdx = 3

tokens.view(0, barIdx).reverse.mkString("|")
res11: String = " |foo|"
tokens.view(0, barIdx).drop(1).reverse.mkString("|")
res12: String = " |foo"
```

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
marcelocenerine: In fact it is not. Given that the method now returns a view, I wanted to make this fact explicit as well as have the return type consistent the `leading*`/`trailing*` methods. It wouldn't cause compilation errors but indeed it would break binary compatibility. I'm gonna change it back to `Seq` to avoid that.

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
```diff
@@ -8,40 +8,29 @@ import scala.meta.tokens.Tokens
...
-    val builder = Seq.newBuilder[Token]
-    var curr = next(from)
-    while (curr.start < to.start) {
-      builder += curr
-      curr = next(curr)
-    }
-    builder.result()
-  }
+  def slice(from: Token, to: Token): Seq[Token] =
+    tokens.view(tok2idx(from), tok2idx(to))
```
marcelocenerine: given that the method no longer says that the returned collection is a view, I was going to change the implementation and delegate to `Tokens.slice` instead which should be slightly faster as it is backed by an array and does not involve any copying. But then I realized that `Tokens.length` is buggy, which breaks `isEmpty`:

```
"foo".tokenize.get.slice(1, 0).isEmpty
res4: Boolean = false
```

I'm gonna file an issue or submit a PR to Scalameta.

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
olafurpg: I see, `drop(1)` was removing the `BOF` token, which is wrong behaviour but probably never surfaced as a problem because it's zero length. Good catch!

scalafix-core/shared/src/main/scala/scalafix/util/TokenList.scala
olafurpg: iirc someone reported a similar problem with `Tokens.length` the other day. Length is implemented like this
```scala
  def length: Int = end - start
```

Should it be this instead?

```scala
  def length: Int = math.max(0, end - start)
```

marcelocenerine: The root cause is in `Tokens.slice`: https://github.com/scalameta/scalameta/pull/1207

PR #538, opened 2018-01-03
Publish SNAPSHOT on merge.

olafurpg:
To make it easier to try out pre-releases, this commit configures the
CI to publish a -SNAPSHOT to sonatype on every merge.


olafurpg:
Let's merge see if this works 😄

olafurpg:
Seems to work 🎉  https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-core_2.12/

PR #540, opened 2018-01-04
Snapshot installation instructions

olafurpg:
I manually tested this and found a bug related to fetching snapshot releases through sbt-scalafix, which should be fixed in this PR as well.


olafurpg:
Merging to unblock the bugfix in 
d93fe27

PR #544, opened 2018-01-04
Fix #543. Use lenient dialect in scalafix-reflect

### LINKED ISSUE #543: scalafix-reflect uses Scala 2.11 dialect for parsing rules
This causes potential issues with rules written and tested against Scala 2.12 that may not compile in Scala 2.11. See https://github.com/typelevel/cats/issues/2136

olafurpg:


scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
```diff
@@ -77,7 +77,7 @@ object ScalafixConfig {
     // Are extractor varargs specified using colons, i.e. is `case Extractor(xs: _*)` legal or not?
     allowColonForExtractorVarargs = true,
     // Are inline vals and defs supported by this dialect?
-    allowInlineMods = true,
+    allowInlineMods = false,
```
gabro: what's this about?

olafurpg: I'm trying to minimize the breaking change by switching the parser. If a rule previously used `inline` as an identifier then it would compile with reflect before this commit but not after. It's the only custom flag in DefaultDialect that removes features from the original dialect, the other flags only enable the dialect to parse more files.

gabro: Clear, thanks!

PR #546, opened 2018-01-06
Docs for DisableUnless and DisableSyntax

vovapolu:


website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -0,0 +1,63 @@
+---
+layout: docs
+title: DisableUnless
+---
+
+** syntactic **
+
+# DisableUnless
+
+_Since 0.5.7-67-5b643194-SNAPSHOT_
```
vovapolu: Don't know what to write here, actually :)

vovapolu:
Maybe I could add docs for `DisableSyntax.regex` also? My next PR adds a bunch of new subrules in `DisableSyntax` that should be documented.

website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -0,0 +1,61 @@
+---
+layout: docs
+title: DisableUnless
+---
+
+# DisableUnless
+
+_Since 0.5.7-67-5b643194-SNAPSHOT_
```
olafurpg: should be since 0.5.8 even if that release is not out yet, otherwise we would have to update the docs when 0.5.8 comes out.

website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -0,0 +1,61 @@
...
+```
+package test
+
+object Test {
+  object IO {
+    def apply[T](run: => T) = ???
+  }
+
+  
+  println("hi") // assert: DisableUnless.println
```
olafurpg: I'm not sure if the user is familiar with the testkit `assert: ` syntax, I think it's fine to use `// not ok` and `// ok` instead

olafurpg:
Regarding Disable.regex, maybe @MasseGuillaume can do that?

Issue #529, opened 2017-12-21
Wartremover migration 


vovapolu:
Migrate all the [wartremover warts](http://www.wartremover.org/doc/warts.html). https://github.com/wartremover/wartremover/tree/master/core/src/test/scala/wartremover/warts

This ticket should also summarise how to use the migrated version.

`Unsafe` warts

- [x] Any (`NoInfer.symbols = ["scala.Any"]`)
- [x] AsInstanceOf (`Disable.symbols = ["scala.Any.asInstanceOf"]`)
- [x] EitherProjectionPartial (`Disable.symbols = ["scala.util.Either.LeftProjection.get", "scala.util.Either.RightProjection.get"]`)
- [x] IsInstanceOf (`Disable.symbols = ["scala.Any.isInstanceOf"]`)
- [ ] TraversableOps (huge list of all overridden methods for now) 
- [ ] NonUnitStatements (https://github.com/scalacenter/scalafix/pull/554) (Blocked by absence of `Term.tpe`)
- [x] Null (`DisableSyntax.keywords = [null]`)
- [x] OptionPartial (`Disable.symbols = ["scala.Option.get", "scala.Some.get", "scala.None.get"]`)
- [x] Product (`NoInfer.symbols = ["scala.Product"]`)
- [x] Return (`DisableSyntax.keywords = [return]`)
- [x] Serializable (`NoInfer.symbols = ["java.io.Serializable"]`)
- [x] StringPlusAny (`NoInfer.symbols = ["scala.Predef.any2stringadd"]`)
- [x] Throw (`DisableSyntax.keywords = [throw]`)
- [x] TryPartial (`Disable.symbols = ["scala.util.Try.get", "scala.util.Failure.get", "scala.util.Success.get"]`)
- [x] Var (`DisableSyntax.keywords = [var]`)

and others

- [x] AnyVal (`NoInfer.symbols = ["scala.AnyVal"]`)
- [ ] ArrayEquals (Blocked by absence of `Term.tpe`)
- [x] DefaultArguments (`DisableSyntax.noDefaultArgs = true`)
- [x] Enumeration (`Disable.symbols = ["scala.Enumeration"]`)
- [x] Equals (`Disable.symbols = ["java.lang.Object.equals", "java.lang.Object#`==`", "java.lang.Object.eq", "java.lang.Object.ne"]`)
[...]

Issue #527, opened 2017-12-21
ban use of implicit object


fommil:
https://github.com/wartremover/wartremover/issues/381

in fact this could be a rewrite rule to rewrite as an `implicit val`.

In ensime we hit implicit resolution bugs when we used `implicit objects`. I think it was because the object has a refined type and is therefore not treated the same as an implicit val override. Perhaps I was pushing the compiler too hard, and this is no longer relevant when using https://gitlab.com/fommil/scalaz-deriving/ but it was certainly enough to let me know that there are dragons and it's worth avoiding.

Issue #525, opened 2017-12-21
ban use of val in abstract things


fommil:
https://github.com/wartremover/wartremover/issues/397

Far too many initialisation bugs. Even in OOP, `val` is an implementation detail, and the interface should use `def`.

vovapolu:
Hmm, can you provide some examples of "abstract things" except `trait`s?

fommil:
`trait` and `abstract class`

vovapolu:
just `val` within `trait`s and `abstract class`es?

PR #548, opened 2018-01-07
New rules

vovapolu:
This PR adds new rules to the scalafix:
- Several sub-rules in `DisableSyntax`
```
DisableSyntax.noDefaultArgs
DisableSyntax.noValInAbstract
DisableSyntax.noImplicitObject
DisableSyntax.noImplicitConversion
```
And rules that actually rewrites syntax:
```
DisableSyntax.noFinalVal
DisableSyntax.noNonFinalCaseClass
```
- `LeakingSealed` rule

See more in docs in this PR, http://www.wartremover.org/doc/warts.html, https://github.com/scalacenter/scalafix/issues/529, https://github.com/scalacenter/scalafix/issues/527, https://github.com/scalacenter/scalafix/issues/525


gabro:
Nice work! A minor thing about naming: I wouldn't differentiate between rules that fix and rule that lint.
E.g. I think it's ok to include in `DisableSyntax` some variants that perform fixes and some variants that don't, without introducing a separate `FixSyntax` rule. I think that was the purpose of the `.atomic` API.
This is consistent with what ESLint (a linter for JS, which inspired the unification of linter and rewrites) does: some rules can be fixed, some can just warn.

 Paging also @MasseGuillaume and @olafurpg that know this area better than me.

MasseGuillaume:
`.atomic` is for `--diff` and `// scalafix:ok` it's a way to describe the relationship between Patches, it's like a transaction in a database, it will rollback the whole transaction if a patch is in a disabled region.

For `addFinalCaseClass` & etc I think the only rule of thumb is that it's ergonomic for the end user. I would put `FixSyntax.removeFinalVal` as `DisableSyntax.noFinalVal`. I would name `NonUnitStatements` as `DiscardedValues`

gabro:
Ah sorry, I had the wrong API in mind.

> For addFinalCaseClass & etc I think the only rule of thumb is that it's ergonomic for the end user. I would put FixSyntax.removeFinalVal as DisableSyntax.noFinalVal.

That's exactly what I had in mind.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +35,125 @@ final case class DisableSyntax(
...
+              if mods.exists(_.is[Mod.Abstract]) =>
+            templ.stats
+          case Defn.Trait(_, _, _, _, templ) => templ.stats
+          case _ => List.empty
+        }
+        val vals = stats.flatMap {
+          case v: Defn.Val => Some(v)
+          case _ => None
+        }
+        if (vals.isEmpty) None else Some(vals)
```
MasseGuillaume: ?

MasseGuillaume: toList

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +35,125 @@ final case class DisableSyntax(
...
+        Seq(
           errorCategory
             .copy(id = "contravariant")
             .at(
               "Contravariant types could lead to error-prone situations.",
               t.pos
             )
-      }
+        )
+      case t @ WithMethods(methods)
```
MasseGuillaume: Well done, I think just this alone could be one PR.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NonUnitStatements.scala
```diff
@@ -0,0 +1,80 @@
...
+case class NonUnitStatements(index: SemanticdbIndex)
+    extends SemanticRule(index, RuleName("NonUnitStatements")) {
+
+  override def description: String =
+    "This rule ensures that statements should only return Unit"
+
+  private lazy val errorCategory: LintCategory =
+    LintCategory.error("Statements should only return Unit")
+
+  def litType(l: Lit): Option[Type] = l match {
```
MasseGuillaume: @olafurpg do we have a utility to do this in scalameta?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NonUnitStatements.scala
```diff
@@ -0,0 +1,80 @@
...
+    case _ => None
+  }
+
+  def statementType(stat: Stat): Option[Type] = stat match {
+    case l: Lit => litType(l)
+    case name: Term.Name => name.symbol.flatMap(_.resultType)
+    case Term.Select(_, name) => statementType(name)
+    case Term.Apply(fun, _) => statementType(fun)
+    case Term.ApplyInfix(_, op, _, _) => statementType(op)
+    case Term.ApplyUnary(op, _) => statementType(op)
```
MasseGuillaume: could this be simply `case term: Term` ?

scalafix-tests/input/src/main/scala/test/NonUnitStatements.scala
```diff
@@ -0,0 +1,66 @@
...
+  f(a)            // assert: NonUnitStatements._root_.scala.Int
+  a + b           // assert: NonUnitStatements._root_.scala.Int
+  123             // assert: NonUnitStatements._root_.scala.Int
+  A.foo.bar(b)    // assert: NonUnitStatements._root_.scala.Int
+  A.foo           // assert: NonUnitStatements._root_.test.NonUnitStatements.A.Foo
+  A.unitFoo       // ok
+
+  a :: List(b)             // assert: NonUnitStatements._root_.scala.collection.immutable.List[B]
+  List(1, 2, 3).map(_ * 2) // assert: NonUnitStatements.That
+  // scala collections ¯\_(ツ)_/¯
```
MasseGuillaume: :laughing:

scalafix-tests/input/src/main/scala/test/NonUnitStatements.scala
```diff
@@ -0,0 +1,66 @@
...
+      def bar(a: Int): Int = 1
+    }
+    def foo: Foo = Foo()
+
+    def unitFoo: Unit = {
+      'symbol        // assert: NonUnitStatements._root_.scala.Symbol
+      println("foo") // ok
+    }
+  }
+
```
MasseGuillaume:
```scala
[trait|class] Foo
new Foo { 1 }
```

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
```diff
@@ -57,4 +61,46 @@ case object DisableSyntax {
 
   trait TraitCo[+T]            // assert: DisableSyntax.covariant
   type TypeCo[+T] = TraitCo[T] // assert: DisableSyntax.covariant
+
```
MasseGuillaume:
```scala
class Buzz(val a = 1)
```

MasseGuillaume:
@vovapolu try to implement only one rule per PR. It makes the review faster. (I'm really bad at this myself :laughing:)

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NonUnitStatements.scala
vovapolu: What do you mean? How I would get `Term.Name` from abstract `term` in that case?

vovapolu:
@MasseGuillaume Yeah, sorry. Olaf was on holidays, so I decided to make a one big PR to go faster without a bunch of dependent branches.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NonUnitStatements.scala
MasseGuillaume: https://github.com/scalacenter/scastie/blob/master/instrumentation/src/main/scala/com.olegych.scastie.instrumentation/Instument.scala#L110

MasseGuillaume: nevermind, you are right.

website/src/main/tut/docs/rules/FixSyntax.md
```diff
@@ -0,0 +1,26 @@
...
+
+This rule contains several sub-rules to rewrite "unsafe" syntax. 
+
+## Configuration
+
+By default, this rule does not rewrite syntax.
+
+It contains the following sub-rules:
+
+* `removeFinalVal` transforms `final val` to `val`
```
martin-g: Why `final` is not desired ? Some extra information about the problems being solved with these rules might increase the usage of this tool!

vovapolu: Yeah, I should add more docs for this.

gabro: I agree, I google it and it's not super obvious why we would want to change from `final val` to `val`. Also the change is not purely syntactic: `final val` gets constant-folded, i.e. if you have

```scala
class Foo {
  final val x = "42"
}
```

anywhere `x` appears, it's directly replaced with "42".

Removing the `final` removes this optimization, since now `x` can be overridden by subclasses.

gabro: Ah, I've seen `FinalVal` in http://www.wartremover.org/doc/warts.html.

Apparently the constant-folding mechanisms breaks incremental compilation, so the purpose of this rule is precisely to avoid it.

Zinc bug: https://github.com/sbt/zinc/issues/227

gabro: This definitely deserves an explanation somewhere :D

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DiscardedValues.scala
```diff
@@ -0,0 +1,80 @@
...
+    case Lit.Byte(_) => Some(t"_root_.scala.Byte")
+    case Lit.Char(_) => Some(t"_root_.scala.Char")
+    case Lit.Long(_) => Some(t"_root_.scala.Long")
+    case Lit.Null() => Some(t"_root_.scala.Null")
+    case Lit.String(_) => Some(t"_root_.scala.String")
+    case Lit.Short(_) => Some(t"_root_.scala.Short")
+    case _ => None
+  }
+
+  def statementType(stat: Stat): Option[Type] = stat match {
```
olafurpg: This is an incomplete solution, there is no way to robustly get the type of a term with the current scalafix API. If there was supported there would be a method for it in the public API 😜  See "Type at position => No" in http://scalameta.org/

You are pretty-much writing your own type-checker here! It's not a bad approach and it will work for some cases, but I think we should do this the Right Way ™ and use the compiler for this kind of stuff.

website/src/main/tut/docs/rules/FixSyntax.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: docs
+title: FixSyntax
+---
+
+# FixSyntax
```
olafurpg: I echo @gabro's comment that we should not distinguish rewrites and linters this much. Linters should be able to provide auto-fixes and rewrites should be able to lint. Would it make sense to bake these into DisableSyntax with `noFinalVal` and `noNonFinalCaseClass`?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/LeakingSealed.scala
```diff
@@ -0,0 +1,38 @@
...
+        |type can be extended in another file through its descendant.""".stripMargin
+    )
+
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+    def isSealed(tpe: Type): Boolean =
+      ctx.index.denotation(tpe).exists(_.isSealed)
+
+    ctx.tree.collect {
+      case t @ Defn.Class(mods, _, _, _, templ)
+          if templ.inits.exists(i => isSealed(i.tpe)) &&
```
olafurpg: Can we abstract this case guard to reuse between classes and traits?

scalafix-tests/input/src/main/scala/test/DisableSyntax.scala
```diff
@@ -57,4 +61,50 @@ case object DisableSyntax {
...
+  }
+
+  val anon = new AbstractThing {
+    val itsalsook = "indeed" // ok
+  }
+
+  implicit object FooImplicit extends Foo {} // assert: DisableSyntax.implicitObject
+
+  implicit def toString(a: Any): String = a.toString // assert: DisableSyntax.implicitConversion
+  implicit def toImplicitString(implicit foo: Foo) = foo.toString // ok
```
olafurpg: I feel like this DisableSyntax.scala test is growing too big, each file should ideally not exceed 50 lines. Small files make it easier to debug breaking tests. Can you please split this into `DisableSyntax2.scala`? (or another name if you can come up with something more creative)

website/src/main/tut/docs/rules/FixSyntax.md
vovapolu: image
```scala
case ((((((((((((a, b), c), d), e), f), g), i), j), k), l), m), n) =>
            DisableSyntaxConfig(a, b, c, d, e, f, g, i, j, k, l, m, n)
```

website/src/main/tut/docs/rules/FixSyntax.md
```diff
@@ -15,8 +15,7 @@ By default, this rule does not rewrite syntax.
 
 It contains the following sub-rules:
```
olafurpg: Remove file?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,145 @@ final case class DisableSyntax(
...
+          if methods.exists(hasDefaultArgs) && config.noDefaultArgs =>
+        methods
+          .filter(hasDefaultArgs)
+          .map(
```
olafurpg: Pro tip, scalafmt dislikes large callbacks with parentheses so it looks a lot nicer with curly braces

olafurpg: some people don't like curly braces for single expressions but I personally disagree!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,145 @@ final case class DisableSyntax(
...
+  }
 
-    treeLintMessages ++ tokensLintMessage ++ regexLintMessages.result()
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case t @ Defn.Class(mods, _, _, _, _)
+          if config.noNonFinalCaseClass &&
+            mods.exists(_.is[Mod.Case]) &&
+            !mods.exists(_.is[Mod.Final]) =>
+        ctx.addLeft(t, "final ")
```
olafurpg: does this handle `@deprecated("blah", "") case class Blah()`? The `final` modifier should appear before `Mod.Case()` but after annotations.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,145 @@ final case class DisableSyntax(
...
+      case t @ Defn.Val(mods, _, _, _)
+          if config.noFinalVal &&
+            mods.exists(_.is[Mod.Final]) =>
+        val finalTokens = mods.find(_.is[Mod.Final]).get.tokens
+        ctx.removeTokens(
+          t.tokens.filter(
+            token =>
+              token.start >= finalTokens(0).start &&
```
olafurpg: Pro tip, `ctx.tokenList.trailingSpaces(Token)` allows you to trim whitespace without random access

olafurpg: we should probably add an overload for `Tree` to avoid repetitive `.tokens.lastOption`!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,145 @@ final case class DisableSyntax(
...
+    ctx.tree.collect {
+      case t @ Defn.Class(mods, _, _, _, _)
+          if config.noNonFinalCaseClass &&
+            mods.exists(_.is[Mod.Case]) &&
+            !mods.exists(_.is[Mod.Final]) =>
+        ctx.addLeft(t, "final ")
+      case t @ Defn.Val(mods, _, _, _)
+          if config.noFinalVal &&
+            mods.exists(_.is[Mod.Final]) =>
+        val finalTokens = mods.find(_.is[Mod.Final]).get.tokens
```
olafurpg: In general, avoid `.get`. Use fold or match instead

website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -48,4 +56,12 @@ DisableSyntax.keywords = [
 DisableSyntax.noTabs = true
 DisableSyntax.noSemicolons = true
 DisableSyntax.noXml = true
+DisableSyntax.noCovariantTypes = true
+DisableSyntax.noContravariantTypes = true
+DisableSyntax.noDefaultArgs = true
+DisableSyntax.noValInAbstract = true
+DisableSyntax.noImplicitObject = true
+DisableSyntax.noImplicitConversion = true
+DisableSyntax.noFinalVal = true
```
olafurpg: It may seem obvious, but it should be explicitly mentioned in the docs that this rule will break valid code if you rely on extending a case class.

website/src/main/tut/docs/rules/FixSyntax.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: docs
+title: FixSyntax
+---
+
+# FixSyntax
```
olafurpg: Remove FixSyntax.md

website/src/main/tut/docs/rules/LeakingSealed.md
```diff
@@ -0,0 +1,25 @@
...
+
+This rule ensures that descendants of a sealed types are final or sealed. 
+
+It disallows the following situation:
+
+`file 1:`
+```scala
+sealed trait t
+trait a extends t
+class c extends t
```
olafurpg: add leading comment where the error message would appear `// error: leaking sealed`

it would also help to have a positive example, where there's a comment `// ok, because final does not leak sealed`

scalafix-tests/input/src/main/scala/test/LeakingSealed.scala
```diff
@@ -0,0 +1,18 @@
+/*
+rules = LeakingSealed
```
olafurpg: Can we add a test case where LeakingSealed and DisableSyntax.noNonFinalCaseClass are both enabled? It seems they might conflict, any idea how we could consolidate them?

website/src/main/tut/docs/rules/FixSyntax.md
vovapolu: Yeah, I forgot. Thanks :)

scalafix-tests/input/src/main/scala/test/LeakingSealed.scala
vovapolu: hmm, `noNonFinalCaseClass` fixes the `LeakingSealed` problem, can we run it in fixed order?

olafurpg: I would recommend baking the two into one rule, something like `MissingFinal`. It can later be extended to add `final` to `def` that are not overridden etc.

vovapolu: Should it add `final` to regular classes when they extend sealed things? And should it ignore traits since they can't be `final`?

vovapolu: @olafurpg what do you think?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
olafurpg: See #553 for a similar case where we addLeft on the `val` modifier tree node instead of enclosing `Defn` tree. The scalameta `List[Mod]` API is annoying but it intentionally treats modifiers as tree nodes in order to accommodate cases like this.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -38,11 +37,10 @@ case class DisableSyntaxConfig(
             c.getField(noImplicitObject) |@|
             c.getField(noImplicitConversion) |@|
             c.getField(noFinalVal) |@|
-            c.getField(noNonFinalCaseClass) |@|
             c.getField(regex)
         ).map {
-          case ((((((((((((a, b), c), d), e), f), g), i), j), k), l), m), n) =>
```
olafurpg: Automatic derivation coming soon! 🙏 https://olafurpg.github.io/metaconfig/#genericderivedecoder

website/src/main/tut/docs/rules/MissingFinal.md
```diff
@@ -0,0 +1,85 @@
...
+* implicit conversions (implicit def with non-implicit parameters) (_Since 0.5.8_)
+
+There're also rules that rewrite syntax:
+
+* transforming `final val` to `val` (It breaks incremental compilation, see more <https://github.com/sbt/zinc/issues/227>) (_Since 0.5.8_)
+ 
+To disallow a syntax:
+
+```
+DisableSyntax.keywords = [
```
olafurpg: Automatic documentation generation coming soon as well! https://olafurpg.github.io/metaconfig/#docs

website/src/main/tut/docs/rules/MissingFinal.md
```diff
@@ -0,0 +1,85 @@
...
+final class e extends t // ok
+```
+
+`file 2:`
+```scala
+trait b extends a // error: leaking sealed
+class d extends c // error: leaking sealed                  
+```
+
+This rule reports errors when a "disallowed" syntax is used. This is a syntactic rule, which means it does not require compilation to run unlike the `Disable` rule.
```
olafurpg: Should this be here? Seems like copy-pasta from DisableSyntax

scalafix-tests/output/src/main/scala/test/MissingFinalRewrite.scala
```diff
@@ -0,0 +1,10 @@
+package test
+
+object MissingFinalRewrite {
+  final case class Bar(i: Int)
+  final case class Ok(ok: String)
+  @deprecated("oh no!", "") final case class BarD(i: Int)
```
olafurpg: Nice 💯

website/src/main/resources/microsite/data/rules.yml
```diff
@@ -13,4 +13,4 @@ rules:
 - Disable
 - DisableSyntax
 - DisableUnless
-- LeakingSealed
```
olafurpg: FYI we'll probably get rid of rules.yml now thanks to tut:passthrough, I will take a look in future PR

scalafix-core/shared/src/main/scala/scalafix/internal/rule/MissingFinal.scala
```diff
@@ -0,0 +1,62 @@
...
+        case Some(mod) => ctx.addLeft(mod, "final ")
+        case None =>
+          mods.lastOption match {
+            case Some(lastMod) => ctx.addRight(lastMod, " final")
+            case None => ctx.addLeft(d, "final ")
+          }
+      }
+
+    ctx.tree.collect {
+      case t @ Defn.Class(mods, _, _, _, _)
```
olafurpg: I suspect we'll want to make it configurable between `case class` vs. leaking sealed. We can delay it for future PR once configuration becomes less boilerplate!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/MissingFinal.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def fix(ctx: RuleCtx): Patch = {
+    def isSealed(tpe: Type): Boolean =
+      ctx.index.denotation(tpe).exists(_.isSealed)
+
+    def leaksSealedParent(mods: List[Mod], templ: Template): Boolean =
+      templ.inits.exists(i => isSealed(i.tpe)) &&
+        mods.forall(m => !m.is[Mod.Final] && !m.is[Mod.Sealed])
+
+    def addFinal(mods: Seq[Mod], d: Defn): Patch =
```
olafurpg: Handy! Might might fit well in the core scalafix. I've written code like this myself a few times.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/MissingFinal.scala
```diff
@@ -0,0 +1,62 @@
...
+import scalafix.{SemanticRule, SemanticdbIndex}
+
+final case class MissingFinal(index: SemanticdbIndex)
+    extends SemanticRule(
+      index,
+      "MissingFinal"
+    ) {
+
+  override def description: String =
+    "Rule that checks for or adds final modifier in corresponding places"
```
olafurpg: Rewrite that inserts `final` to case classes and classes that extend a sealed parent

vovapolu: It inserts `final` only to case classes. I don't think inserting `final` to classes and not inserting anything to traits is a good idea since classes and traits are in the same position.

website/src/main/tut/docs/rules/MissingFinal.md
vovapolu: lol, oops

website/src/main/tut/docs/rules/MissingFinal.md
```diff
@@ -0,0 +1,28 @@
...
+title: MissingFinal
+---
+
+# MissingFinal
+
+_Since 0.5.8_
+
+This rule checks for or adds final modifier in the corresponding places.
+It has two use cases:
+- Check that descendants of a sealed types are final or sealed.
```
olafurpg: > Reports an error on classes that descend a sealed type but are not final.

BTW. Is there a reason why we shouldn't automatically add `final` to these cases?

website/src/main/tut/docs/rules/MissingFinal.md
```diff
@@ -0,0 +1,28 @@
...
+`file 1:`
+```scala
+sealed trait t
+trait a extends t // error: leaking sealed
+class c extends t // error: leaking sealed
+```
+
+`file 2:`
+```scala
+trait b extends a // error: leaking sealed
```
olafurpg: Missing example for case class rewrite.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,139 @@ final case class DisableSyntax(
...
+          }
+      case t @ AbstractWithVals(vals) if config.noValInAbstract =>
+        vals.map { v =>
+          errorCategory
+            .copy(id = "valInAbstract")
+            .at(
+              "val definitions in traits/abstract classes may cause initialization bugs",
+              v.pos)
+        }
+      case t @ Defn.Object(mods, _, _)
```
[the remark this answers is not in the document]
vovapolu: Well, I don't trust `.parent`. :) Intuitively I prefer to traverse tree from top to down and don't look to the parents. Is it important?

website/src/main/tut/docs/rules/MissingFinal.md
vovapolu: Yeah, I also did it initially :) But there's some imbalance if we add `final` to classes, but not `sealed` to traits, although they're almost identical from this point of view.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,139 @@ final case class DisableSyntax(
...
+        methods
+          .filter(hasDefaultArgs)
+          .map { m =>
+            errorCategory
+              .copy(id = "defaultArgs")
+              .at(
+                "Default args makes it hard to use methods as functions.",
+                m.pos)
+          }
+      case t @ AbstractWithVals(vals) if config.noValInAbstract =>
```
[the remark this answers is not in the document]
olafurpg: avoiding parent is a good idea indeed, I agree your PR is a better implementation 👍

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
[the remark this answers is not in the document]
olafurpg: You're actually right, let's not use parent. good call!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,139 @@ final case class DisableSyntax(
...
+        Seq(
           errorCategory
             .copy(id = "contravariant")
             .at(
               "Contravariant types could lead to error-prone situations.",
               t.pos
             )
-      }
+        )
+      case t @ WithMethods(methods)
```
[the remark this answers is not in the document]
olafurpg: Term.Parameter is for method definition site only, for callsite `foo(arg = b)` there is another tree node Term.Arg.Named.
```scala
@ q"def foo(a: Int = 2) = a"
res1: Defn.Def = Defn.Def(
  List(),
  Term.Name("foo"),
  List(),
  List(List(Term.Param(List(), Term.Name("a"), Some(Type.Name("Int")), Some(Lit.Int(2))))),
  None,
  Term.Name("a")
)

@ q"foo(a = 3)"
res2: Term.Apply = Term.Apply(Term.Name("foo"), List(Term.Arg.Named(Term.Name("a"), Lit.Int(3))))
```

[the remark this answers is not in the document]
olafurpg: Aaah, so you only want to match methods global symbols and not local defs inside of other blocks. Now I get it, this implementation is great then 👍

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -35,38 +37,139 @@ final case class DisableSyntax(
...
+          case _ => None
+        }
+        if (vals.isEmpty) None else Some(vals)
+      }
+    }
+
+    object WithMethods {
+      def unapply(t: Tree): Option[List[Defn.Def]] = {
+        val stats = t match {
+          case Defn.Class(_, _, _, _, templ) => templ.stats
```
olafurpg: Why not match on `Template`? This implementation currently does not handle defs in `object`

vovapolu: Mmm, because I didn't know about it :D

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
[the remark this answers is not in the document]
olafurpg: OK, I see there is a test for
```scala
object a {
  def bar(x: Int = 42) = 5 // ok
}
```
Why is it OK to use default args for defs on objects?

olafurpg:
LGTM 👍 Let's address comments in separate PRs to unblock the release. In the future @vovapolu , please keep the number of new features small in each PR to simplify review.

Issue #542, opened 2018-01-04
Make val field in implicit class private


olafurpg:
I just observed that the `val` field in an implicit class is accesible as an extension method
```scala
scala> implicit class XtensionString(val str: String) extends AnyVal {
     |   def doubled = str + str
     | }
defined class XtensionString

scala> "message".str
res0: String = message
```
A workaround to prevent this is to make the field private
```scala
scala> implicit class XtensionString(private val str: String) extends AnyVal
```
It would be nice to have a scalafix rule to automatically fix this.

Context: https://gitter.im/scala/contributors?at=5a4e57685355812e57502b5b

LeonardMeyer:
I took a stab at this. Can't really test for now because of #547 but will do.

I do have one question though : Why does `Ctor.Primary` of ScalaMeta has a `List[List[Term.Param]]` ? Why two levels ?

olafurpg:
Cool @LeonardMeyer ! I will try to take a look at #547 in the coming week.

Primary constructors can be curried which are represented as `List[List[Term.Param]]`
```scala

@ show(q"class A(a: Int)(b: String)".ctor)
Ctor.Primary(
  List(),
  _,
  List(
    List(Term.Param(List(), Term.Name("a"), Some(Type.Name("Int")), None)),
    List(Term.Param(List(), Term.Name("b"), Some(Type.Name("String")), None))
  )
)
```

LeonardMeyer:
I literally realized this right before coming back here 😄 By the way, even private won't help with something like this : 

```
  class A(val str: String) {
    def doubled: String = str + str
  }

  implicit class B(override val str: String) extends A(str)

  "test".str  //visible
```

LeonardMeyer:
I have a question but nobody seemed active on Gitter so I'll repeat it here : can someone explain to me how tests are configured ? I understand comments play a role in it but I can't figure it out. All I saw was [this](https://scalacenter.github.io/scalafix/docs/rule-authors/setup#scalafix-testkit) but it's pretty vague about configuration...

PR #553, opened 2018-01-13
New ImplicitClassPrivateParams rule #542

LeonardMeyer:


scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateParams.scala
```diff
@@ -0,0 +1,52 @@
+package scalafix.internal.rule
+
+import scala.meta.Term.Param
+import scala.meta.{Defn, Mod}
+import scalafix.rule.RuleCtx
+import scalafix.{Patch, Rule}
+
+case object ImplicitClassPrivateParams
+    extends Rule("ImplicitClassPrivateParams") {
```
olafurpg: `ImplicitClassPrivateVal`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateParams.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  override def description: String =
+    "Add private access modifier to val or var parameters of implicit classes in order to prevent public access"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case Defn.Class(mods, _, _, primCtor, _) if hasImplicitMod(mods) =>
```
olafurpg: I think we can refine the condition to only apply for AnyVals that are missing `private`

```scala
        t.collect {
          case Defn.Class(
              cmods,
              _,
              _,
              Ctor.Primary(_, _, (Term.Param(pmods, _, _, _) :: Nil) :: Nil),
              Template(_, init"AnyVal" :: Nil, _, _)
              ) =>
            for {
              _ <- cmods.find(_.is[Mod.Implicit])
              if !pmods.exists(_.is[Mod.Private])
              valmod <- pmods.find(_.is[Mod.ValParam])
            } yield ctx.addLeft(valmod, "private ")
        }
```

For non-anyvals it's possible to remove the `val` parameter, `implicit class Foo(x: Int)  { .. }`. Classes that `extends AnyVal` are however forced to add the `val` modifier which makes  it easy to accidentally forget `private`

scalafix-tests/input/src/main/scala/test/ImplicitClassPrivateParams.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  implicit class XtensionAnnotatedVal(@transient val str: String) {
+    def doubled: String = str + str
+  }
+
+  implicit class XtensionAnnotatedFinalVal(@transient final val str: String) {
+    def doubled: String = str + str
+  }
+
+  implicit class XtensionVar(var str: String) {
```
olafurpg: We can ignore `var` cases since those are less commonly (if ever) used for implicit classes.

scalafix-tests/input/src/main/scala/test/ImplicitClassPrivateParams.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  implicit class XtensionAnnotatedProtectedVal(@transient protected val int: Int) {
+    def doubled: Int = int + int
+  }
+
+  implicit class XtensionAnnotatedProtectedVar(@transient protected var int: Int) {
+    def doubled: Int = int + int
+  }
+
+  implicit class Xtension(int: Int) {
```
olafurpg: We can ignore non-AnyVals since those fields are private by default unless they're case classes, but case classes are rarely used for implicit classes.

scalafix-tests/input/src/main/scala/test/ImplicitClassPrivateParams.scala
LeonardMeyer: Isn't it best to keep these as non reg tests ? Unless you think this is being overzealous.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateParams.scala
LeonardMeyer: Got it. Can you explain to me how `is()` is working here ? I noticed it but thought it only worked with tokens. Where are the instances for `Classifiable`/`Classifier` ?

olafurpg: The instances are generated by the scalameta `@ast` macro annotations. It's a flexible mechanism to guard against categories of trees/tokens both via `.is[T]` or in pattern matching with `case T()`. You can build custom groups of classes like we do in scalafix https://github.com/scalacenter/scalafix/blob/00199145261b958fb12cf256773c2ed08f336359/scalafix-core/shared/src/main/scala/scalafix/util/TokenClasses.scala

scalafix-tests/input/src/main/scala/test/ImplicitClassPrivateParams.scala
olafurpg: Negative tests are good :) I probably don't write enough of them. We don't need too many here however since the implementation guards against `val`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateParams.scala
LeonardMeyer: Ok thanks. One last question : your implementation ignores cases where `final` or `protected` is present. Don't we care ?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateParams.scala
```diff
@@ -0,0 +1,25 @@
...
+  override def description: String =
+    "Add private access modifier to val or var parameters of implicit classes in order to prevent public access"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case Defn.Class(cMods, _, _, Ctor.Primary(_, _, (Term.Param(pMods, _, _, _) :: Nil) :: Nil), Template(_, init"AnyVal" :: Nil, _, _)) =>
+        val optPatch = for {
+          _ <- cMods.find(_.is[Mod.Implicit])
+          if !pMods.exists(_.is[Mod.Private])
+          valMod <- pMods.find(_.is[Mod.ValParam])
```
olafurpg: You are correct that this does not handle `final val param: T`. We can use `.find(!_.is[Mod.Annot])` instead, but adding private can still break cases like `override val param: T`. We should probably exclude cases that contain `override`.

LeonardMeyer: I think `override` isn't an issue since we're only looking for implicit value classes and it can't extend both `AnyVal` and a base class.

olafurpg: AnyVals can inherit from "universal traits"

```scala
@ trait Foo extends Any {
    def bar: Int
    def plusOne = bar + 1
  }
defined trait Foo

@ implicit class UInt(override val bar: Int) extends AnyVal with Foo
defined class UInt

@ 2.plusOne
res4: Int = 3
```

See http://docs.scala-lang.org/overviews/core/value-classes.html

LeonardMeyer: Indeed. This case won't be taken into account though in your implem because you explicitly target a sole `Init` (`AnyVal` here). Should we remove this limitation and account for this : 
```scala
trait BaseClass {
  def bar: Int
}

implicit class XtensionOverrideVal(val bar: Int) extends AnyVal with BaseClass
```

LeonardMeyer: Indeed. This case won't be taken into account though, because in your implem you explicitly target a sole Init (AnyVal here).

website/src/main/tut/docs/rules/ImplicitClassPrivateVal.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: docs
+title: ImplicitClassPrivateVal
+---
+
+# ImplicitClassPrivateVal
+
+`val` fields of implicit classes are accessible as an extension.
```
olafurpg: Non-private `val` fields of implicit classes leak as publicly accessible extension methods.

website/src/main/tut/docs/rules/ImplicitClassPrivateVal.md
```diff
@@ -0,0 +1,25 @@
...
+"message".str // compiles
+
+// after
+implicit class XtensionValFixed(private val str: String) extends AnyVal {
+  def doubled: String = str + str
+}
+"message".str // does not compile
+```
+
+This rule safely ignores all other modifiers.
```
olafurpg: Note. This rule only triggers for `val` fields, it ignores other modifiers such as `var`.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ImplicitClassPrivateVal.scala
```diff
@@ -0,0 +1,33 @@
...
+    "Add private access modifier to val parameters of implicit value classes in order to prevent public access"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case Defn.Class(
+          cMods,
+          _,
+          _,
+          Ctor.Primary(_, _, (Term.Param(pMods, _, _, _) :: Nil) :: Nil),
+          Template(_, init"AnyVal" :: Nil, _, _))
```
olafurpg: we can lift the restriction that it cannot mix in universal classes, if you are feeling more ambitious 😄

LeonardMeyer: I think it's better to wait and see if people actually need that, because as you said, the error is more likely with implicit value classes. Speaking of, should I rename the rule to `ImplicitValueClassPrivateVal` ?

olafurpg: > I think it's better to wait and see if people actually need that because as you said, the error is more likely to happen with implicit value classes.

I agree.

> Speaking of, should I rename the rule to ImplicitValueClassPrivateVal ?

Good question. How about `LeakingImplicitClassVal`? That gives us room to later support non-AnyVal in case there's demand. I think it's good to hint in the name what the problem is and the `Leaking*` prefix is consistent with other upcoming rules like `LeakingSealed`

LeonardMeyer:
@olafurpg Fixed it

PR #556, opened 2018-01-17
Make DottyAutoTuplingFunctionArgs semantic

### LINKED ISSUE #521: DottyAutoTuplingFunctionArgs needs to be semantic
@allanrenucci pointed out that DottyAutoTuplingFunctionArgs can't be syntactic https://github.com/scalacenter/scalafix/pull/513#issuecomment-353049179

I believe we have sufficient data to correctly handle tricker cases, but it's far from easy to do with the current APIs. As I understand it correctly, we need to add an additional guard against cases where tuples are deconstructed `case (a1, a2,..., aN)`  and the expected argument type is `Function1[A, B]` and A is a subtype of `ProductN`. Here is an example where the  rewrite currently triggers when it should not do anything

```scala
scala> def foo(xs: List[Any]) = xs.map { case (x, y) => 1 } 
def foo(xs: List[Any]): List[Int]
scala> def foo(xs: List[Any]) = xs.map((x, y) => 1 ) 
1 |def foo(xs: List[Any]) = xs.map((x, y) => 1 )
  |                                ^^^^^^^^^^^
  |                                wrong number of parameters, expected: 1
```

olafurpg:
Fixes #521. While thinking about the problem more I realized we are able to tell with the current API if all of the deconstructed arguments are `scala.Any`. This is sufficient for us to guard against the cases where `Any` or a tparam `T` is pattern matched into `(Any, Any)`. 

Review @marcelocenerine

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
@@ -1,0 +1,50 @@
+package scalafix.internal.rule
+
+import scala.meta._
+import scala.meta.tokens.Token._
+import scalafix.Patch
+import scalafix.rule.{Rule, RuleCtx}
+
+case object DottyAutoTuplingFunctionArgs
+    extends Rule("DottyAutoTuplingFunctionArgs") {
+  override def description: String =
+    "Rewrite that removes pattern-matching decomposition if function arguments can be automatically tupled in Dotty."
+
+  override def fix(ctx: RuleCtx): Patch =
+    ctx.tree.collect {
+      case Term.Apply((_, List(pf: Term.PartialFunction)))
+          if canBeAutoTupled(pf) =>
+        val pfTokens = pf.tokens
+
+        (for {
+          leftBrace <- pfTokens.headOption if leftBrace.is[LeftBrace]
+          rightBrace <- pfTokens.lastOption if rightBrace.is[RightBrace]
+          kwCase <- pfTokens.find(_.is[KwCase])
+        } yield {
+          val tkList = ctx.tokenList
+          val replacedBraces =
+            if (canReplaceBracesWithParens(pf)) {
+              ctx.replaceToken(leftBrace, "(") +
+                ctx.replaceToken(rightBrace, ")") +
+                ctx.removeTokens(tkList.leadingSpaces(leftBrace)) +
+                ctx.removeTokens(tkList.trailingSpaces(leftBrace)) +
+                ctx.removeTokens(tkList.leadingSpaces(rightBrace))
+            } else Patch.empty
+
+          replacedBraces +
+            ctx.removeToken(kwCase) +
+            ctx.removeTokens(tkList.trailingSpaces(kwCase))
+        }).asPatch
+    }.asPatch
+
+  private def canBeAutoTupled(pf: Term.PartialFunction): Boolean =
+    pf.cases match {
+      case Case(Pat.Tuple(args), None, _) :: Nil
+          if args.forall(a => a.is[Pat.Var] || a.is[Pat.Wildcard]) =>
+        true
+      case _ => false
+    }
+
+  private def canReplaceBracesWithParens(pf: Term.PartialFunction): Boolean =
+    pf.pos.startLine == pf.pos.endLine && pf.cases.head.stats.size == 1
+}
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
@@ -2,11 +2,12 @@ package scalafix.internal.rule
 
 import scala.meta._
 import scala.meta.tokens.Token._
-import scalafix.Patch
-import scalafix.rule.{Rule, RuleCtx}
+import scalafix._
+import scalafix.syntax._
+import scalafix.util.SymbolMatcher
 
-case object DottyAutoTuplingFunctionArgs
-    extends Rule("DottyAutoTuplingFunctionArgs") {
+case class DottyAutoTuplingFunctionArgs(index: SemanticdbIndex)
+    extends SemanticRule(index, "DottyAutoTuplingFunctionArgs") {
   override def description: String =
     "Rewrite that removes pattern-matching decomposition if function arguments can be automatically tupled in Dotty."
 
@@ -37,11 +38,28 @@ case object DottyAutoTuplingFunctionArgs
         }).asPatch
     }.asPatch
 
+  private val ScalaAny = SymbolMatcher.exact(Symbol("_root_.scala.Any#"))
+  private def isMaybeAny(name: Term.Name): Boolean =
+    (for {
+      symbol <- name.symbol
+      tpe <- symbol.resultType
+    } yield ScalaAny.matches(tpe)).getOrElse(true)
+
+  // See https://github.com/scalacenter/scalafix/issues/521 why we
+  // skip the rewrite in this case.
+  private def allNamesAreAny(args: List[Pat]): Boolean = args.forall {
+    case Pat.Var(name) => isMaybeAny(name)
+    case _ => true
+  }
+
   private def canBeAutoTupled(pf: Term.PartialFunction): Boolean =
     pf.cases match {
-      case Case(Pat.Tuple(args), None, _) :: Nil
-          if args.forall(a => a.is[Pat.Var] || a.is[Pat.Wildcard]) =>
-        true
+      case Case(Pat.Tuple(args), None, _) :: Nil =>
+        args.forall {
+          case Pat.Var(_) => true
+          case Pat.Wildcard() => true
+          case _ => false
+        } && !allNamesAreAny(args)
       case _ => false
     }
 
--- a/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
@@ -1,0 +1,36 @@
+/*
+rules = DottyAutoTuplingFunctionArgs
+ */
+package test
+
+object DottyAutoTuplingFunctionArgs {
+  Some((1, 2)).map { case (a, b) => a + b }
+  List((1, 2, 3), (4, 5, 6)).map { case (a, b, c) => println(a); a + b + c }
+  List((1, 2), (3, 4)).foreach { case (a, b) => { val c = a + b; println(c) } }
+  Set((1, 2, 3)).map { case (a, b, c) =>
+    println(a + b + c)
+  }
+  List((1, 1, 1, 1, 1), (2, 2, 2, 2, 2)).flatMap{
+    case (_, _, _, _, n) => List(n)
+  }
+  val xs = List("3", "1", "0").zipWithIndex
+  xs.filter { case (w, i) => w == i.toString }
+
+  // should not rewrite the below
+  xs.map { case (w, i) if i > 1 => s"$w-$i" } // case with guards
+  List((Some(1), Some(1))).map { case (Some(a), Some(b)) => a + b } // extraction
+  xs.map {
+    case (a, b) => a + b
+    case _ => 0
+  } // multiple cases
+  xs.find { case (_, x @ b) => x - b == 0 } // binding
+  xs.map { case ("0", 0) => "zero" } // literals
+  List(("a", 1, false, 1.0), ("b", 2, true, 2.0)).exists{
+    case (_: String, _: Int, c: Boolean, _: Double) => c
+  } // typed parameters
+  class A
+  class B extends A
+  def foo(xs: List[(A, A)]) = xs.map {
+    case (x: B, y: A) => 1
+  } // typed parameters not matching tuple element types
+}
--- a/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
+++ b/scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
@@ -33,4 +33,8 @@ object DottyAutoTuplingFunctionArgs {
   def foo(xs: List[(A, A)]) = xs.map {
     case (x: B, y: A) => 1
   } // typed parameters not matching tuple element types
+  def any(xs: List[Any]): List[Int] = xs.map { case (x, y) => 1 }
+  def anyIntPair(xs: List[(Int, Any)]): List[Int] = xs.map { case (x, y) => 1 }
+  def anyAnyPair(xs: List[(Any, Any)]): List[Int] = xs.map { case (x, y) => 1 } // false negative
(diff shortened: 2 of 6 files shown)


scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,40 @@
+/*
+rules = DottyAutoTuplingFunctionArgs
+ */
+package test
+
+object DottyAutoTuplingFunctionArgs {
```
MasseGuillaume: I would also add:
```scala
def foo[T](xs: List[T]) = xs.map { case (x, y) => 1 }
```

https://github.com/scalacenter/scalafix/issues/521#issuecomment-353063569

```scala
List(1, 2, 3).zipWithIndex.drop(1).take(1).reverse.map { case (x, y) => 1 }
  {
    if (Random.nextInt() > 0) Option(("a", "a"))
    else Option(("b", "b"))
  }.map { case (a, b) => a + b }
class Foo[T](val a: T)
implicit def foo2Opt[T](f: Foo[T]): Option[(T, T)] = Some((f.a, f.a))
new Foo(1).map { case (x, y) => 1 }
```

https://github.com/scalacenter/scalafix/issues/521#issuecomment-353363529

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
```diff
@@ -0,0 +1,23 @@
...
+
+_Since v0.5.8_
+
+# DottyAutoTuplingFunctionArgs
+
+Removes pattern-matching decomposition if function arguments can be automatically tupled. See [http://dotty.epfl.ch/docs/reference/auto-parameter-tupling.html](http://dotty.epfl.ch/docs/reference/auto-parameter-tupling.html)
+
+```scala
+// before
+xs.map { case (x, y) => x + y }
```
MasseGuillaume: what is the type of `xs` ?

MasseGuillaume: `val xs: List[(Int, Int)]`

MasseGuillaume: Maybe put a concrete example like:

```scala
val ranges: List[(Int, Int)] = ((1, 5), (9, 11))
ranges.flatMap((start, end) => start to end)
```

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
```diff
@@ -0,0 +1,23 @@
...
+// after
+xs.map((x, y) => x + y)
+```
+
+Auto-tupling is a language feature only supported in Dotty. Therefore, the code produced by this rewrite does not compile with Scala 2.x.
+
+## Caveats
+
+- the rewrite has false negatives, meaning it may not trigger for cases where the rewrite would still be safe. In particular, the rewrite ignores cases when all deconstructed arguments are `Any`.
```
MasseGuillaume: I don't see false negatives flagged in the unit test above.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,68 @@
...
+  private def isMaybeAny(name: Term.Name): Boolean =
+    (for {
+      symbol <- name.symbol
+      tpe <- symbol.resultType
+    } yield ScalaAny.matches(tpe)).getOrElse(true)
+
+  // See https://github.com/scalacenter/scalafix/issues/521 why we
+  // skip the rewrite in this case.
+  private def allNamesAreAny(args: List[Pat]): Boolean = args.forall {
+    case Pat.Var(name) => isMaybeAny(name)
```
MasseGuillaume: just `Pat.Var` ?

olafurpg: Yes, this is the only case where the rewrite triggers.

scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
olafurpg: The first one is already covered the bottom of the file, I don't see what new condition the other cases are supposed to cover.

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
olafurpg: Done

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
olafurpg: There is a unit test with trailing `// false negative` comment.

MasseGuillaume: I'm blind :+1:

website/src/main/tut/docs/rules/DottyAutoTuplingFunctionArgs.md
MasseGuillaume: I forgot **List**
```scala
val ranges: List[(Int, Int)] = List((1, 5), (9, 11))
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,68 @@
...
+  private def canBeAutoTupled(pf: Term.PartialFunction): Boolean =
+    pf.cases match {
+      case Case(Pat.Tuple(args), None, _) :: Nil =>
+        args.forall {
+          case Pat.Var(_) => true
+          case Pat.Wildcard() => true
+          case _ => false
+        } && !allNamesAreAny(args)
+      case _ => false
+    }
```
marcelocenerine: there is one scenario that was pointed out here https://github.com/scalacenter/scalafix/pull/513#issuecomment-353062915 that should be skipped:

```scala
  def foo2(f: PartialFunction[(Int, Int), Int]): Int = ???
  foo2 { case (x, y) => 1 }
```

this should do the job:
```scala
  override def fix(ctx: RuleCtx): Patch = {
    ctx.tree.collect {
      case Term.Apply((Term.Select(_, name), List(pf: Term.PartialFunction)))
    // omitted
  }

  private def canBeAutoTupled(method: Term.Name, pf: Term.PartialFunction): Boolean =
    if (!hasPartialFunctionParam(method)) {
      // omitted
    } else false

  private val PartialFunction = SymbolMatcher.exact(Symbol("_root_.scala.PartialFunction#"))
  private def hasPartialFunctionParam(methodName: Term.Name): Boolean =
    (for {
      denotation <- index.denotation(methodName)
      paramTypeName <- denotation.names.headOption
      paramTypeSymbol = paramTypeName.symbol
    } yield PartialFunction.matches(paramTypeSymbol)).getOrElse(false)
```

scalafix-tests/output-dotty/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,37 @@
...
+  } // typed parameters
+  class A
+  class B extends A
+  def foo(xs: List[(A, A)]) = xs.map {
+    case (x: B, y: A) => 1
+  } // typed parameters not matching tuple element types
+  def any(xs: List[Any]): List[Int] = xs.map { case (x, y) => 1 }
+  def anyIntPair(xs: List[(Int, Any)]): List[Int] = xs.map((x, y) => 1)
+  def anyAnyPair(xs: List[(Any, Any)]): List[Int] = xs.map { case (x, y) => 1 } // false negative
+  def tparam[T](xs: List[T]): List[Int] = xs.map { case (x, y) => 1 }
```
marcelocenerine: there is one comment in line 16 which separates the scenarios that should be rewritten from the ones that should not. Maybe move the line 34 to somewhere before that comment?

scalafix-tests/input/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
marcelocenerine: yeah, the other cases have no importance given the solution that Olafur has implemented.

allanrenucci:
What about?
```scala
def product2(xs: List[Product2[Int, Int]]): List[Int] = xs.map { case (x, y) => x + y } 
```

marcelocenerine:
I tested the rewrite for this in the dotty repl last night and it worked

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
olafurpg: If I understand correctly, auto-tupling currently works fine for partial functions. In the dotty repl
```scala
@ def foo2(f: PartialFunction[(Int, Int), Int]): Int = f(1 -> 2)
defined function foo2

@   foo2 { case (x, y) => 1 }
res3: Int = 1
```

Has there been any change @allanrenucci ?

scalafix-tests/output-dotty/src/main/scala/test/DottyAutoTuplingFunctionArgs.scala
olafurpg: Good idea, moved these tests above that comment.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
olafurpg: That repl session was actually ammonite, but dotty is the same
```scala
scala> def foo2(f: PartialFunction[(Int, Int), Int]): Int = f(1 -> 2)
def foo2(f: PartialFunction[(Int, Int), Int]): Int
scala> foo2 { case (x, y) => 1 }
val res0: Int = 1
scala> foo2((x, y) => 1)
val res1: Int = 1
```

allanrenucci:
```scala
def product2(xs: List[Product2[Int, Int]]): List[Int] = xs.map((x, y) => 1)
```
is supposed to compile in Dotty but the semantic is not the same:
```scala
class Foo extends Product2[Int, Int] {
  def _1 = 1
  def _2 = 2
  def canEqual(that: Any) = ???
}

val xs: List[Product2[Int, Int]] = List(new Foo)
xs.map((x, y) => 1) // List(1)
xs.map { case (x, y) => 1 } // MatchError
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
allanrenucci: auto-tupling currently works for any subtype of `FunctionN` and SAM types. There is an open issue that hasn't been discussed: lampepfl/dotty#3682

olafurpg:
Interesting. Is there a possibility that by triggering the rewrite in such a case that scalafix turns a non-crashing program (no MatchError) into another non-crashing program with different semantics? If the underlying Product2 is a Tuple2 then the semantics are the same. The rewrite only triggers if there is one `case` branch so non-TupleN values would unconditionally fail with MatchError.

allanrenucci:
```scala
xs.collect { case (x, y) => x + y } // List()
xs.collect((x, y) => x + y) // List(3)
```
One more example that convinces me that you shouldn't rewrite when it is a partial function

olafurpg:
Nice catch! MatchError is an observable effect so it's possible to devise cases for non-PartialFunction that exhibit the same problem

```scala
scala> def fakeCollect(p: Product2[Int, Int])(f: Product2[Int, Int] => String) = try f(p) catch { case _: MatchError => "ERROR" }
def fakeCollect(p: Product2[Int, Int])(f: Product2[Int, Int] => String): String
scala> class Foo extends Product2[Int, Int] {
           def _1 = 1
           def _2 = 2
           def canEqual(that: Any) = ???
         }
// defined class Foo
scala> fakeCollect(new Foo) { case (a, b) => "OK" }
val res5: String = "ERROR"
scala> fakeCollect(new Foo)((a, b) => "OK")
val res6: String = "OK"
```
Do you know why deconstructing ProductN with `case (a, b, c)` is not reported as a compiler error? This seems like a bug the compiler should complain about like it does for other similar cases
```scala
scala> val List(1) = 2
1 |val List(1) = 2
  |    ^^^^^^^
  |    cannot test if value types are references
```

Either way, it seems the workaround in this PR is insufficient due to how deconstructed names on ProductN can infer non-Any types. We learned a bit more at least :) Thanks a lot @allanrenucci !

allanrenucci:
> Do you know why deconstructing ProductN with `case (a, b, c)` is not reported as a compiler error? This seems like a bug the compiler should complain about like it does for other similar cases

I am not sure to understand. Why would this be an error? `TupleN` is a subtype of `ProductN`. We emit an error when the types are unrelated.

olafurpg:
For what type `T <: Product2` that is not a subtype of `TupleN` is it safe to deconstruct with the `case (a, b,... n)` syntax? If it's only safe to use that syntax with `TupleN` then why does the compiler permit deconstructing `ProductN`?

olafurpg:
Nevermind my question, the current compiler behavior makes sense to me now.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyAutoTuplingFunctionArgs.scala
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private def canBeAutoTupled(pf: Term.PartialFunction): Boolean =
+    pf.cases match {
+      case Case(Pat.Tuple(args), None, _) :: Nil =>
+        args.forall {
+          case Pat.Var(_) => true
+          case Pat.Wildcard() => true
+          case _ => false
+        } && !allNamesAreAny(args)
```
allanrenucci: Thinking more about it, I think the rewrite would be safe if you substitute `!allNamesAreAny(args)` by this condition:

`pf.tpe` of the form `A => B` where `A =:= TupleN` for `N = pf.cases.length`

olafurpg: thanks @allanrenucci , that is worth exploring!

marcelocenerine: In order to figure out what `A` is, I guess we will need the `tpe` semantic operator you mentioned here https://github.com/scalacenter/scalafix/issues/521#issuecomment-358018225

allanrenucci: >by N = pf.cases.length I believe @allanrenucci probably meant N = args.length, where args comes from:

Yep! Thanks for the correction

PR #561, opened 2018-01-19
Add documentation for DisableSyntax.regex

MasseGuillaume:


website/src/main/tut/docs/rules/DisableSyntax.md
```diff
@@ -49,4 +49,11 @@ DisableSyntax.noSemicolons = true
 DisableSyntax.noXml = true
 DisableSyntax.noCovariantTypes = true
 DisableSyntax.noContravariantTypes = true
-```
\ No newline at end of file
+DisableSyntax.regex = [
```
olafurpg: LGTM 👍

olafurpg:
Pro tip, if you push to the main origin you trigger two travis jobs which makes CI run slower. We should also keep the main origin only contain `master` and `gh-pages` in order to simplify cloning the repo for contributors. I don't like cloning new repos and get 100s of stale WIP/topic/feature branches that will likely never get merged.

PR #562, opened 2018-01-19
Fix several ExplicitResultTypes bugs.

olafurpg:
- unary functions were printed as Function0[T] instead of () => T
- local symbols caused MatchError in TypeSyntax.prettify
- upgrade to scalameta 2.1.6 to to enjoy https://github.com/scalameta/scalameta/pull/1216

These fixes combined make ExplicitResultTypes safe to run on over 60k loc in
the akka build. It still has some bugs left, see
https://github.com/scalameta/scalameta/issues/1222

### DIFF
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -250,6 +250,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       else if (isSnapshot.value) Opts.resolver.sonatypeSnapshots
       else Opts.resolver.sonatypeStaging
     },
+    resolvers += Opts.resolver.sonatypeSnapshots,
     scmInfo := Some(
       ScmInfo(
         url("https://github.com/scalacenter/scalafix"),
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -250,7 +250,6 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       else if (isSnapshot.value) Opts.resolver.sonatypeSnapshots
       else Opts.resolver.sonatypeStaging
     },
-    resolvers += Opts.resolver.sonatypeSnapshots,
     scmInfo := Some(
       ScmInfo(
         url("https://github.com/scalacenter/scalafix"),
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
@@ -13,7 +13,7 @@ object TypeSyntax {
       implicit index: SemanticdbIndex): (Type, Patch) = {
 
     val functionN: SymbolMatcher = SymbolMatcher.exact(
-      1.to(22).map(i => Symbol(s"_root_.scala.Function$i#")): _*
+      0.to(22).map(i => Symbol(s"_root_.scala.Function$i#")): _*
     )
     val tupleN: SymbolMatcher = SymbolMatcher.exact(
       1.to(22).map(i => Symbol(s"_root_.scala.Tuple$i#")): _*
@@ -33,7 +33,7 @@ object TypeSyntax {
       }
     }
 
-    def stableRef(sym: Symbol): (Patch, Tree) = {
+    def stableRef(sym: Symbol.Global): (Patch, Tree) = {
       var patch = Patch.empty
       def loop[T: ClassTag](symbol: Symbol): T = {
         val result = symbol match {
@@ -87,10 +87,12 @@ object TypeSyntax {
         case Type.Apply(tupleN(_), args) =>
           val rargs = apply_![Type](args)
           Type.Tuple(rargs)
-        case Name(_) :&&: index.Symbol(sym) =>
+        case Name(_) :&&: index.Symbol(sym: Symbol.Global) =>
           val (addImport, t) = stableRef(sym)
           patch += addImport
           t
+        case Type.Select(Term.This(_), _) =>
+          tree
         case Type.Select(qual, name) =>
           Type.Select(apply_![Term.Ref](qual), name)
         case els =>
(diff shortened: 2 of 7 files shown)


olafurpg:
The full diff from running the rewrite on akka can be seen here https://github.com/akka/akka/compare/master...olafurpg:explicit-result-types?expand=1

2,232 additions and 2,223 deletions with a single discovered bug causing 4 compile errors reported in https://github.com/scalameta/scalameta/issues/1222

olafurpg:
Pending fix for the last bug https://github.com/scalameta/scalameta/pull/1224

project/ScalafixBuild.scala
```diff
@@ -250,6 +250,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       else if (isSnapshot.value) Opts.resolver.sonatypeSnapshots
       else Opts.resolver.sonatypeStaging
     },
+    resolvers += Opts.resolver.sonatypeSnapshots,
```
MasseGuillaume: Hum, why?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -87,12 +87,16 @@ object TypeSyntax {
         case Type.Apply(tupleN(_), args) =>
           val rargs = apply_![Type](args)
           Type.Tuple(rargs)
-        case Name(_) :&&: index.Symbol(sym) =>
+        case Name(_) :&&: index.Symbol(sym: Symbol.Global) =>
```
MasseGuillaume: what is `:&&:` ?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -33,7 +33,7 @@ object TypeSyntax {
       }
     }
 
-    def stableRef(sym: Symbol): (Patch, Tree) = {
+    def stableRef(sym: Symbol.Global): (Patch, Tree) = {
```
MasseGuillaume: missing scaladoc :D

MasseGuillaume: should be private

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -13,7 +13,7 @@ object TypeSyntax {
       implicit index: SemanticdbIndex): (Type, Patch) = {
```
MasseGuillaume: This could be useful:
https://github.com/lihaoyi/PPrint/blob/master/pprint/shared/src/main/scala/pprint/TPrint.scala
https://github.com/lihaoyi/PPrint/blob/master/build.sbt#L53

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -33,7 +33,7 @@ object TypeSyntax {
       }
     }
 
-    def stableRef(sym: Symbol): (Patch, Tree) = {
+    def stableRef(sym: Symbol.Global): (Patch, Tree) = {
       var patch = Patch.empty
       def loop[T: ClassTag](symbol: Symbol): T = {
```
MasseGuillaume: I would like to see an example of each case.

scalafix-tests/output/src/main/scala/test/explicitResultTypes/ExplicitResultTypesPathDependent.scala
```diff
@@ -3,9 +3,14 @@ package test.explicitResultTypes
...
+    implicit val x: Path.this.B = new B
     implicit val y: x.C = new x.C
     def gimme(yy: x.C) = ???; gimme(y)
   }
   implicit val b: _root_.test.explicitResultTypes.ExplicitResultTypesPathDependent.Path#B = new Path().x
+  trait Foo[T] {
+    type Self
+    def bar: Self
+  }
+  implicit def foo[T]: _root_.test.explicitResultTypes.ExplicitResultTypesPathDependent.Foo[T]#Self = null.asInstanceOf[Foo[T]].bar
```
MasseGuillaume: Why it's not possible to handle this case?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -87,12 +87,16 @@ object TypeSyntax {
...
+        case Name(_) :&&: index.Symbol(sym: Symbol.Global) =>
           val (addImport, t) = stableRef(sym)
           patch += addImport
           t
+        case Type.Select(Term.This(_), _) =>
+          tree
         case Type.Select(qual, name) =>
           Type.Select(apply_![Term.Ref](qual), name)
+        case Type.Project(qual, name) =>
+          Type.Project(apply_![Type](qual), name)
```
MasseGuillaume: the `typeMismatch` branch is not tested.

scalafix-tests/output/src/main/scala/test/explicitResultTypes/ExplicitResultTypesPathDependent.scala
olafurpg: What do you mean? We correctly handle this case now. Previously we would discard the `[T]` parameter creating broken code.

This test puts only fully _root_ qualified names. The `Path.this` example above is as fully qualified as it can be, prefixing `Path` with `_root_...` will make the code not compile.

MasseGuillaume: Why it cannot be:

```scala
implicit def foo[T]: Foo[T]#Self = null.asInstanceOf[Foo[T]].bar
```

olafurpg: Because the default setting fully qualifies all names with `_root_`. This is totally  unrelated to this PR.

PR #563, opened 2018-01-20
New format for DisableUnlessConfig 

vovapolu:


scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -6,19 +6,34 @@ import org.langmeta.Symbol
...
+object UnlessConfigSymbol {
+  implicit val decoder: ConfDecoder[UnlessConfigSymbol] =
+    ConfDecoder.instanceF[UnlessConfigSymbol] {
+      case c: Conf.Obj =>
+        (c.get[Symbol.Global]("symbol") |@|
+          c.getOption[String]("message")).map {
+          case (a, b) => UnlessConfigSymbol(a, b)
+        }
+      case els =>
+        implicitly[ConfDecoder[Symbol.Global]]
```
gabro: Random thought not related to this PR, but adding an implicit apply on ConfDecoder would save these implicitly calls

website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -14,26 +14,36 @@ within the given "safe" blocks are banned again, to avoid leakage.
 
 ## Configuration
 
-By default, this rule does allows all symbols. To disallow a symbol in a block:
+By default, this rule does allows all symbols. To disallow a symbols in a block:
```
gabro: "a symbols" seems wrong

vovapolu: oops :D

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -6,19 +6,34 @@ import org.langmeta.Symbol
 import scalafix.internal.config.MetaconfigPendingUpstream.XtensionConfScalafix
 import scalafix.internal.util._
 
+case class UnlessConfigSymbol(symbol: Symbol.Global, message: Option[String])
```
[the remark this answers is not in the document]
vovapolu: hehe, nope :) I'll try then.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -3,52 +3,57 @@ package scalafix.internal.config
...
+    CustomMessage.decoder(field = "symbol")
+
+  implicit val reader: ConfDecoder[UnlessConfig] =
     ConfDecoder.instanceF[UnlessConfig] {
       case c: Conf.Obj =>
-        (c.get[Symbol.Global]("block") |@|
-          c.get[Symbol.Global]("symbol") |@|
-          c.getOption[String]("message")).map {
-          case ((a, b), c) => UnlessConfig(a, b, c)
+        (c.get[Symbol.Global]("unless") |@|
```
olafurpg: Pro tip, with `getField` you don't have to specify the type parameter. But doesn't matter much anyways, very soon we can use generic derivation to avoid this boilerplate in the first place.

website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -14,26 +14,36 @@ within the given "safe" blocks are banned again, to avoid leakage.
...
 
-By default, this rule does allows all symbols. To disallow a symbol in a block:
+By default, this rule does allows all symbols. To disallow symbols in a block:
 ```
 DisableUnless.symbols = [
   {
-    block = "scala.Option"
-    symbol = "dangerousFunction"
-    message = "the function may return null"
+      unless = "scala.Option"
```
olafurpg: I have a feeling will eventually want to support a list of symbols here, but we can handle that case once the need arise without breaking existing functionality.

PR #565, opened 2018-01-24
Upgrade to metaconfig v0.6 with generic derivation for decoders.

olafurpg:
Biggest improvements are automatic derivation of ConfDecoder instances.
See https://olafurpg.github.io/metaconfig/ to learn more.

Keeping the documentation on configuration settings and their default
values is a growing problem. Metaconfig v0.6 adds support for both
deriving automatic decoders as well as automatic documentation
generation from configuration settings.

This commit adds a proof-of-concept implementation that it's possible
to automatically generate docs from a rule configuration class. The
implementation is hacky, but it's still a big improvement over the old
way where default values and descriptions easily got off-sync with the
actual code.

screen shot 2018-01-24 at 14 57 20

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/Class2Hocon.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/Class2Hocon.scala
@@ -1,7 +1,19 @@
 package scalafix.internal.config
 
 import metaconfig.HasFields
-import metaconfig.String2AnyMap
+
+object String2AnyMap {
+  def unapply(arg: Any): Option[Map[String, Any]] = arg match {
+    case someMap: Map[_, _] =>
+      try {
+        Some(someMap.asInstanceOf[Map[String, Any]])
+      } catch {
+        case _: ClassCastException =>
+          None
+      }
+    case _ => None
+  }
+}
 
 object Class2Hocon {
 
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/ExplicitResultTypesConfig.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/ExplicitResultTypesConfig.scala
@@ -1,36 +1,37 @@
 package scalafix.internal.config
 
 import metaconfig._
-import MetaconfigPendingUpstream.XtensionConfScalafix
+import metaconfig.generic.Surface
+import metaconfig.annotation._
 
 case class ExplicitResultTypesConfig(
+    @Description("Enable/disable this rule for defs, vals or vars.")
+    @ExampleValue("[Def, Val, Var]")
     memberKind: List[MemberKind] = Nil,
+    @Description("Enable/disable this rule for private/protected members.")
+    @ExampleValue("[Public, Protected]")
     memberVisibility: List[MemberVisibility] = Nil,
+    @Description(
+      "If false, insert explicit result types even for simple definitions like `val x = 2`")
     skipSimpleDefinitions: Boolean = true,
+    @Description(
+      "If false, insert explicit result types even for locally defined implicit vals")
     skipLocalImplicits: Boolean = true,
     // Experimental, still blocked by https://github.com/scalameta/scalameta/issues/1099
     // to work for defs. May insert names that conflicts with existing names in scope.
     // Use at your own risk.
+    @Description(
+      "If true, does a best-effort at inserting short names and add missing imports. " +
+        "WARNING. This feature is currently implemented in a naive way and it contains many bugs.")
     unsafeShortenNames: Boolean = false
-) {
-  implicit val reader: ConfDecoder[ExplicitResultTypesConfig] =
-    ConfDecoder.instanceF[ExplicitResultTypesConfig] { c =>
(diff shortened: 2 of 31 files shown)


website/src/main/tut/docs/rules/ExplicitResultTypes.md
```diff
@@ -30,16 +30,15 @@ def foo = {
...
-ExplicitResultTypes.unsafeShortenNames = true // false by default.
+This rule has several different options.
+
+```tut:invisible
+import scalafix.internal.config._
+```
+```tut:passthrough
+println(
+scalafix.website.rule("ExplicitResultTypes", ExplicitResultTypesConfig.default)
+)
```
gabro: ❤️

olafurpg:
Whoa, scalafix reported error that I didn't type annotate the `implicit val` 😎 

```diff
 import metaconfig.generic
+import metaconfig.generic.Surface
 
 case class NoInferConfig(symbols: List[Symbol.Global] = Nil) {}
 
 object NoInferConfig {
-  implicit val surface = generic.deriveSurface[NoInferConfig]
+  implicit val surface: Surface[NoInferConfig] = generic.deriveSurface[NoInferConfig]
   val default: NoInferConfig = NoInferConfig()
   implicit val decoder: ConfDecoder[NoInferConfig] =
```

website/src/main/tut/docs/rules/ExplicitResultTypes.md
olafurpg: tut:passthrough is awesome!

xeno-by:
This is really neat! 😍

scalafix-core/shared/src/main/scala/scalafix/internal/config/Class2Hocon.scala
```diff
@@ -1,7 +1,19 @@
 package scalafix.internal.config
 
 import metaconfig.HasFields
-import metaconfig.String2AnyMap
+
+object String2AnyMap {
```
vovapolu: :cold_sweat:

scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
```diff
@@ -2,17 +2,14 @@ package scalafix.internal.config
...
 
-import MetaconfigPendingUpstream.XtensionConfScalafix
-
-case class NoInferConfig(symbols: List[Symbol.Global] = Nil) {
-  implicit val reader: ConfDecoder[NoInferConfig] =
-    ConfDecoder.instanceF[NoInferConfig](
-      _.getField(symbols).map(NoInferConfig(_))
-    )
-}
+case class NoInferConfig(symbols: List[Symbol.Global] = Nil) {}
```
vovapolu: Why this case class has empty body?

scalafix-core/shared/src/main/scala/scalafix/internal/config/Class2Hocon.scala
olafurpg: This is a total hack that was previously in metaconfig and I deprecated it 😅

scalafix-core/shared/src/main/scala/scalafix/internal/config/NoInferConfig.scala
olafurpg: no reason, should be removed.

scalafix-core/shared/src/main/scala/scalafix/internal/config/ExplicitResultTypesConfig.scala
```diff
@@ -1,36 +1,37 @@
 package scalafix.internal.config
 
 import metaconfig._
-import MetaconfigPendingUpstream.XtensionConfScalafix
+import metaconfig.generic.Surface
+import metaconfig.annotation._
 
 case class ExplicitResultTypesConfig(
```
MasseGuillaume: What about the other rules` configuration?

olafurpg: I wanted to keep this PR small, prefer to update the other rule configurations in separate PRs.

MasseGuillaume: I think it makes more sense if it's all in one PR. It's confusing to have haft the new way and haft the old way.

olafurpg: Fair enough, I'll migrate more configs. I will leave DisableSyntax alone to avoid conflicts with #548

olafurpg:
I've ported all docs except DisableSyntax to the new format. DisableSyntax can be updated after #548 is merged. Here is how DisableUnless looks like now

screen shot 2018-01-25 at 13 31 31

There is still a bit more polish that can be done but I'd like to make some upstream improvements in metaconfig first.

MasseGuillaume:
@olafurpg thanks for your patience!

PR #567, opened 2018-01-25
DisableSyntax.noDefaultArgs for any method (e.g. any `def`)

vovapolu:


scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -114,17 +98,14 @@ final case class DisableSyntax(
...
-        methods
-          .filter(hasDefaultArgs)
-          .map { m =>
-            errorCategory
-              .copy(id = "defaultArgs")
-              .at(
-                "Default args makes it hard to use methods as functions.",
-                m.pos)
-          }
+      case d: Defn.Def if hasDefaultArgs(d) && config.noDefaultArgs =>
```
olafurpg: How about matching on `Term.Parameter` and point the caret at the parameter name instead of the def name?

scalafix-tests/input/src/main/scala/test/DisableSyntaxMoreRules.scala
```diff
@@ -29,6 +29,10 @@ object DisableSyntaxMoreRules {
     def foobar(a: Int, b: Int = 1) = ??? // assert: DisableSyntax.defaultArgs
   }
 
+  object FooO {
+    def bar(x: Int = 42) = ??? // assert: DisableSyntax.defaultArgs
```
olafurpg: Can we test the caret position and message? See https://github.com/scalacenter/scalafix/blob/50e0771a8939be5a9e8ee8a14cd64c4cbf04c2b7/scalafix-tests/input/src/main/scala/test/Disable.scala#L51 for an example

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
vovapolu: We don't want to match default args in class constructor, for example. And I don't want to call `.parent` to check that the default arg is in the `def` :)

olafurpg:
Fair enough, but can we point the caret at the parameter name instead of
the def?

scalafix-tests/input/src/main/scala/test/DisableSyntaxMoreRules.scala
MasseGuillaume: https://github.com/scalacenter/scalafix/pull/569

MasseGuillaume:
Done in: https://github.com/scalacenter/scalafix/pull/569

PR #422, opened 2017-11-07
Add descriptions to rules

ag91:
Closes #260 

This is my first contribution to scalafix, so huge amounts of feedback are welcome!

PR #568, opened 2018-01-26
Update docs.

olafurpg:
- Port DisableSyntax and MissingFinal to the new doc format.
- Make MissingFinal configurable so that it can be documented like other rules.
- Automatically render rules.md page with all rule names + description
- Deprecate NoFinalize in favor of DisableSyntax.noFinalize. I did this here because I noticed `NoFinalize` was missing a description which made rules.md ugly. 
- Best-effort pretty-printing of default values.

New rules overview

screen shot 2018-01-26 at 12 01 16


olafurpg:
cc/ @vovapolu DisableSyntax page has grown a lot!
screen shot 2018-01-26 at 12 01 29

olafurpg:
@vovapolu I made MissingFinal configurable, 
screen shot 2018-01-26 at 12 05 14

olafurpg:
The new overview page is inspired by https://eslint.org/docs/rules/

olafurpg:
@ag91 FYI, your #422 PR is now being used to render the rules.md page!

MasseGuillaume:
@olafurpg That was productive, we should update docs more often!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala
```diff
@@ -1,23 +1,22 @@
...
 
 import scala.meta._
-
 import scalafix.rule.Rule
 import scalafix.rule.RuleCtx
 import scalafix.lint.LintMessage
-import scalafix.lint.LintCategory
+import scalafix.rule.RuleName
+
+case object NoFinalize
```
MasseGuillaume: @olafurpg We can delete this one, it was never released.

website/src/main/scala/scalafix/website/package.scala
```diff
@@ -1,11 +1,58 @@
...
+          th("Name"),
+          th("Description")
+        )
+      ),
+      tbody(rows)
+    )
+    html.toString()
+  }
+
+  // TODO(olafur) replace this hack with ConfEncoder[T] typeclass.
```
MasseGuillaume: I'm going to add a scalafix Rule to disallow TODO in comments ;-)

MasseGuillaume:
Nevermind `NoFinalize` was added in 0.5.7 https://github.com/scalacenter/scalafix/blob/v0.5.7/scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoFinalize.scala

website/src/main/scala/scalafix/website/package.scala
MasseGuillaume: I guess we can do this with the `DisableSyntax.regex`

vovapolu:
Very cool!

PR #569, opened 2018-01-26
DisableSyntax.noDefaultArgs point carret on the default method

MasseGuillaume:


scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -72,25 +72,26 @@ final case class DisableSyntax(
...
+    object DefaultArgs {
+      def unapply(t: Tree): Option[List[Token]] = {
+        t match {
+          case d: Defn.Def => {
+            val defaults =
+              for {
+                params <- d.paramss
+                param <- params
+                default <- {
+                  param.tokens.collect { case eqs: Token.Equals => eqs }
```
olafurpg: What is the motivation to go via `param.tokens`?

olafurpg: It's possible find a (contrived) counter-example where the caret will not point to the right place here
```scala
@ class Bar { type Foo = Int; def foo = 42 }
defined class Bar

@ def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo
defined function foo

@ foo()
res6: Int = 42
```
We can point to the position of the parameter name or the default value instead.

MasseGuillaume: Hum, good point, or we can point to the default value and do prev to get the `=`. Sorry for merging to fast, I wanted to release and clear the PR queue.

vovapolu: Why don't just point caret to `param`? I don't think it would be hard to understand where the problem is in that case.

MasseGuillaume: @vovapolu When you have a lot of errors/warning you don't want to lose this extra 100ms to get the real position.

```scala
def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo
//      ^
// Default args makes it hard to use methods as functions.

def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo
//                                          ^
// Default args makes it hard to use methods as functions.
```

What is actionable includes the = and everything after it

```scala
def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo
//                                         ^^^^^^^^^^
// Default args makes it hard to use methods as functions.
```

vovapolu: I simply suggested an easy solution to current wrong behavior :) Do you have any thoughts how to fix it?

olafurpg: The position of the default argument value is the best choice IMO. If we start highlighting range positions it will make even more sense compared to highlighting the parameter name.

PR #572, opened 2018-01-26
DisableSyntax.noDefaultArgs add corner case with token position

MasseGuillaume:


scalafix-tests/input/src/main/scala/test/DisableSyntaxMoreRules.scala
```diff
@@ -53,6 +53,12 @@ Default args makes it hard to use methods as functions.
     def foobarfoo(x: Int = 42) = 6 // assert: DisableSyntax.defaultArgs
   }
 
+  class Bar { type Foo = Int; def foo = 42 }
+  def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo /* assert: DisableSyntax.defaultArgs
```
MasseGuillaume:
```scala
def foo(a: { type Foo = Int; def foo: Foo } = new Bar): Int = a.foo
                                              ^ (param.default)
                                            ^ (token)
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -81,7 +81,10 @@ final case class DisableSyntax(
                 params <- d.paramss
                 param <- params
                 default <- {
-                  param.tokens.collect { case eqs: Token.Equals => eqs }
+                  param.tokens.reverse.find {
```
vovapolu:
```scala
def foo(a: { type Foo = Int; def foo: Foo } = {val a = new Bar, a}): Int = a.foo
```
:D

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -81,7 +81,10 @@ final case class DisableSyntax(
...
+                    case eqs: Token.Equals => true
+                    case _ => false
+                  }.toList
```
MasseGuillaume: or `param.default.toList`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
MasseGuillaume: dam

olafurpg: Just use `param.default.pos`

MasseGuillaume: aww guys that extra 2 chars :p

scalafix-tests/input/src/main/scala/test/DisableSyntaxMoreRules.scala
olafurpg: There is no benefit to pointing the caret at `=` over `new`, and with range positions it will be way clearer if `new Bar` is highlighted instead of only `=`

PR #579, opened 2018-01-27
Adds missing  `symbols = [` to the config file example

d1egoaz:


scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableUnlessConfig.scala
```diff
@@ -38,20 +38,22 @@ case class DisableUnlessConfig(
...
+                    |    ]
                     |  }
                     |  {
-                    |    unless = "scala.util.Try"
-                    |    {
-                    |      symbol = "scala.Option.get"
-                    |      message = "the function may throw an exception"
-                    |    }
+                    |     unless = "scala.util.Try"
+                    |     symbols = [
```
d1egoaz: missing `symbols` from the conf example

olafurpg: Good catch! Thanks a lot ❤️

Issue #461, opened 2017-11-25
configureForConfigurations is a bit clunky for large projects


fommil:
It seems to assume that every `Project` has all the configurations.

It would be easier to use if there was a `scalafixSettings` method that could be called `inConfig`, which could be part of the standard boilerplate when adding a test config to a project.

olafurpg:
Did you try ?

```scala
lazy val projectA = project.settings(scalafixConfigure(Test))
lazy val projectB = project.settings(scalafixConfigure(Compile))
```

`scalafixConfigure` returns a list of settings
```
    def scalafixConfigure(configs: Configuration*): Seq[Setting[_]]
```

it works this way instead of through `inConfig()` because I wanted `scalafix` to have the following behavior

- unscoped `scalafix` runs on all enabled  configs
- scoped `scalafix:config` runs only for `config`

I wasn't able to get that working with `inConfig`.

fommil:
In fact... I wonder if it's possible to create a `addCommandAlias` that knows about all non-hidden configurations in a build. That would allow you to easily ship the `fix` command alias as stock.

fommil:
some more details... it's pretty normal in a build to have something like this

```
implicit class IntegrationTestOps(p: Project) {
    def enableIntegrationTests: Project =
      p.configs(It)
        .settings(
          inConfig(It)(
            Defaults.testSettings ++ sensibleTestSettings ++ Seq(
              javaOptions ++= Seq("-Xms1400m", "-Xmx1400m")
            )
          )
        )
  }
```

and the developers can put everything in here that integration tests should enable. There are similar methods for other test configs, and it is in here that the scalafix settings would be applied.

This is all, of course, a workaround to the fact that nobady cares about https://github.com/sbt/sbt/issues/2534, except me.

Issue #559, opened 2018-01-18
Feedback from Sam Halliday


olafurpg:
Before I forget, feedback from @fommil on gitter https://gitter.im/scalacenter/scalafix?at=5a60da3bce68c3bc74d3d282

Some of these are duplicates to existing issues 

- [ ] ability to ignore generated (expanded macros) code, sbt-scalafix
- [ ] ignore imports for Disable and DisableUnless, cc/ @vovapolu 
- [x] better config format for DisableUnless
- [x] canonical way to add scalafix to alternative test configs
- [ ] output in sbt should look exactly like scalac output, including `[error]` gutter
- [ ] support exclude / include format in symbol lists (disable and disableunless). e.g. exclude "scala" (and its children) but then allow "scala.Int"

fommil:
final two are

- https://github.com/scalacenter/scalafix/issues/461
- https://github.com/scalacenter/scalafix/issues/491

MasseGuillaume:
I renamed to Feedback from Sam Halliday to avoid confusion with the Single Abstract Mehod (SAM) rule :P.

fommil:
I broke everything out into separate tickets to track what is and isn't part of a potential final 2 weeks of sponsorship.

PR #587, opened 2018-01-31
Canonical sbt plugin 

vovapolu:
See https://github.com/scalacenter/scalafix/issues/461. 
This PR breaks the compatibility completely. :D I suggest to discuss how to support (or delete) `configureForConfigurations` safely.


scalafix-sbt/src/sbt-test/sbt-scalafix/cross-build/test
```diff
@@ -1,15 +1,33 @@
...
+-> test:scalafixTest ProcedureSyntax
+-> it:scalafixTest ProcedureSyntax
+
+> compile:scalafix ProcedureSyntax
+> test:scalafix ProcedureSyntax
+> it:scalafix ProcedureSyntax
+
+> compile:scalafixTest ProcedureSyntax
+> test:scalafixTest ProcedureSyntax
+> it:scalafixTest ProcedureSyntax
```
MasseGuillaume: This does not test that `compile:scalafix` is independent from `test:scalafix`

What about:
```
-> compile:scalafixTest ProcedureSyntax
-> test:scalafixTest ProcedureSyntax
-> it:scalafixTest ProcedureSyntax

> compile:scalafix ProcedureSyntax
> compile:scalafixTest ProcedureSyntax
-> test:scalafixTest ProcedureSyntax
-> it:scalafixTest ProcedureSyntax

> test:scalafix ProcedureSyntax
> test:scalafixTest ProcedureSyntax
-> it:scalafixTest ProcedureSyntax

> it:scalafix ProcedureSyntax
> it:scalafixTest ProcedureSyntax
```

MasseGuillaume:
Missing: Update https://github.com/scalacenter/scalafix/blob/master/website/src/main/tut/docs/users/installation.md

olafurpg:
I'm in favor of breaking stuff in sbt-scalafix and cut a v0.6. Based on feedback, I think this PR is a good improvement. 

One more thing I would like to change is stop enabling semanticdb-scalac by default. The auto-installation does not work for all builds, so I prefer to require users to manually `addCompilerPlugin(..); scalacOptions += "-Yrangepos"` or use `scalafixEnable` from the shell.

I documented the breaking changes I'd like to make for the next release https://github.com/scalacenter/scalafix/milestone/6

Are there any other breaking changes we want to make?

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -200,78 +168,34 @@ object ScalafixPlugin extends AutoPlugin {
...
-
-  /** Configure scalafix/scalafixTest tasks for given configurations */
-  def configureForConfigurations(
-      configurations: Seq[Configuration],
-      task: InputKey[Unit],
-      impl: Seq[Configuration] => Def.Initialize[InputTask[Unit]]
-  ): Seq[Def.Setting[InputTask[Unit]]] =
-    (task := impl(configurations).evaluated) +:
-      configurations.map(c => task.in(c) := impl(Seq(c)).evaluated)
+  lazy val scalafixConfigSettings: Seq[Def.Setting[_]] = scalafixSettings ++
```
olafurpg: Can we move this to `autoImport` so users don't have to prefix `ScalafixPlugin.`?

website/src/main/tut/docs/users/installation.md
```diff
@@ -12,6 +12,13 @@ Currently, Scalafix does not provide any IDE integrations with IntelliJ/ENSIME.
 
 ## sbt-scalafix
 
+_Since 0.6.0_
+
+_Later this plugin has `scalafixConfigure` method to configure scalafix for custom configuration like IntegrationTest. 
+Since 0.6.0 version it now sticks to more canonical way for such configurations. 
+The plugin has `scalafixConfigSettings` that can be imported to every configuration by `inConfig(_)`.
```
olafurpg: Can we remove all updates to the sbt-scalafix installation docs? I will can do it in a followup PR.

A general note, sentences like

> The plugin has `scalafixConfigSettings` that can be imported to every configuration by `inConfig(_)`. 

are very helpful if you are not aware of what `inConfig(_)` does. I generally prefer concrete examples like below since that is useful for both expert and beginner users.

website/src/main/tut/docs/users/installation.md
```diff
@@ -21,15 +28,16 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "{{ site.stableVersion }}")
...
-> scalafixCli                                            // Run all rules configured in .scalafix.conf
-> scalafixCli --rule RemoveUnusedImports                 // Run only RemoveUnusedImports rule
-> myProject/scalafixCli --rule RemoveUnusedImports       // Run rule in one project only
-> test:scalafixCli --rule RemoveUnusedImports            // Run rule in single configuration
-> scalafixCli --rule ExplicitR<TAB>                      // Use tab completion
-> scalafixCli --rule replace:com.foobar/com.buzbaz       // Refactor (experimental)
-> scalafixCli --rule file:rules/MyRule.scala             // Run local custom rule
-> scalafixCli --rule github:org/repo/v1                  // Run library migration rule
-> scalafixCli --test                                     // Make sure no files needs to be fixed
+> compile:scalafixCli                                            // Run all rules configured in .scalafix.conf
```
olafurpg: Is the `compile:` prefix necessary? I would expect it to default to `compile:`

vovapolu: Yeah, you're right :)

vovapolu:
I added Olaf `format` branch here. It should resolve point about `[error]` gutter from https://github.com/scalacenter/scalafix/issues/559.

MasseGuillaume:
* [ ] Missing: Update website/src/main/tut/docs/users/installation.md
* [ ] What happens with the original task? It's not backward compatible.
* [ ] maybe add a task: `scalafixAll` to do:  `it:scalafix`, `test:scalafix`, `compile:scalafix`

vovapolu:
@MasseGuillaume 
>Missing: Update website/src/main/tut/docs/users/installation.md

I included the update doc in the previous commit, Olaf said 
>Can we remove all updates to the sbt-scalafix installation docs? I will can do it in a followup PR.

> What happens with the original task? It's not backward compatible.

Well, yes, that's right :) Let's discuss what we can do with the old tasks. 

> maybe add a task: scalafixAll to do: it:scalafix, test:scalafix, compile:scalafix

Don't know. It's easy to add, but is it really necessary?

olafurpg:
@MasseGuillaume docs will need to be updated separately. The next release will be breaking so it's fine to remove the old task.

olafurpg:
> maybe add a task: scalafixAll to do: it:scalafix, test:scalafix, compile:scalafix

This can be left for a separate PR.

fommil:
`scalafixAll` is unnecessary in my opinion... you're assuming a lot about the user's project structure. On larger project with complex test configurations it is normal to set these sorts of things up on a per-project basis.

PR #594, opened 2018-02-01
New Rule: SingleAbstractMethod

MasseGuillaume:
* [ ] Documentation

Specs: http://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#sam-conversion
SAM conversion

An expression (p1, ..., pN) => body of function type (T1, ..., TN) => T is 
sam-convertible to the expected type S if the following holds:

* [ ] SAM1: the class C of S declares an abstract method m with signature (p1: A1, ..., pN: AN): R;
* [ ] SAM2: besides m, C must not declare or inherit any other deferred value members;
* [ ] SAM3: the method m must have a single argument list;
* [ ] SAM4: there must be a type U that is a subtype of S, so that the expression 
  new U { final def m(p1: A1, ..., pN: AN): R = body } is well-typed 
  (conforming to the expected type S);
* [ ] SAM5: for the purpose of scoping, m should be considered a static member
  (U's members are not in scope in body);
* [ ] SAM6: (A1, ..., AN) => R is a subtype of (T1, ..., TN) => T
  (satisfying this condition drives type inference of unknown type parameters in S);

Note that a function literal that targets a SAM is not necessarily compiled to
the above instance creation expression. This is platform-dependent.

It follows that:

* [ ] SAM7: if class C defines a constructor, it must be accessible and must define exactly one,
  empty, argument list;
* [ ] SAM8: class C cannot be final or sealed (for simplicity we ignore the possibility of SAM
  conversion in the same compilation unit as the sealed class);
* [ ] SAM9: m cannot be polymorphic;
* [ ] SAM10: it must be possible to derive a fully-defined type U from S by inferring any unknown
  type parameters of C.

Finally, we impose some implementation restrictions (these may be lifted in future releases):

* [ ] SAM11: C must not be nested or local (it must not capture its environment, as that results in a zero-argument constructor)
* [ ] SAM12: C's constructor must not have an implicit argument list (this simplifies type inference);
* [ ] SAM13: C must not declare a self type (this simplifies type inference);
* [ ] SAM14: C must not be `@specialized`.


build.sbt
```diff
@@ -110,7 +110,9 @@ val core = MultiScalaCrossProject(
     buildInfoSettings,
     libraryDependencies ++= List(
       scalameta.value,
-      "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided
+      "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided,
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value,
```
olafurpg: Do we use scala-compiler?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,174 @@
...
+      case tree @ Sam(lambda, _) =>
+        ctx.replaceTree(
+          tree,
+          lambda.show[Syntax]
+        )
+
+    }.asPatch
+  }
+
+  private def collectOnce[T](tree: Tree)(
```
olafurpg: Why is this necessary? collect also uses `.lift`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,174 @@
...
+import scala.meta._
+import scalafix.Patch
+import scalafix.SemanticdbIndex
+import scalafix.rule.RuleCtx
+import scalafix.rule.SemanticRule
+
+/*
+http://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#sam-conversion
+
+SAM conversion
```
olafurpg: Is it necessary to inline this comment? A link is enough IMO, this docstring should explain scalafix specific logic.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,174 @@
...
+ * C must not be @specialized.
+
+ */
+case class SingleAbstractMethod(index: SemanticdbIndex)
+    extends SemanticRule(index, "SingleAbstractMethod") {
+  override def description: String = ???
+  override def fix(ctx: RuleCtx): Patch = {
+
+    // extract the top name position java.util.TimerTask => TimerTask
+    def parentNamePosition(tpe: Type): Option[Position] = {
```
olafurpg: This should already be handled by `index.symbol`. In my scope branch I've added more cases so `index.symbol` just works. Should maybe PR those immediately.

olafurpg: Anything that is a subtype of `Ref` should be automatically handled by `index.symbol`, that includes `Init`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,174 @@
...
+          term,
+          Defn
+            .Def(
+              mods,
+              name,
+              tparams,
+              paramss,
+              Some(decltpe.getOrElse(tpe)),
+              lambda)
+            .show[Syntax]
```
olafurpg: Please use `.syntax` instead of `show[Syntax]`

build.sbt
MasseGuillaume: I forgot to remove it. I wanted to keep a trace of the experimentation in the git history: https://github.com/scalacenter/scalafix/pull/594/commits/3e11aeac950f2246e7a52a04c319218ccbbe2c97

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
MasseGuillaume: I wanted to do a checklist.

scalafix-testkit/src/main/scala/scalafix/testkit/DiffTest.scala
```diff
@@ -25,49 +25,61 @@ object DiffTest {
...
+              isOnly = code.contains("ONLY")
+            ))
         }
+        case Input.VirtualFile(label, _) =>
+          println(s"ignoring: $label")
+          // I don't know why, we get those as VirtualFile:
+          // ignoring: scalafix-tests/shared/src/main/scala/test/SymbolMatcherTest.scala
+          // ignoring: scalafix-tests/shared/src/main/scala/test/DenotationOpsTest.scala
+          // ignoring: scalafix-tests/shared/src/main/scala/test/SemanticdbIndexTest.scala
+          // ignoring: scala/sbtfix/Sbt1.scala
```
MasseGuillaume: @olafurpg Do you know why this has changed? I think this is since I bumped the scalameta version.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
MasseGuillaume: for example:

```scala
import java.lang.Runnable
class SingleAbstractMethod {
  val runnable1 = new Runnable(){ def run(): Unit = println("Run!")}
  new Thread(new Runnable(){ def run(): Unit = println("Hello, Thread!")})
}
```

```scala
ctx.tree.collect {
      case term @ Defn.Val(mods, List(Pat.Var(name)), tpe0, Sam(lambda, tpe)) => ...
      case tree @ Sam(lambda, _) => ... // [B]
}
```

`val runnable1` will be collected twice. first matching the Defn.Val then second matching the Sam.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
MasseGuillaume: Yep it look like it changed since I started this PR. `index.symbol(clazz)` now works

scalafix-testkit/src/main/scala/scalafix/testkit/DiffTest.scala
MasseGuillaume: this was because of the slim mode!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
MasseGuillaume: However, here is a counterexample:

```scala
  val runnable4 = new Runnable(){
    def run(): Unit = {
      val runnable5 = new Runnable(){ 
        def run(): Unit = println("runnable5!")
      }
    }
  }
```

ShaneDelmore: @olafurpg can you point to an example of this technique. I can’t envision it from your description alone but it seems like a technique I should know.

MasseGuillaume:
@olafurpg ready for review.

scalafix-tests/unit/src/test/scala/scalafix/tests/rule/SemanticTests.scala
```diff
@@ -38,11 +38,16 @@ object SemanticTests {
...
     )
   )
-  def classpath: Classpath = Classpath(
-    List(
-      AbsolutePath(BuildInfo.semanticSbtClasspath),
-      AbsolutePath(BuildInfo.semanticClasspath)
-    )
-  )
+  def classpath: Classpath = {
+    val dependencies = BuildInfo.dependencyClasspath
```
olafurpg: Why is this necessary? This would explain why  `test/shared/` started appearing

website/src/main/tut/docs/rules/SingleAbstractMethod.md
```diff
@@ -0,0 +1,12 @@
+---
+layout: docs
+title: SingleAbstractMethod
+---
+
+# SingleAbstractMethod
+
+http://www.scala-lang.org/files/archive/spec/2.12/06-expressions.html#sam-conversion
+
+"-P:semanticdb:members:all",
```
olafurpg: I think we can wait documenting the installation since we can change sbt-scalafix to configure the default flags so SAM works out of the box. `members:all` won't be necessary either once my Scope PR gets merged, same for `denotations:all`. I suspect we'll make overrides enabled by default

scalafix-tests/output/src/main/scala-2.12/test/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,38 @@
...
+trait C extends B
+trait D { def m: String }
+
+class SingleAbstractMethod {
+  val runnable1: Runnable = () => println("runnable1!") 
+  var runnable2: Runnable = () => println("runnable2!") 
+  def runnable3: Runnable = () => println("runnable3!") 
+  val runnable4: Runnable =
+    () => {
+      val runnable5: Runnable =
```
olafurpg: Wow 👏

scalafix-testkit/src/main/scala/scalafix/testkit/DiffTest.scala
```diff
@@ -25,34 +25,39 @@ object DiffTest {
...
-              ScalafixConfig
-                .fromInput(
-                  Input.VirtualFile(label, stripPrefix(comment)),
-                  LazySemanticdbIndex(_ => Some(index)))(decoder)
-                .get
-          }
-          .getOrElse(throw new TestFailedException(
-            s"Missing scalafix configuration inside comment at top of file $relpath",
-            0))
+      if (!label.startsWith("scalafix-tests")) {
```
olafurpg: This seems to be required by the introduction of `dependencies` to the classpath, if we remove that we can probably revert this guard.

scalafix-core/shared/src/main/scala/scalafix/rule/ScalafixRules.scala
```diff
@@ -26,7 +26,8 @@ object ScalafixRules {
     NoAutoTupling(index),
     Disable(index, DisableConfig.default),
     DisableUnless(index, DisableUnlessConfig.default),
-    MissingFinal(index)
+    MissingFinal(index),
+    SingleAbstractMethod(index)
```
olafurpg: What about `PreferSAM` or `SAM`? Not really sure if those are any better though. Suggestions for another name are welcome! cc/ @gabro 😄

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,273 @@
...
+      removeOptional[LeftParen]
+      removeOptional[RightParen]
+
+      remove[LeftBrace]
+      remove[KwDef]
+      remove[Space]
+      remove[Ident] // method name
+
+      def patchParams(hasParams: Boolean): (Token, Boolean) = {
+        find(
```
olafurpg: Please use `find { t => ... }`. I should make scalafmt replace `(` with `{` for large arguments!!

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,273 @@
...
+      }
+      def next: T = {
+        if (!hasNext) throw new NoSuchElementException()
+        val out = slot.getOrElse(iterator.next)
+        slot = None
+        out
+      }
+    }
+
+    class PatchBuilder(tokens: Tokens) {
```
olafurpg: Can we move this out of the SAM rewrite? Somewhere in `scalafix.internal.util`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
```diff
@@ -0,0 +1,273 @@
...
+        case Term.NewAnonymous(
+            Template(
+              _,
+              List(Init(clazz, _, _)),
+              _,
+              List(Defn.Def(_, method, _, paramss, _, _))
+            )
+            )
+            if !visited.contains(tree) && (paramss.size == 1 || paramss.size == 0) =>
+          visited += tree
```
olafurpg: I think this would be more readable with a 

```scala
object SAMCandidate { def unapply(tree: Term.NewAnonymous): Option[(A, B, C)]


var patch = Patch.empty
new Traverser { 
  def apply(tree: Tree) = tree match {
    case SAMCandidate(a, b, c) =>
      patch += samPatch(a, b, c)
      super.apply(c)
    case _ => super.apply(t)
  }
}
patch
```
instead of using `visited: mutable.Set` and `.collect`

olafurpg:
To clarify the example above, the SAM conversion here

```scala
@ val animals: Animals = () => new Dog
```
is unsafe because it looses the refinement in the non-SAM anonymous class that `def animal()`  returns a `Dog`. We should guard against these cases where a `Term.NewAnonymous` is bound to a val/var/def with an inferred type that is a refinement of `Template.init.tpe`

scalafix-core/shared/src/main/scala/scalafix/rule/ScalafixRules.scala
MasseGuillaume: `SingleAbstractMethodConversion` or `SAMConversion`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/SingleAbstractMethod.scala
olafurpg: @ShaneDelmore see https://github.com/scalacenter/scalafix/pull/594#discussion_r166322766 for an example. The trick is to do `super.apply` on the children that you know won't match against again against your pattern

scalafix-core/shared/src/main/scala/scalafix/rule/ScalafixRules.scala
olafurpg: In general, the names are `VerbSomething` like `RemoveUnusedImports`. In scalafix, we use `PreferX` and `AvoidX` for several rewrites like `AvoidInfix`

gabro: I like `PreferSAM` and I bet @fommil likes it too.

fommil: This SAM approved by Sam.

MasseGuillaume: `ConvertSingleAbstractMethod` ?

MasseGuillaume:
@olafurpg I should not have dropped out of FOS, lol!

MasseGuillaume:
@olafurpg

given

```
trait S { def m (p1: A1, ..., pn: AN): R }
type U <: S
val x = new U { def m(x1: T1, ..., xn: TN): T }
```

SAM6 says:

```
FunctionN[A1, ..., AN, R] <: FunctionN[T1, ..., TN, T]
```

this implies

`T <: R` (last parameter is contravariant)

with you Dog/Animal example we have:

```
val y : S' = new U { def m(x1: T1, ..., xn: TN): T' }
```

and `S' <: S` and `S`` is causing problems later on.

We just need to add a constraint to make sure that  `T =:= R`

MasseGuillaume:
And guess what, we have `T =:= R` because we are looking at the overrides. To be an override, you need to have `A1 =:= T1`, ..., `AN =:= TN`, `T =:= R`.

MasseGuillaume:
I added a test to show that your counter example is handled.

MasseGuillaume:
```
 val animals = new Animals { def animal(): Dog = new Dog }
animals: AnyRef with Animals{def animal(): ammonite.$sess.cmd21.Dog} = 
```

this is a new type `AnyRef with Animals{def animal(): ammonite.$sess.cmd21.Dog}` not an override.

MasseGuillaume:
This is also another restriction for SAM rewriting:
https://docs.google.com/document/d/1P_wmH3U356f079AYgSsN53HKixuNdxSEvo8nw_tgLgM/edit#heading=h.q3d7cyx3frv3

olafurpg:
Here is another tricky example
```scala
@ trait Functional { def inc(a: Int): Int }
defined trait Functional

@ class Foo {
    def foo(func: Functional) = func.inc(1)
    def foo(func: Int => Int) = func(2)
  }
defined class Foo

@ new Foo().foo(x => x + 1)
res4: Int = 3

@ new Foo().foo(new Functional { def inc(x: Int) = x + 1 })
res5: Int = 2
```

In this case, SAM conversion for `res5` would produce a compiling program with different semantics. We need to check that the calling function doesn't have an overload for `FunctionN`

MasseGuillaume:
Here is my solution with scala reflect:

```scala
import scala.reflect.runtime.universe._

trait Functional { def inc(a: Int): Int }
class Foo {
  def foo(func: Functional): Int = func.inc(1) // M1
  def foo(func: Int => Int): Int = func(2)     // M2
}

val foos = typeOf[Foo].members.filter(_.name.toString == "foo").map(_.asMethod)
val inc = typeOf[Functional].members.filter(_.name.toString == "inc").map(_.asMethod).head

def getEtaExpandedMethodType(methodSymbol: MethodSymbol): Option[Type] = {
  methodSymbol.info.paramLists match {
    case List(params) => {
      val typesList = params.map(x => x.typeSignature) :+ methodSymbol.returnType
      val arity = methodSymbol.paramLists(0).size
      Some(appliedType(definitions.FunctionClass(arity), typesList))
    }
    case _ => None
  }
}

val hasPolymorphicCandidate = 
  getEtaExpandedMethodType(inc) match {
    case Some(incFun) => { // "scala.Int => scala.Int"
      foos.map(_.info.paramLists).exists {
        case List(List(param)) => param.typeSignature <:< incFun
        case _ => false
      }
    }
    case _ => false
  }
```

fommil:
somewhat related would be a rule to use kind-projector syntax over type constructors

```scala
val convert = new (IList ~> List) {
  def apply[A](fa: IList[A]): List[A] = fa.toList
}
```

to

```scala
val convert = λ[IList ~> List](_.toList)
```

olafurpg:
Here is how I would approach implementing the rewrite @MasseGuillaume https://github.com/scalacenter/scalafix/commit/77d58df1d05f021432df066ff2a5467ba0bf14b8

olafurpg:
Closing until we're unblocked by upstream changes in semanticdb to clear the PR queue.

olafurpg:
To clarify, this rule will be implementable soon thanks to recent work in semanticdb!

@fommil Is that a SAM conversion or something library specific? Feel free to follow up in a separate ticket.

Issue #598, opened 2018-02-04
New DisableConfig format


vovapolu:
Currently `Disable` rule is quite simple and has several issues (see https://github.com/scalacenter/scalafix/issues/559). We can change its behavior so that it would combine the old one with flexible settings and `DisableUnless`. 

Olaf suggested the following format for the updated rule:
```
Disable.symbols = [
  { 
    symbols = [
      {
         symbol = "blah"
         message = "blah blah blah"
      }, 
     "foo",
     "bar"
    ]
    unlessInsideBlock = "IO"   // just new name for unless from DisableUnless
    unlessSynthetic = true     // this option turns on/off the blocking of generated code
  }
]
```

olafurpg:
Fixed in #601

PR #601, opened 2018-02-05
New Disable format

vovapolu:
Initial issue https://github.com/scalacenter/scalafix/issues/598. 

The `DisableUnless` rule was merged into `Disable`. New `unlessSynthetic` option was introduced. Imports are now ignored in the default mode, e.g. without `unlessSynthetic`.


scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,48 +1,139 @@
...
+                    |    unlessSynthetic = true
+                    |    symbols = [
+                    |      {
+                    |        symbol = "scala.Predef.any2stringadd"
+                    |        message = "Use explicit toString be calling +"
+                    |      }
+                    |    ]
+                    |  }
+                    |  { // unlessInsideBlock mode
+                    |    unlessInsideBlock = "com.IO"
```
olafurpg: Can we remove this example? I think it's enough to have one example and `scala.util.Try` seems like the most easy to understand

scalafix-tests/input/src/main/scala/test/DisableSimple.scala
```diff
@@ -49,7 +59,7 @@ case object Disable {
   }
   val yy = AA.asInstanceOf // OK, no errors
   Option(1).get /* assert: Disable.Option.get
-            ^
+  ^
```
olafurpg: Why this change? The caret should point to `get`

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,48 +1,139 @@
...
-        |    symbol = "scala.Any"
-        |    message = "Explicitly type annotate Any if this is intentional"
-        |  }
-        |  # Without custom message (discouraged)
-        |  "com.Lib.implicitConversion"
-        |]""".stripMargin)
-    symbols: List[CustomMessage[Symbol.Global]] = Nil
+    @Description("The list of disable configurations")
+    @ExampleValue("""[
+                    |  { // default mode
```
olafurpg: We should preserve the old syntax for `Disable.symbols`, the new layout is not user-friendly
```
Disable.parts = [
  { symbols = [
    ...
  ]}
]
```

Instead of `parts`, I would prefer
```
Disable.symbols = [...]
Disable.unlessSynthetic = [ ... ]
Disable.unlessInsideBlock = [ ... ]
```

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -5,10 +5,22 @@ title: Disable
 
 # Disable
 
-_Since 0.5.0_
+_Since 0.6.0_
```
olafurpg: Let's keep this unchanged and create a new "v0.6.0" section at the bottom that describes the new configuration options.

The website goes live as soon as this is merged so we should not immediately replace the v0.5 documentation, that would be confusing for users.

website/src/main/tut/docs/rules/DisableUnless.md
```diff
@@ -1,62 +0,0 @@
----
-layout: docs
```
olafurpg: Instead of removing this file we should add a note saying this rule has been merged with `Disable` and include migration instructions. People may have linked to this page so we should not return 404.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
fommil: I'm confused by how `unlessSynthetic` works here. Is it a "disable this even if it is synthetic" setting? Neither `symbols` nor `unless` should operate on synthetic symbols.

scalafix-tests/input/src/main/scala/test/DisableSimple.scala
vovapolu: I use `ctx.index.symbol(tree)` to get symbol during tree traversal. Looking at sources for that method we have ` case Term.Select(_, name @ Name(_)) => symbol(name)` :)

olafurpg: We should report only on the `Name`, if you are in a pattern match the best way to do it is with `case index.Symbol(tree: Name) => ...`

website/src/main/tut/docs/rules/Disable.md
vovapolu: Hmm, how can we keep the old version if we generate it automatically?

olafurpg: The stable version is currently hardcoded for 0.5.10 https://scalacenter.github.io/scalafix/docs/users/installation had to do that before releasing 0.6.0-M1 #610

vovapolu: And how we can describe the new format then? :D

olafurpg: Copy paste the current generated markdown file and inline it. You'll need to generate the site first.

fommil:
Any more comments on this?

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -29,4 +117,3 @@ println(
 scalafix.website.rule("Disable", DisableConfig.default)
```
olafurpg: This will print out the latest version of the config, we only want to copy paste the old settings.

scalafix-tests/input/src/main/scala/test/DisableSimple.scala
```diff
@@ -14,15 +14,17 @@ Disable.symbols = [
          |...
          |// scalafix:on Option.get"""
   }
-  "scala.collection.mutable.ListBuffer"
+  "scala.collection.mutable"
+]
+Disable.unlessSynthetic = [
```
olafurpg: This reads weird, "disable unless synthetic" should mean this symbol is enabled/allowed when it's synthetic but the test indicates it's the opposite: symbol is disabled only if it's synthetic. Should this be `Disable.ifSynthetic` instead?

scalafix-tests/input/src/main/scala/test/DisableUnlessInsideBlock.scala
```diff
@@ -1,11 +1,11 @@
...
 rules = [
-  DisableUnless
+  Disable
 ]
 
-DisableUnless.symbols = [
+Disable.unlessInsideBlock = [
   {
-    unless = "test.DisableUnless.IO"
+    unless = "test.DisableUnlessInsideBlock.IO"
```
olafurpg: Should this field be called "safeBlock" or "block" instead of "unless"?

```
Disable.unlessInsideBlock = [
  { safeBlock = "IO",
    disabledSymbols = ... }
]
```
?

Any thoughts @fommil ? I personally haven't used this rule, but "unless = X, symbols = YY" seems weird.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -21,25 +62,87 @@ final case class Disable(index: SemanticdbIndex, config: DisableConfig)
...
+    errorCategory
+      .copy(id = id)
+      .at(message, pos)
+  }
+
+  private def checkTree(ctx: RuleCtx): Seq[LintMessage] = {
+    def filterBlockedSymbolsInBlock(
+        blockedSymbols: List[Symbol.Global],
+        block: Tree): List[Symbol.Global] = {
+      val Some(symbolBlock: Symbol.Global) = ctx.index.symbol(block)
```
olafurpg: This can crash with MatchError, please refactor to avoid unsafe code.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -21,25 +62,87 @@ final case class Disable(index: SemanticdbIndex, config: DisableConfig)
...
+          Term.Apply(Term.Select(disabledBlock(block), Term.Name("apply")), _),
+          blockedSymbols) =>
+        Right(filterBlockedSymbolsInBlock(blockedSymbols, block)) // <Block>.apply
+      case (Term.Apply(disabledBlock(block), _), blockedSymbols) =>
+        Right(filterBlockedSymbolsInBlock(blockedSymbols, block)) // <Block>(...)
+      case (_: Defn.Def, _) =>
+        Right(config.allSymbols) // reset blocked symbols in def
+      case (_: Term.Function, _) =>
+        Right(config.allSymbols) // reset blocked symbols in (...) => (...)
+      case (t: Name, blockedSymbols) if treeIsBlocked(t, blockedSymbols) =>
```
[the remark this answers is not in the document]
olafurpg: As a general note, the scalafix API typically tries to make it possible to guard against stuff in the pattern instead of the guard.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -1,18 +1,59 @@
...
+import scalafix.lint.{LintCategory, LintMessage}
+import scalafix.rule.{Rule, RuleCtx, SemanticRule}
+import scalafix.util.{SemanticdbIndex, SymbolMatcher}
+
+object Disable {
+
+  /**
+    * Searches for specific nodes in Tree and adds result of searching to the list.
+    * Also passes context through tree traversal.
+    * fn can either return a value. that stops the searching, or change the context.
```
olafurpg:
```
@param fn If fn returns Left then XXX. If fn returns Right then YYY.
```

It's not immediately clear from the signature what `fn` does.

olafurpg: I know this PR is just moving code around but given that `Disable` is covering more use-cases after this PR we should try to document better what SearcherWithContext does.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -1,18 +1,59 @@
...
+    */
+  class SearcherWithContext[T, U](initContext: U)(
```
olafurpg: `ContextTraverser`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -1,18 +1,59 @@
...
-import scalafix.util.SymbolMatcher
+import scalafix.internal.util.SymbolOps
+import scalafix.lint.{LintCategory, LintMessage}
+import scalafix.rule.{Rule, RuleCtx, SemanticRule}
+import scalafix.util.{SemanticdbIndex, SymbolMatcher}
+
+object Disable {
+
+  /**
+    * Searches for specific nodes in Tree and adds result of searching to the list.
```
olafurpg: > A tree traverser to collect values with a custom context.
> 
> At every tree node, either builds a new context `U` or returns a new value `T` to accumulate.
> To collect all accumulated values, use `result(Tree)`.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
olafurpg: s/T/Value/
s/U/Context/

?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
vovapolu: :D I'm not very experienced in writing docs, unfortunately.

scalafix-tests/input/src/main/scala/test/DisableUnlessInsideBlock.scala
fommil: the names of these things are starting to get quite big, but I think a lot of symbols will have a shared `safeBlock` so it makes sense for them to be grouped like this. Each symbol may have a different explanation, though, so as long as that is supported we're good.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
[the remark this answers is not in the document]
vovapolu: @olafurpg `blockedSymbols` isn't static, it changes during tree traversal.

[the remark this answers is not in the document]
olafurpg: You can  construct a new SymbolMatcher whenever it changes https://github.com/scalacenter/scalafix/blob/master/scalafix-core/shared/src/main/scala/scalafix/util/SymbolMatcher.scala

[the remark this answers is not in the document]
vovapolu: And what then? :) I receive a tuple `(Tree, List[Symbol.Global])` and want to test if `tree` is blocked by these symbols.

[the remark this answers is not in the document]
olafurpg:
```scala
val isBlocked = SymbolMatcher.normalized(badSymbols:_*)
tree match {
  case isBlocked(n: Name) =>
}
```
?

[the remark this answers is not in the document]
vovapolu: :thinking: But why is it better than guard? I can place these code in some function :)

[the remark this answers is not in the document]
olafurpg: Unless I'm seeing something wrong, it looks like `treeIsBlocked` is reimplementing the SymbolMatcher logic. You could replace it with SymbolMatcher and it would be clearer what is happening.

> It's even worse, I want to ignore cases when tree isn't blocked.

Why can't you ignore cases when the tree isn't blocked with SymbolMatcher?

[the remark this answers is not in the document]
olafurpg: `val Some(...) = ...` is a sign that you were not able to prove to the compiler this deconstruction can't fail. By using `SymbolMatcher.unapply` you don't need to outsmart the typechecker.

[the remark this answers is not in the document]
vovapolu: Yeah, I agree with `val Some(...) = ...`. 

>Why can't you ignore cases when the tree isn't blocked with SymbolMatcher?
 
Well, in `SearcherWithContext` I need to return new context or `LintError`, ignoring means no changing in the context. I could just return an old context... I didn't want to do this :)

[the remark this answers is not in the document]
olafurpg: I'm not sure I follow why SymbolMatcher is not suitable here but I trust your judgment 😉 we can keep it unchanged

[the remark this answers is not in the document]
vovapolu: I can change it to what you want, maybe it will be more understandable.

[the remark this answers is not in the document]
vovapolu: Hehe, IDEA can't parse `case (ctx.index.Symbol(t: Name), ...`

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -114,6 +75,31 @@ import scalafix.internal.config._
...
 ```
+
+The the example configuration above, Scalafix will report the following warnings:
+```scala
+import scala.util.Try
+
+object Test {
+  val z = 1.asInstanceOf[String] // default mode, not ok
+
+  List(1) + "any2stringadd" // unlessSynthetic mode, not ok
```
olafurpg: s/unlessSynthetic/ifSynthetic/

olafurpg: Non-executable docs are difficult to maintain, would be nice if we could somehow reuse testkit for docs 🤔

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,48 +1,111 @@
...
+                    |    symbols = [
+                    |      {
+                    |        symbol = "scala.Option.get"
+                    |        message = "the function may throw an exception"
+                    |      }
+                    |    ]
+                    |  }
+                    |]
+      """.stripMargin)
+    unlessInsideBlock: List[UnlessInsideBlock] = Nil
```
olafurpg: What do you think about renaming this to "unlessInsideSafeBlock"? I think it makes sense to be consistent betweeen `.unlessInsideBlock` and `safeBlock =`. Standalone "Block" is too generic so I think "safeBlock" may be clearer, but I'll leave it to you and @fommil to decide what makes most sense.

fommil: the concept of "safe"ness is perhaps limited to my usecase. I can imagine people wanting to use this for other purposes, such as logging or monitoring, in which case "safe" will be a misnomer.

olafurpg: Fair enough. What should the "safe symbol" be called then?

`Disable.unlessInside = [ ??? ]`

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,48 +1,111 @@
...
-        |  }
-        |  {
-        |    symbol = "scala.Any"
-        |    message = "Explicitly type annotate Any if this is intentional"
-        |  }
-        |  # Without custom message (discouraged)
-        |  "com.Lib.implicitConversion"
-        |]""".stripMargin)
-    symbols: List[CustomMessage[Symbol.Global]] = Nil
+    @Description("The list of symbols to disable only in the actual sources.")
```
olafurpg: > List of symbols to disable if written explicitly in source. Does not report an error for inferred symbols in macro expanded code or implicits.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -1,48 +1,111 @@
...
+    @ExampleValue("""
+                    |[
+                    |  {
+                    |    symbol = "scala.Any.asInstanceOf"
+                    |    message = "use pattern-matching instead"
+                    |  }
+                    |]""".stripMargin)
+    symbols: List[CustomMessage[Symbol.Global]] = Nil,
+    @Description(
+      "The list of symbols to disable, also blocks symbols in the generated code")
```
olafurpg: List of symbols to disable if inferred. Does not report an error for symbols written explicitly in source code.

vovapolu:
>Is there any nuke setting to disable a symbol regardless if it's synthetic or not?

Add it to the `ifSynthetic` and to `symbols`?

>Also, should we deprecate NoInfer in favor of Disable.ifSynthetic?

Why not? :)

fommil:
can we please do any more refinements in a follow up? I'm really keen to try this out in the wild once you cut a release and I fear we will run out of Vladimir's sponsored time by the time I discover anything.

olafurpg:
> Add it to the ifSynthetic and to symbols?

That seems a bit tedious for a seemingly common use-case. Would be good to have an option to enable both synthetic and non-synthetic checks and `Disable.symbols` seems to be the best choice for that.

> can we please do any more refinements in a follow up? 

Sure, we can do that esp. since we are kind of in between major releases. Keep in mind that the current master is a bit in a limbo in a transition to the new `Type` and method symbol syntax in semanticdb (https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#type). The current semanticdb-scalac default options produce both old and new types, which incurs some unnecessary overhead.  We released Scalameta v3.4.0 today (https://github.com/scalameta/scalameta/releases/tag/v3.4.0) which flips the defaults to much saner default which will give be a nice speedup since we soon no longer need to persist signatures for all referenced symbols, only definitions in each compilation unit. But it will take ~2 more weeks to complete the refactoring, which includes updating ExplicitResultTypes and some other rewrites.

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -21,12 +21,85 @@ MyCode.scala:7: error: [DisallowSymbol.asInstanceOf] asInstanceOf is disabled.
...
+]
+```
+
+_Since 0.6.0_
+
+This rule reports errors when a "disallowed" symbol is referenced.
+
+It has several different modes:
+- Default mode. In this mode symbols are banned in the actual source code.
+- `unlessInsideBlock`
```
olafurpg: unlessInside

website/src/main/tut/docs/rules/Disable.md
```diff
@@ -21,12 +21,85 @@ MyCode.scala:7: error: [DisallowSymbol.asInstanceOf] asInstanceOf is disabled.
...
+The the example configuration above, Scalafix will report the following warnings:
+```scala
+import scala.util.Try
+
+object Test {
+  val z = 1.asInstanceOf[String] // default mode, not ok
+
+  List(1) + "any2stringadd" // ifSynthetic mode, not ok
+
+  Option.empty[Int].get //  unlessInsideBlock mode, not ok
```
olafurpg: unlessInside

olafurpg:
@fommil I recommend adding scalacOptions `-P:semanticdb:signatures:old` and `-P:semanticdb:denotations:all` when you give it a spin. This will no longer be necessary once v0.6.0 final is out.

@vovapolu I can take care of deprecating NoInfer 😄

olafurpg:
I opened https://github.com/scalacenter/scalafix/issues/641 to track followup changes on this PR @vovapolu

PR #616, opened 2018-02-18
Fix ProcedureSyntax rule for type params

eed3si9n:
### steps

run ProcedureSyntax on the following:

```scala
  def baz[A] { println("baz[A]") }
```

### problem

ProcedureSyntax generates uncompilable code.

```scala
  def baz: Unit =[A] { println("baz[A]") }
```

### expectation

```scala
  def baz[A]: Unit = { println("baz[A]") }
```

### notes

Previously the rule was looking for the right paren. That's not enough if the procedure has  a type parameter.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
@@ -4,6 +4,7 @@ import scala.collection.immutable.Seq
 import scala.meta._
 import scala.meta.tokens.Token.LeftBrace
 import scala.meta.tokens.Token.RightParen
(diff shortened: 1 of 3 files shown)


eed3si9n:
I caught this one when I ran `coursier launch ch.epfl.scala:scalafix-cli_2.12.3:0.5.3 -- -r ProcedureSyntax test` on scala/scala. Lots of corner cases in there, so you might find other things.

smarter:
does the rewrite handle abstract methods such as in `trait Foo { def bar(x: Int) }` ?

eed3si9n:
@smarter It will after https://github.com/scalacenter/scalafix/pull/618, but I am not sure if there's a release out yet that includes this or that.

PR #618, opened 2018-02-18
Modernize ProcedureSyntax.

### LINKED ISSUE #617: ExplicitUnit should be merged into ProcedureSyntax
### steps

Run ProcedureSyntax

```scala
  abstract class Structural extends { def f() }
```

### problem

No changes.

### expectation

```scala
  abstract class Structural extends { def f(): Unit }
```

olafurpg:
- Use TokenList.leading to simplify ProcedureSyntax logic.
  ProcedureSyntax was the first scalafix rewrite rule and back then
  TokenList did not exist.
- Merge ExplicitUnit into ProcedureSyntax.  ExplicitUnit was rewriting
  procedure syntax for declarations (abstract methods), there was no good
  reason to use separate rules. Fixes #617.

I ran ProcedureSyntax on scala/scala and it seems to work correctly,
full diff is here
https://gist.github.com/olafurpg/94d77311532cc538cc1f1241df8f5901

Review @eed3si9n


scalafix-core/shared/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -1,40 +1,37 @@
...
-                .find(_.is[RightParen])
-                .getOrElse(defEnd)
-            )
-        ctx.addRight(closingBracketOrParen, s": Unit =").atomic
-    }
-    patches.asPatch
+        val fixed = for {
+          bodyStart <- t.body.tokens.headOption
+          toAdd <- ctx.tokenList.leading(bodyStart).find(!_.is[Trivia])
+        } yield ctx.addRight(toAdd, s": Unit =").atomic
```
eed3si9n: I am not familiar with these tokens, so this change looks magic to me, but since the diff on test data looks fine so LGTM.

olafurpg: The trick is `tokenList.leading` which returns a view of the tokens in reverse order starting from the `{` of the method body. `Trivia` is a category for whitespace and comment tokens.

Issue #442, opened 2017-11-17
Respect @SuppressWarnings to disable rules


olafurpg:
We recently added support to disable rules over regions of code using comments in #415. Comments were chosen for several reasons,

- they're used the same way in many other linters, including scalastyle, eslint and pylint
- comments allow you to target code more precisely than annotations, since some things cannot be annotated (like a class name, you have to disable the rule for the entire class).
- comments can target code that cannot be annotated, like a group of 10 methods out of 20 methods in a class, here a real world example from the scalafix codebase to disable scalafmt https://github.com/scalacenter/scalafix/blob/d4ccaec624005ce51bc96ee76ea66a871bb4578c/scalafix-core/shared/src/main/scala/scalafix/internal/config/ScalafixReporter.scala#L44-L50)
- comments can be robustly used by syntactic linters (which includes scalafix, it's both syntactic and semantic), while annotations require compilation to be 100% confident the annotation resolved to your expected annotation. A syntactic linter is free to assume that `@SuppressWarnings` means `@java.lang.SuppressWarnings`, but it requires compilation to be 100% guarantee it's true.

However, java.lang.SupppressWarnings is a standard way to suppress messages from the compiler and it's used by linters like WartRemover. I think scalafix should respect `@SuppressWarnings` for those users who still prefer to use annotations.

marcelocenerine:
@olafurpg, what behavior do you expect when comments and annotations are mixed and they overlap?

I took a look at what Checkstyle does, given that it has filters supporting both approaches which can be combined. From what I could see, suppression via annotation takes precedence over comments. Overlapping on/off comments within annotated blocks are still honored but only take effect outside of the enclosing scope. Examples:

- Rule is **disabled** for all lines in `foo1` (annotation) and **enabled** for all lines in `foo2` (comment within `foo1`):
```java
//CHECKSTYLE:OFF
@SuppressWarnings("checkstyle:needBraces")
public void foo1(int n) {
    if (n % 3 == 0) System.out.println("foo");
    //CHECKSTYLE:ON
    if (n % 5 == 0) System.out.println("bar");
}

public void foo2(int n) {
    if (n % 3 == 0) System.out.println("foo");
    if (n % 5 == 0) System.out.println("bar");
}
```

- Rule is **disabled** for all lines in `foo1` (annotation) as well as for `foo2` (comment within `foo1`):
```java
@SuppressWarnings("checkstyle:needBraces")
public void foo1(int n) {
[...]

PR #620, opened 2018-02-21
Respect @SuppressWarnings to disable rules

marcelocenerine:
Reviewers: @olafurpg & @MasseGuillaume 
&nbsp;

This PR adds support to the `@SuppressWarnings` annotation as a hierarchical mechanism to disable Scalafix rules (#442).

This new mechanism is supported in all locations where Scala allows the annotation to be placed:
  - class
  - trait
  - object
  - type
  - methods
  - constructors (primary and secondary)
  - method/constructor parameters
  - val and vars

One or more rules can be specified in each annotation (more than one annotation can be applied to a given definition):

```scala
@SuppressWarnings(Array("rule1"))
@SuppressWarnings(Array("rule2"))
def foo(): Unit = ???
```

or

```scala
@SuppressWarnings(Array("rule1", "rule2"))
def foo(): Unit = ???
```

Rule names can be optionally prefixed with `scalafix:`. If they are, then Scalafix is able to report warnings for unused/redundant rules.

The standard wildcard value `all` can be used to disable all rules and is equivalent to `// scalafix: off`.

`@SuppressWarnings` can be combined with comment anchors on the same source file. If there is any overlapping, the annotation will take precedence. In that case, the overlapping anchor is **only ignored within the scope of the annotated definition**.  (see discussion on #442 about this. Also, see example scenarios in the `AnchorAnnotationMixed` test file).

**NOTE:** The documentation has not been updated yet. I will wait until reviewers are happy with the code changes then I'll do it.

Implementation details:
----------------------

### `SuppressWarnings` vs. `@java.lang.SuppressWarnings`

As @olafurpg  mentioned on #442, annotations require compilation in order to be 100% confident about their types (`@java.lang.SuppressWarnings` in this case). Given that Scalafix supports both syntactic and semantic rules and the type information is only available for the latter, we need to assume that `SuppressWarnings` (when in a `Mod.Annot`) resolves to `@java.lang.SuppressWarnings`.

### Refactoring

In order to accomodate the implementation of this feature, I had to move some things around and do some cleanup. This may make `EscapeHatch` a little hard to review and it may be a good idea take a look at each commit individually, or else I can try to split it into multiple PRs. The main changes were:

- moved all anchor-related code into a new inner class `AnchoredEscapes`. The anchor parsing logic was consolidated in the `AnchoredEscapes` companion object.
- eliminated one step in the anchor parsing logic. Currently it transforms tree/tokens -> `Escape` -> `EscapeFilter`. The intermediary step didn't seem to add much benefit, so I merged it with the last one. This helped to eliminate a plenty of lines of code without compromising readability (my opinion). I may reconsider this if you don't agree. 
- all code related to `SuppressWarnings` handling was implemented in a new inner class `AnnotatedEscapes` and its corresponding companion object.
- each `EscapeFilter` now corresponds to only a single rule rather than all the rules in one comment/annotation. The reason for that is explained in the improvement below.
- Now `EscapeHatch` only cares about whether a rule is enabled or not. In order to decide that, it delegates to the 2 classes above according to the order of precedence.
 

### Improvement: Warnings to point to the exact position of unused rules

Scalafix shows warnings for enable/disable rules that are unused. However, the warning messages don't mention the rule name nor the caret ("^") points to location of the unused rule. Instead, they point to the beginning of the anchor comment. When there are multiple unused rules in one anchor, the exact same warning is printed multiple times which does not help much. Please see example here: https://gist.github.com/marcelocenerine/619a52ef03f3872fbcd7e29dc8d921f3

I thought it would be more helpful to know the exact rule that is causing the warning so that users can go and remove it rather than try to manually figure out which one is causing the warning. This is probably even more relevant for the `@SuppressWarnings` case. See example: https://gist.github.com/marcelocenerine/8ab0358781826078d10a0acee786b4b7

Not sure if you agree but I made this improvement (I can take it out if you don't like). Now the caret points to the exact position of the unused rules when there is any, otherwise to the beginning of the comment. I also changed the warning messages to be more meaningful. See example: https://gist.github.com/marcelocenerine/b22bb526d012b783a8ad40aa41d54bba

Besides making the warning messages more helpful, this improvement also helped to uncovered 2 potential bugs.

### Bugs

* comma (`,`) not a separator for rule names in comments (https://scalacenter.github.io/scalafix/docs/users/configuration#selectively-disabling-rules):

```scala
{
  val x = null
  val aDummy = 0
  a.get
} // scalafix:ok EscapeHatchNoNulls, EscapeHatchDummyLinter, Disable.get
```

Scalafix does not ignore the first 2 rules:

```
scala/test/escapeHatch/Baz.scala:25:13: error: [EscapeHatchNoNulls]:
Nulls are not allowed.
    val x = null
            ^

---------------------------------------

scala/test/escapeHatch/Baz.scala:26:9: error: [EscapeHatchDummyLinter]:
Dummy!
    val aDummy = 0
        ^
```

* `NoInfer.any2stringadd`:
The code below does not trigger the rule:

```scala
1 + "foo"
```

whereas this does:

```scala
Some(1) + "foo"
```

There was a test relying on this, which should have failed but was passing because of the 1st bug. I did not investigate this one further, so if you think there is really something wrong then let's file a bug.


scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorExpression.scala
```diff
@@ -37,7 +37,7 @@ object EscapeHatchExpression {
 
   (
     null,
-    1 + "foo",
+    Some(1) + "foo",
     a.get
   ) // scalafix:ok NoInfer.any2stringadd, Disable.get
```
marcelocenerine: this is the test where `NoInfer.any2stringadd` does not get triggered but still passes due to the bug in the regex

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorMultipleRules.scala
```diff
@@ -19,10 +19,10 @@ NoInfer.symbols = [
...
 package test.escapeHatch
 
-object EscapeHatchMultipleRules {
+object AnchorMultipleRules {
 
   // scalafix:off NoInfer.any2stringadd, Disable.get
-  1 + "foo"
+  Some(1) + "foo"
 
   val aDummy = 0 // assert: EscapeHatchDummyLinter
```
marcelocenerine: same here

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,396 @@ class EscapeHatch(
...
-    val disableRules = TreeMap.newBuilder[EscapeOffset, EscapeFilter]
-
-    def enable(
-        offset: EscapeOffset,
-        anchor: AnchorPosition,
-        rules: String): Unit = {
-      enableRules += (offset -> EscapeFilter(matcher(rules), anchor, offset))
-    }
+      def splitRules(rules: String): List[String] =
+        rules.split(",\\s*").map(_.trim).toList
```
marcelocenerine: Because of the `trim`, this regex can be simplified to `, ` ....

olafurpg:
Thank you for this contribution @marcelocenerine nice work! 😄 

I'll let @MasseGuillaume take a first pass since he developed EscapeHatch where the biggest diff seems to be

MasseGuillaume:
Really good PR, I'm really impressed by the quality of this work.

I found another bug on `scalafix:ok` while reviewing this:

```scala
object A {
  val a = null
} // scalafix:ok DisableSyntax.keywords.null
```

```
[fast2] Running DisableSyntax
/home/gui/fast2/src/main/scala/a.scala:4:11: error: [DisableSyntax.keywords.null] null is disabled
  val a = null
          ^
/home/gui/fast2/src/main/scala/a.scala:5:3: warn: [UnusedScalafixSupression.Disable] This comment does not disable any rule
} // scalafix:ok DisableSyntax.keywords.null
  ^
```

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorAnnotationMixed.scala
```diff
@@ -0,0 +1,87 @@
...
+  val dDummy = 0 // assert: EscapeHatchDummyLinter
+
+
+
+  // Overlap - anchors enclosed by annotated method
+  @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+  def aDummy4(x: Option[Any]): Unit = {
+    /* scalafix:off EscapeHatchDummyLinter */ // assert: UnusedScalafixSuppression.Disable
+    val bDummy = 0
+    // scalafix:on EscapeHatchDummyLinter
```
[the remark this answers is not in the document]
marcelocenerine: this is because the annotation takes precedence when there is overlapping. See my comment https://github.com/scalacenter/scalafix/issues/442#issuecomment-366519138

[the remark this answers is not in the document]
marcelocenerine: (there is a counterexample in the comment I mentioned above)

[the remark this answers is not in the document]
MasseGuillaume: Yup, I got that. I think we can handle the case where the anchor range position is completely inside the annotation range position.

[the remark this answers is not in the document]
MasseGuillaume: Also:

```scala
  // scalafix:off EscapeHatchDummyLinter
  def aDummy = {}
  // scalafix:on EscapeHatchDummyLinter

  @SuppressWarnings(Array("EscapeHatchDummyLinter"))
  def bDummy = {}

  @SuppressWarnings(Array("EscapeHatchDummyLinter"))
  def cDummy = {}
```

Could have the same effect as:

```scala
  // scalafix:off EscapeHatchDummyLinter
  def aDummy = {}
  // scalafix:on EscapeHatchDummyLinter

  // scalafix:off EscapeHatchDummyLinter
  def bDummy = {}
  // scalafix:on EscapeHatchDummyLinter

  // scalafix:off EscapeHatchDummyLinter
  def cDummy = {}
  // scalafix:on EscapeHatchDummyLinter
```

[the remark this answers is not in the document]
marcelocenerine: ok, let me put some thoughts on this. @olafurpg, would you like to chip in?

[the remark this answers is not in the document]
olafurpg: I agree with @marcelocenerine 's assessment in https://github.com/scalacenter/scalafix/issues/442#issuecomment-363778884 that it is counter-intuitive if annotations are treated as comments. Annotations should have push/pop semantics and they should override comments.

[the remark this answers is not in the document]
marcelocenerine: so should I keep the current behavior then? I updated the docs last night.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorExpression.scala
olafurpg: Aaah, lol. Nice catch!

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationMultipleRules.scala
```diff
@@ -0,0 +1,45 @@
...
+  "class:scalafix.test.EscapeHatchNoNulls"
+]
+
+Disable.symbols = [
+  "scala.Option.get"
+]
+*/
+package test.escapeHatch
+
+object AnnotationMultipleRules {
```
olafurpg: Very nice 👍

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
```diff
@@ -0,0 +1,273 @@
...
+      val jDummy = 0
+      var hDummy = 0
+    }
+  }
+
+  // Primary constructor
+  class DummyClass_2 @SuppressWarnings(Array("EscapeHatchDummyLinter"))( // assert: EscapeHatchDummyLinter
+                 val aDummy: Int, val bDummy: Int) {
+
+    val cDummy = 0 // assert: EscapeHatchDummyLinter
```
olafurpg: I think it's sufficient to have only a single `val cDummy = 0` in this template to test the scoping of the annotations. The other expressions are redundant.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
```diff
@@ -0,0 +1,273 @@
...
+      var hDummy = 0 // assert: EscapeHatchDummyLinter
+    }
+  }
+
+  // Primary constructor parameter
+  class DummyClass_3( // assert: EscapeHatchDummyLinter
+                 @SuppressWarnings(Array("EscapeHatchDummyLinter")) val aDummy: Int,
+                 val bDummy: Int) { // assert: EscapeHatchDummyLinter
+
+    val cDummy = 0 // assert: EscapeHatchDummyLinter
```
olafurpg: Ditto. I think it's sufficient to have only a single `val cDummy = 0` in this template to test the scoping of the annotations. The other expressions are redundant.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
```diff
@@ -0,0 +1,273 @@
...
+  class DummyClass_4( // assert: EscapeHatchDummyLinter
+                 val aDummy: Int, // assert: EscapeHatchDummyLinter
+                 val bDummy: Int) { // assert: EscapeHatchDummyLinter
+
+    @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+    val cDummy = 0
+    @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+    var dDummy = 0
+
+    def this(eDummy: String, // assert: EscapeHatchDummyLinter
```
olafurpg: Ditto.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
```diff
@@ -0,0 +1,273 @@
...
+                iDummy: Int): Unit = { // assert: EscapeHatchDummyLinter
+      @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+      val jDummy = 0
+      @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+      var hDummy = 0
+    }
+  }
+
+  // nested scopes
+  def outter = {
```
olafurpg: Nice  👍

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
```diff
@@ -0,0 +1,273 @@
...
+    def gDummy( // assert: EscapeHatchDummyLinter
+                @SuppressWarnings(Array("EscapeHatchDummyLinter")) hDummy: Int,
+                iDummy: Int): Unit = { // assert: EscapeHatchDummyLinter
+      val jDummy = 0 // assert: EscapeHatchDummyLinter
+      var hDummy = 0 // assert: EscapeHatchDummyLinter
+    }
+  }
+
+  // Local variable
+  class DummyClass_9( // assert: EscapeHatchDummyLinter
```
olafurpg: Ditto. You can assume the position of the annotation enclosing jDummy below is correct, otherwise it would be a scalameta bug.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationUnused.scala
```diff
@@ -0,0 +1,30 @@
...
+  def a = ()
+
+  @SuppressWarnings(Array("scalafix:EscapeHatchDummyLinter", "EscapeHatchNoNulls")) // assert: UnusedScalafixSuppression.Disable
+  def b = ()
+
+  @SuppressWarnings(Array("scalafix:EscapeHatchDummyLinter")) // assert: UnusedScalafixSuppression.Disable
+  @SuppressWarnings(Array("scalafix:EscapeHatchNoNulls")) // assert: UnusedScalafixSuppression.Disable
+  val c = 0
+
+  @SuppressWarnings(Array("EscapeHatchDummyLinter", "EscapeHatchNoNulls")) // OK, not prefixed
```
olafurpg: Great!

website/src/main/tut/docs/users/configuration.md
```diff
@@ -157,4 +196,10 @@ foo(null)
...
 ```scala
 List(1, "") // scalafix:ok NoInfer.any
-```
\ No newline at end of file
+```
+
+&nbsp;
+
+**Note:** Suppression via comments and `@SuppressWarnings` can be combined in the same source file. Be mindful not to 
+introduce overlaps between the two as it can cause confusion. Scalafix gracefully handles overlaps by given
```
olafurpg: s/gracefully//

Should we add an example here? Might also want to add "it can cause confusion and counter-intuitive behavior".

website/src/main/tut/docs/users/configuration.md
```diff
@@ -121,32 +120,72 @@ debug.printSymbols = true
 
 To build custom rules see {% doc_ref Rule Authors %}.
 
-## Per source file
 
-It's possible to supress false positives linter message over source file regions using comments. This provides fine grained control over what to report
+## Suppressing rules
```
olafurpg: Great writing! You have a knack for docs

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,396 @@ class EscapeHatch(
...
-        case Toogle.Enable =>
-          val enabledRules = splitRules(rules)
-          val enabledNotDisabled = enabledRules -- currentlyDisabledRules
-
-          enabledNotDisabled.foreach(
-            rule => unusedEnable += unusedEnableWarning.at(anchorPosition.value)
-          )
+      def hasSuppressWarnings(mods: List[Mod]): Boolean =
+        mods.exists {
+          case Mod.Annot(Init(Type.Name(SuppressWarnings), _, _)) => true
```
olafurpg: What happens for `@java.lang.SuppressWarnings`?  I agree it makes sense to do this syntactically in order to enable syntactic rules like DisableSyntax to support annotations. Can we add an entry in the docs explaining the  detection of `@SuppressWarnings`?

> The `@SuppressWarnings` annotation is detected without compilation. Any annotation matching the syntax `@SuppressWarnings(..)` regardless if it's `java.lang.SupressWarnings` or not will trigger the suppression mechanism. This is done in order to support `@SuppressWarnings` for syntactic rules like `DisableSyntax`.


To support qualifying the annotation, I recommend we do the following:

```scala
case Mod.Annot(Init(Type.Name(SuppressWarnings), _, List(List(_)))) => true
case Mod.Annot(Init(Type.Select(_, Type.Name(SuppressWarnings)), _, List(List(_)))) => true // `@java.lang.SuppressWarnings(..)`
```

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorExpression.scala
marcelocenerine: Yes indeed. But the bug in the regex was preventing it from happening

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorAnnotationMixed.scala
```diff
@@ -0,0 +1,87 @@
+/*
+rules = [
+  "class:scalafix.test.EscapeHatchDummyLinter"
```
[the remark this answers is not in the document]
marcelocenerine: done (the shorter option)

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: >What happens for @java.lang.SuppressWarnings?

Nice catch! Fixed

>Can we add an entry in the docs explaining the detection of @SuppressWarnings?

Done

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorAnnotationMixed.scala
```diff
@@ -0,0 +1,87 @@
...
+
+
+  // Overlap - anchor ON before and OFF within annotated method
+  // scalafix:off EscapeHatchDummyLinter
+  def aDummy5(): Unit = ???
+
+  @SuppressWarnings(Array("EscapeHatchDummyLinter"))
+  def aDummy6(x: Option[Any]): Unit = {
+    val bDummy = 0
+    // scalafix:on EscapeHatchDummyLinter
```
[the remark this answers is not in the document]
marcelocenerine: 😅

website/src/main/tut/docs/users/configuration.md
marcelocenerine: I just copied most of this stuff from your comment in #442 🤣

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/escapeHatch/AnnotationScopes.scala
marcelocenerine: done

website/src/main/tut/docs/users/configuration.md
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorAnnotationMixed.scala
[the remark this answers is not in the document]
marcelocenerine: That is definitely doable. The only thing is that it would require the comment type (on+off/ok) to be preserved. Currently that information gets lost very early as `// scalafix:ok` is transformed into a pair of off/on `EscapeFilter`s. If you think this is a nice to have, then I can invest some time on it.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
-
-        case Toogle.Enable =>
-          val enabledRules = splitRules(rules)
-          val enabledNotDisabled = enabledRules -- currentlyDisabledRules
-
-          enabledNotDisabled.foreach(
-            rule => unusedEnable += unusedEnableWarning.at(anchorPosition.value)
-          )
+      tree.foreach {
+        case t @ Defn.Class(mods, _, _, _, _) if hasSuppressWarnings(mods) =>
```
olafurpg: I've been meaning for a long time to add `mods: List[Mod]` on `Member` to avoid boilerplate like this, even more motivated now!

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        case Mod.Annot(
+            Init(Type.Select(_, Type.Name(SuppressWarnings)), _, _)) =>
+          true
+        case _ => false
+      }
 
-        case _ => ()
+    private def extractRules(mods: List[Mod]): List[(String, Position)] = {
+      def process(rules: List[Term]) = rules.map {
+        case lit @ Lit.String(rule) =>
```
olafurpg: What happens if this is not a string literal?

```
val x = "foo"
@SuppressWarnings(Array(x))
```

I think it's fair to require string literals, but we should at least not crash if there is a variable.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
olafurpg: Actually, could we add a `Mods` extractor in `TreeExtractors`? https://github.com/scalacenter/scalafix/blob/67a1f287618ad11144e85af3d79a3c240e0268de/scalafix-core/shared/src/main/scala/scalafix/util/TreeOps.scala#L11

```scala
object Mods {
  def unapply(tree: Tree): Option[List[Mod]] = ...
}
```

Seems like `Decl.{Val,Var,Def,Type}` are unhandled, it should be possible to annotate those as well.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        case t @ Defn.Def(mods, _, _, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Defn.Val(mods, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Defn.Var(mods, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Term.Param(mods, _, _, _) if hasSuppressWarnings(mods) =>
```
olafurpg: Missing Type.Param

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        case t @ Defn.Object(mods, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Defn.Trait(mods, _, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Defn.Type(mods, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Defn.Def(mods, _, _, _, _, _) if hasSuppressWarnings(mods) =>
```
olafurpg: Missing Defn.Macro

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        case t @ Defn.Var(mods, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Term.Param(mods, _, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Ctor.Primary(mods, _, _) if hasSuppressWarnings(mods) =>
+          addAnnotatedEscape(t, mods)
+
+        case t @ Ctor.Secondary(mods, _, _, _, _)
```
olafurpg: Missing `Pkg.Object`

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
-        .map(unused => unusedDisableWarning.at(unused.anchorPosition.value))
-
-    val filteredLints = lintMessages.result()
-
-    (patches, filteredLints ++ unusedEnable ++ unusedEscapesWarning)
+      patchesByName.map { case (name, patch) => loop(name, patch) }.asPatch
+
+    val unusedDisableWarning = LintCategory
+      .warning(
+        "Disable",
```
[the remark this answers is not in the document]
marcelocenerine: I like it. Done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        "Rule(s) don't need to be disabled at this position. This can be removed!")
```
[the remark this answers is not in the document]
marcelocenerine: done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+      .withOwner(UnusedScalafixSuppression)
+    val unusedEnableWarning = LintCategory
+      .warning(
+        "Enable",
```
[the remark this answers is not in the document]
marcelocenerine: done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +79,409 @@ class EscapeHatch(
...
+        "Rule(s) not disabled at this position. This can be removed!")
```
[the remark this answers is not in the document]
marcelocenerine: done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: >Actually, could we add a Mods extractor in TreeExtractors?

Nice! Done

>Seems like Decl.{Val,Var,Def,Type} are unhandled, it should be possible to annotate those as well.

For some reason I missed these ones. Thanks for catching this

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: done

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: done. However, the annotation is not permitted on a package object

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: oops. That's another corner case that I did not consider. I think there is no a way to lookup the value. Am I right? Just avoiding crashing for now unless you have some better idea.

marcelocenerine:
@olafurpg thanks for taking a look at this PR again and for providing great suggestions :)

I addressed all comments and updated the PR, so it's ready for another round. Please let me know when you are happy with it so that I can squash all the commits before merging it (unless gh let you do it yourself). Thanks

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -118,261 +80,369 @@ class EscapeHatch(
...
   }
 }
 
 object EscapeHatch {
-  private[EscapeHatch] case class EscapeFilter(
+
+  private val UnusedScalafixSuppression = RuleName("UnusedScalafixSuppression")
+  private val UnusedWarning = LintCategory
+    .warning("", "This comment has no effect and can be removed")
+    .withOwner(UnusedScalafixSuppression)
```
marcelocenerine: Actually, the term "comment" is not 100% accurate here given that this warning message can also refer to a rule name in the annotation. Maybe we can say  "escape" or "suppression" instead?

marcelocenerine: Please see if the updated warning message is generic enough for all positions/scenarios where it can refer to:

- specific rule in an annotation:
```
scala/test/escapeHatch/AnnotationPrefixedRules.scala:23:47: error: [UnusedScalafixSuppression]:
Unused Scalafix suppression. This can be removed
  @SuppressWarnings(Array("scalafix:NoNull", "scalafix:NoDummy"))
                                              ^
```

- wildcard ON comment:
```
scala/test/escapeHatch/AnchorMultipleRules.scala:31:3: error: [UnusedScalafixSuppression]:
Unused Scalafix suppression. This can be removed
  // scalafix:on
  ^
```

- wildcard OFF comment:
```
scala/test/escapeHatch/AnchorWildcard.scala:16:3: error: [UnusedScalafixSuppression]:
Unused Scalafix suppression. This can be removed
  // scalafix:off
  ^
```

- specific rule in an ON comment:
```
scala/test/escapeHatch/AnchorMultipleRules.scala:30:18: error: [UnusedScalafixSuppression]:
Unused Scalafix suppression. This can be removed
  /* scalafix:on NoInfer.any2stringadd, Disable.get */
                 ^
```

- specific rule in an OFF comment:
```
scala/test/escapeHatch/AnchorMultipleRules.scala:22:42: error: [UnusedScalafixSuppression]:
Unused Scalafix suppression. This can be removed
 // scalafix:off NoInfer.any2stringadd, Disable.get
                                        ^
```

olafurpg: Good catch! "comment" is indeed misleading, I like the new phrasing better 👍

scalafix-core/shared/src/main/scala/scalafix/util/TreeOps.scala
```diff
@@ -15,6 +15,29 @@ object TreeExtractors {
       case _ => None
     }
   }
+
+  object Mods {
```
olafurpg: 😍

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -251,11 +210,13 @@ object EscapeHatch {
       }
 
     private def extractRules(mods: List[Mod]): List[(String, Position)] = {
-      def process(rules: List[Term]) = rules.map {
+      def process(rules: List[Term]) = rules.flatMap {
```
olafurpg: FYI `.collect` is a good fit here

marcelocenerine: that's true... too late now. I can update it as part of a future PR

marcelocenerine:
Awesome! Thanks @olafurpg and @MasseGuillaume for your great input on this PR 👏

PR #634, opened 2018-02-25
Regex and hierarchy disabling 

vovapolu:
PR depends on https://github.com/scalacenter/scalafix/pull/601. It adds new ways to block things in `Disable` rule: disable symbols by regexes and by hierarchy.


scalafix-tests/input/src/main/scala/test/DisableHierarchy.scala
```diff
@@ -0,0 +1,30 @@
...
+]
+ */
+package test
+
+object DisableHierarchy {
+  Array(1, 2, 3).equals(Array(5, 6, 7)) // assert: Disable.equals
+  class A(i: Int) {
+    override def equals(o: scala.Any): Boolean = super.equals(o) // assert: Disable.equals
+  }
+  new A(1).equals(new A(2)) // ok
```
vovapolu: That's not an intended behavior. `Denotation.overrides` is always empty.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -1,18 +1,88 @@
...
+
+  def matchDisabledSymbol(disabledSymbol: DisabledSymbol, symbol: Symbol)(
+      implicit index: SemanticdbIndex): Boolean =
+    disabledSymbol match {
+      case DisabledSymbol(Some(s), _, _, banHierarchy, _) =>
+        SymbolOps.isSameNormalized(symbol, s) || (
+          banHierarchy &&
+            index
+              .denotation(symbol)
+              .exists(_.overrides.exists(SymbolOps.isSameNormalized(symbol, _)))
```
vovapolu: Here, `ovverides` is empty list.

MasseGuillaume: I suspect it's because you need to add a `scalacOption`

https://github.com/MasseGuillaume/scalafix/blob/09666ec5406759737b80c4ef0abc5252c2f556e0/build.sbt#L263-L273

vovapolu: @MasseGuillaume It doesn't changes anything...

fommil:
I'm really looking forward to this one. Ban all the java.* and scala.* !

Does it have exclusion lists, to allow a few things through?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,67 @@ package scalafix
...
+
+case class DisabledSymbol(
+    @Description("Symbol to ban.")
+    symbol: Option[Symbol.Global],
+    @Description("Custom message.")
+    message: Option[String],
+    @Description("Custom id for error messages.")
+    id: Option[String],
+    @Description(
+      "When true all overrides of the symbol are also banned. Works only with symbol field, not regex. Doesn't work in unlessSynthetic mode.")
```
olafurpg: Can you please break long sentences like this down into multiple string literals?

```scala
 +      "When true all overrides of the symbol are also banned. " +
          "Works only with symbol field, not regex. " +
          "Doesn't work in unlessSynthetic mode.")

File: scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala [hunk 4]
```diff
@@ -2,18 +2,67 @@ package scalafix
...
+    banHierarchy: Boolean,
+    @Description(
+      "Regex to ban instead of one symbol. When symbol field is specified, this field is ignored.")
+    regex: Option[String]) {
```
olafurpg: Let's make this `Regex` so pattern errors are caught when loading rules instead of during fixing time. I think there's a `ConfDecoder[Regex]` somewhere

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
olafurpg: iirc, you need both
```
-P:semanticdb:overrides:all
-P:semanticdb:denotations:all
```

However, denotations:all should no longer be necessary soon thanks to metacp. I think it's best to wait with depending on `overrides` for now

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -46,6 +46,36 @@ object Disable {
       buf.toList
     }
   }
+
+  def matchDisabledSymbol(disabledSymbol: DisabledSymbol, symbol: Symbol)(
```
olafurpg: Would it be better to move this method to `DisabledSymbol.matches(Symbol)(Index)`?

I'm not a fan of pattern matching case classes with so many fields.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -46,6 +46,36 @@ object Disable {
...
+        SymbolOps.isSameNormalized(symbol, s) || (
+          banHierarchy &&
+            index
+              .denotation(symbol)
+              .exists(_.overrides.exists(SymbolOps.isSameNormalized(symbol, _)))
+        )
+      case d @ DisabledSymbol(None, _, _, _, Some(_)) =>
+        d.compiledRegex.exists(_.findFirstIn(symbol.toString).isDefined)
+      case DisabledSymbol(None, _, _, _, None) =>
+        true // weird case, but it has some logic
```
olafurpg: Should this be handled by the ConfDecoder? If so then we might as well `sys.error("impossible")`.

olafurpg: Comments like `// weird case, but it has some logic` are not very helpful. It's better to explain **why** this is a weird case and **what** "some logic" means.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
olafurpg: s/unlessSynthetic/ifSynthetic/

olafurpg: I propose s/banHierarchy/banOverrides` and alternative phrase the description to

> If true, additionally ban references to methods that override of this symbol.
> Works only with the `symbol` option, not `regex`.

I would then guard against mixing this option with regex in ConfDecoder.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
olafurpg: Actually, this should be a `FilterMatcher` so we can support includes and excludes. You can use the `FilterMatches.matches(String): Boolean` method to test if a string matches the filter.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,85 @@ package scalafix
...
+    message: Option[String],
+    @Description("Custom id for error messages.")
+    id: Option[String],
+    @Description(
+      "Regex to ban instead of one symbol. " +
+        "Supports exclude and include regexes." +
+        "Symbol option is forbidden when regex is specified.")
+    @ExampleValue("""
+                    |{
+                    |  include = "java.io.*"
```
olafurpg: Is this example copy-pasteable? `include` is a HOCON keyword.

It might also be good to give an example with lists
```
includes = [
  "java.io.*"
  "scala.io.*"
]
excludes = [
  scala.io.Foo
  // ...
]
```

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,85 @@ package scalafix
...
+            case (((Some(_), b), c), Some(_)) =>
+              Configured.notOk(
+                ConfError.message("Symbol and regex are both specified."))
+            case (((a @ Some(_), b), c), None) =>
+              Configured.ok(DisabledSymbol(a, b.map(normalizeMessage), c, None))
+            case (((None, b), c), d @ Some(_)) =>
+              Configured.ok(DisabledSymbol(None, b.map(normalizeMessage), c, d))
+            case (((None, b), c), None) =>
+              Configured.notOk(
+                ConfError.message("Symbol and regex are both not specified."))
```
olafurpg: Either `symbol` or `regex` must be specified.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,85 @@ package scalafix
...
+    ConfDecoder.instanceF[DisabledSymbol] {
+      case c: Conf.Obj =>
+        (c.getOption[Symbol.Global]("symbol") |@|
+          c.getOption[String]("message") |@|
+          c.getOption[String]("id") |@|
+          c.getOption[FilterMatcher]("regex"))
+          .andThen {
+            case (((Some(_), b), c), Some(_)) =>
+              Configured.notOk(
+                ConfError.message("Symbol and regex are both specified."))
```
olafurpg: Cannot specify both `symbol` and `regex`, only one of them is allowed.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -46,6 +45,20 @@ object Disable {
       buf.toList
     }
   }
+
+  final class DisableSymbolMatcher(symbols: List[DisabledSymbol])(
```
olafurpg: 👍

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
fommil: would this allow the use of `Foo` the type and the `Foo` constructor whilst banning use of methods on `Foo`?

olafurpg: You can probably encode those rules with more advanced regexps. At the end of the day we have a bunch of "symbol occurences" where symbols are opaque string IDs following a well-specified format https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#scala-symbol

I think the current implementation uses "normalized" symbol matching, which blurs between type/term equality but IIRC we can make "exact" symbol matching opt-in so you can target the type explicitly with `Foo#` instead of `Foo.` (term)

olafurpg: In SemanticDB, a `Symbol` is nothing more than a string ID. `SymbolInformation` holds metadata like types/modifiers/.

Every class/object/def/val/type parameter/.. has a globally unique symbol that should also be quite readable and ergonomic to write by hand.

olafurpg:
Left a review on commits from this PR in #635. To keep things simple let's close this PR in favor of #635 given that it already contains all commits from this PR.

Issue #526, opened 2017-12-21
ban orphan implicits


fommil:
https://github.com/wartremover/wartremover/issues/396

It would be good to exclude definitions of implicits `F[G]` unless they are on the companion of the `F` or the `G`. We might want to consider other HKT shapes too.

Of course, people can still opt out of this with the comment rules, but this makes it more obvious that they are doing something dangerous.

olafurpg:
This is a good idea. After the PR https://github.com/scalacenter/scalafix/pull/827, we are moving most built-in rules outside of the scalacenter/scalafix repository so that they can evolve at a different pace than Scalafix APIs. As long as a custom rule meets a certain level of quality, it can still 

- be automatically be included by sbt-scalafix
- have tab completion in the sbt shell like built-in rules
- be documented on the main Scalafix website

Ping us on Gitter if you believe your rule is good enough for inclusion. Don't hesitate to ask for help around code review or general guidance on writing/sharing custom rules.

PR #635, opened 2018-02-25
OrphanImplicits

vovapolu:
Depends on https://github.com/scalacenter/scalafix/pull/634. Fixes https://github.com/scalacenter/scalafix/issues/526.


olafurpg:
Can you please rebase this PR and cleanup the WIP commits?

website/src/main/tut/docs/rules/OrphanImplicits.md
```diff
@@ -0,0 +1,27 @@
...
+  trait Foo
+  trait Bar[T]
+
+  object Foo {
+    implicit val foo: Foo = ??? // ok
+    implicit val listFoo: List[Foo] = ??? // ok
+  }
+
+  object Bar {
+    implicit val foo: Foo = ??? // error
```
olafurpg: Why is this an error? `Foo` is not of shape `F[G]`

olafurpg: It's not a violation of

> This rule bans definitions of implicits F[G] unless they are on the companion of the F or the G.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+import scalafix.lint.LintCategory
+import scalafix.rule.RuleCtx
+import scalafix.util.SemanticdbIndex
+import scalafix.{LintMessage, Rule}
+
+final case class OrphanImplicits(index: SemanticdbIndex)
+    extends Rule("OrphanImplicits") {
+
+  override def description: String =
+    "Linter bans definitions of implicits F[G] unless they are on the companion of the F or the G"
```
olafurpg: > Linter that reports an error on `implicit` instances of shape `F[G]` or `F` that do not belong in the companion objects of `F` or `G`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+              s"""Orphan implicits are not allowed.
+                 |You should put this definition to one of the following objects:
+                 |${symbols.mkString(", ")}
+                """.stripMargin,
+              pos
+            ))
+      }
+    }
+    new ContextTraverser[LintMessage, List[Symbol.Global]](List.empty)({
+      case (Defn.Object(_, name, _), objs) =>
```
olafurpg: what about implicit instances outside of objects? For example traits/classes/term.block/...

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+        index.symbol(name) match {
+          case Some(s @ Symbol.Global(_, _)) => Right(s :: objs)
```
olafurpg: Does this report an error?

```scala
trait Foo[T]
object Foo {
  object Bar {
    implicit val x: Foo[Int] = null
  }
}
```

It should be an error IMO because the implicit is not a member of the `Foo` companion object.

website/src/main/tut/docs/rules/OrphanImplicits.md
fommil: there is a move to avoid implicits that are not of shape `F[_]` in scalaz, so I'd be happy with this as an error!

fommil: (but it's not an orphan problem... so this should not error under this rule)

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
fommil: this is definitely about things of shape `F[G]`, not about implicits of `F` on an `F`.

In fact I'd like a rule to ban implicits for anything except typeclasses :smile:

fommil: (`ReaderT` is a far superior alternative to implicit configuration objects)

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
fommil: the `trait` rule be fairly complex to implement. It could be enough to say that it's only allowed on a `trait` if it's in the same compilation unit as the `F` or `G`. To combat implicit ordering, you end up with crazy scaffolding like this https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/EitherT.scala#L297-L373

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
olafurpg: The docs had an example of `implicit val f: Foo = null` so I assumed that was intended behavior too. Drop the "or `F`" part then

website/src/main/tut/docs/rules/OrphanImplicits.md
olafurpg: But is this an "orphan implicit error" or something else? I'll leave the final decision to you, just wanted to point out that the tests were not in sync with the documentation description.

fommil: This is not an orphan implicit

vovapolu: @fommil what about F[A, B] form? Is it should be ignored or not?

fommil: `F[A, B]` can look like `F[_]` if one of the types is fixed. If you can test if this implicit is defined on `F`, `A` or `B` that would be good.

fommil:
in hindsight, I think this is very much related to #661 ... 661 is about *what* is allowed to be implicit and this is about where the implicits can be defined.

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,88 @@ package scalafix
...
+import metaconfig.annotation.{Description, ExampleValue}
+import metaconfig.generic.Surface
 import org.langmeta.Symbol
 
 import scalafix.internal.util.SymbolOps
-import metaconfig.annotation.{Description, ExampleValue}
-import metaconfig.generic.Surface
+
+case class DisabledSymbol(
+    @Description("Symbol to ban.")
```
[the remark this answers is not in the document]
olafurpg: Can we give examples here?

* `scala.Predef.any2stringadd`
* `scala.sys.process.Process`

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,88 @@ package scalafix
...
+    symbol: Option[Symbol.Global],
+    @Description("Custom message.")
+    message: Option[String],
+    @Description("Custom id for error messages.")
+    id: Option[String],
+    @Description(
+      "Regex to ban instead of one symbol. " +
```
olafurpg: > Regex to ban symbols matching a given include/exclude patterns.
> Cannot be used together with the `symbol` option.
> Include and exclude filters can be either a list of regex strings or a single regex string.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+import scalafix.rule.RuleCtx
+import scalafix.util.SemanticdbIndex
+import scalafix.{LintMessage, Rule}
+
+final case class OrphanImplicits(index: SemanticdbIndex)
+    extends Rule("OrphanImplicits") {
+
+  override def description: String =
+    "Linter that reports an error on implicit instances of shape F[G] " +
+      "that do not belong in the companion objects of F or G."
```
olafurpg: > that do not belong to the companion object

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  private lazy val errorCategory: LintCategory =
+    LintCategory.error("Orphan implicits should be avoided")
```
olafurpg: I think this description can be `""` because you provide a more detailed message below.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      if (tpeSymbols.nonEmpty && !tpeSymbols.exists(
+          SymbolOps.isSameNormalized(obj, _)
+        )) {
+        Some(
+          errorCategory
+            .at(
+              s"""Orphan implicits are not allowed.
+                 |You should put this definition to one of the following objects:
```
olafurpg: > This definition is only allowed in one of the following objects:

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+                """.stripMargin,
+              pos
+            ))
+      } else {
+        None
+      }
+    }
+
+    ctx.tree.collect {
+      case Defn.Object(_, name, templ) =>
```
olafurpg: `case Defn.Object(_, name @ index.Symbol(obj), templ) =>`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+        index.symbol(name) match {
+          case Some(obj @ Symbol.Global(_, _)) =>
+            templ.stats.collect {
+              case t @ Defn.Val(mods, _, Some(tpe), _)
```
olafurpg: This won't catch implicits with inferred types

```scala
object Foo {
  implicit val x = List(1)
}
```

The robust way to implement this is to query the type of a definition is to use the `.resultType` extension method as is done here https://github.com/scalacenter/scalafix/blob/e46a0401c78d955a9ee218555cfffc7eff2c8230/scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L75-L78

`.resultTypes` will soon be refactored to use `Denotation.tpe: semanticdb3.Type`, but I can take care of that refactoring once that happens.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  override def check(ctx: RuleCtx): Seq[LintMessage] = {
+
+    def handleImplicit(
+        tpe: Type,
+        obj: Symbol.Global,
+        pos: Position): Option[LintMessage] = {
+
+      val tpeSymbols = tpe match {
+        case Type.Apply(t, Seq(arg)) =>
```
olafurpg: This won't handle cases with multiple type arguments

```scala
object Foo {
  implicit val x: Map[String, Int] = Map.empty
}
```

website/src/main/tut/docs/rules/OrphanImplicits.md
olafurpg: Is this example up-to-date?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
vovapolu: Yeah, it won't :) As I understand @fommil wants only F[G] form or some complicated detection of fixed type-vars.

fommil: in hindsight, #661 is a prerequisite for doing orphan implicits properly... it bans everything except typeclasses. And then with what's left we can seach on all the type parameters to see if it's defined in a good place or not.

olafurpg:
> in hindsight, #661 is a prerequisite for doing orphan implicits properly... it bans everything except typeclasses. And then with what's left we can seach on all the type parameters to see if it's defined in a good place or not.

Should the two features be under the same rule? If yes, what should it be named and what are the configuration options?

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -2,18 +2,90 @@ package scalafix
...
+              Configured.ok(DisabledSymbol(None, b.map(normalizeMessage), c, d))
+            case (((None, b), c), None) =>
+              Configured.notOk(
+                ConfError.message("Either symbol or regex must be specified."))
+          }
+      case s: Conf.Str =>
+        symbolGlobalReader
+          .read(s)
+          .map(sym => DisabledSymbol(Some(sym), None, None, None))
+      case _ => Configured.NotOk(ConfError.message("Wrong config format"))
```
olafurpg: `.instance` allows you to pass a partial function and metaconfig produces a slightly more helpful error message

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+    def handleDefn(
+        mods: List[Mod],
+        tpe: Option[Type],
+        d: Defn,
+        obj: Symbol.Global): Option[LintMessage] = {
+      if (mods.exists(_.is[Mod.Implicit])) {
+        tpe match {
+          case Some(t) => handleImplicit(t, obj, d.pos)
```
olafurpg: This would report false positives on situations like
```scala
trait Writer[T]
object Writer {
  type IntWriter = Writer[Int]
  implicit val x: IntWriter = ...
}
```
I would not rely on the syntactic structure of the type in the source.

scalafix-tests/input/src/main/scala/test/DisableRegex.scala
```diff
@@ -0,0 +1,49 @@
...
+]
+ */
+package test
+
+object DisableRegex {
+  import scala.collection.mutable._ // ok
+  import scala.collection.mutable   // ok
+
+  @SuppressWarnings(Array("Disable.ListBuffer"))
+  @SuppressWarnings(Array("Disable.<init>"))
```
olafurpg: I think we should not report duplicate errors on `new X` if `X` is disabled. I think it would be better to skip `<init>` because there is no way to target a constructor without referencing the explicit type.

Not sure what to do if the user wants to disable a specific constructor overload.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
```diff
@@ -0,0 +1,85 @@
...
+              res <- handleImplicit(tpe, obj, d.pos)
+            } yield res
+        }
+      } else {
+        None
+      }
+    }
+
+    ctx.tree.collect {
+      case Defn.Object(_, name @ index.Symbol(sym), templ) =>
```
olafurpg:
```scala
case Defn.Object(_, name @ index.Symbol(sym: Symbol.Global), templ) => 
```

scalafix-core/shared/src/main/scala/scalafix/internal/util/ContextTraverser.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.meta.Tree
+import scala.meta.transversers.Traverser
+
+/**
+  * A tree traverser to collect values with a custom context.
+  * At every tree node, either builds a new Context or returns a new Value to accumulate.
+  * To collect all accumulated values, use result(Tree).
+  */
+class ContextTraverser[Value, Context](initContext: Context)(
```
olafurpg: Is this still used by OrphanImplicits?

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OrphanImplicits.scala
olafurpg: Ahh, you're right! The new semanticdb Type does handle that case however, you can leave it unchanged for now 😉

scalafix-tests/input/src/main/scala/test/DisableRegex.scala
vovapolu: In this case I'm disabling regex `"scala.collection.mutable.*"`. I'm not sure there's a good solution to ignore `<init>`  in regexes. I'd say regex initially is kind of hack, everyone who uses it should expect it's not a silver bullet.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala
```diff
@@ -131,9 +132,11 @@ final case class Disable(index: SemanticdbIndex, config: DisableConfig)
         val isBlocked = new DisableSymbolMatcher(blockedSymbols)
         ctx.index.symbol(t) match {
           case Some(isBlocked(s: Symbol.Global, disabled)) =>
-            Left(
-              createLintMessage(s, disabled, t.pos)
-            )
+            SymbolOps.normalize(s) match {
```
vovapolu: Is it better than the old version? It doesn't affect tests. I added this because of your suggestion to disable normalized symbols.

scalafix-tests/input/src/main/scala/test/DisableRegex.scala
fommil: a good usecase is to allow creating `File`, but not doing anything with it. I'm fine with this meaning that overloads are treated equally. I custom syntax in the future might be nice, but it's a second order feature for sure.

olafurpg:
Superseded by  #669

Issue #643, opened 2018-03-04
Please add support for sbt 1.0


petro-semeniuk-da:
Hi there

I tried to use scalafix by adding just two plugins (as per documentation):
```
addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.10")
addCompilerPlugin("org.scalameta" % "semanticdb-sbt" % "0.6.0" cross CrossVersion.full)
```

however on sbt compile step I'm getting:
```
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
[warn]     ::          UNRESOLVED DEPENDENCIES         ::
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
[warn]     :: org.scalameta#semanticdb-sbt_2.12.4;0.6.0: not found
[warn]     ::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn]     Note: Unresolved dependencies path:
[warn]         org.scalameta:semanticdb-sbt_2.12.4:0.6.0 (/Users/petrosemeniuk/src/da/spider/da-scala/project/plugins.sbt#L15-16)
[warn]           +- default:da-scala-build:0.1.0-SNAPSHOT (scalaVersion=2.12, sbtVersion=1.0)
[error] sbt.librarymanagement.ResolveException: unresolved dependency: org.scalameta#semanticdb-sbt_2.12.4;0.6.0: not found
[error]     at sbt.internal.librarymanagement.IvyActions$.resolveAndRetrieve(IvyActions.scala:331)
```

gabro:
The documentation may be a bit miseleading, but you don't need `semanticdb-sbt` unless you're trying to migrate from sbt 0.13 to sbt 1.0.

It seems you're already using sbt 1.0, so you can't just remove the `semanticdb-sbt` plugin, as it's essentially an experimental workaround for the lack of Scalameta support in Scala 2.10 (used by sbt 0.13)

gabro:
@olafurpg do you think it's worth to add a note in the documentation saying something along these lines

> You only need `semanticdb-sbt` if you intend to use the `Sbt1` rule to migrate from sbt 0.13 to sbt 1.0

?

gabro:
@olafurpg do you think it's worth to add a note in the documentation saying something along these lines

> You only need `semanticdb-sbt` if you intend to use the `Sbt1` rule to migrate from sbt 0.13 to sbt 1.0

?

[...]

PR #644, opened 2018-03-05
Remove sbtfix and semanticdb-sbt from the installation docs.

olafurpg:
sbtfix was an interesting experiment but given that it's incomplete
and no longer being improved it's doing more harm being listed on
the website (see #643) than it's doing good.


gabro:
Should this be removed also from the implementation?

olafurpg:
Removing the implementation would take more time than I have right now, I opened https://github.com/scalacenter/scalafix/issues/645 to track removing the implementation.

gabro:
Perfect, as long as it's tracked, I'm fine with removing it at least from the docs.

Issue #599, opened 2018-02-05
migration mode to add scalafix:ok instead of failing


fommil:
When enabling scalafix on an existing codebase, or introducing a new rule, it is often necessary to temporarily ignore rules until time can be set aside to clean the legacy code.

It would be very useful to be able to run a special scalafix task and have the necessary comments added automatically, such that a subsequent run of the scalafix task will succeed.

vovapolu:
Should it put `scalafix:ok` to prevent rewriting? Or only on linter errors?

fommil:
just linter errors

olafurpg:
Should we generically implement this for all rules or if linter rules should provide rewrites to disable each message under a specific flag? I'm personally inclined for the second option since rules have access to trees which means then can target the comments more precisely. Linter messages have only a `Position`, we could blindly put a `// scalafix:ok` at the end of those lines but it might not work perfectly in all cases.

olafurpg:
Fixed in #646

PR #646, opened 2018-03-05
Suppress mode to add scalafix:ok instead of failing

vovapolu:
See https://github.com/scalacenter/scalafix/issues/599.


scalafix-sbt/src/sbt-test/sbt-scalafix/suppress/build.sbt
```diff
@@ -0,0 +1,19 @@
+import _root_.scalafix.Versions
+
+val suppress = project.settings(
+  scalaVersion := Versions.scala212
+)
+
+TaskKey[Unit]("check") := {
+  val s = streams.value
+  val assertContentMatches: ((String, String) => Boolean) =
```
olafurpg: Is this necessary? I added it in another file to get IntelliJ support when there were many assertions, but here there is only one assertion.

scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
```diff
@@ -52,6 +52,7 @@ sealed abstract case class CliRunner(
   val sbtConfig: ScalafixConfig = config.copy(dialect = dialects.Sbt0137)
   val writeMode: WriteMode =
     if (cli.stdout) WriteMode.Stdout
+    else if (cli.suppress) WriteMode.Supress
```
olafurpg: s/Supress/Suppress/

scalafix-core/shared/src/main/scala/scalafix/util/SuppressOps.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.langmeta.inputs.Position
+
+import scala.annotation.tailrec
+import scala.meta.tokens.Token
+import scalafix.Patch
+import scalafix.rule.RuleCtx
+
+object SuppressOps {
+  @tailrec
+  private def findTokenWithPos(
```
olafurpg:
```scala
/** Binary searches for the last leading token before `pos` */
def findClosestLeadingToken(...)
```

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/SuppressTests.scala
```diff
@@ -0,0 +1,53 @@
...
+    name = "suppress vars",
+    originalFile = """
+                     |object A {
+                     |  var a = 1
+                     |}
+                   """.stripMargin,
+    rule = "scala:scalafix.tests.cli.SuppressRules.NoVars",
+    expectedFile = """
+                     |object A {
+                     |  var/* scalafix:ok */ a = 1
```
olafurpg: I'm a bit concerned about the strategy to place the comment right after the position of the reported message.  Firstly, I'm not sure how excited people would be to merge diffs like this. Secondly, I'm also not sure whether the comment will disable the given linter message in all cases.

I have a feeling it might be better to retrofit rules to support this mode. Something like `ctx.lint(LintMessage, autoSupresss = Patch): Patch` could do the job. Any thoughts @fommil ?

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -20,6 +20,8 @@ object ScalafixPlugin extends AutoPlugin {
       inputKey[Unit]("Run scalafix rule.")
     val scalafixTest: InputKey[Unit] =
       inputKey[Unit]("Run scalafix as a test(without modifying sources).")
+    val scalafixSuppress: InputKey[Unit] =
```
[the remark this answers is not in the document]
olafurpg: PS. I'm not sure there is a need to add an sbt InputTask for this give it can be accomplished with `scalafixCli --suppress`. It might require an update on the sbt-scalafix completions.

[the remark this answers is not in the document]
fommil: I have no strong preference. Naming stuff is hard.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/SuppressTests.scala
olafurpg: Sounds good 👍 this generic approach should be OK then.

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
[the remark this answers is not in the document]
vovapolu: I'd prefer the longest name since it shouldn't be used often :)

scalafix-sbt/src/sbt-test/sbt-scalafix/suppress/project/plugins.sbt
```diff
@@ -0,0 +1,3 @@
+libraryDependencies += "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0"
```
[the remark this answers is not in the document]
vovapolu: It doesn't work without that line :)

fommil:
wh0000p!

PR #647, opened 2018-03-07
typo in docs

fommil:
I think this is the reason for the mistake in the docs at

https://scalacenter.github.io/scalafix/docs/rules/Disable#examples-1


olafurpg:
@vovapolu ping

PR #655, opened 2018-03-10
Ban overrides

vovapolu:
Based on https://github.com/scalacenter/scalafix/pull/634. Moved `banOverrides` option to the separate PR since there will be new way to detect overries n the semanticDb soon.


olafurpg:
Is there a ticket tracking this feature? It would be good to label it as "blocked".

PR #656, opened 2018-03-11
Fix #653

### LINKED ISSUE #653: RemoveUnusedImports with comma-separated imports
I have a line of imports like so:

```scala
import eu.timepit.refined, refined.api.Refined, refined.string._, refined.auto._, refined.collection._
```

Now imagine `eu.timepit.refined.string._` is unused.

Unfortunately, Scalafix is a little eager in removing characters it seems as it turns the above into the following:

```scala
import eu.timepit.refined, refined.api.Refinedrefined.auto._, refined.collection._
```

I have to manually place back a comma (and space) or (what I am going to do now) keep the imports on separate lines altogether.

gabro:
I think I've identified the problem, but this is still WIP (for now, I've simply added a failing test)

In short: we currently remove the comma leading an importer in any case, but we should probably limit it to when the importer lives in a binpack block.

E.g.

```scala
import a.b.{ c, d } // it's ok to remove the leading comma when removing `c`

import a.b, b.c._, b.d // it's not ok to remove the leading comma when removing `b.c._`
```

This boils down to fixing the logic here:

https://github.com/scalacenter/scalafix/blob/f5e0a1c084256e42cf6725208d5c13113f3f4821/scalafix-core/shared/src/main/scala/scalafix/internal/patch/ImportPatchOps.scala#L206-L207

which I still haven't figured out how.

### DIFF
--- a/scalafix-tests/input/src/main/scala/test/RemoveUnusedImports.scala
+++ b/scalafix-tests/input/src/main/scala/test/RemoveUnusedImports.scala
@@ -23,6 +23,7 @@ import scala.concurrent.{ // formatting caveat
 //  , ExecutionException
 //  , TimeoutException // ERROR
 //}
+import scala.util, util.Sorting._, util.Sorting
 
 object RemoveUnusedImports {
   import Future._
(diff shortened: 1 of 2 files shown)


fommil:
Oh, I hit this recently! Good to see you're on it.

olafurpg:
Thanks for the repro @gabro ! I left a comment in #653 saying I'll try to take a stab at this once v0.6.0 is out, until then we can close this PR.

PR #664, opened 2018-03-22
Remove support for semantic rewriting of .sbt files.

### LINKED ISSUE #645: Remove semanticdb-sbt integration
Currently, scalafix supports semantic rewriting of sbt sources through semanticdb-sbt. This functionality does not work so well and its implementation is no longer being improved. Given that it's still a maintenance burden to upgrade semanticdb-sbt https://github.com/scalameta/semanticdb-sbt/pulls?q=is%3Apr+is%3Aclosed
screen shot 2018-03-05 at 10 52 23

we should remove this integration.

* [ ] upgrade to sbt 0.13.17

olafurpg:
Support for semantic rules for .sbt files never worked robustly.
It's not on the roadmap to improve the situation so it's best to remove
this functionality altogether.

Fixes #645.


olafurpg:
A lot of removed code today 😎 🔥

PR #669, opened 2018-04-01
Disable regex

vovapolu:
Removing `OrphanImplicits` from https://github.com/scalacenter/scalafix/pull/635, since it's not going to be merged soon...


scalafix-tests/input/src/main/scala/test/DisableRegex.scala
```diff
@@ -0,0 +1,49 @@
...
+]
+ */
+package test
+
+object DisableRegex {
+  import scala.collection.mutable._ // ok
+  import scala.collection.mutable   // ok
+
+  @SuppressWarnings(Array("Disable.ListBuffer"))
+  @SuppressWarnings(Array("Disable.<init>"))
```
olafurpg: I propose we either

* ignore constructors (skip when name is `<init>`), or
* give `Disable.<init>`a more helpful name such as `ListBuffer.<init>`

I vote for the first option since you can't reference a constructor without referencing the enclosing type. Having to disable twice is annoying.

scalafix-tests/input/src/main/scala/test/DisableRegex.scala
```diff
@@ -0,0 +1,49 @@
...
+  val buffer = new ListBuffer[Int]()
+
+  @SuppressWarnings(Array("Disable.mutable"))
+  @SuppressWarnings(Array("Disable.HashMap"))
+  @SuppressWarnings(Array("Disable.<init>"))
+  val abc = new mutable.HashMap[String, String]()
```
olafurpg: Intuitively, there is only one symbol being referenced here `HashMap.<init>` so I think it's a bit weird that three errors are reported.

I think it would make sense to skip the qualifier of a `Term.Select` if it only contains `Term.Select` or `Term.Name`.

```scala
a.b.C // only report on C
d.E // only report on E
f.g(1).h // report on h and g
```

It should be simple to do this check with something like
```scala
def skipQualifier(qual: Term) Boolean = qual match {
  case _: Term.Name => true
  case Term.Select(q, _) => skipQualifier(q)
}
```

scalafix-tests/input/src/main/scala/test/DisableRegex.scala
```diff
@@ -29,12 +29,9 @@ object DisableRegex {
   import scala.collection.mutable   // ok
 
   @SuppressWarnings(Array("Disable.ListBuffer"))
-  @SuppressWarnings(Array("Disable.<init>"))
   val buffer = new ListBuffer[Int]()
```
olafurpg: What happens if you have a more complex qualifier in a select?

```
a.b(1).c
```
and `a` and `c` are banned? I would expect there to be two lint errors
```
a.b(1).c
^

a.b(1).c
       ^
```

fommil: Colombo: one more thing

olafurpg: I opened https://github.com/scalacenter/scalafix/issues/679 and assigned the ticket to you @vovapolu . I will go ahead and merge to unblock a release, but please be sure to follow up on that ticket as I think the current change results in surprising behavior.

Issue #674, opened 2018-04-04
ExplicitResultTypes crashes on java paths


olafurpg:
```scala
scala/test/explicitResultTypes/ExplicitResultTypesBase.scala:55:16: error: java.lang.ClassCastException: scala.meta.Type$Select$TypeSelectImpl cannot be cast to scala.meta.Term$Ref
	at scalafix.internal.util.TypeSyntax.loop$2(TypeSyntax.scala:183)
	at scalafix.internal.util.TypeSyntax.symbolToTree(TypeSyntax.scala:196)
	at scalafix.internal.util.TypeSyntax.toTypeRef(TypeSyntax.scala:287)
	at scalafix.internal.util.TypeSyntax.toType(TypeSyntax.scala:341)
	at scalafix.internal.util.TypeSyntax.$anonfun$toType$1(TypeSyntax.scala:297)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:448)
	at scala.collection.Iterator.foreach(Iterator.scala:929)
	at scala.collection.Iterator.foreach$(Iterator.scala:929)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
	at scala.collection.generic.Growable.$plus$plus$eq(Growable.scala:59)
	at scala.collection.generic.Growable.$plus$plus$eq$(Growable.scala:50)
	at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:186)
	at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:44)
	at scala.collection.TraversableOnce.to(TraversableOnce.scala:310)
	at scala.collection.TraversableOnce.to$(TraversableOnce.scala:308)
	at scala.collection.AbstractIterator.to(Iterator.scala:1417)
	at scala.collection.TraversableOnce.toList(TraversableOnce.scala:294)
	at scala.collection.TraversableOnce.toList$(TraversableOnce.scala:294)
	at scala.collection.AbstractIterator.toList(Iterator.scala:1417)
	at scalafix.internal.util.TypeSyntax.targs$1(TypeSyntax.scala:297)
	at scalafix.internal.util.TypeSyntax.toType(TypeSyntax.scala:347)
	at scalafix.internal.util.TypeSyntax.prettify(TypeSyntax.scala:449)
	at scalafix.internal.util.TypeSyntax$.prettify(TypeSyntax.scala:28)
[...]

PR #675, opened 2018-04-04
Upgrade to scalameta 3.7.0 with default settings.

olafurpg:
This commit is a major milestone in that Scalafix now uses the latest
SemanticDB with default settings. Previously, scalafix relied on two
features that are now no longer default:

* -P:semanticdb:denotations:all, to lookup information about any symbol
  scalafix relied on semanticdb-scalac to persist full information for
  all referenced symbols in each compilation unit. Now, scalafix
  can load information directly from the classpath thanks to metacp
  and the new --dependency-classpath command line flag.
* -P:semanticdb:signatures:old, these were pretty-printed versions of
  types that have been replaced with the new specification of `Type` in
  https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md#type

There's quite a bit happening in this PR, it's roughly broken down by:

* New --dependency-classpath cli flag be processed by metacp.
* New `SymbolTable` to lookup information about symbols from the metacp
  processed classpath.
* Refactor ExplicitResultTypes to use s.Type instead of the deprecated
  s.SymbolInformation.signatur
* Remove tests for Symbol.Multi, which are no longer emitted by
  semanticdb-scalac.

These changes required updates to testkit, the cli and sbt-scalafix.

Future work:

* The tests pass but I noticed a couple errors while running
  ExplicitResultTypes on Slick. Given that this PR already does quite a
  lot I prefer to get it merged asap and fix those errors later.
* https://github.com/scalacenter/scalafix/issues/674
* --dependency-classpath may not be necessary. We might be able to away
  with only --classpath and be smarter about which files to load from
  the provided files to fix.


build.sbt
```diff
@@ -207,10 +207,18 @@ val scalafixSbt = MultiSbtProject(
...
     .disablePlugins(ScalafixPlugin)
 )
-lazy val scalafixSbt1 =
-  scalafixSbt(scala212, sbt1, _.dependsOn(testUtils212 % Test))
-lazy val scalafixSbt013 =
-  scalafixSbt(scala210, sbt013, _.dependsOn(testUtils210 % Test))
+lazy val scalafixSbt1 = scalafixSbt(
+  scala212,
+  sbt1,
+  _.dependsOn(testUtils212 % Test)
```
ShaneDelmore: Was this formatting change intended?

scalafix-cli/src/main/scala/scalafix/cli/CliRunner.scala
```diff
@@ -315,8 +319,8 @@ object CliRunner {
         .map(path => AbsolutePath(path)(common.workingPath))
         .toList
 
-    def resolveClasspath: Configured[Classpath] =
-      classpath match {
+    def resolveClasspath: Configured[Classpath] = {
+      classDirectory match {
```
ShaneDelmore: Not a blocker but I don't like that this name hints that the classpath is a single directory entry, which is not what I usually see for a classpath.

scalafix-cli/src/main/scala/scalafix/internal/cli/ClasspathOps.scala
```diff
@@ -1,8 +1,41 @@
...
+    case (k, v) if k.endsWith(".boot.class.path") => Classpath(v)
+  }
+
+  /**
+    * Process classpath with metacp to build semanticdbs of global symbols.
+    *
+    * @param sclasspath Regular classpath to process.
+    * @param out The output stream to print out error messages.
+    */
+  def toMetaClasspath(sclasspath: Classpath, out: PrintStream): Classpath = {
```
xeno-by: The fact that these methods take a printstream looks really awkward.

scalafix-cli/src/main/scala/scalafix/internal/cli/ClasspathOps.scala
```diff
@@ -1,8 +1,41 @@
...
+    val withJDK = Classpath(
+      bootClasspath.fold(sclasspath.shallow)(_.shallow ::: sclasspath.shallow))
+    val settings = metacp
+      .Settings()
+      .withClasspath(withJDK)
+      .withScalaLibrarySynthetics(true)
+    val reporter = metacp.Reporter().withOut(out)
+    val mclasspath = scala.meta.cli.Metacp.process(settings, reporter).get
+    mclasspath
```
xeno-by: Is it okay to crash if metacp returns None?

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
```diff
@@ -55,12 +55,20 @@ case class ScalafixOptions(
     sourceroot: Option[String] = None,
     @HelpMessage(
       "java.io.File.pathSeparator separated list of directories or jars containing " +
```
xeno-by: Is this a list or a single directory?

scalafix-core/shared/src/main/scala/scalafix/internal/util/CombinedSymbolTable.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.langmeta.internal.semanticdb._
+import scala.meta.internal.{semanticdb3 => s}
+
+/**
+  * Combination of a SymbolTable and SemanticdbIndex.
+  *
+  * This utility is necessary because SymbolTable contains only global symbols and
+  * SemanticdbIndex has local symbols.
+  */
+class CombinedSymbolTable(index: SemanticdbIndex, table: SymbolTable) {
```
xeno-by: Why do we need to pass both `index` and `table` here, given that `EagerInMemorySemanticdbIndex` already takes a symbol table?

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
```diff
@@ -0,0 +1,117 @@
...
+/**
+  * Implementation of SymbolTable that lazily loads symbols on demand.
+  *
+  * @param mclasspath The classpath to load symbols from. Needs to be pre-processed by metacp,
+  *                   see ClasspathOps.toMclasspath in scalafix-cli.
+  */
+class LazySymbolTable(mclasspath: Classpath) extends SymbolTable {
+
+  // Map from symbols to the paths to their corresponding semanticdb file where they are stored.
+  private val unloadedSymbols = TrieMap.empty[String, AbsolutePath]
```
xeno-by: "unloaded" as in "not yet loaded" or as in "previously loaded, but no longer available"?

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
```diff
@@ -0,0 +1,117 @@
...
+      throw new IllegalArgumentException(root.toString())
+    }
+  }
+
+  private def withJarFileSystem[T](path: AbsolutePath)(
+      f: AbsolutePath => T): T = {
+    val fs = newJarFileSystem(path)
+    // NOTE(olafur): We don't fs.close() because that can affect another place where `FileSystems.getFileSystems`
+    // was used due to a `FileSystemAlreadyExistsException`. I don't know what the best solution is for reading the
+    // same zip file from multiple concurrent places.
```
xeno-by: Maybe this is/was the source of deadlocks?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,451 @@
...
 object TypeSyntax {
-  def prettify(tpe: Type, ctx: RuleCtx, shortenNames: Boolean)(
-      implicit index: SemanticdbIndex): (Type, Patch) = {
+  def prettify(
+      stpe: s.Type,
+      ctx: RuleCtx,
+      shortenNames: Boolean,
+      pos: Position)(implicit index: SemanticdbIndex): Option[(Type, Patch)] = {
+    try {
+      val table = index.asInstanceOf[EagerInMemorySemanticdbIndex].table
```
xeno-by: Woah, this cast looks scary.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,451 @@
...
         }
+      // TODO: handle non-singleton widening.
+      case _ => tpe
+    }
+  }
+
+  /*
+   * Returns a scala.meta.Tree given a scala.meta.Symbol.
+   *
+   * NOTE: this method has become an utter mess and is in need of a clean reimplementation.
```
xeno-by: Okay, not gonna review then.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -74,9 +75,16 @@ case class ExplicitResultTypes(
...
-        typ <- symbol.resultType
-      } yield TypeSyntax.prettify(typ, ctx, config.unsafeShortenNames)
+        denot <- name.denotation
+        tpe <- denot.tpeInternal
+        // Skip existential types for now since they cause problems.
+        if !tpe.tag.isExistentialType
+        result <- TypeSyntax.prettify(
+          tpe,
+          ctx,
+          config.unsafeShortenNames,
```
ShaneDelmore: How is this config passed? https://scalacenter.github.io/scalafix/docs/users/configuration shows how to add rules to the scalafix config, but does not show if it is possible to pass rule specific config.

scalafix-core/shared/src/main/scala/scalafix/internal/util/CombinedSymbolTable.scala
ShaneDelmore: Is combined symbol table needed if shortenNames is false?

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
```diff
@@ -0,0 +1,113 @@
...
+  * @param mclasspath The classpath to load symbols from. Needs to be pre-processed by metacp,
+  *                   see ClasspathOps.toMclasspath in scalafix-cli.
+  */
+class LazySymbolTable(mclasspath: Classpath) extends SymbolTable {
+
+  // Map from symbols to the paths to their corresponding semanticdb file where they are stored.
+  private val unloadedSymbols = TrieMap.empty[String, AbsolutePath]
+  private val loadedSymbols = TrieMap.empty[String, s.SymbolInformation]
+  private val semanticdb = "META-INF/semanticdb"
+  private val semanticIdx = "META-INF/semanticdb.semanticidx"
```
ShaneDelmore: would this be a good place to add
```
private val manifest = "META-INF/MANIFEST.MF"
```
in case there is a classpath manifest in a jar?

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
```diff
@@ -0,0 +1,113 @@
...
+        toplevel.symbol,
+        root.resolve(semanticdb).resolve(toplevel.uri))
+    }
+  }
+
+  private def loadSemanticdbIndex(root: AbsolutePath): Unit = {
+    if (root.isDirectory) {
+      loadIndex(root, Files.newInputStream(root.resolve(semanticIdx).toNIO))
+    } else if (PathIO.extension(root.toNIO) == "jar") {
+      withJarFileSystem(root) { jarRoot =>
```
ShaneDelmore: I am not familiar enough with the NIO api to know the answer to this, but are you sure that these apis handle symlinks?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
ShaneDelmore: Seconded, why do we know more than the compiler here?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,451 @@
...
 import scala.meta._
+import scala.meta.internal.{semanticdb3 => s}
 import scala.reflect.ClassTag
+import scala.util.control.NonFatal
 import scalafix._
 import scalafix.internal.util.SymbolOps.SymbolType
-import scalafix.util.SymbolMatcher
-import scalafix.util.TreeExtractors._
 import org.langmeta.semanticdb.Symbol
+import scala.meta.internal.semanticdb3.SingletonType.{Tag => x}
```
ShaneDelmore: This saves two characters and makes the code harder to read.  Why not just Tag.?

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,451 @@
...
       }
+      def unapply(tpe: s.Type): Boolean = matches(tpe)
+    }
+    object AnyRef extends TypeRefExtractor("scala.AnyRef#")
+    object Nothing extends TypeRefExtractor("scala.Nothing#")
+    object Any extends TypeRefExtractor("scala.Any#")
+  }
+
+  def isFunctionN(symbol: String): Boolean = {
+    symbol.startsWith("scala.Function") &&
```
ShaneDelmore: I don't know if there this is a perforamance issue, but from an accuracy issue how about "scala.Function(\\d)+#".r to make sure we capture Function1 but not FunctionPure (not that such a thing currently exists).

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,451 @@
...
+  def fail(any: GeneratedMessage): Nothing = sys.error(any.toProtoString)
+
+  def isStable(symbol: Symbol): Boolean = {
+    def loop(symbol: Symbol): Boolean = symbol match {
+      case Symbol.None => true
+      case Symbol.Global(owner, Signature.Term(_)) => loop(owner)
+      case _ => false
     }
+    symbol match {
+      case Symbol.Global(owner, Signature.Term(_) | Signature.Type(_)) =>
```
ShaneDelmore: I read this as "A symbol is stable only if it is A Type or Term owned by Symbol.None or owned by a chain of n Global Terms".  Is that correct?

scalafix-tests/output/src/main/scala/test/explicitResultTypes/ExplicitResultTypesPathDependent.scala
```diff
@@ -4,7 +4,7 @@ object ExplicitResultTypesPathDependent {
   class Path {
     class B { class C }
     implicit val x: Path.this.B = new B
-    implicit val y: x.C = new x.C
+    implicit val y: Path.this.x.C = new x.C
     def gimme(yy: x.C) = ???; gimme(y)
```
ShaneDelmore: ??? indeed.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
olafurpg: It's configuration for ExplicitResultTypes documented here https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes

scalafix-core/shared/src/main/scala/scalafix/internal/util/CombinedSymbolTable.scala
olafurpg: Good catch, I've merged this into `EagerSemanticdbIndex` avoiding the need for `CombinedSymbolTable`.

olafurpg: It's needed both when shortenNames is true and false.

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
olafurpg: Renamed to `notYetLoadedSymbols`

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
olafurpg: I think that should be handled elsewhere, for example `scala.meta.Classpath` should be able to handle.

scalafix-core/shared/src/main/scala/scalafix/internal/util/LazySymbolTable.scala
olafurpg: That was my first suspicion as well! I continuously ran the test suite to see problems. This wasn't the cause of the deadlock, I checked the stack traces and  they led to metacp.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: I've replaced `x` with `Tag` but kept the import local, `Tag` is used in `s.Type.Tag`/`s.SingletonType.Tag` and a few others I think.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: The `SemanticdbIndex` trait should only have a single implementation, I've kept the abstract trait to separate the public API from the implementation. I admit this is not an ideal situation but the alternative is to expose implementation details in the public API. We have the same situation in other places, for example for linter suppression. `private[scalafix]` does not bite enough IMO compared to forcing a cast.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: there are no other `scala.Function*` symbols so I thought this would be enough, regexes are such a heavy machine.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: This method suffers from the same problems as symbolToTree, it can be ignored.

olafurpg: With the introduction of SymbolInformation one should no longer infer any information from the structure of symbols. Symbols are purely an ID and nothing more.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: The implementation is unchanged from the previous TypeSyntax, the diff is mostly reformatting with minor changes.

olafurpg:
I'm investigating the CI build failures in https://github.com/scalameta/scalameta/pull/1476 I suspect they are caused by even more incorrect usage of nio FileSystems.

The Travis 2.11 failures seem to be caused by differences is positions compared to 3.2.0

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
ShaneDelmore: I wasn’t able to put that together from the docs for some reason but I get it now, thanks.

scalafix-cli/src/main/scala/scalafix/internal/cli/ClasspathOps.scala
xeno-by: Use System.out directly.

olafurpg: Scalafix runs via nailgun, in IDEs with custom logging, etc, 

It's not OK to use System.out

scalafix-cli/src/main/scala/scalafix/internal/cli/ClasspathOps.scala
```diff
@@ -1,8 +1,54 @@
...
+    val withJDK = Classpath(
+      bootClasspath.fold(sclasspath.shallow)(_.shallow ::: sclasspath.shallow))
+    val default = metacp.Settings()
+    val settings = default
+      .withClasspath(withJDK)
+      .withScalaLibrarySynthetics(true)
+      .withCacheDir(cacheDirectory.getOrElse(default.cacheDir))
+    val reporter = metacp
+      .Reporter()
+      .withOut(devNull) // out prints classpath of proccessed classpath, which is not relevant for scalafix.
```
xeno-by: Huh, this is unfortunate, given that we're calling `process` directly.

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
```diff
@@ -55,12 +55,28 @@ case class ScalafixOptions(
...
+    @ValueDescription("entry1.jar:entry2.jar:target/scala-2.12/classes/")
+    dependencyClasspath: Option[String] = None,
+    @HelpMessage(
+      "Global cache location to persist metacp artifacts produced by analyzing --dependency-classpath. " +
+        "The default location depends on the OS and is computed with https://github.com/soc/directories-jvm " +
+        "using the project name 'semanticdb'. " +
+        "On macOS the default cache directory is ~/Library/Caches/semanticdb. ")
+    metacpCacheDir: Option[String] = None,
+    @HelpMessage("Set this flag to disable parallel processing with metacp.")
+    metacpNoPar: Boolean = false,
```
xeno-by: I think that the help message should explain that there is a danger of deadlocks.

xeno-by: @ShaneDelmore We'll need to set this flag when we'll be upgrading Scalafix internally.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
```diff
@@ -1,138 +1,452 @@
...
+
+  object FunctionN {
+    def unapply(symbol: String): Boolean = isFunctionN(symbol)
+  }
+  object TupleN {
+    def unapply(symbol: String): Boolean = isTupleN(symbol)
+  }
+}
+
+class TypeSyntax private (
```
xeno-by: Would you ultimately want parts of this code to end up in Metap? https://github.com/scalameta/scalameta/issues/1479

scalafix-cli/src/main/scala/scalafix/internal/cli/ScalafixOptions.scala
olafurpg: Done.

scalafix-core/shared/src/main/scala/scalafix/internal/util/TypeSyntax.scala
olafurpg: I'd like to rewrite this file from scratch first before considering that. I am not happy with the code quality here, but I am more eager to upgrade to scalameta v3.7 and this PR is already doing too much.

olafurpg:
Merging and releasing 0.6.0-M2 :shipit: 

Note that ExplicitResultTypes in M2 will be a fairly broken until TypeSyntax gets a rewrite that I'm planning for this week.

PR #690, opened 2018-04-11
Cache jars and classloaders in sbt-scalafix.

olafurpg:
There were reported heap-space usage problems that are potentially
caused by leaking classloaders in the sbt plugin. I thought that the sbt
task engine cached task results if the depended on settings/tasks don't
change.  This commit is a quick fix to test the hypothesis whether the
sbt plugin is causing problems

### DIFF
--- a/scalafix-sbt/src/main/scala/scalafix/internal/sbt/CliWrapperPlugin.scala
+++ b/scalafix-sbt/src/main/scala/scalafix/internal/sbt/CliWrapperPlugin.scala
@@ -1,6 +1,5 @@
 package scalafix.internal.sbt
 
-import java.net.URL
 import java.net.URLClassLoader
 import java.util
 import sbt.AutoPlugin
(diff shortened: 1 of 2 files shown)


scalafix-sbt/src/main/scala/scalafix/internal/sbt/CliWrapperPlugin.scala
```diff
@@ -29,15 +32,30 @@ object CliWrapperPlugin extends AutoPlugin {
       taskKey[HasMain]("Classloaded instance of main")
   }
   import autoImport._
+  private val cachedScalafixMain =
+    util.Collections.synchronizedMap(
+      new util.HashMap[(Seq[URL], String), HasMain]())
```
fommil: prefer `String` to `URL` if you rely on hashcodes or equals working. It hits the network.

I have a refined type for this, but I'm guessing you can't depend on Refined :smile:  https://gitlab.com/fommil/drone-dynamic-agents/blob/master/src/main/scala/http/client/AsciiUrl.scala

fommil: yeah, but make sure you canonicalise. And catch the exceptions.

Incidentally, one of the biggest code cleanup efforts in ensime was ensuring we did this everywhere we could be given a `File`. Hence, https://github.com/ensime/ensime-server/blob/3.0/api/src/main/scala/org/ensime/io/Canon.scala

Issue #679, opened 2018-04-11
Disable should report error on banned symbols in select chains 


olafurpg:
Consider the snippet `a.banned1(1, 2).c.banned2`. If the user has configured `Disable` to report an error on symbols `banned1` and `banned2` then I would expect two errors to be reported. Changes in #669 may introduce a bug so that only one error is reported, I am not 100% but we should at least add a test case for this scenario.

```
a.banned1(1, 2).c.banned2
  ^

a.banned1(1, 2).c.banned2
                  ^

```

vovapolu:
Everything works fine :) Current implementation skips only `a.b.c.d` cases, e.g. reports only on the last disabled symbol. For example if we disable `a` and `c`, then it will report only on `c`, if we disable `a`, `b`, `d` it will report only on `d`. It doesn't affect on more complex `Term.Select` chains.

olafurpg:
Can you send a PR adding a test for `a.banned1(2).banned2`?

vovapolu:
:D done

olafurpg:
Fixed in #691

PR #691, opened 2018-04-11
Tests for new behavior on Term.Select in Disable rule

vovapolu:
See #679


scalafix-tests/input/src/main/scala/test/DisableRegex.scala
```diff
@@ -34,6 +37,9 @@ object DisableRegex {
   @SuppressWarnings(Array("Disable.HashMap"))
   val abc = new mutable.HashMap[String, String]()
 
+  @SuppressWarnings(Array("Disable.ListBuffer", "Disable.iterator", "Disable.toString"))
```
olafurpg: 😍

PR #697, opened 2018-04-12
Parallel tags

fommil:
now users can

```
concurrentRestrictions in Global += Tags.limit(Linting, 1)
```

to limit how many tasks are spawned at once.

Unfortunately, in my case, one of my projects uses 12GB of heap so I don't benefit.


project/Dependencies.scala
```diff
@@ -14,7 +14,7 @@ object Dependencies {
   def scala212 = "2.12.4"
   val currentScalaVersion = scala212
 
-  def sbt013 = "0.13.16"
+  def sbt013 = "0.13.17"
```
olafurpg: 👍 this upgrade was blocked until the recent removal of semanticdb-sbt

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
```diff
@@ -15,6 +15,8 @@ object ScalafixPlugin extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
   override def requires: Plugins = JvmPlugin
   object autoImport {
+    val Linting = Tags.Tag("linting")
```
olafurpg: `Linting` is a bit misleading considering scalafix does linting and refactoring. Would it be OK to call it `Scalafix` instead?

olafurpg:
`./scalafmt --diff` fixes the CI complaints

scalafix-sbt/src/main/scala/scalafix/sbt/ScalafixPlugin.scala
fommil: the sbt manual recommends that the tag be named after the purpose of the work, not the specific implementation. I was nervous with calling it `Scalafix` incase it clashed with some other symbol of the same name... I can make the change if you're sure about it.

fommil:
@gabro if you can point me to the relevant file to edit, I can send a small comment.

gabro:
@fommil I think either the [installation](https://scalacenter.github.io/scalafix/docs/users/installation) or the [FAQ](https://scalacenter.github.io/scalafix/docs/users/faq) are good pages where to squeeze this in. The easiest way to contribute this is to click on *EDIT THIS PAGE* on the page.

PR #701, opened 2018-04-13
regexes describe exact matches

### LINKED ISSUE #700: regexes should be interpreted as "match exactly"
The following regexes

```scala
includes = "\\Qjava.\\E.*"
excludes = "\\Qjava.\\E"
```

result in a no-op because they are interpreted as sub pattern matches, which came up in #680 

This is very unsurprising, I think it would be best if regexes were interpreted as "match exactly" which can be implemented by adding `^` / `$` to the front/end of the string (if not already present).

fommil:
close #700


olafurpg:
Closing in favor #715

Issue #396, opened 2017-10-12
Website css on mobile


olafurpg:
Should code listing be wrapped? I personally prefer horizontal scrolling.
![img_3444](https://user-images.githubusercontent.com/1408093/31488842-4ae40990-af3f-11e7-8579-e073ea4e75a9.png)

Long titles exceed width
![img_3443](https://user-images.githubusercontent.com/1408093/31488843-4b01bdf0-af3f-11e7-9e30-6c560bc536de.png)

Long inline code exceeds width
![img_3442](https://user-images.githubusercontent.com/1408093/31488844-4b20a4f4-af3f-11e7-91e2-3702863d50c6.png)

Should tables get horizonal scrolling?
![img_3441](https://user-images.githubusercontent.com/1408093/31488845-4bbcf494-af3f-11e7-9e30-8ea993b22e84.png)

Also, should we maybe disable the chat button on mobile? It's quite demanding IMO on smaller screens.

olafurpg:
Fixed in https://github.com/scalacenter/scalafix/pull/808 by moving to docusaurus.

PR #711, opened 2018-04-22
#396

skyrylyuk:
Added horizontal scrolling for tables.
Fixed long titles, texts and inline codes who exceed width.


olafurpg:
Note we currently don't auto-publish the site after CI merge so it will take a while until this change goes live. We are currently blocked by an internal refactoring for v0.6

PR #712, opened 2018-04-23
Make Patch.merge public.

### LINKED ISSUE #709: Make `Patch.merge` public
Currently, `Patch.merge` is private https://github.com/scalacenter/scalafix/blob/57061f728e69fb422928beed478de00f43f4c80f/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala#L111

We use this method in metals to convert scalafix `Patch` into LSP `TextEdit`. We currently copy-paste the implementation, but it would be great to call the scalafix API directly.

olafurpg:
Fixes #709

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
@@ -108,7 +108,7 @@ object Patch {
   /** A patch that does no diff/rule */
   val empty: Patch = EmptyPatch
 
-  private def merge(a: TokenPatch, b: TokenPatch): TokenPatch = (a, b) match {
+  def merge(a: TokenPatch, b: TokenPatch): TokenPatch = (a, b) match {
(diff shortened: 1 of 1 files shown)


scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -108,7 +108,7 @@ object Patch {
   /** A patch that does no diff/rule */
   val empty: Patch = EmptyPatch
 
-  private def merge(a: TokenPatch, b: TokenPatch): TokenPatch = (a, b) match {
+  def merge(a: TokenPatch, b: TokenPatch): TokenPatch = (a, b) match {
```
gabro: Seeing `empty`, `merge` and no instance of `Monoid` makes me sad :(

Maybe one day :P

olafurpg: haha, I know, the design is not coincidental 😉

Issue #694, opened 2018-04-11
multiple scalafix:ok will fail


fommil:
because there are two `scalafix:ok` markups, both get ignored.

```scala
  val map: mutable.Map /* scalafix:ok */ [String, Data] = mutable.Map.empty /* scalafix:ok */

```

olafurpg:
Can you please include more details even if it feels like spelling out the problem. 

* What is your configuration?
* What is the expected output?
* What is the obtained output?

fommil:
yeah, sorry, it's difficult because this is a proprieatry codebase. Roughly the config is just `Exclude.symbols` but I think this is relevant to any two rules on the same line. I've also seen it with `DisableSyntax.keywords` for e.g. a var that is set to null like

```
var foo: String = null
```

with the `scalafix:ok` after `var` and `null`.

olafurpg:
I was able to reproduce with this test case

```scala
/* ONLY
rules = Disable
Disable.ifSynthetic = [
  "scala.Any"
]
*/
package test

class DisableIfSynthetic {
  val x =  // assert Disable.Any
    null.asInstanceOf[Any]
}
```

There seems to be a bug in how ContextTraverser uses `allDisabledSymbols` here https://github.com/scalacenter/scalafix/blob/6d48e0c419a63956274cfcc8e9bb23601c75ccfa/scalafix-core/shared/src/main/scala/scalafix/internal/rule/Disable.scala#L138

Could you take a look @vovapolu ?

olafurpg:
Whoops, wrong ticket, sry 😅 Removing comment above

marcelocenerine:
The root cause is basically the same as in #693. The following test reproduces it:

```scala
/*
rules = Disable
Disable.symbols = [
  scala.collection.mutable
]
*/
package test.escapeHatch

import scala.collection.mutable

object OverlappingExpressionToggles {

  val map: mutable.Map[String, String] /* scalafix:ok */ = mutable.Map.empty /* scalafix:ok */
}

```

output:

```
scala/test/escapeHatch/OverlappingExpressionToggles.scala:13:60: error: [Disable.mutable]:
mutable is disabled
  val map: mutable.Map[String, String] /* scalafix:ok */ = mutable.Map.empty /* scalafix:ok */
                                                           ^
```

Both switches overlap and interfere with each other.

@fommil as a potential workaround, you can keep only the last `/* scalafix:ok */` as it disables the rules for the entire expression.

I'm working on a fix which should address both issues.

Issue #693, opened 2018-04-11
scalafix:ok disables scalafix:off


fommil:
anecdotally...

```scala
  // scalafix:off DisableSyntax.keywords.null,DisableSyntax.keywords.var,DisableSyntax.keywords.while
```

followed by some code like

```scala
    var value: String = null
    val decodedValue = mutable.Set /* scalafix:ok */ [String]()
```

will fail the check on the `null`. If everything is moved into the `scalafix:off` it is ok.

marcelocenerine:
this is probably caused by the same bug I fixed [here](https://github.com/scalacenter/scalafix/pull/620):  (search for "bugs"). See if it gets resolved with 0.6.0-M2

marcelocenerine:
ok, I found out what is going wrong here. The following test reproduces the issue:

```scala
/*
rules = DisableSyntax
DisableSyntax.keywords = [
  var
  while
  null
]
*/
package test.escapeHatch

object Foo {

  // scalafix:off DisableSyntax.keywords.null,DisableSyntax.keywords.var,DisableSyntax.keywords.while
  val x = null // scalafix:ok
  var y = null

  while (true) {
    println("foo")
  }
}
```

when you add `// scalafix:ok DisableSyntax`, scalafix disable the given rule before the expression starts and enables it back again at the end of the expression. However, if you don't specify any rule, scalafix will then disable all rules before the expression and (incorrectly) enables everything at the end of the expression. It should enable only the rules that were disabled before the expression.

olafurpg:
Nice job tracking down the root cause of this ticket. Would you be interested in taking on this issue @marcelocenerine ?

marcelocenerine:
@olafurpg yeah, definitely

olafurpg:
Fixed in #718

PR #718, opened 2018-05-01
Bug fix: 'scalafix:ok' must not re-enable rules already suppressed

marcelocenerine:
Fixes:
- #693: `scalafix:ok` (wildcard) incorrectly re-enables all rules after the end of its corresponding expression. This includes rules that were already suppressed before the expression (example [here](https://github.com/scalacenter/scalafix/issues/693#issuecomment-381344041)). With this fix, `scalafix:ok` now only re-enables rules suppressed by itself

- #694: a `scalafix:ok` comment is currently interpreted as two independent `off` and `on` instructions whose end offset is EOF. This causes unpredictable results when +2  expressions marked with `scalafix:ok` overlap (see example [here](https://github.com/scalacenter/scalafix/issues/694#issuecomment-384107792)). With this fix, `scalafix:ok` comments are now treated as a single `off` instruction whose end offset is the last position of the corresponding expression (this is similar to how `@SuppressWarnings` is handled).

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
@@ -1,6 +1,7 @@
 package scalafix
 package internal.patch
 
+import scala.annotation.tailrec
 import scala.collection.immutable.TreeMap
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
@@ -114,7 +115,7 @@ object EscapeHatch {
       matcher: FilterMatcher,
       cause: Position,
       startOffset: EscapeOffset,
-      endOffset: Option[EscapeOffset] = None) {
+      endOffset: Option[EscapeOffset]) {
     def matches(id: RuleName): Boolean =
       id.identifiers.exists(i => matcher.matches(i.value))
   }
@@ -259,29 +260,33 @@ object EscapeHatch {
     def isEnabled(
         ruleName: RuleName,
         position: Int): (Boolean, Option[EscapeFilter]) = {
-      var culprit = Option.empty[EscapeFilter]
-      val isDisabled = {
-        val disablesUpToPos =
-          disabling.to(EscapeOffset(position)).values.flatten
-        disablesUpToPos.exists { disableFilter =>
-          val isDisabled = disableFilter.matches(ruleName)
-          if (isDisabled) culprit = Some(disableFilter)
-
-          isDisabled && {
-            val enablesRange = enabling
-              .range(disableFilter.startOffset, EscapeOffset(position))
-              .values
-              .flatten
-            !enablesRange.exists { enableFilter =>
-              val isEnabled = enableFilter.matches(ruleName)
-              if (isEnabled) culprit = Some(enableFilter)
-              isEnabled
-            }
-          }
-        }
+      def findEnableInRange(from: EscapeOffset, to: EscapeOffset) = {
+        val enables = enabling.range(from, to).valuesIterator.flatten
+        enables.find(_.matches(ruleName))
       }
 
-      (!isDisabled, culprit)
+      @tailrec
+      def loop(
+          disables: List[EscapeFilter],
+          culprit: Option[EscapeFilter]): (Boolean, Option[EscapeFilter]) =
+        disables match {
+          case head :: tail if head.matches(ruleName) =>
+            head.endOffset match {
+              case Some(end) =>
+                if (position < end.offset) (false, Some(head))
+                else loop(tail, culprit)
+              case None => // EOF
+                val maybeEnable =
+                  findEnableInRange(head.startOffset, EscapeOffset(position))
+                if (!maybeEnable.isDefined) (false, Some(head))
+                else loop(tail, maybeEnable)
+            }
+          case _ :: tail => loop(tail, culprit)
+          case Nil => (true, culprit)
+        }
+
+      val disables = disabling.to(EscapeOffset(position)).valuesIterator.flatten
+      loop(disables.toList, None)
     }
 
     def disableEscapes: Iterable[EscapeFilter] = disabling.values.flatten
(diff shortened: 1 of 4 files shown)


scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorOverlapping.scala
```diff
@@ -0,0 +1,185 @@
...
+  // Test case: a 'on' should not re-enable a rule within the range of a 'ok' and
+  // it should be reported as unused.
+  // -----------------------------------------------------------------------------
+
+  {
+    /* scalafix:on */ // FIXME bug - this should be reported as unused
+    val dummy = 0
+    null
+  } // scalafix:ok
+
```
marcelocenerine: there is a bug in the logic that tracks unused `on` comments. In the test above, it should have reported the comment as unused. I have a fix for that already, but will submit it as part of a follow-up PR.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -108,23 +109,18 @@ object EscapeHatch {
     .warning("", "Unused Scalafix suppression. This can be removed")
     .withOwner(RuleName("UnusedScalafixSuppression"))
 
+  private type EscapeOffset = Int
```
olafurpg: Out of curiosity, is there a motivation besides saving allocations to make this a type alias? In general, I'm in favor of wrapper classes since they provide compile-time guarantees but I don't mind it much.

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorOverlapping.scala
olafurpg: Nice catch, looking forward to the PR!

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorWildcard.scala
```diff
@@ -6,18 +6,37 @@ rules = [
...
+  {
+    null
+    val dummy = 0
+  } // scalafix:ok
+
+  null // assert: NoNull
+  val eDummy = 0 // assert: NoDummy
+
+
+  // after a `scalafix:ok`, only rules that were previously enabled should be re-enabled
```
olafurpg: I had not thought of this situation 👍

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorOverlapping.scala
```diff
@@ -0,0 +1,185 @@
+/*
+rules = [
+  "class:scalafix.test.NoDummy"
+  "class:scalafix.test.NoNull"
+]
+*/
+package test.escapeHatch
+
+object AnchorOverlapping {
```
olafurpg: Can we add a test suite named `Issue694` to reproduce the bug from https://github.com/scalacenter/scalafix/issues/694 ?

```scala
  val map: mutable.Map /* scalafix:ok */ [String, Data] = mutable.Map.empty /* scalafix:ok */
```

That would be a good sanity check to make sure the bug is fixed.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: I thought it was a bit overkill because when we need to look up something in the `TreeMap`, first we need to wrap `Position.start|end` (which are `Int`s) into the case class, which adds some boilerplate:

```scala
escapeTree.to(EscapeOffset(position))
```

There is only one exception to that. Besides, `EscapeOffset` is private.

It is not a big deal though... I actually favor wrapper/value classes too, but didn't see much benefit for this particular case. I can revert the change if you prefer 😄

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorWildcard.scala
MasseGuillaume: haha ops good one.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
MasseGuillaume: @marcelocenerine I would prefer to revert the change. Making EscapeOffset a value class helped me a lot to solve the problem.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -302,38 +300,39 @@ object EscapeHatch {
...
+          end: Option[EscapeOffset],
           anchor: Token.Comment,
           rules: String): List[EscapeFilter] = {
-        val splittedRules = splitRules(rules)
+        val splitRules0 = splitRules(rules)
 
-        if (splittedRules.isEmpty) { // wildcard
-          List(EscapeFilter(FilterMatcher.matchEverything, anchor.pos, offset))
+        if (splitRules0.isEmpty) { // wildcard
+          EscapeFilter(FilterMatcher.matchEverything, anchor.pos, start, end) :: Nil
```
MasseGuillaume: I prefer `List(x)` over `x :: Nil` it makes it easier to see the return type.

marcelocenerine: 👍

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: ok, I will revert it

scalafix-tests/input/src/main/scala/test/escapeHatch/AnchorOverlapping.scala
marcelocenerine: done

scalafix-tests/input/src/main/scala/test/escapeHatch/Issue694.scala
```diff
@@ -0,0 +1,16 @@
...
+]
+*/
+package test.escapeHatch
+
+import scala.collection.mutable
+
+class Issue694 {
+
+  // overlapping switches must not interfere with each other. The one with wider range wins
+  val map: mutable.Map[String, String] /* scalafix:ok */ = // assert: UnusedScalafixSuppression
```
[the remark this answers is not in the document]
fommil: that's fine, great stuff!

Issue #622, opened 2018-02-21
Require all built-in rewrites to respect // scalafix: off


Mharlin:
When running semantic rules the exclusion comments like `// scalafix:off` or `scalafix:ok` are ignored and the rules are applied anyway.

On the positive side it tells you in the output that it has ignored the exclusion comments.

olafurpg:
Thank you for reporting! Can you please include which rules don't respect the exclusion comments? Currently, rules opt-into supporting this functionality by documenting which patches should atomically be applied together.

Mharlin:
The only rule that I have tried it for is MissingFinal.

olafurpg:
This ticket can be fixed in two steps

* [ ] use .atomic in all built-in scalafix rewrites (easy)
* [ ] update testkit to fail if a rewrite does not use `.atomic` (more advanced). This involved validating `patch` below contains `Atomic` nodes

https://github.com/scalacenter/scalafix/blob/dcaa5320321972aa99f0683176c6e759a83b5f8b/scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala#L82

olafurpg:
`MissingFinal` has now been moved to a separate repo and all scalafix rules have been updated to use `.atomic`. I'm not sure if it's a good idea anymore to assert in testkit that all patches are wrapped in `.atomic` so I think the current behavior is fine.

PR #720, opened 2018-05-01
Scalafix off fix

joan38:
Attempt on fixing https://github.com/scalacenter/scalafix/issues/622


scalafix-core/shared/src/main/scala/scalafix/internal/rule/MissingFinal.scala
```diff
@@ -31,11 +31,11 @@ final case class MissingFinal(index: SemanticdbIndex)
...
-        case Some(mod) => ctx.addLeft(mod, "final ")
+        case Some(mod) => ctx.addLeft(mod, "final ").atomic
         case None =>
           mods.lastOption match {
-            case Some(lastMod) => ctx.addRight(lastMod, " final")
-            case None => ctx.addLeft(d, "final ")
+            case Some(lastMod) => ctx.addRight(lastMod, " final").atomic
+            case None => ctx.addLeft(d, "final ").atomic
           }
       }).atomic
```
marcelocenerine: this rule already emits atomic patches (line 40). In order to avoid redundancy, I'd keep it as is or only your changes if you believe they make it clearer

scalafix-core/shared/src/main/scala/scalafix/internal/rule/LeakingImplicitClassVal.scala
```diff
@@ -22,7 +22,7 @@ case object LeakingImplicitClassVal extends Rule("LeakingImplicitClassVal") {
         val optPatch = for {
           anchorMod <- pMods.find(!_.is[Mod.Annot])
           if !pMods.exists(m => m.is[Mod.Private] || m.is[Mod.Protected])
-        } yield ctx.addLeft(anchorMod, "private ")
+        } yield ctx.addLeft(anchorMod, "private ").atomic
         optPatch.asPatch
 
     }.asPatch
```
marcelocenerine: could you add a test for this? see https://github.com/scalacenter/scalafix/blob/master/scalafix-tests/input/src/main/scala/test/MissingFinalRewrite.scala#L14 and https://github.com/scalacenter/scalafix/blob/master/scalafix-tests/output/src/main/scala/test/MissingFinalRewrite.scala#L11 as an example

marcelocenerine: I think we need to make similar change to `DottyVarArgPattern` and `DottyVolatileLazyVal`. Could you double check?

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
```diff
@@ -3,11 +3,13 @@ package testkit
 
 import scalafix.syntax._
 import scala.meta._
+
 import scalafix.internal.util.EagerInMemorySemanticdbIndex
 import org.scalameta.logger
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.FunSuite
 import org.scalatest.exceptions.TestFailedException
+import scalafix.patch.{AtomicPatch, Concat, LintPatch}
```
marcelocenerine: I guess these are no longer needed 😄

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
```diff
@@ -98,7 +100,7 @@ abstract class SemanticRuleSuite(
         .collectFirst { case f if f.isFile => f.readAllBytes }
         .map(new String(_))
         .getOrElse {
-          if (Patch.isOnlyLintMessages(patch)) obtained // linter
+          if (Patch.isAllAtomic(patch) && Patch.isOnlyLintMessages(patch)) obtained // linter
           else {
```
marcelocenerine: I'm not familiar with this area but it seems that we need to implement this validation somewhere outside of this IF. Otherwise tests would fail with `java.lang.RuntimeException: Missing expected output file for test ...`, which is not the correct reason (see build results)

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
joan38: Indeed

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
joan38: I'm not sure where should I put the check here @olafurpg pointed around here.
To me it sounds like I should just skip all the code after line 87, can anyone help on this?

marcelocenerine: Hey @joan38, sorry for not replying to your comment earlier.

I believe we can implement this validation after line 108. It seems fair to me to validate the output of rules after we know the test suite is properly set up (see validation in line 104). I think we can do something like:

```scala
if (!Patch.isAllAtomic(patch)) {
  sys.error(
    s"One or more rules used by test ${diffTest.filename} emit non-atomic " +
      "top level patches. To fix this, please use `.atomic` on individual or group " +
      "of patches that need to be atomically applied.")
}
```

This would fail a test when a non-atomic top level patch is detected. Please feel free to change the error message if you come up with something better.
It would be nice to print the exact rule name so that users could spot the faulty one straight away. However, I observed that when multiple rules are used by a given test, their names are concatenated with `+`, which wouldn't make the error message more helpful I suppose.

I hope it helps 😉

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
```diff
@@ -98,13 +99,19 @@ abstract class SemanticRuleSuite(
...
                  |$tried""".stripMargin)
           }
+          if (!Patch.isAllAtomic(patch))
+            sys.error(
+              s"One or more rules used by test ${diffTest.filename} emit non-atomic " +
+                "top level patches. To fix this, please use `.atomic` on individual or group " +
+                "of patches that need to be atomically applied.")
+
+          obtained // linter
         }
```
marcelocenerine: it needs to be "after line 108" (outside of the `.getOrElse` block and before the `val` below) 😉

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -198,6 +198,12 @@ object Patch {
     patch.isEmpty || hasLintMessage && onlyLint
   }
 
+  private[scalafix] def isAllAtomic(patch: Patch): Boolean = patch match {
+    case Concat(a, b) => isAllAtomic(a) && isAllAtomic(b)
+    case AtomicPatch(_) => true
+    case _ => false
+  }
+
```
marcelocenerine: I just realized that `isAllAtomic` would return false for `EmptyPatch`, `LintPatch`, ... which is probably not correct. In fact what we want to detect is the presence of top-level `TokenPatch`es which would cause rules not to respect the suppression comments/annotation. I think we can change it a bit to:

```scala
  private[scalafix] def isAllTokenPatchAtomic(patch: Patch): Boolean = patch match {
    case Concat(a, b) => isAllTokenPatchAtomic(a) && isAllTokenPatchAtomic(b)
    case _: TokenPatch => false
    case _ => true
  }
```

@olafurpg , what do you think?

joan38: Indeed reversing the cases should help

scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala
joan38: Indeed that doesn't make sense in the `.getOrElse` :)

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyKeywords.scala
```diff
@@ -14,8 +14,8 @@ case object DottyKeywords extends Rule("DottyKeywords") {
   override def fix(ctx: RuleCtx): Patch =
     ctx.tree.collect {
       case name @ Name("enum") =>
-        ctx.replaceTree(name, s"`enum`")
+        ctx.replaceTree(name, s"`enum`").atomic
       case name @ Name("inline") if !name.parents.exists(_.is[Mod.Annot]) =>
-        ctx.replaceTree(name, s"`inline`")
+        ctx.replaceTree(name, s"`inline`").atomic
     }.asPatch
```
joan38: @marcelocenerine Surely I can do the `.atomic`s right after `.asPatch` instead of in the match cases?

marcelocenerine: That's a very good question. The answer is 'no'. The single patch returned by `Rule.fix` groups all the rewrites for the entire source file. If we make it atomic, then all the rewrites for a given file would be rolled back despite users might wanted to suppress only one particular occurrence of code that triggers the rule.
Instead, what we need to do is to make sure that each occurrence of a rewrite is atomic. For example, if you wanted to rewrite usage of infix operators you would have to combine multiple patches, like:

input:
```scala
obj doSomething 1
```

expected output:
```scala
obj.doSomething(1)
```

rule (probably incomplete):
```
ctx.removeTokens(tokenList.trailingSpaces(op))
ctx.removeTokens(tokenList.leadingSpaces(op))
ctx.addLeft(op, ".")
ctx.addRight(op, "(")
ctx.addRight(args, ")")
```

In this case, if users wanted to suppress this rewrite for one particular occurrence of the infix operator in their code, scalafix should suppress all the 5 patches together.

We need to run the tests and check what rules are not emiting atomic top level patches for rewrites. Then for each rule we need to see what makes sense to be grouped in an atomic patch and put the `.atomic` call there.

marcelocenerine: @olafurpg btw, I think such rule 👆would be a good addition 😄

joan38:
@marcelocenerine Alright this looks like good now

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -185,7 +185,7 @@ final case class DisableSyntax(
   override def fix(ctx: RuleCtx): Patch = {
     val lints =
       (checkTree(ctx) ++ checkTokens(ctx) ++ checkRegex(ctx)).map(ctx.lint)
-    fixTree(ctx) ++ lints
+    (fixTree(ctx) ++ lints).atomic
```
olafurpg: This would disable the rule for a full file if a single `// scalafix:ok` exists in the file. `.atomic` should be targeted towards a specific fix, for example on line 181 above

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyKeywords.scala
olafurpg: What rule exactly do you have in mind @marcelocenerine ? I'm wondering if it makes sense to automatically insert `.atomic` for every value in `.asPatch`

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -198,6 +198,13 @@ object Patch {
     patch.isEmpty || hasLintMessage && onlyLint
   }
 
+  private[scalafix] def isAllTokenPatchAtomic(patch: Patch): Boolean =
+    patch match {
+      case Concat(a, b) => isAllTokenPatchAtomic(a) && isAllTokenPatchAtomic(b)
+      case _: TokenPatch => false
```
olafurpg:
```scala
case _: AtomicPatch|  _: LintPatch  => true
case _ => false
```
?

scalafix-tests/input/src/main/scala/test/ScalafixOk.scala
```diff
@@ -0,0 +1,10 @@
+/*
+rules = MissingFinal
+ */
+package test
+
+object ScalafixOk {
```
olafurpg: Let's name this test suite `MissingFinalAtomic`

scalafix-tests/unit/src/main/scala/scalafix/test/FqnRule.scala
```diff
@@ -7,23 +7,23 @@ import scalafix._
 case class FqnRule(index: SemanticdbIndex)
     extends SemanticRule(index, "FqnRule") {
   override def fix(ctx: RuleCtx): Patch =
-    ctx.addGlobalImport(importer"scala.collection.immutable")
+    ctx.addGlobalImport(importer"scala.collection.immutable").atomic
```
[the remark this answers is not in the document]
olafurpg: I would be careful to automatically infer atomic for import patches or rename. It's possible to imagine rules that rename AND add an import. More explicit is better IMO.

When I think of it, we should not change `.asPatch` either. Instead, I think it's better to add a helper method `.eachAtomic` (name up for debate) for `Iterable[Patch]` so that it's possible to do `tree.collect { ... }.eachAtomic.asPatch`.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/DottyKeywords.scala
marcelocenerine: @olafurpg it would be probably ok for most cases where `.asPatch` is used at the end of `.collect`. But maybe for some uncommon use cases it could be as dangerous as https://github.com/scalacenter/scalafix/pull/706#discussion_r182087220 since an individual `TokenPatch` would become "suppress-able". Do you think that is a good compromise? I was thinking the same for `RuleCtx.replaceTree` btw.

olafurpg: I noticed that `.asPatch` is used in several places in the scalafix repo where automatically inserting `.atomic` to each element would not make sense. I think it's best to force people to be explicit with `.map(_.atomic)` with potentially a `.mapAtomic` extension method for convenience.

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
marcelocenerine: @olafurpg it would return `false` for `EmptyPatch` and `TreePatch` (imports + symbols). Please see my comments: https://github.com/scalacenter/scalafix/pull/720#discussion_r186379813 / https://github.com/scalacenter/scalafix/pull/720#discussion_r185717259 . wdyt?

scalafix-tests/unit/src/main/scala/scalafix/test/FqnRule.scala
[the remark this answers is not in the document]
marcelocenerine: I guess we both did the review at the same time so our comments are all intertwined 😄

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
olafurpg: It should return `true` for `EmptyPatch` but `false` for `TreePatch`, Tree patches are intended to be combined into larger atomic units.

joan38:
```scala
case _: AtomicPatch |  _: LintPatch | _: EmptyPatch => true
case _ => false
```
?

joan38:
See comments answered.

scalafix-core/shared/src/main/scala/scalafix/patch/Patch.scala
olafurpg: Correct @joan38

scalafix-tests/unit/src/main/scala/scalafix/test/FqnRule.scala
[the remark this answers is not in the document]
joan38: I didn't quiet get this, should I make it atomic or not? If I don't then we'll getting our error `One or more rules used by test scala/test/FqnRule.scala emit non-atomic top level patches`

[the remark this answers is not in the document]
joan38: I made the change

joan38:
I'm having difficulties to understand where is the missing atomic.
I've tried to add some arround ScalafixMetaconfigReaders.scala but it's not good.

olafurpg:
@joan38 what happens if you put `.atomic` inside `classloadRuleDecoder` before calling `Rule.constant`?

joan38:
This is what I did, see https://github.com/scalacenter/scalafix/pull/720/commits/34614d694bc3cb08f412a2eab6f9e102fb1e3c70#diff-6a2f8c431d97adfc75396e1073f145b1R159

joan38:
Guys I didn't loose faith in this one. I need some help on understanding where is the missing `.atomic`

olafurpg:
Hey @joan38 no worries, I fetched your branch and rebase on top of master here https://github.com/olafurpg/scalafix/tree/scalafix-off It does not compile still since I'm not 100% sure where is the best location to add the Patch assertion. The current handling of `suppress: Boolean` feels unsatisfying, it seems we can do abstract over both auto-suppression and asserting all patches have `.atomic`. I will think about it more and get back to you 😉

joan38:
Thanks man. Let me know.

Issue #721, opened 2018-05-02
Rewrite that replace Some() constructor by Option() constructor


joan38:
```
val a = Some(hey)
```
should be rewrote:
```
val a = Option(hey)
```
Since the `Some()` constructor is not safe.

I'm happy to take a shot to this one when I will have time.

olafurpg:
This is a good idea! I'll copy-paste my review from #722 

I am a bit on the fence to include this rule into the core. In the long term, I envision that domain specific rules like this rule will live as 3rd party modules and the core contains only the most generic rules like ExplicitResulTypes and OrganizeImports (once it's implemented).

I admit the process has been unclear so far on what rules are added to core repo and which ones are not. I have probably been too relaxed in the past and merged more than I should have, which is not sustainable.

I'm sorry, but I recommend you host this rule in a separate repo using the g8 template https://scalacenter.github.io/scalafix/docs/rule-authors/setup#scalacenterscalafixg8 I know it's more inconvenient to share/run such rules than have them builtin but I hope to make that workflow more ergonomic in the future.

PR #722, opened 2018-05-02
Rewrite rule for Some() to Option()

joan38:
For https://github.com/scalacenter/scalafix/issues/721

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
@@ -32,7 +32,7 @@ case class DisableSyntaxConfig(
     noValInAbstract: Boolean = false,
     @Description("Report error on object with `implicit` modifier.")
     noImplicitObject: Boolean = false,
-    @Description("Report error on method that define an implicit conversion. ")
+    @Description("Report error on method that define an implicit conversion.")
     noImplicitConversion: Boolean = false,
     @Description(
       "Remove `final` modifier for val definitions, " +
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoSomeConstructor.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/NoSomeConstructor.scala
@@ -1,0 +1,25 @@
+package scalafix.internal.rule
+
+import scala.meta._
+
+import scalafix.{Patch, SemanticRule, SemanticdbIndex}
+import scalafix.rule.{RuleCtx, RuleName}
+import scalafix.util.SymbolMatcher
+
+final case class NoSomeConstructor(index: SemanticdbIndex)
+    extends SemanticRule(index, RuleName("NoSomeConstructor")) {
+
+  override def description: String =
+    "Rewrite that replaces Some() constructor by Option() constructor"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    val someConstructor = SymbolMatcher.normalized(
+      Symbol("_root_.scala.Some.")
+    )
+
+    ctx.tree.collect {
+      case Term.Apply(someConstructor(fun), _) =>
+        ctx.replaceTree(fun, "Option").atomic
+    }.asPatch
+  }
+}
--- a/scalafix-core/shared/src/main/scala/scalafix/rule/ScalafixRules.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/rule/ScalafixRules.scala
@@ -23,7 +23,8 @@ object ScalafixRules {
     RemoveUnusedTerms(index),
     NoAutoTupling(index),
     Disable(index, DisableConfig.default),
-    MissingFinal(index)
+    MissingFinal(index),
+    NoSomeConstructor(index)
   )
   def all(index: SemanticdbIndex): List[Rule] =
     syntax ++ semantic(index)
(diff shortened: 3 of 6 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
```diff
@@ -0,0 +1,28 @@
...
+      )
+    ) {
+
+  override def description: String =
+    "Rewrite that replaces Some() constructor by Option() constructor"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case b @ q"Some($a)" =>
+        ctx.replaceTree(b, q"Option($name)").atomic // Any way of replacing entire Trees?
```
joan38: Help needed here

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
```diff
@@ -0,0 +1,28 @@
...
+        "0.6.0"
+      )
+    ) {
+
+  override def description: String =
+    "Rewrite that replaces Some() constructor by Option() constructor"
+
+  override def fix(ctx: RuleCtx): Patch = {
+    ctx.tree.collect {
+      case b @ q"Some($a)" =>
```
gabro: This will match `Some` syntactically, which is not safe since we're not sure it's `scala.Some`.

This has to be a `SemanticRule` and use `index.symbol(tree)` to query for the resolved symbol and check that is `scala.Some`.

A good test for this is 

```scala
import scala.{Some => MySome}

MySome(42) // the rule must be able to replace this with Option(42)
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
```diff
@@ -0,0 +1,28 @@
...
+import scalafix.Patch
+import scalafix.rule.{Rule, RuleCtx, RuleName}
+import scalafix.util.Trivia
+
+case object OptionConstructor
+    extends Rule(
+      RuleName("OptionConstructor").withDeprecatedName(
+        "OptionConstructor",
+        "Use the Option constructor instead",
+        "0.6.0"
```
gabro: this looks wrong, `withDeprecatedName` is used for rules that have been renamed. You don't need this since this is a new rule.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -12,7 +12,7 @@ case object ProcedureSyntax
     extends Rule(
       RuleName("ProcedureSyntax").withDeprecatedName(
         "ExplicitUnit",
-        "Use ProcedureSyntax instead",
+        "Use explicit : Unit = instead",
```
gabro: same as above, this `withDeprecatedName` is used to provide alternate deprecated names for rules.

joan38:
This works now thanks for the guidance.
Let me know if this is done the proper way tho.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
```diff
@@ -0,0 +1,23 @@
...
+import scalafix.rule.{RuleCtx, RuleName}
+
+final case class OptionConstructor(index: SemanticdbIndex)
+    extends SemanticRule(index, RuleName("OptionConstructor")) {
+
+  override def description: String =
+    "Rewrite that replaces Some() constructor by Option() constructor"
+
+  private def isSome(term: Term) =
+    index.symbol(term).contains(Symbol("_root_.scala.Some."))
```
gabro: I would advise to use a SymbolMatcher here, which defines an `unapply` that can be used in the pattern match below. Here's an example to use as a reference:

https://github.com/typelevel/cats/blob/49374a4137823ec82782b17fac8aafcea1979d96/scalafix/rules/src/main/scala/fix/Cats_v1_0_0.scala#L143-L145

You would then use it as

```scala
  val someConstructor = //... SymbolMatcher
   
  // ...

  case Term.Apply(some(fun), _) => // ... replace fun
```

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
```diff
@@ -0,0 +1,23 @@
+package scalafix.internal.rule
+
+import scala.meta._
+
+import scalafix.{Patch, SemanticRule, SemanticdbIndex}
+import scalafix.rule.{RuleCtx, RuleName}
+
+final case class OptionConstructor(index: SemanticdbIndex)
```
gabro: Minor, but I think something like `NoSomeConstructor` or `PreferOptionConstructor` would be more inline with other rules names.

/cc @olafurpg

scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala
```diff
@@ -32,8 +32,10 @@ case class DisableSyntaxConfig(
     noValInAbstract: Boolean = false,
     @Description("Report error on object with `implicit` modifier.")
     noImplicitObject: Boolean = false,
-    @Description("Report error on method that define an implicit conversion. ")
+    @Description("Report error on method that define an implicit conversion.")
     noImplicitConversion: Boolean = false,
+    @Description("Report error on usage of the constructor Some.")
+    noSome: Boolean = false,
```
gabro: I think this is spurious. You're adding a config option to the `DisableSyntax` rule, but this then goes unused by the rule itself.

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
joan38: That's way much better.
Little question: Do you know what the dot at the end is? Ex: `_root_.scala.Some.`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
joan38: This will make it much more consistant with the `DisableSyntaxConfig.noSomeConstructor`

scalafix-core/shared/src/main/scala/scalafix/internal/rule/OptionConstructor.scala
gabro: IIRC It's to distinguish the type from the object (# vs .)

joan38:
https://twitter.com/JoanG38/status/991975906334924800 😃

Issue #626, opened 2018-02-21
RemoveUnusedTerms does not work on class/object


MasseGuillaume:
RemoveUnusedTerms does not produce a diff on stats:

```scala
class A {
  private val a = 1
}
object B {
  private val a = 1
}
package object C {
  private val a = 1
}
trait A {
  private val a = 1
}
```

joan38:
This can be kept open since #723 didn't solve this issue.
The issue is in SemanticDB not forwarding the compiler warning message.
@olafurpg Do you have an idea of how we can fix this?

olafurpg:
We can open a ticket in scalameta/scalameta with the reproduction in https://github.com/scalacenter/scalafix/pull/723#issuecomment-387017310

I'm sorry I won't have time to investigate this much in the coming weeks as I need to prioritize #704 and it seems it will require a fairly deep refactoring.

joan38:
Don't worry I'll open the issue on Scalameta and we can see later.

kubukoz:
The unused val warning is actually there for the last case, btw

joan38:
This should be solved with the latest Scalameta

gabro:
So, I've tried to tackle this with Scalameta 4 and here's what I found.

Given the 4 cases from the OP, the trait one works correctly in Scala 2.12, whereas the other 3 are still not working with the current implementation of `RemoveUnused` and none of them is working in Scala 2.11.

This is due to the positions of the warning. In Scala 2.12:

https://github.com/scalameta/scalameta/blob/7cdcba551547d72a4ac7ec3063e86ef2c861d574/tests/jvm/src/test/scala-2.12/scala/meta/tests/semanticdb/TargetedSuite2_12.scala#L21-L25

the reported position is precisely on the identifier being defined, except for the trait case, where the warning spans on the whole statement.

gabro:
Well, I've tried to work around it in #889

PR #723, opened 2018-05-02
Fix RemoveUnusedTerms for private defs

joan38:
Attempt to fix https://github.com/scalacenter/scalafix/issues/626

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
@@ -13,10 +13,9 @@ case class RemoveUnusedTerms(index: SemanticdbIndex)
     "Rewrite that removes unused locals or privates by -Ywarn-unused:locals,privates"
 
   private val unusedTerms = {
-    val UnusedLocalVal = """local (.*) is never used""".r
-    index.messages.toIterator.collect {
-      case Message(pos, _, UnusedLocalVal(_*)) =>
-        pos
+    val UnusedPrivateLocalVal = """(local|private) (.*) is never used""".r
+    index.messages.collect {
+      case Message(pos, _, UnusedPrivateLocalVal(_*)) => pos
     }.toSet
   }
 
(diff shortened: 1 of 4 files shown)


joan38:
Sure will do that this evening

olafurpg:
Thanks for this contribution @joan38 ! I am busy organizing a spree today, will try to take a closer look later this week.

joan38:
No worries @olafurpg I know you are usually very quick ;)

@gabro I've added some tests.

gabro:
@joan38 thanks! They seem to be failing though, can you take a look?

joan38:
Super weird indeed the regex doesn't match on messages like:
```
private method f in object RemoveUnusedTerms is never used
```

joan38:
I don't seem to be able to get the message for the `private val` from `index.messages`.
All though it works fine for the `private def`s.
Do you have any idea what's going on here? Is it a SementicDB issue?

gabro:
Is the message there? Can you print `index.messages` and see how it looks like?

joan38:
I've tried to:
```
index.messages.collect {
  case m @ Message(pos, _, UnusedPrivateLocalVal(_*)) =>
    println("m: " + m)
    pos
}.toSet
```
and then ctrl+f `private val ff` in there but can't find it.

joan38:
@olafurpg Any idea on the `index.messages` not reporting the warning for private vals?

olafurpg:
I am able to reproduce that the message for the private val is not included in the SemanticDB
```
➜  target git:(v1) cat unused.scala
object a {
  private val unused = 1
  private def unusedDef = 1
  def app() = println(1)
}
➜  target git:(v1) metac -Ywarn-unused:privates unused.scala
     unused.scala:2: warning: private val unused in object a is never used
  private val unused = 1
              ^
unused.scala:3: warning: private method unusedDef in object a is never used
  private def unusedDef = 1
              ^
two warnings found
➜  target git:(v1) metap META-INF/semanticdb/unused.scala.semanticdb
unused.scala
------------

...
Diagnostics:
[2:2..2:27)[warning] private method unusedDef in object a is never used
```

It could be that the message is reported in a later compile phase. I recommend we limit this PR to unused defs and follow up with a ticket in scalameta/scalameta :)

joan38:
Done. Only private vals are missing. I left the regex modification for later when the semantic DB will be fixed.

joan38:
@olafurpg @gabro Ready to go with this?

joan38:
Cheers @olafurpg

Issue #716, opened 2018-04-23
Syntactic rule to ban non-tuple val patterns


fommil:
```scala
val Right(noproblemo) = ... thing that returns `Either`...
```

is a bang at runtime waiting to happen. Should be banned.

Tuples should be special cased and allowed. The general case that should be allowed is a **total** pattern match, but there is no way to know that in scala.

olafurpg:
This is a great idea! I think this can be part of the `DisableSyntax` rule under option `DisableSyntax.noValPatterns: Boolean`.

The configuration option can be added here 
https://github.com/scalacenter/scalafix/blob/6d48e0c419a63956274cfcc8e9bb23601c75ccfa/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableSyntaxConfig.scala#L18

and the check can be added somewhere around here
https://github.com/scalacenter/scalafix/blob/6d48e0c419a63956274cfcc8e9bb23601c75ccfa/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala#L57

A good place to test this would be here https://github.com/scalacenter/scalafix/blob/6d48e0c419a63956274cfcc8e9bb23601c75ccfa/scalafix-tests/input/src/main/scala/test/DisableSyntax.scala

mcanlas:
What do you mean by "total pattern match"?

fommil:
total in the sense of a total function, as opposed to a partial function.

Actually we can do better than just tuples in the "allowed" list... we can also allow synthetic case class unapply since they are also total. We may be able to detect this by seeing if the `unapply` returns a `Some` (as opposed to some other structure like `Option` or a structural thing).

olafurpg:
If you want more sophisticated type analysis then this will require more expertise. I think a syntactic rule that prohibits any non-tuple pattern in vals would be useful in itself

olafurpg:
I updated the PR to be about adding a syntactic rule. The more advanced case which works on types boils down to implementing exhaustivity checking, which is a research topic in itself and is already implemented in the compiler. I'm sure you can get those warnings in the compiler by writing a compiler plugin like https://github.com/oleg-py/better-monadic-for that removed the synthetic `@unchecked` annotation for val desctructuring

```scala
➜  target git:(v1) cat foo.scala
object a {
  val List(1) = List(3)
}
➜  target git:(v1) scalac -Xprint:typer foo.scala
[[syntax trees at end of                     typer]] // foo.scala
[...]

PR #725, opened 2018-05-02
Add syntactic rule for matching on assignment #716

perok:
Done on the Open Source Spree for flatMap(Oslo) 2018.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
@@ -90,10 +90,26 @@ final case class DisableSyntax(
       }
     }
 
+    object NoValPatterns {
+      def unapply(t: Tree): Option[Tree] = t match {
+        case v: Defn.Val =>
+          v.pats.find(prohibited)
+        case _ => None
+      }
+
+      def prohibited(v: Pat): Boolean = v match {
+        case _: Pat.Tuple => false
+        case _: Pat.Var => false
+        case _ => true
+      }
+    }
+
     def hasNonImplicitParam(d: Defn.Def): Boolean =
       d.paramss.exists(_.exists(_.mods.forall(!_.is[Mod.Implicit])))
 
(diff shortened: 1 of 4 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -90,10 +90,33 @@ final case class DisableSyntax(
...
     def hasNonImplicitParam(d: Defn.Def): Boolean =
       d.paramss.exists(_.exists(_.mods.forall(!_.is[Mod.Implicit])))
 
     val DefaultMatcher: PartialFunction[Tree, Seq[LintMessage]] = {
+      case NoValPatterns(v) if config.noValPatterns =>
+        Seq(
+          errorCategory
+            .copy(id = "noValPatterns")
+            .at(
+              "Pattern matching in val assignment is not guaranteed to be total.",
```
olafurpg: > Pattern matching in val assignment can result in match error, use "syntax of code using `x match { ... }` instead".

?

scalafix-tests/input/src/main/scala/test/disableSyntax/DisableSyntaxNoValPatterns.scala
```diff
@@ -0,0 +1,14 @@
+/*
+rules = DisableSyntax
+DisableSyntax.noValPatterns = true
+*/
+package test.disableSyntax
+
+case object DisableSyntaxNoValPatterns {
+  val Right(notFound) = 1.asInstanceOf[Either[String, String]] // assert: DisableSyntax.noValPatterns
```
olafurpg: Can we add a test case in the basic DisableSyntax to check that this error is not reported unless `DisableSyntax.noValPatterns = true`?

perok: I have added the same line of code in DisableSyntaxBase.

olafurpg:
Great catch with `Defn.Var` 🔥

joan38:
Great one @perok!

PR #727, opened 2018-05-02
Act on "pattern is never used" warning from scalac, fixes #555

### LINKED ISSUE #555: "Pattern var is never used" autofix
I'm not sure about potential edge cases, but I think this is a pretty safe rewrite:

From:

```scala
opt match {
  case Some(x) => 42
//          ^
//          pattern var x in value $anonfun is never used; `x@_' suppresses this warning
}
```

to

```scala
opt match {
  case Some(_) => 42
}
```

Similarly:

```scala
try {
  whatever()
} catch {
  case e: SomeException => println("boom!")
}
```

to

```scala
try {
  whatever()
} catch {
  case _: SomeException => println("boom!")
}
```

Maybe this could be an optional feature of `RemoveUnused`

lucav76:


scalafix-core/shared/src/main/scala/scalafix/internal/rule/RemoveUnusedTerms.scala
```diff
@@ -37,9 +48,15 @@ case class RemoveUnusedTerms(index: SemanticdbIndex)
     case _ => None
   }
 
+  private def addBinding(term: Pat.Var): Option[Tokens] = term match {
```
olafurpg: Remove?

lucav76:
Shouldn't we try to handle both 2.11 and 2.12? If it is just a matter of supporting two patterns.

olafurpg:
It would be great to support 2.11 as well if possible

olafurpg:
Closing due to inactivity and to avoid conflicts with #737, I may pick this up again to complete the 2.11 support.

PR #731, opened 2018-05-07
Fix: EscapeHatch is not reporting unused 'scalafix:on' in some cases

marcelocenerine:
- fix missing cases where `scalafix:on` is not reported as unused:
  - sequence of multiple `scalafix:on`
  - after a rule is turned OFF and then ON again
- report unused rule when it is wrapped by triple quotes in `@SuppressWarnings`
- minor refactorings in `EscapeHatch` to save some lines of code

Each change is in a separate commit.

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
@@ -82,21 +82,9 @@ class EscapeHatch private (
 
     val patches =
       patchesByName.map { case (name, patch) => loop(name, patch) }.asPatch
-
-    // we don't want to show unused warnings for non-Scalafix rules
-    val prefixedEscapes =
-      annotatedEscapes.disableEscapes
-        .filter { escape =>
-          val ruleName = escape.cause.text
-          ruleName.startsWith(AnnotatedEscapes.OptionalRulePrefix)
-        }
-    val unusedDisable =
-      (prefixedEscapes ++ anchoredEscapes.disableEscapes)
-        .filterNot(usedEscapes)
-        .map(_.cause)
-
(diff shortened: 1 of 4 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -300,53 +289,39 @@ object EscapeHatch {
...
           }
         }
       }
 
-      def splitRules(rules: String): List[String] =
-        rules.trim.split("\\s*,\\s*").toList
+      def splitRules(rules: String): List[String] = {
+        val trimmed = rules.trim
+        if (trimmed.isEmpty) Nil else trimmed.split("\\s*,\\s*").toList
+      }
```
marcelocenerine: this method had a bug that was accidentally making the unused reporting logic work for some cases.

scalafix-core/shared/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -357,28 +332,6 @@ object EscapeHatch {
...
-              case (rule, pos) if !disabledRules(rule) =>
-                unusedAnchoredEnable += pos
-              case _ => ()
-            }
-          }
-        } else {
-          currentlyDisabledRules ++= rulesToPos
-        }
-      }
-
```
marcelocenerine: the main change is here. `trackUnusedRules` is being replaced by `OnOffTracker` (see below).

In the old code once a rule is added to `currentlyDisabledRules`, it is never removed from it. This prevents potential unused `scalafix:on` comments further down the file from being reported as unused.

marcelocenerine:
No worries @olafurpg. This fix is not too important and can wait 😉

Issue #702, opened 2018-04-13
allow descriptions in scalafix:ok


fommil:
I'd like to allow user-provided descriptions of why a rule is overridden. e.g.

```
    URLEncoder.encode(s, "UTF-8") // scalafix:ok UTF-8 always succeeds
```

but this makes the suppression be ignored

marcelocenerine:
yeah, this is because you can specify the exact rules you want to suppress following `scalafix:ok`. If it is followed by nothing then all rules are suppressed (wildcard). If you specify something, scalafix will parse it and disable the rules that there is a match.

At the moment you can add descriptions as long as you use the following format:

```
URLEncoder.encode(s, "UTF-8") /* scalafix:ok */ // UTF-8 always succeeds
```

fommil:
I wonder if perhaps some marker, like another `//`, or a semicolon should allow arbitrary comments

marcelocenerine:
yeah, could be

olafurpg:
I think this is a good idea, do you have any suggestion how this can be supported alongside the syntax to target a single rule? For example, `replace(",.*", "")` could support something like

```scala
var x: String = null // scalafix:ok DisableSyntax.var, I need mutability
var x: String = null // scalafix:ok, I need mutability
```

fommil:
comma is confusing because it is overloaded with the meaning in the `scalafix:off` syntax

olafurpg:
do you have any suggestion how this can be supported alongside the syntax to target a single rule?

fommil:
colon or semicolon instead of comma would be good

olafurpg:
Can you give an example how that would look like?

fommil:
```
 URLEncoder.encode(s, "UTF-8") // scalafix:ok; UTF-8 always succeeds
```

marcelocenerine:
closed via #733

PR #733, opened 2018-05-20
Allow arbitrary description in scalafix:on|off|ok comments

marcelocenerine:
Addressing #702

/cc @fommil


website/src/main/tut/docs/users/configuration.md
```diff
@@ -202,6 +202,17 @@ foo(null)
 List(1, "") // scalafix:ok NoInfer.any
 ```
 
+Optionally, you can include an arbitrary description at the end of any suppression comment, following a semicolon. The
```
olafurpg: 👍

PR #737, opened 2018-05-24
Clean up internals 

olafurpg:
This PR does a massive cleanup of Scalafix internals. The main objective with this refactoring is to

1. allow existing rules to continue to work mostly unchanged. Critically, *existing rules will continue to work with only a few minor changes** (see below).
2. enable us to create a new semantic api that reflects the lessons learned in the past year with SemanticDB. The new API is 
    - closer to the SemanticDB layer, avoiding leaky abstractions
    - lower memory + cpu overhead

Roughly, this PR makes the following changes:

- load one SemanticdbIndex per source file instead of on SemanticdbIndex globally. This should significantly reduce memory usage for larger projects.
- new package `scalafix.v1` that is a clean slate implementation of the rule + semantic APIs.
- new package `scalafix.v0` that contains the current scalafix rule API + inlined legacy APIs from scala.meta such as `scala.meta.{Database,Document,Synthetic,ResolvedName,ResolvedSymbol,...}`.
- new `scalafix.v1.RuleDecoder` utility to go from string to a `scalafix.v1.Rule`. Previously this logic to load rules from configuration was spread through multiple files+modules  and was tricky to understand.
- re-implementation of the command-line interface. The old cli had gone through ~4-5 different generations of the semantic API, where each refactoring made it slightly more difficult to understand the code. Instead of adapting the old cli to the new "one SemanticIndex per source file" architecture adding further technical debt, this PR rewrote the cli from scratch.

I spent several days to break this cleanup into smaller chunks before I gave up and decided to bundle everything in one PR. Observe however that there are only trivial changes in the rule implementations (mostly imports). The required steps to make an existing rule compile with the new API are:

* Remove any imports from the following packages: `scala.meta`, `org.langmeta`, `scalafix`
* Add `import scalafix.v0._`

This PR is still a work in progress. TODOs

- [x] error on stale semanticdb errors, seems like there were no unit tests for this functionality
- [x] update sbt-scalafix
- [x] handle `file:Path.scala`
- [x] remove dead code
- [x] --help
- [x] --zsh
- [x] --bash
- [x] move as much as possible to scalafix.internal
- [x] remove dead code

### DIFF
--- a/scalafix-sbt/src/test/scala/scalafix/internal/sbt/SbtCompletionsSuite.scala
+++ b/scalafix-sbt/src/test/scala/scalafix/internal/sbt/SbtCompletionsSuite.scala
@@ -100,30 +100,22 @@ class SbtCompletionsSuite extends FunSuite {
     val obtained = completions.map(_.display).toSet
     val expected = Set(
       "",
-      "--bash",
       "--classpath",
-      "--classpath-auto-roots",
       "--config",
-      "--config-str",
       "--diff",
       "--diff-base",
       "--exclude",
       "--help",
-      "--no-strict-semanticdb",
       "--non-interactive",
       "--out-from",
       "--out-to",
-      "--quiet-parse-errors",
       "--rules",
-      "--single-thread",
       "--sourceroot",
       "--stdout",
       "--test",
       "--tool-classpath",
-      "--usage",
       "--verbose",
-      "--version",
-      "--zsh"
+      "--version"
     )
 
     assert((expected -- obtained).isEmpty)
@@ -137,10 +129,6 @@ class SbtCompletionsSuite extends FunSuite {
     assert(appends.contains(":"))
   }
   check2("--classpath bar:", SkipWindows) { (appends, displays) =>
-    assert(displays.contains("foo"))
-    assert(appends.contains("foo"))
-  }
-  check2("--classpath-auto-roots" + space, SkipWindows) { (appends, displays) =>
     assert(displays.contains("foo"))
     assert(appends.contains("foo"))
   }


scalafix-core/shared/src/main/scala/scalafix/v0/Database.scala
```diff
@@ -0,0 +1,27 @@
+package scalafix.v0
+
+import scala.compat.Platform.EOL
+import scala.meta.internal.io.PathIO
+
+final case class Database(documents: Seq[Document]) {
```
xeno-by: This is copy/pasted from Scalameta 3.x, right? If that's the case, it would be a good idea to update NOTICE.md.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
```diff
@@ -0,0 +1,149 @@
...
+      (props & property.value) != 0
+  }
+
+  final class Accessibility private[Sym] (a: s.Accessibility) {
+    def isPrivate: Boolean = a.tag.isPrivate
+    def isPrivateThis: Boolean = a.tag.isPrivateThis
+    def isProtected: Boolean = a.tag.isProtected
+    def isProtectedThis: Boolean = a.tag.isProtectedThis
+    def isPublic: Boolean = a.tag.isPublic
+    def within: Sym = new Sym(a.symbol)
```
xeno-by: How does one check if something is `private[foo]` in this API?

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
```diff
@@ -0,0 +1,149 @@
+package scalafix.v1
+
+import scala.meta.Tree
+import scala.meta.internal.{semanticdb3 => s}
+import scala.meta.internal.semanticdb3.SymbolInformation.{Property => p}
+import scala.meta.internal.semanticdb3.Scala._
+
+final class Sym private (val value: String) {
```
xeno-by: Can we move `scala.meta.Symbol` to `v0` and rename this to `v1.Symbol`?

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
```diff
@@ -0,0 +1,149 @@
...
+    if (sym.isEmpty) Sym.None
+    else {
+      if (!sym.startsWith("local")) {
+        sym.desc // assert that it parses as a symbol
+      }
+      new Sym(sym)
+    }
+  }
+
+  object Root {
```
xeno-by: RootPackage?

xeno-by: Also, is there an extractor for EmptyPackage?

scalafix-core/shared/src/main/scala/scalafix/v1/Doc.scala
```diff
@@ -0,0 +1,59 @@
+package scalafix.v1
```
xeno-by: Am I right in thinking that scalafix.v0 is still the default? Or do you plan to flip the switch to use scalafix.v1 in 0.6-final?

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
```diff
@@ -0,0 +1,144 @@
...
+import scala.meta.contrib.AssociatedComments
+import scalafix.internal.util.SymbolTable
+import scalafix.util.MatchingParens
+import scalafix.util.TokenList
+import scala.meta.internal.{semanticdb3 => s}
+import scalafix.internal.patch.DeprecatedSemanticdbIndex
+import scalafix.internal.v1._
+import scalafix.util.SemanticdbIndex
+
+trait SemanticContext
```
xeno-by: Why is there a need in a trait without any methods?

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
```diff
@@ -0,0 +1,144 @@
...
+  // Semantic API
+  // ============
+  def symbol(tree: Tree): Sym = {
+    val result = symbols(TreePos.symbol(tree))
+    if (!result.hasNext) Sym.None
+    else result.next() // Discard multi symbols
+  }
+
+  def info(tree: Tree): Sym.Info = info(symbol(tree))
+  def info(sym: Sym): Sym.Info = {
```
xeno-by: How about just `Info`? That'll save some typing, seemingly without the loss of modularity.

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
```diff
@@ -0,0 +1,144 @@
...
+  // ============
+  // Semantic API
+  // ============
+  def symbol(tree: Tree): Sym = {
+    val result = symbols(TreePos.symbol(tree))
+    if (!result.hasNext) Sym.None
+    else result.next() // Discard multi symbols
+  }
+
+  def info(tree: Tree): Sym.Info = info(symbol(tree))
```
xeno-by: How much do we need this method? Strictly speaking, a tree doesn't have an info, so it was always a bit strange to me that there is a method like that. In comparison, in scalac, every tree does have a tpe, so `Tree.tpe` is appropriate, imo.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
```diff
@@ -0,0 +1,149 @@
...
+    def kind: Kind = new Kind(info)
+    def props: Properties = new Properties(info.properties)
+    def access: Accessibility =
+      new Accessibility(info.accessibility.getOrElse(s.Accessibility()))
+
+    override def toString: String = s"Sym.Info(${info.symbol})"
+
+    // privates
+    private[scalafix] def tpe: s.Type =
+      info.tpe.getOrElse(s.Type())
```
xeno-by: Do you plan to eventually expose `tpe` to Scalafix rules?

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
```diff
@@ -0,0 +1,149 @@
...
+    def isTrait: Boolean = k.isTrait
+    def isMethod: Boolean = k.isMethod
+    def isField: Boolean = k.isField
+    def isMacro: Boolean = k.isMacro
+    def isConstructor: Boolean = k.isConstructor
+    def isType: Boolean = k.isType
+    def isParameter: Boolean = k.isParameter
+    def isTypeParameter: Boolean = k.isTypeParameter
+    def isPackage: Boolean = k.isPackage
+    def isPackageObject: Boolean = k.isPackageObject
```
xeno-by: Looks like there are a few SemanticDB kinds missing from this API. Is this intentional?

scalafix-core/shared/src/main/scala/scalafix/v0/Database.scala
olafurpg: Good idea, done.

scalafix-core/shared/src/main/scala/scalafix/v0/Flags.scala
```diff
@@ -0,0 +1,47 @@
+package scalafix.v0
+
+trait Flags {
+  final val VAL: Long = 1L << 0
+  final val VAR: Long = 1L << 1
+  @deprecated("Use `METHOD` instead.", "3.6.0")
```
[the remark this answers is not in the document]
olafurpg: Removed deprecated members.

scalafix-core/shared/src/main/scala/scalafix/v1/Doc.scala
olafurpg: This PR makes `scalafix.v1` the default, the hard work was making it so seamless that 0 changes are required for existing rules that use `scalafix.v0`.

The cli and testkit support rules written against both v1 and v0.

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
olafurpg: Semantic patches like `Patch.replaceSymbols` take an implicit evidence that you have access to semantic APIs. It has no methods since it's extended by both v1.SemanticDoc and v0.SemanticdbIndex and used in `Patch._` methods. Moved to separate file and added docstring.

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
olafurpg: I agree, this method should not exist. Removed

(PS. because no code is written against v1 yet we can tweak it to our desires 😄 )

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
olafurpg: Can you elaborate? You mean `Info` instead of `Sym.Info`?

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
olafurpg: My name is @olafurpg and I approve this message 👍

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
olafurpg: Removed in favor of 
```
case `RootPackage` =>
```

since there is already a `val RootPackage = ...`

olafurpg: No need, only `Local` and `Global` extractors.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
olafurpg: Eventually yes once we figure out a more ergonomic API for it. Maybe we have classes that wrap it just like SymbolInformation 🤔

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
olafurpg: Good catch, updated to match 100% with semanticdb3.proto (including order).

Only missing Kind.LOCAL which is going away.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
olafurpg: I think `isPrivate && within == FooSymbol`? But haven't had any time to test the ergonomics of this new API

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
xeno-by: Yes.

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticContext.scala
```diff
@@ -0,0 +1,7 @@
+package scalafix.v1
+
+/** An implicit instance of SemanticContext implies the call-site has access to semantic APIs.
+  *
+  * Useful to guard users from calling semantic operations without access to semantic APIs.
```
xeno-by: Oh. This explains why we need an empty trait.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
xeno-by: I don't think that `isPrivate` covers this. There is a separate `PRIVATE_WITHIN` tag in `Accessibility`.

scalafix-core/shared/src/main/scala/scalafix/v1/Sym.scala
xeno-by: Alright. No rush, just wondering.

olafurpg:
Testkit changes are addressed in #739, will cut a milestone release and update the scalafix.g8 template once that is released.

PR #762, opened 2018-06-26
707: Disable.unlessInside safeBlock array

### LINKED ISSUE #707: Disable.unlessInside safeBlock array
It would be very useful if `Disable.unlessInside`'s `safeBlock` could be an array. For example, to include all suitable `IO` or `fromNullable` things.

The rules in `Disable.unlessInside` are expected to get rather long in scalazzi, so copying the configuration will be extremely awkward and repetitive.

nicodelpiano:
This is an attempt to fix #707

@fommil

### DIFF
--- a/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
+++ b/scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
@@ -79,8 +79,8 @@ object DisabledSymbol {
 }
 
 case class UnlessInsideBlock(
-    @Description("The symbol that indicates a 'safe' block.")
-    safeBlock: DisabledSymbol,
+    @Description("The symbols that represent 'safe' blocks.")
+    safeBlocks: List[DisabledSymbol],
     @Description(
       "The unsafe symbols that are banned unless inside a 'safe' block")
     symbols: List[DisabledSymbol])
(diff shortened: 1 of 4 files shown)


scalafix-core/shared/src/main/scala/scalafix/internal/config/DisableConfig.scala
```diff
@@ -93,7 +93,7 @@ object UnlessInsideBlock {
   implicit val reader: ConfDecoder[UnlessInsideBlock] =
     ConfDecoder.instanceF[UnlessInsideBlock] {
       case c: Conf.Obj =>
-        (c.get[DisabledSymbol]("safeBlock") |@|
+        (c.get[List[DisabledSymbol]]("safeBlock") |@|
```
olafurpg: This change breaks existing configuration where a single safeBlock is defined `safeBlock = "a.B"`. We should allow users to either declare a single string or an array. We can do this by writing a custom decoder, something like
```
  val safeBlockReader = ConfDecoder.instanceF[List[DisabledSymbol]] { conf =>
    val toRead = conf match {
      case str @ Conf.Str(_) => Conf.Lst(str)
      case e => e
    }
    ConfDecoder[List[DisabledSymbol]].read(toRead)
  }
  implicit val reader: ConfDecoder[UnlessInsideBlock] =
    ConfDecoder.instanceF[UnlessInsideBlock] {
      case c: Conf.Obj =>
        (
          c.get[DisabledSymbol]("safeBlock") |@|
            c.get[List[DisabledSymbol]]("symbols")(safeBlockReader) // <- pass in custom decoder
```

fommil: I'm fine with the config breaking, but sure it would be useful if it supported both formats.

olafurpg: I agree it's not critical to keep compat in this particular case but I would like to keep stability when possible, and having both formats would be nice indeed.

fommil: perhaps the field should be renamed to `safeBlocks` though... then I think it's clear that it expects a list.

nicodelpiano: Hi @olafurpg and @fommil, thanks for your comments! To recap, we agree on:

- ensuring compatibility
- renaming to `safeBlocks`

Is there something else we should take into account?

nicodelpiano:
Hi @olafurpg @fommil , I already made the changes requested.
I changed the name of `safeBlock` to be `safeBlocks`, so the configuration files now should use that new name. Also let me know if we need to write a new test for this new list behavior (recall I removed the list in the tests and just left the old usage).

fommil:
Yeah this is super useful!

PR #770, opened 2018-07-12
Support sourceroot in testkit

ShaneDelmore:
This is not ready for merge, it is meant only to demonstrate a small set of changes that would fix scalafix-testkit support for build tools using a sourceroot format like Pants/Bazel/Buck.  This PR currently breaks SBT usage.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -317,13 +317,14 @@ def unit(
         // copy-pasted code from ScalafixTestkitPlugin to avoid cyclic dependencies between build and sbt-scalafix.
         val props = new java.util.Properties()
 
-        def put(key: String, files: Seq[File]): Unit =
-          props.put(
-            key,
-            files.iterator
-              .filter(_.exists())
-              .mkString(java.io.File.pathSeparator)
-          )
+        def put(key: String, files: Seq[File]): Unit = {
+          val value = files.iterator
+            .filter(_.exists())
+            .mkString(java.io.File.pathSeparator)
+          val scalaName = s"scala-${scalaBinaryVersion.value}"
+          val adjustedValue = value.replace("scala-2.12", scalaName)
+          props.put(key, adjustedValue)
+        }
 
         put(
           "inputClasspath",
--- a/build.sbt
+++ b/build.sbt
@@ -197,7 +197,6 @@ val testkit = MultiScalaProject(
     scalafixSettings,
     libraryDependencies ++= Seq(
       semanticdb,
-      ammonite,
       googleDiff,
       scalacheck,
       scalatest
--- a/build.sbt
+++ b/build.sbt
@@ -227,10 +227,6 @@ val testsInput = TestProject(
     project.settings(
       noPublish,
       semanticdbSettings,
-      scalacOptions += {
-        val sourceroot = baseDirectory.in(ThisBuild).value / srcMain / "scala"
-        s"-P:semanticdb:sourceroot:$sourceroot"
-      },
       scalacOptions ~= (_.filterNot(_ == "-Yno-adapted-args")),
       scalacOptions += "-Ywarn-adapted-args", // For NoAutoTupling
       scalacOptions += "-Ywarn-unused-import", // For RemoveUnusedImports
--- a/build.sbt
+++ b/build.sbt
@@ -320,8 +320,12 @@ def unit(
           val value = files.iterator
             .filter(_.exists())
             .mkString(java.io.File.pathSeparator)
-          val scalaName = s"scala-${scalaBinaryVersion.value}"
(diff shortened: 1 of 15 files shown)


scalafix-testkit/src/main/scala/scalafix/testkit/TestkitProperties.scala
```diff
@@ -21,17 +22,21 @@ import scala.meta.Classpath
...
     val inputSourceDirectories: List[AbsolutePath],
-    val outputSourceDirectories: List[AbsolutePath]
+    val outputSourceDirectories: List[AbsolutePath],
+    val sourceRoot: AbsolutePath
 ) {
   def inputSourceDirectory: AbsolutePath =
     inputSourceDirectories.head
   def outputSourceDirectory: AbsolutePath =
     outputSourceDirectories.head
+  def inputOffset: RelativePath = inputSourceDirectory.toRelative(sourceRoot)
```
olafurpg: Would it be possible for you to add the flag `-P:semanticdb:sourceroot:<path>` in pants to compile the input source directories? This is what the scalafix.g8 template generates https://github.com/scalacenter/scalafix.g8/blob/1bc1e8643666350d93b5096be4a7830bba36dc01/scalafix/build.sbt#L17-L20

Otherwise, it's actually a cleaner solution to use the default sourceroot and handle this offset in testkit instead. That would avoid the custom sbt config in every scalafix.g8 project 🤔

ShaneDelmore: Pants does pass -P:semanticdb:sourceroot correctly both when compiling and when calling scalafix-cli.

ShaneDelmore:
This only deals with TestKit regression, not cli.

scalafix-testkit/src/main/scala/scalafix/testkit/TestkitProperties.scala
ShaneDelmore: To be more specific I believe the issue is that pants passes the sourceroot param, which causes semanticdb to store the semanticdb uri relative to a specific path, but that relative path offset is never passed to testkit.

ShaneDelmore:
With the latest build this fixes all of the pants issues we had.  I will use an internally published version until your next public release so no rush on the next version.

scalafix-core/shared/src/main/scala/scalafix/v1/SemanticDoc.scala
```diff
@@ -109,34 +109,31 @@ final class SemanticDoc private[scalafix] (
...
-    val semanticDbRelativeLocation = s"META-INF/semanticdb/$path.semanticdb"
-    Option(semanticdbLoader.getResourceAsStream(semanticDbRelativeLocation))
-      .map{inputStream =>
-        val sdocs = s.TextDocuments.parseFrom(inputStream).documents
+    val semanticdbReluri = s"META-INF/semanticdb/$path.semanticdb"
+    Option(classLoader.getResourceAsStream(semanticdbReluri)) match {
+      case Some(inputStream) =>
+        val sdocs =
+          try s.TextDocuments.parseFrom(inputStream).documents
+          finally inputStream.close()
```
ShaneDelmore: 👍

scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
```diff
@@ -98,4 +99,9 @@ object ClasspathOps {
     roots.foreach(x => Files.walkFileTree(x.toNIO, visitor))
     Classpath(buffer.result())
   }
+
+  def toClassLoader(classpath: Classpath): ClassLoader = {
```
ShaneDelmore: This is handy

build.sbt
```diff
@@ -317,13 +317,14 @@ def unit(
...
-            files.iterator
-              .filter(_.exists())
-              .mkString(java.io.File.pathSeparator)
-          )
+        def put(key: String, files: Seq[File]): Unit = {
+          val value = files.iterator
+            .filter(_.exists())
+            .mkString(java.io.File.pathSeparator)
+          val scalaName = s"scala-${scalaBinaryVersion.value}"
+          val adjustedValue = value.replace("scala-2.12", scalaName)
```
ShaneDelmore: Why hardcode 2.12?  Will this break when updating to 2.13?

ShaneDelmore:
These changes all look great to me. The addition of TestKitPath really cleaned a few bits up.  I’ll test this out in a few hours.

build.sbt
olafurpg: Updated to use `scalaBinaryVersion.in(testsInput).value`

PR #783, opened 2018-08-07
Add scalafix-interfaces with Java APIs for reflective invocation

olafurpg:
This commit adds a new 8kb module scalafix-interfaces with Java-only
interfaces exposing a public API to invoke Scalafix via reflection.
This API is intended for build tools such as sbt, gradle or IDEs like
IntelliJ.

This API is inspired by how zinc Java APIs and Dotty Java interfaces
are designed. A document describing alternative approaches can be
seen here
https://docs.google.com/document/d/1Y1MVMjVQ8P25YEI3uvh86gg3n61WZng0hr1qRUS_S6I/edit#heading=h.1fe4wp3hxrrj

This PR is large for two reasons

1. Several large files were split into multiple files with almost no code changes
2. This commit makes several improvements to the linter and reporting pipeline.

The old reporting pipeline had several problems:

- it was difficult to get the rule name and category id separately
  because they were merged into a single string.
- based on my personal observation, the old `LintCategory` and `LintMessage`
  were not intuitive and several contributed linter rules used them
  in weird ways. I hope the new `LintMessage` interface and `LintID`
  case class makes it more intuitive what the role of category IDs and
  explanations is. Rule authors are able to create custom classes
  that extend `LintMessage`. This design is similar to how Dotty
  error messages are designed and expensive values such as explanations
  can be computed on-demand.
- the internal ScalafixReporter trait contains a lot of complex methods with
  questionable value. Now it contains a single method matching
  the Java interface API.
- the cli did not have an extensible way to hook into the reporting
  of linter messages. Now it's possible to collect linter messages
  programmatically the same way compiler diagnostics are reported
  via zinc.

This commit introduces *almost* no breaking changes to the public API.
Existing rules and tests in this repo are unchanged.  Users that
previously called `LintMessage.copy` (should be rare) must migrate to
use `LintMessage.EagerLintMessage.copy` instead.


scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -0,0 +1,190 @@
...
+  test("error") {
+    val cl = new URLClassLoader(Array())
+    val ex = intercept[ScalafixException] {
+      i.Scalafix.classloadInstance(cl)
+    }
+    assert(ex.getCause.isInstanceOf[ClassNotFoundException])
+  }
+
+  test("runMain") {
+    // This is a full integration test that stresses the full breadth of the scalafix-interfaces API
```
olafurpg: Here is a full integration test for the new reflective API. Gives an idea how it will be used.

scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * JVM runtime reflection method helper to classload an instance of {@link Scalafix}.
+     *
+     * @param classLoader Classloader containing the full Scalafix classpath, including the scalafix-cli module.
+     * @return An implementation of the {@link Scalafix} interface.
+     * @throws ScalafixException in case of errors during classloading, most likely caused
+     * by an incorrect classloader argument.
+     */
+    static Scalafix classloadInstance(ClassLoader classLoader) throws ScalafixException {
```
olafurpg: This is the entrypoint for reflectively loading a `Scalafix` instance.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainArgs.java
```diff
@@ -0,0 +1,99 @@
...
+ * Wrapper around arguments for invoking the Scalafix command-line interface main method.
+ * <p>
+ * To obtain an instance of MainArgs, use {@link scalafix.interfaces.Scalafix#newMainArgs()}.
+ * Instances of MainArgs are immutable and thread safe. It is safe to re-use the same
+ * MainArgs instance for multiple Scalafix invocations.
+ *
+ * @implNote This interface is not intended for extension, the only implementation of this interface
+ * should live in the Scalafix repository.
+ */
+public interface ScalafixMainArgs {
```
olafurpg: Builder interface to construct arguments for the scalafix cli.

scalafix-core/src/main/scala/scalafix/lint/LintMessage.scala
```diff
@@ -3,34 +3,88 @@ package scalafix.lint
...
-    val explanation =
-      if (explain)
-        s"""
-           |Explanation:
-           |${category.explanation}
-           |""".stripMargin
-      else ""
-
-    s"[${category.id}] $message$explanation"
+trait LintMessage {
```
olafurpg: `LintMessage` became a trait so that it's possible (but not necessary) to implement custom message types

```scala
class UnusedCode(val position: Position) extends LintMessage { ... }
```

olafurpg: Question whether basic values should have default implementations 🤔 For example `categoryID` and `explanation` could be empty string and `severity` could default to error. I'll go ahead and update that ASAP.

olafurpg: Done.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixLintID.java
```diff
@@ -0,0 +1,24 @@
...
+ * The contract of id is that all diagnostics of the same "category" will have the same id.
+ * For example, the DisableSyntax rule has a unique ID for each category such as "noSemicolon"
+ * or "noTabs".
+ */
+public interface ScalafixLintID {
+
+    /**
+     * @return The name of the rule that produced this diagnostic.
+     */
+    String ruleName();
```
MasseGuillaume: For example: `RemoveUnusedImport`

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixLintID.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * @return the sub-category within this rule, if any.
+     * Empty if the rule only reports diagnostics of a single
+     * category.
+     */
+    String categoryID();
```
MasseGuillaume: Example?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainArgs.java
```diff
@@ -0,0 +1,99 @@
...
+     *                      Must be a URLClassLoader (not regular ClassLoader) to support
+     *                      compiling sources.
+     */
+    ScalafixMainArgs withToolClasspath(URLClassLoader toolClasspath);
+
+    /**
+     * @param paths Files and directories to run Scalafix on. Directories are recursively expanded
+     *              for files matching the patterns <code>*.scala</code> and <code>*.sbt</code>.
+     */
+    ScalafixMainArgs withPaths(List<Path> paths);
```
MasseGuillaume: Hum, how is it expanded? How do you exclude directories?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainCallback.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Callback handler for events that happen in the command-line interface.
+ */
+public interface ScalafixMainCallback {
+    /**
+     * Handle a diagnostic event reported by the Scalafix reporter.
+     *
+     * @param diagnostic the reported diagnostic.
+     */
+    void reportDiagnostic(ScalafixDiagnostic diagnostic);
```
MasseGuillaume: This is a bit limited compared to what scalafix can do.

It would be useful to have a list of files that were modified. Ideally, it would be useful to have interactive rewrites where the user can make a choice. For example for collection.Seq in the new collection design.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PositionSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.meta.inputs.Input
+import scala.meta.inputs.Position
+import scalafix.testkit.DiffAssertions
+import scalafix.internal.util.PositionSyntax._
+
+class PositionSuite extends FunSuite with DiffAssertions {
+
+  def check(name: String, original: String, expected: String): Unit = {
+    test(name) {
+      val marker = "@@".r
```
MasseGuillaume: Can we reuse this: https://github.com/scalameta/scalameta/blob/master/tests/shared/src/test/scala/scala/meta/tests/tokenizers/BaseTokenizerCoverageSuite.scala ?

It looks really neat:
https://github.com/scalameta/scalameta/blob/master/tests/shared/src/test/scala/scala/meta/tests/tokenizers/TokenizerCoverageSuite.scala

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
```diff
@@ -44,15 +44,22 @@ class CliSemanticSuite extends BaseCliSuite {
     }
   )
 
-  checkSemantic(
-    name = "MissingSemanticDB",
-    args = Array(), // no --classpath
```
MasseGuillaume: can we keep this comment?

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -0,0 +1,192 @@
...
+    val api = i.Scalafix.classloadInstance(this.getClass.getClassLoader)
+    val cwd = Files.createTempDirectory("scalafix")
+    val d = cwd.resolve("out")
+    val src = cwd.resolve("src")
+    Files.createDirectories(d)
+    Files.createDirectories(src)
+    val semicolon = src.resolve("Semicolon.scala")
+    // This rule is published to Maven Central to simplify testing --tool-classpath.
```
MasseGuillaume: I would prefer to see this in the build.

I would: add example-scalafix-rule project and make the test task depends on publishLocal of example-scalafix-rule

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -0,0 +1,192 @@
...
+import scalafix.interfaces.ScalafixException
+import scalafix.interfaces.ScalafixMainCallback
+import scalafix.interfaces.ScalafixMainMode
+import scalafix.internal.reflect.ClasspathOps
+import scalafix.internal.reflect.RuleCompiler
+import scalafix.internal.sbt.ScalafixJarFetcher
+import scalafix.testkit.DiffAssertions
+
+class ScalafixImplSuite extends FunSuite with DiffAssertions {
+  def semanticdbPluginPath(): String = {
```
MasseGuillaume: Can we get this from BuildInfo?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixLintID.java
MasseGuillaume: I guess in `Disable.get`, `get` is the `categoryID`.

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
olafurpg: BuildInfo doesn't robustly support task generators, we could make it a resource generator but that would make the test harder to read IMO. I inlined it here since that makes the test more self-contained (instead of magically pulling in info from the build) and avoids the need to regenerate this information on every compile/run instead of only when we run `ScalafixImplSuite`.

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
olafurpg: I agree it would be good to bake this into the build but I opted for this because it's closer to the environment end-users will use this functionality (I've been bitten by different in custom build environments and end-user environments).

I want this integration test to be as self-contained as possible so that readers don't have to read build.sbt to understand what's going on. I can imagine this test case will be a good reference point for people who want to write unit tests when integrating with Scalafix.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
olafurpg: Done.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PositionSuite.scala
olafurpg: I personally dislike unicode characters since they're inconvenient to write. However, I refactored the test to use
```
  val startMarker = '→'
  val stopMarker = '←'
```
for consistency with the Scalameta tests. The extraction implementation stayed unchanged.

olafurpg:
@ShaneDelmore @xeno-by I am curious to hear your thoughts about the new public Java interface API. The internal changes for `ScalafixReporter` are less interesting and require less scrutiny, most of the work there was to synchronize the internal implementation with the public Java interface API.

olafurpg:
Gonna merge to unblock a new milestone release, happy to address feedback in followup PRs.

PR #786, opened 2018-08-09
Upgrade to scalameta v4.0-M8

olafurpg:
This is most likely the last milestone release.  There were two major
breaking changes that are addressed in this commit.

- new synthetics are now an ADT so we implement a code printer to
  produce the old synthetics
- accessibility is now an ADT


olafurpg:
Appveyor is failing with 

> Error making request with Error Code GatewayTimeout and Http Status Code GatewayTimeout. No further error information was returned by the service

Will merge regardless.

PR #801, opened 2018-08-16
Add names to v0.Denotations (fix #792)

### LINKED ISSUE #792: scalafix.v0.Denotation.names are empty

MasseGuillaume:


scalafix-core/src/main/scala/scalafix/internal/v0/LegacyCodePrinter.scala
```diff
@@ -30,21 +31,58 @@ class LegacyCodePrinter() {
...
+      if (scope.symbols.nonEmpty) {
+        mkString("[", scope.symbols, "]")(emit)
+        text.append(" => ")
+      }
+    }
+  }
+
+  private def pprint(signature: s.Signature): Unit = {
+    signature match {
+      case sig: s.ClassSignature =>
```
MasseGuillaume: @olafurpg should we support those? I gave it a try and it was not so easy, this is why I left a stub here.

PR #802, opened 2018-08-16
Fix legacy sythetics (fix #795)

### LINKED ISSUE #795: v0 Sythetics cannot parse.
Here is what I get:

```scala
*(canBuildFrom[IntInt, ])
*[Vector[Int]Tuple2[IntInt, ], Map[IntInt, ], ]
*[Seq[Int]Tuple2[IntInt, ], Map[IntInt, ], ]
```

MasseGuillaume:


scalafix-core/src/main/scala/scalafix/internal/v0/LegacyCodePrinter.scala
```diff
@@ -30,12 +31,12 @@ class LegacyCodePrinter() {
       text.append(start)
       var first = true
       trees.foreach { tree =>
-        fn(tree)
```
MasseGuillaume: This was the bug.

PR #804, opened 2018-08-20
Followup from recent PRs

olafurpg:
See commit log.

Review @MasseGuillaume


olafurpg:
> bash: [[ -v CI_COVERAGE ]] && bash <(curl -s https://codecov.io/bash): No such file or directory

Any idea what's wrong @MasseGuillaume ?

scalafix-testkit/src/main/scala/scalafix/testkit/DiffAssertions.scala
```diff
@@ -70,13 +75,15 @@ trait DiffAssertions extends FunSuiteLike {
...
-      title: String = ""): Boolean = {
+      title: String = ""
+  ): Boolean = {
     val result = compareContents(obtained, expected)
     if (result.isEmpty) true
     else if (obtained.lines.toList == expected.lines.toList) {
       // Ignore \r\n and \n differences
       true
     } else {
+      println(obtained)
```
MasseGuillaume: remove ?

olafurpg: Good catch, done. I accidentally force pushed so it's not clear from diff that was the only change, sry 😓

PR #805, opened 2018-08-20
Aggregate code coverage

MasseGuillaume:


olafurpg:
Still

> bash: [[ -v CI_COVERAGE ]] && bash <(curl -s https://codecov.io/bash): No such file or directory

:/

PR #808, opened 2018-08-22
Migrate website to docusaurus

olafurpg:
This commit migrates the site to use docusaurus, which has the following
benefits:

- around 10x faster preview while typing documentation. Changes usually
  render within ~500ms with automatic browser reload after file save
  compared to ~10-20 second latency in the previous setup.
- splashy landing page

Big thanks to @gabro for setting up the first Docusaurus website for Metals https://scalameta.org/metals/


website/pages/en/index.js
```diff
@@ -0,0 +1,154 @@
...
+    padding={["bottom", "top"]}
+    id={props.id}
+    background={props.background}
+  >
+    <GridBlock align="left" contents={props.children} layout={props.layout} />
+  </Container>
+);
+
+const Features = props => {
+  const features = [
```
MasseGuillaume: img should have `alt`

MasseGuillaume: `img` should have `alt`

build.sbt
```diff
@@ -219,18 +219,18 @@ lazy val unit = project
...
-    noPublish,
-    websiteSettings,
-    unidocSettings,
-    libraryDependencies += "com.geirsson" %% "metaconfig-docs" % metaconfigV,
-    unidocProjectFilter in (ScalaUnidoc, unidoc) := inProjects(
-      testkit,
-      core
+    libraryDependencies ++= List(
+      "com.geirsson" %% "metaconfig-docs" % metaconfigV,
+      "com.geirsson" % "mdoc" % "0.4.0" cross CrossVersion.full,
```
gabro: do you need the explicit `mdoc` dependency even when using the `DocusaursPlugin`?

website/sidebars.json
```diff
@@ -0,0 +1,35 @@
+{
+  "docs": {
+    "Users": ["users/installation", "users/configuration", "users/faq"],
+    "Rule Authors": [
+      "rule-authors/setup",
+      "rule-authors/tutorial",
+      "rule-authors/glossary"
+    ]
+  },
+  "rules": {
```
MasseGuillaume: Can we generate this?

docs/rule-authors/setup.md
```diff
@@ -38,21 +42,28 @@ scalafix
...
-- Each individual *.scala file in the input project can have a custom .scalafix.conf configuration provided in a comment at the top of the file.
+- Each individual \*.scala file in the input project can have a custom
+  .scalafix.conf configuration provided in a comment at the top of the file.
 
-- Test failures are reported as unified diffs from the obtained output of the rule and the expected output in the `output` project.
+- Test failures are reported as unified diffs from the obtained output of the
+  rule and the expected output in the `output` project.
 
-- Assert that a {% glossary_ref LintMessage %} is expected at a particular line by suffixing the line with the comment `// assert: <LintCategory>`. For an example, see the NoInfer test suite:
+- Assert that a {% glossary_ref LintMessage %} is expected at a particular line
```
gabro: These refs are now unhandled.

docs/rules/DisableSyntax.md
```diff
@@ -1,15 +1,12 @@
 ---
 layout: docs
```
gabro: it doesn't harm, but you don't need `layout` anymore md files

scalafix-docs/src/main/scala/docs/RulesModifier.scala
```diff
@@ -0,0 +1,16 @@
...
+import scala.meta.inputs.Input
+
+class RulesModifier extends StringModifier {
+  val name = "scalafix-rules"
+  override def process(
+      info: String,
+      code: Input,
+      reporter: Reporter
+  ): String = {
+    scalafix.website.allRulesTable(reporter)
```
gabro: As discussed offline, I would suggest to make this a variable, since it doesn't depend on the code block content (which is, in fact, empty)

website/pages/en/index.js
```diff
@@ -0,0 +1,154 @@
...
+    {
+      title: "Catch bugs",
+      content: "Automatically enforce coding conventions of your organization.",
+      image: imgUrl("linting.png"),
+      imageAlign: "left"
+    },
+    {
+      title: "Extensible",
+      content: "Extend Scalafix with custom rewrite and linting rules",
+      image: imgUrl("extensible.png"),
```
MasseGuillaume: I would prefer we avoid using image for this feature. We cannot copy paste, it's not super accessible. I can do a followup PR for this.

website/siteConfig.js
```diff
@@ -0,0 +1,64 @@
...
+  tagline: "Refactoring and linting tool for Scala",
+  url: "https://scalacenter.github.io/",
+  baseUrl: "/scalafix/",
+
+  // Used for publishing and more
+  projectName: "scalafix",
+  organizationName: "scalacenter",
+
+  algolia: {
+    apiKey: "???",
```
gabro: you may want to remove the `algolia` section until you have a valid API, so that the search bar doesn't appear

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+---
+
+To run scalafix on your project, you must first install the Scalafix sbt plugin
+or command line interface. Currently, Scalafix does not provide any IDE
+integrations.
+
+| Scalafix Version | Scalameta(SemanticDB) Version | Scala Version           |
+| ---------------- | ----------------------------- | ----------------------- |
+| 0.5.10           | 2.1.7                         | 2.11.12 / 2.12.4        |
+| @VERSION@        | @SCALAMETA@                   | @SCALA211@ / @SCALA212@ |
```
MasseGuillaume: This is pain point number one. I would prefer we release a Scalameta version for each Scala Version. I think we discussed this offline and you said it would be to much work. We should make it really clear here what is supported.

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+
+## sbt
+
+The quickest way to get started is to run a **syntactic** rule like
+`ProcedureSyntax`
+
+```scala
+// project/plugins.sbt
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "@VERSION@")
```
MasseGuillaume: This is incorrect. Right now I see: 0.5.10 + Scala 2.12.6.

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+// sbt shell
+> myproject/scalafix RemoveUnusedImports      // run rule on main sources
+> myproject/test:scalafix RemoveUnusedImports // run rule on test sources
+```
+
+Notes.
+
+- The `semanticdb-scalac` compiler plugin produces
+  [SemanticDB](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md)
+  files in the target directory `META-INF/semanticdb/`. These files are used by
```
MasseGuillaume: Implementation details?

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+### Example project
+
+For a minimal example project using sbt-scalafix, see the
+[scalacenter/scalafix-sbt-example](https://github.com/scalacenter/scalafix-sbt-example)
+repository.
+
+```scala
+git clone https://github.com/olafurpg/scalafix-sbt-example
```
MasseGuillaume: Should be under `scalacenter/scalafix-sbt-example`.

In the front page we have gitter8, We should stick with only one example. I prefer `git clone` since it's more simple.

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+sbt "scalafix RemoveUnusedImports"
+git diff // should produce a diff
+```
+
+### Settings and tasks
+
+| Name                    | Type           | Description                                                                                                                               |
+| ----------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `scalafix <rule>..`     | `Unit`         | Run scalafix on project sources. See [Rules](rules/overview.md) or use tab completion to explore supported rules.                         |
+| `scalafixCli <args>`    | `Unit`         | Invoke scalafix command line interface directly. See {% doc_ref Installation, cli %} or use tab completion to explore supported arguments |
```
MasseGuillaume: Broken ref

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+// build.sbt
+lazy val myproject = project.settings(
+  scalaVersion := "@SCALA212@", // or @SCALA211@
+  addCompilerPlugin(scalafixSemanticdb),
+  scalacOptions ++= List(
+    "-Yrangepos",          // required by semanticdb-scalac compiler plugin
+    "-Ywarn-unused-import" // required by `RemoveUnusedImports` rule
+  )
+)
+
```
MasseGuillaume: split code block

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+- How to exclude/add files that are processed by the `scalafix` task?
+
+```
+unmanagedSources.in(Compile, scalafix) := unmanagedSources.in(Compile).value.filter(file => ...)
+```
+
+## Command line
+
+The recommended way to install the scalafix command-line interface is with
```
MasseGuillaume: I would remove recommended, since it's the only sane way to install the cli.

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+### --help
+
+```scala mdoc:--help
+
+```
+
+## Maven
+
+It is possible to use scalafix with scala-maven-plugin but it requires a custom
+setup since there is no scalafix specific Maven plugin.
```
MasseGuillaume: Help wanted?

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+snapshot release has a unique version number, jars don't get overwritten. To
+find the latest snapshot version number, go to
+<https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-core_2.12/>
+and select the version number at the bottom, the one with the latest "Last
+Modified". Once you have found the version number, adapting the version number
+
+If using the sbt plugin
+
+```scala
+// project/plugins.sbt
```
MasseGuillaume: We can get this version with a small javascript snippet: https://github.com/scalacenter/bloop/pull/374

docs/users/installation.md
```diff
@@ -0,0 +1,225 @@
...
+
+```sh
+// coursier
+coursier bootstrap ch.epfl.scala:scalafix-cli_@SCALA212@:@VERSION@ -f --main scalafix.cli.Cli -o scalafix
+./scalafix --help
+```
+
+### --help
+
+```scala mdoc:--help
```
MasseGuillaume: --help is not formatted properly since it's a scala code block.

gabro: As discussed offline, I would make this a variable instead of a code block.

docs/users/installation.md
olafurpg: Good catch, the repo is already under `scalacenter/`, was a bad link.

I agree confusing to advertise two different examples. The front page links to the template for rule authors while this is an example for end users. It's best to remove the reference to `scalafix.g8` on the landing page and use a better image for "Extensible". It's not possible to have git clone workflow for rule authors since filenames dependent on the project name.

docs/users/installation.md
olafurpg: Fixed.

docs/users/installation.md
olafurpg: Good catch, previously there were several alternatives to install the cli (homebrew, wget). Coursier only is simpler.

docs/users/installation.md
olafurpg: @MasseGuillaume the rendered block will not have the `scala` language. mdoc should support any language so it's possible to either skip the language prefix or do `sh mdoc`. I opened https://github.com/olafurpg/mdoc/issues/57

@gabro If we make `--help` a site variable then positions will be messed up for the rest of the document since we search/replace site variables before parsing the markdown document with flexmark. Site variables should be single-line only. Maybe we can come up with a better solution, but until then I think having a custom modifier is best.

docs/users/installation.md
olafurpg: Updated. Also improved the Maven section a bit more.

docs/users/installation.md
olafurpg: Cool, we can do that in a separate PR

scalafix-docs/src/main/scala/docs/RulesModifier.scala
olafurpg: This uses the reporter to warn about missing documentation so I prefer to keep it as a modifier

scalafix-docs/src/main/scala/docs/website.scala
```diff
@@ -17,25 +18,19 @@ package object website {
...
-        .resolve("tut")
         .resolve("docs")
         .resolve("rules")
         .resolve(rule.name.value + ".md")
-      if (!Files.exists(docPath.toNIO)) {
-        sys.error(
-          s"Missing documentation for rule ${rule.name.value} in path $docPath"
-        )
+      if (!Files.isRegularFile(docPath.toNIO)) {
+        reporter.warning(s"Missing $docPath")
```
olafurpg: cc/ @gabro

website/pages/en/index.js
olafurpg: Let's improve the landing page in a followup PR

website/sidebars.json
olafurpg: We can, but it's simplest to keep it manual for now. Ideally, I want `docs.Main` to take care of generating data like this and also start the live reload server.

olafurpg:
Merging to unblock a PR updating the v1 API. We can iterate on the contents and structure separately, nothing goes live here.

PR #809, opened 2018-08-22
Second steps towards scalafix.v1

olafurpg:
This PR makes progress towards migrating the codebase to use scalafix.v1 instead of scalafix.v0

- Migrate most syntactic rules to v1 API, the diff is minimal
- s/v1.Sym/v1.Symbol/ now that `scala.meta.Symbol` is gone
- New kids on the block: v1.Type and v1.Signature. Binary stable front-ends for their protobuf equivalents. Their only divergence from the protobuf data structures is that `Scope` is replaced with `List[SymbolInfo]` which is possible thanks to having a symbol table around. Most of the pain was writing `equals` and `hashCode` by hand since I think `Type` and `Signature` should be "data classes" with reasonable equality
- Remove dead rules: xml literals, no finalize, no infer, disable unless
- Remove references to v0 in v1 package

TODO:

- [x] ExplicitResultTypes
- [x] MissingFinal
- [x] Disable, it will stay unchanged on v0
- [x] RemoveUnusedTerms
- [x] RemoveUnusedImports
- [x] NoAutoTupling
- [x] s/v1.Type/v1.Tpe/
- [x] s/LintMessage/Diagnostic/


scalafix-core/src/main/scala/scalafix/v1/SymbolInfo.scala
```diff
@@ -0,0 +1,21 @@
...
+
+final class SymbolInfo private[scalafix] (
+    private[scalafix] val info: s.SymbolInformation
+) {
+  def isNone: Boolean = info.symbol.isEmpty
+  def sym: Symbol = Symbol(info.symbol)
+  def owner: Symbol = Symbol(info.symbol).owner
+  def name: String = info.name
+  def kind: SymbolKind = new SymbolKind(info)
+  def props: SymbolProperties = new SymbolProperties(info.properties)
```
xeno-by: How about inlining `kind.isXXX` and `props.isXXX` into `SymbolInfo`, so that it's `info.isFinal` and `info.isMethod` instead of `info.props.isFinal` and `info.kind.isMethod`?

scalafix-core/src/main/scala/scalafix/v1/SymbolAccess.scala
```diff
@@ -0,0 +1,20 @@
...
+    case _ => scala.None
+  }
+  def isProtected: Boolean = a.isInstanceOf[s.ProtectedAccess]
+  def isProtectedThis: Boolean = a.isInstanceOf[s.ProtectedThisAccess]
+  def protectedWithin: Option[Symbol] = a match {
+    case s.ProtectedWithinAccess(symbol) => Some(Symbol(symbol))
+    case _ => scala.None
+  }
+  def isPublic: Boolean = a.isInstanceOf[s.PublicAccess]
+  def isNone: Boolean = a == s.NoAccess
```
xeno-by: An alternative approach to this API would be to replace `privateWithin` and `protectedWithin` with `isPrivateWithin`, `isProtectedWithin` and `within`. I think I like that better than the current API.

scalafix-core/src/main/scala/scalafix/v1/SymbolProperties.scala
```diff
@@ -21,4 +24,28 @@ final class SymbolProperties private[scalafix] (props: Int) {
...
+    if (isCase) buf += "case"
+    if (isContravariant) buf += "covariant"
+    if (isContravariant) buf += "contravariant"
+    if (isStatic) buf += "static"
+    if (isPrimary) buf += "primary"
+    if (isEnum) buf += "enum"
+    if (isVal) buf += "val"
+    if (isVar) buf += "var"
+    buf.mkString("SymbolProperties(", " ", ")")
+  }
```
xeno-by: I wonder if we can use metap for that.

scalafix-core/src/main/scala/scalafix/internal/v1/FromProtobuf.scala
```diff
@@ -23,4 +23,9 @@ object FromProtobuf {
       PublicAccess
   }
 
+//  def constant(c: s.Constant): Constant = c match {
+//    case s.NoConstant => throw new IllegalArgumentException(c.toString)
+//    case s.NoConstant => throw new IllegalArgumentException(c.toString)
+//  }
```
xeno-by: Is this used anywhere?

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
```diff
@@ -9,7 +9,7 @@ final class Symbol private (val value: String) {
   def isGlobal: Boolean = value.isGlobal
   def isLocal: Boolean = value.isLocal
   def owner: Symbol = Symbol(value.owner)
-  def info(doc: SemanticDoc): SymbolInfo = doc.info(this)
+  def info(doc: SemanticDoc): Option[SymbolInfo] = doc.info(this)
```
xeno-by: Looks like there are two ways of doing the same thing: `doc.info` and `symbol.info`. How about we keep only the former?

scalafix-core/src/main/scala/scalafix/v1/SymbolInfo.scala
xeno-by: https://github.com/scalameta/scalameta/pull/1750

scalafix-core/src/main/scala/scalafix/v1/SymbolAccess.scala
xeno-by: https://github.com/scalameta/scalameta/pull/1750

scalafix-core/src/main/scala/scalafix/v1/SymbolProperties.scala
olafurpg: We can, but it requires making metap more flexible: https://github.com/scalameta/scalameta/pull/1751

scalafix-core/src/main/scala/scalafix/internal/v1/FromProtobuf.scala
olafurpg: It's not, removed.

olafurpg:
Gonna merge to unblock a PR from @MasseGuillaume on the website, happy to iterate further on this.

I believe it will be more interesting to discuss the new v1 API once the docs are in, which is next on my queue :)

PR #812, opened 2018-08-24
Performance improvements

### LINKED ISSUE #810: Scalafix cli doesn't release memory of files processed.
I attempted to run a syntactic scalafix lint on a large number of files using the coursier bootstrap and noticed that memory usage kept increasing until at 10GB of RAM used I killed it.  

The rule I ran was a silly test lint shown here:
```
//file DisablePrintln.scala
import scala.meta._
import scalafix.lint.LintMessage
import scalafix.v0._

/*
class DisablePrintln() extends Rule("DisablePrintln") {
  val errorSeverityRequired = 
    LintCategory.error("Remember to remove debugging printlns before committing code")

    ctx.tree.collect {
      case t @ Term.Name("println") =>
        ctx.lint(LintMessage("use logging instead of println", t.pos, errorSeverityRequired))
    }.asPatch
}

```

Then run via scalafix cli installed via coursier
```
coursier bootstrap ch.epfl.scala:scalafix-cli_2.11.12:0.6.0-M12 -f --main scalafix.cli.Cli -o scalafix
```

Then run against a large repo using:
```
./scalafix --rules=file:DisablePrintln.scala --test .
```

I was able to successfully run the rule on all code using the workaround of calling scalafix cli once per top level directory in the repo:
```
ls -d */ | xargs -I _ ./scalafix --rules=file:DisablePrintln.scala --test _
```

### LINKED ISSUE #811: Short circuit suppression mechanism when there is no // scalafix comment
Follow up of #810, memory profiles indicate that `EscapeHatch` is an allocation hotspot. We can eliminate that expensive analysis if there are no comment suppressions of `SuppressWarnings` annotations in the file. That should eliminate the cost for the majority of files without micro-optimizing the analysis itself.

olafurpg:
- prevent the heap from growing with number of processed files
- always skip parsing the tree if possible (previously it was under a flag because it was unsafe)
- skip suppression analysis if there are no `scalafix:` or `SuppressWarnings` appearances in the text

Fixes #810 
Fixes #811


olafurpg:
Manually tested by running against https://github.com/olafurpg/scala-repos. The heap snapshot no longer shows massive (+400mb) references to `Token.Space`. Running with this branch compared to M15 was noticeably faster.

marcelocenerine:
Hey @olafurpg 
>skip suppression analysis if there are no `scalafix:` or `SuppressWarnings` appearances in the text

I think the right place for the optimization would be in `EscapeHatch.apply` where most of the heavy lifting is performed. I can still see ~30% of CPU time spent in `EscapeHatch` when running a dummy rule. Maybe the quickest thing to do now would be to turn the constructor parameters into by-name params. Once I come back from holidays I will do some profiling and find ways to optimize that code.

PR #814, opened 2018-08-24
Inline scalameta-structure into repo

olafurpg:
Based on feedback from @MasseGuillaume and @ShaneDelmore, there's been missing a utility to inspect the structure of trees. `Tree.structure` is too dense to be helpful. This commit inlines the sources from the repo https://github.com/masseguillaume/scalameta-structure. Thank you @MasseGuillaume for implementing scalameta-structure! I did my best to preserve the git history, but I suspect I messed it up somehow 😅

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/util/Inspect.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/util/Inspect.scala
@@ -5,15 +5,16 @@ import scala.meta._
 import org.typelevel.paiges._
 
 object Inspect {
-  def pretty(tree: Tree, showFieldNames: Boolean): String = {
-    prettyTree(tree, showFieldNames).render(1)
-  }
 
-  private def prettyList(tree: List[Tree], showFieldNames: Boolean): Doc = {
+  def prettyList(tree: List[Tree], showFieldNames: Boolean): Doc = {
     wrapList(tree.map(t => prettyTree(t, showFieldNames)))
   }
 
-  private def prettyTree(tree: Tree, showFieldNames: Boolean): Doc = {
+  def prettyOption(tree: Option[Tree], showFieldNames: Boolean): Doc = {
+    wrapOption(tree.map(t => prettyTree(t, showFieldNames)))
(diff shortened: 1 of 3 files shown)


scalafix-core/src/main/scala/scalafix/internal/util/Inspect.scala
```diff
@@ -0,0 +1,84 @@
...
+  private def prettyTree(tree: Tree, showFieldNames: Boolean): Doc = {
+    tree match {
+      case _ if tree.tokens.isEmpty =>
+        Doc.empty
+      case v: Term.Name =>
+        Doc.text(v.structure)
+      case t: Type.Name =>
+        Doc.text(t.structure)
+      case _ =>
+        val args = tree.productFields.zip(tree.productIterator.toList).map {
```
olafurpg: cc/ @gabro `productFields` sees more usage!

scalafix-core/src/main/scala/scalafix/util/Api.scala
```diff
@@ -16,6 +19,11 @@ trait Api {
     def asPatch: Patch = patch.getOrElse(Patch.empty)
   }
 
+  implicit class XtensionScalafixTreeInspect(tree: Tree) {
+    def inspect: String = Inspect.pretty(tree, showFieldNames = false)
+    def inspectLabeled: String = Inspect.pretty(tree, showFieldNames = true)
+  }
+
```
MasseGuillaume: I would also add an extension for `Option[Tree]` and `List[Tree]`.

scalafix-core/src/main/scala/scalafix/internal/util/Inspect.scala
gabro: 🎉

scalafix-core/src/main/scala/scalafix/util/Api.scala
olafurpg: Sure thing, done.

PR #825, opened 2018-08-26
Several fixes

### LINKED ISSUE #788: Expose version numbers via resources in scalafix-interfaces
It would be helpful to have a property files in resources of the scalafix-interfaces module. This would allow sbt-scalafix to depend on scalafix-interfaces and find the latest scalafix version

### LINKED ISSUE #817: Allow semantic rules to require specific scalacOptions
Some rules like `RemoveUnusedImports` require -Ywarn-unused-imports to function. It would be good if those rules could document that they require a particular scalac setting in order to function. In case the setting is not enabled, then scalafix should fail at the start with an error message saying what setting is required.

### LINKED ISSUE #821: Expose available rule names in scalafix-interfaces
scalafix-interfaces should expose a method to get a `List[String]` of valid rules names. This list should include rules that are loaded via service loaders. This list can be used by sbt-scalafix to provide completions from the sbt shell.

olafurpg:
See independent commits. Travis CI was freezing for a couple hours so I couldn't open separate PRs for each fix.

Fixes #821
Fixes #817
Fixes #788

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/v1/Rules.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/v1/Rules.scala
@@ -1,18 +1,15 @@
 package scalafix.internal.v1
 
+import java.util.ServiceLoader
 import metaconfig.Conf
 import metaconfig.Configured
 import scala.meta.tokens.Tokens
-import scalafix.internal.config.DisableConfig
 import scalafix.internal.config.MetaconfigPendingUpstream
-import scalafix.internal.rule._
 import scalafix.internal.util.SuppressOps
-import scalafix.internal.v0.LegacySemanticRule
 import scalafix.lint.RuleDiagnostic
 import scalafix.lint.Diagnostic
 import scalafix.patch.Patch
 import scalafix.rule.RuleName
-import scalafix.v0
 import scalafix.v1.Doc
 import scalafix.v1.Rule
 import scalafix.v1.SemanticDoc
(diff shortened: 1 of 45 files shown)


scalafix-core/src/main/scala/scalafix/internal/v1/Rules.scala
```diff
@@ -71,35 +68,15 @@ case class Rules(rules: List[Rule] = Nil) {
...
-      val name = fn(v0.SemanticdbIndex.empty).name
-      new LegacySemanticRule(name, fn)
-    }
+  def all(classLoader: ClassLoader): List[Rule] = {
+    import scala.collection.JavaConverters._
+    ServiceLoader
+      .load(classOf[Rule], classLoader)
+      .iterator()
+      .asScala
+      .toList
```
gabro: beautiful 😍

scalafix-core/src/main/scala/scalafix/rule/RuleName.scala
```diff
@@ -25,7 +25,7 @@ final case class RuleName(identifiers: List[RuleIdentifier]) {
   def value: String =
     if (isDeprecated) identifiers.mkString("+")
     else if (nonDeprecated.isEmpty) "empty"
-    else nonDeprecated.mkString("+")
+    else nonDeprecated.head.value
```
gabro: If you feel like

```scala
if (isDeprecated) identifiers.mkString("+")
else nonDeprecated.headOption.map(_.value).getOrElse("empty")
```

olafurpg: Refactored into pattern match to avoid `head` ;)

PR #828, opened 2018-08-27
Expose synthetics in v1 API

olafurpg:
This commit exposes the new SemanticDB Synthetics in the public Scalafix
API. For reference, there is an example ExplicitSynthetic rule
that inserts explicit `.apply` and inferred type parameters for infix
operators. This kind of refactoring would have been significantly more
difficult with the v0 API.

Other fixes:
- Ignore non-existing files instead of throwing NoSuchFileException:
  This comes up when reading stale semanticdb files of removed files.
- s/NoTpe/NoType/
- render `.inspect` with width 80 instead of 1.
- make `diagnostics` an iterator instead of eager list

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/util/PositionSyntax.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/util/PositionSyntax.scala
@@ -9,6 +9,11 @@ object PositionSyntax {
   implicit class XtensionPositionsScalafix(private val pos: Position)
       extends AnyVal {
 
+    def contains(other: Position): Boolean = {
+      pos.start <= other.start &&
+      pos.end >= other.end
+    }
+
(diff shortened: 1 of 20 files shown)


scalafix-core/src/main/scala/scalafix/util/Api.scala
```diff
@@ -27,23 +27,23 @@ trait Api {
 
   implicit class XtensionScalafixTreeInspect(tree: Tree) {
     def inspect: String =
```
MasseGuillaume:
```scala
def inspect: String = inspect(1)
def inspect(maxColumns: Int): String = // ...
``

olafurpg: I'll revert this change.

PR #830, opened 2018-08-27
Document why Patch.isEmpty returns false for replaceSymbols (fix #772)

### LINKED ISSUE #772: Make Patch.isEmpty return false if the patch produces an empty diff
I'm using `ctx.replaceSymbols` to replace some symbol. I would like to know if the patch replaced something to add an additional import. 

`Patch.nonEmpty` only check if the `Patch` is the `EmptyPatch`, this is a bit counter-intuitive. I would rename this `isEmptyPatch`. I think the semantic of `isEmpty` should be: did it changed something?

MasseGuillaume:


scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -61,7 +61,16 @@ sealed abstract class Patch {
   def +(other: Option[Patch]): Patch =
     this.+(other.getOrElse(Patch.empty))
   def ++(other: Iterable[Patch]): Patch = other.foldLeft(this)(_ + _)
+
+  /**
+    * Check if the patch is Patch.empty. This is an inexpensive computation.
```
olafurpg: The main point of this operation is not that it's cheap so I would rather formulate it this way

> Returns true if this patch is `Patch.empty`, false otherwise.
>
> Note, may return false for patches that produce empty diffs. For example, a `Patch.replaceSymbols(from, to)` returns false even if the symbol `from` is not referenced in the code resulting in an empty diff.

PR #833, opened 2018-08-28
Unimport unused renames instead of removing them

### LINKED ISSUE #614: RemoveUnusedImports does not handle wildcard imports correctly.
If I have the code
```
import scala.util.{Try => UnusedTry, _}
```
and UnusedTry is actually unused, then rewriting the code to 
```
import scala.util._ 
```
does not actually remove the import.  

A better rewrite here would be to explicitly unimport
```
import scala.util.{Try => _, _}
```

olafurpg:
Unused renames may still have significant meaning so removing them may
cause compile errors.

Fixes #614


ShaneDelmore:
Looks great, thank you!

PR #835, opened 2018-08-28
Polish remove unused rule based on real-world experiment

olafurpg:
I ran the rule on akka and found a couple low hanging fruit improvements.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -6,7 +6,7 @@ import scala.collection.mutable
 import scala.meta._
 import scalafix.v1._
 
-class RemoveUnused
+class RemoveUnused(config: RemoveUnusedConfig)
     extends SemanticRule(
       RuleName("RemoveUnused")
         .withDeprecatedName(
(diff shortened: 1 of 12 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -29,27 +31,42 @@ class RemoveUnused
...
     }
 
-    doc.tree.collect {
-      case i: Importee if isUnusedImport(importPosition(i)) =>
-        Patch.removeImportee(i).atomic
-      case i: Defn if isUnusedTerm(i.pos) =>
-        defnTokensToRemove(i).map(Patch.removeTokens).asPatch.atomic
-    }.asPatch
+    if (isUnusedImport.isEmpty && isUnusedTerm.isEmpty) {
+      // Optimization: don't traverse if there are no diagnostics to act on.
```
olafurpg: cc/ @ShaneDelmore making use of lazy trees here :) running `RemoveUnused` on the entire akka codebase (~220k loc) takes ~20 seconds on my laptop. That includes loading up all semanticdbs and producing a 7k loc diff.

ShaneDelmore: That is so cool to see 👍

PR #838, opened 2018-08-29
Implement holistic pretty-printing strategy for scalafix.v1 types

olafurpg:
- `toString` pretty-prints psuedo syntax, trying where possible to mimic
  actual code while still printing synthetic modifiers like `<default>`.
- `structure` pretty-prints the raw structure, showing class names.
- `structureLabeled` pretty-prints the raw structure, showing class and field names.
  Field names are particularly helpful in scalafix.v1 since types like
  STree and SType don't expose `unapply` for pattern matching.

This commit introduces a fixture to run "expect tests" to assert that
the pretty-printer does not regress in future releases. Expect tests
have been very helpful for us in scalameta so I'm excited to bring them
to scalafix.

It's a shame that we can't use case classes, but they are unfortunately
unusable if we ever add a new field to the public types. This
boilerplate is cumbersome to write but it allows us to evolve without
binary breaking changes.

All pretty-printing methods implement line-wrapping out of the box
thanks to the excellent Paiges library.

This PR upgrades to the latest Scalameta milestone and aligns the field
names that got renamed in https://github.com/scalameta/scalameta/pull/1760.


scalafix-core/src/main/scala/scalafix/util/Api.scala
```diff
@@ -25,25 +25,19 @@ trait Api {
...
-      Inspect.prettyList(tree, showFieldNames = true).render(1)
+  implicit class XtensionScalafixProductInspect(product: Product) {
+    // NOTE(olafurpg) I can't explain why, but fortunately these methods shadow
+    // `scala.meta.Tree.structure` regardless of how `import scalafix.v1, scala.meta._`.
+    // These structure methods work the same except they implement line wrapping which
+    // makes the output more readable.
+    def structure: String =
+      ProductStructure.structure(product).render(80)
+    def structure(printWidth: Int): String =
+      ProductStructure.structure(product).render(printWidth)
```
olafurpg: cc/ @MasseGuillaume @ShaneDelmore This PR replaces the recently introduced `.inspect` with `.structure` and `.structureLabeled`. The implementation is generic for any `Product` and supports field names for `scalafix.v1.{SType,STree,Signature}`. These overloads can be used to control the column width of the output.

scalafix-tests/unit/src/main/resources/expect/Pretty.expect
```diff
@@ -0,0 +1,159 @@
+[6:7..6:17]:      /* test/PrettyTest# */ class PrettyTest extends AnyRef { +4 decls }
+[6:18..6:18]:     /* test/PrettyTest#`<init>`(). */ <primary> def this
+[7:3..7:21]:      *(canBuildFrom[Int])
+[7:3..7:14]:      *[Int,List[Int]]
+[7:3..7:7]:       *.apply[Int]
+[8:3..12:12]:     orig(1.to(10)).flatMap[Int,IndexedSeq[Int]](
```
olafurpg: cc/ @maxov, synthetics printed with fancy line wrapping

PR #839, opened 2018-08-29
Rename Doc and SemanticDoc to SyntacticDocument and SemanticDocument

### LINKED ISSUE #379: Consider renaming RuleCtx
Following up on discussion with @gabro. It's not clear that RuleCtx is the context for a single source file. Some alternative names that might be clearer

- FileContext
- DocumentContext

olafurpg:
I started using `Doc` to avoid ambiguous references to
`scala.meta.Document`. That is no longer a problem so let's use the more
self-descriptive names. This has a nice synchrony with SyntacticRule and
SemanticRule.

Fixes #379


olafurpg:
I'll give you the honors @gabro to merge 😉

PR #840, opened 2018-08-29
One task to rule them all: scalafix

MasseGuillaume:
Related to: https://github.com/scalacenter/sbt-scalafix/pull/22


olafurpg:
It would be good to document how to use --test and create aliases to run in both `Compile` and `Test` config, but we can leave that for a separate PR

PR #845, opened 2018-08-30
Refine sbt installation docs

olafurpg:


docs/users/installation.md
```diff
@@ -87,17 +113,71 @@ git diff // should produce a diff
...
+
+```scala
+// top of build.sbt
+addCommandAlias("fix", "all compile:scalafix test:scalafix")
+```
+
+### Enforce in CI
+
+To automatically enforce that Scalafix has been run on all sources, use the
+`scalafixTest` instead of `scalafix`. This task fails the build if running
```
olafurpg: This task does not exist yet, I will add it later https://github.com/scalacenter/scalafix/issues/846

MasseGuillaume: @olafurpg can you explain me again why we need `scalafixTest` and we cannot use `scalafix --test`?

website/pages/en/index.js
```diff
@@ -233,8 +233,8 @@ class Index extends React.Component {
           <div className="split-right">
```
MasseGuillaume: Can you quote the `'` in don't ?

MasseGuillaume: Can you add the sbt command in the Travis terminal window ?

docs/users/installation.md
olafurpg: `scalafixTest` is needed in order to support the parallel `all scalafixTest test:scalafixTest`. Otherwise, you would need to run the sequential `; scalafix --test; test:scalafix --test`.

docs/users/installation.md
```diff
@@ -3,69 +3,95 @@ id: installation
...
 ```
-> myproject/scalafix ProcedureSyntax      // run rule on main sources
-> myproject/test:scalafix ProcedureSyntax // run rule on test sources
+
+Next, if we run a **semantic** rule like `RemoveUnusedImports` then we get an
+error
+
+```
+> myproject/scalafix RemoveUnusedImports
+[error] SemanticDB not found: META-INF/semanticdb/src/main/scala/...
```
MasseGuillaume: Can we put the sbt instruction in the error message?

docs/users/installation.md
```diff
@@ -3,69 +3,95 @@ id: installation
...
--Xmx4G
-```
+
+Great! You are all set to use Scalafix with sbt :)
+
+> Beware that the `semanticdb-scalac` compiler plugin adds around 7-25% overhead
+> to compilation. It's recommended to provide generous JVM memory and stack
+> settings in `.jvmopts`:
+>
+> ```scala
```
MasseGuillaume: not scala

docs/users/installation.md
```diff
@@ -87,17 +113,71 @@ git diff // should produce a diff
...
+| ----------------- | -------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| `scalafix <args>` | `TaskKey[Unit]`            | Invoke scalafix command line interface directly. Use tab completion to explore supported arguments or consult [--help](#help)     |
+| `scalafixConfig`  | `SettingKey[Option[File]]` | .scalafix.conf file to specify which scalafix rules should run. Defaults to `.scalafix.conf` in the root directory, if it exists. |
+
+### Main and test sources
+
+The task `myproject/scalafix` runs only in the project `myproject` for main
+sources, excluding test sources. To run Scalafix on test source, execute
+`myproject/test:scalafix` instead. To run on both main and test sources, execute
+`all myproject/scalafix myproject/test:scalafix`
```
MasseGuillaume: I never seen `all`. What does it do? Run two task in parallel?

docs/users/installation.md
gabro: I would link to the docs, but I agree with the spirit of this comment!

docs/users/installation.md
```diff
@@ -3,69 +3,95 @@ id: installation
...
+The error message "SemanticDB not found" means the
+[SemanticDB](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md)
+compiler plugin is not enabled in the build. Let's fix that by adding the
+following settings to `build.sbt`
 
 ```scala
 // build.sbt
 lazy val myproject = project.settings(
   scalaVersion := "@SCALA212@", // or @SCALA211@
   addCompilerPlugin(scalafixSemanticdb),
```
gabro: It would be nice to render this code block as a diff, to highlight what changed.

docs/users/installation.md
gabro: the `scalafixTest` _task_ instead of `scalafix`.?

docs/users/installation.md
olafurpg: Let's discuss this in https://github.com/scalacenter/scalafix/issues/847 I agree with you this is not great. Good thing about documenting the UX is you see what can be improved :) Will update the docs once the situation improves.

docs/users/installation.md
olafurpg: Good idea! Done.

docs/users/installation.md
olafurpg: Fair enough, I added it since it still rendered more nicely with the comment. Removed

docs/users/installation.md
olafurpg: Yes, it's like executing two tasks at once.

docs/users/installation.md
olafurpg: I've removed `scalafixTest` in favor of `scalafix --test`. I'm convinced having a single `scalafix` task is the best option.

olafurpg: Fixed.

website/pages/en/index.js
olafurpg: This is a syntax highlighting bug in GitHub, "don't" renders fine

screen shot 2018-08-31 at 16 13 07

> Can you add the sbt command in the Travis terminal window ?

There's not enough vertical space to fit another line, I tried tweaking the css but couldn't figure it out so that the diff doesn't get hidden under the travis bar.
screen shot 2018-08-31 at 16 14 05

PR #852, opened 2018-09-03
Avoid scrollbar on website landing page (#Fix 842)

MasseGuillaume:


olafurpg:
I will follow up with a PR reverting back to the original docusaurus landing page.

PR #861, opened 2018-09-07
Improve documentation and polish several details

### LINKED ISSUE #432: Document + Report when a rules does not have a default constructor.
I'm running `unit/testOnly scalafix.tests.rule.SemanticTests` and a test fails with:

```
[info] - scala/test/DisableSyntax.scala *** FAILED *** (7 milliseconds)
[info]   java.lang.IllegalStateException: Type mismatch;
[info]   found    : String (value: "DisableSyntax")
[info]   expected : scalafix.rule.Rule
```

This is due to not having an empty constructor:

```scala
final case class DisableSyntax(config: DisableSyntaxConfig) extends Rule("DisableSyntax") with Product {
  // ...
}
```

The error reported should be something along those lines:

```
Cannot instantiate `DisableSyntax`. `Rule`s need default parameters for all fields and `SemanticRule`s need default parameters for all fields except `SemanticdbIndex`.
```

### LINKED ISSUE #796: Add scaladoc
I would like to see some scaladoc, even for private/internal classes. As the codebase grows in complexity, it's important to have a quick summary of the class responsibility.

### LINKED ISSUE #819: java.util.ServiceConfigurationError: scalafix.v1.Rule: Provider fix.MyRule_v1 not a subtype
This error happens when we have `object MyRule_V1 extends Rule` and it should be `class MyRule_v1 extends SemanticRule`. We should document this since it took me a moment to understand what's going on.

### LINKED ISSUE #860: Simplify github: expansion rules
While writing the new tutorial for rule authors I got frustrated by the complicated expansion rules for the `github:org/repo` scheme

- doing `org/repo` without a version should work but it currently doesn't
- the filename for `org/repo/v1` becomes `Repo_v1.scala`, using a mix of g8-flavored camel case for the repo name and g8-flavored snake case for the version number.

I propose to simplify this

- `org/some-repo` expands by default to regular camel case `SomeRepo.scala`, with a fallback to g8-flavored `Somerepo.scala` in case of a 404.
- `org/some-repo/v1` expands to `v1.scala`, using the version number verbatim without automatic CamelCase/snake_case conversion

This makes the short syntax `org/some-repo` useful, and the `org/repo/RuleName` syntax simple to understand since there is no complicated casing rule to infer.

olafurpg:
- Simplify `github:` expansion (fixes #860)
- New "Developer guide" section with a comprehensive tutorial for
  writing a syntactic linters and semantic rewrite. (fixes #819, fixes #432)
- Upgrade to scalafmt v1.5.1 and use javadoc formatting for docstrings
- Docstrings for SymbolInfo, SType and STree (fixes #796)


olafurpg:
The website can be previewed here https://olafurpg.github.io/scalafix/

docs/rules/overview.md
```diff
@@ -1,14 +1,22 @@
...
+while other are **semantic**.
+
+**Syntactic**: the rule can run directly on source code without compilation.
+Syntactic rules are simple to run and they are typically fast but they do not
+have access to important information such as symbols and types.
+
+**Semantic**: the rule can requires the source code to be compiled with the
+Scala compiler beforehand with the
+[SemanticDB](https://github.com/scalameta/scalameta/blob/master/semanticdb/semanticdb3/semanticdb3.md)
+compiler plugin. Semantic rulesa are more complicated to run and they are
```
gabro: typo: rulesa

docs/rules/DisableSyntax.md
```diff
@@ -11,15 +11,16 @@ rule.
 Example:
 
 ```scala
-MyCode.scala:7: error: [DisableSyntax.xml] xml is disabled.
-  <a>xml</a>
-  ^
+MyCode.scala:7: error: [DisableSyntax.xml] xml literals is disabled.
+  var x = 2
+  ^^^
```
gabro: this looks wrong (message vs actual code)

olafurpg: Good catch! I removed the var example, `throw` is enough to get the idea. Thanks

PR #863, opened 2018-09-12
Rename --test to --check

### LINKED ISSUE #862: Consider renaming --test to --check
--test is a bit confusing when explaining it in the context of sbt scopes. 

I propose to introduce the --check alias and deprecate (or remove) --test

olafurpg:
The --test flag will continue to work but it's hidden from
documentation. Fixes #862


olafurpg:
whoops

PR #867, opened 2018-09-17
Second round of documentation improvements

olafurpg:
This commit adds documentation for 

- SymbolMatcher
- Patch
- ScalaType (old SType)
- SyntheticTree (old STree)
- SymbolInformation (old SymbolInfo)

Live demo: https://olafurpg.github.io/scalafix/docs/developers/patch.html

While working on the docs and hosting a workshop at Scala Italy I saw several low-hanging fruits to improve the API. The changes are breaking however so the next release will need to be v0.7.0-RC1 instead of v0.6.0-RC2.


docs/developers/symbol-matcher.md
```diff
@@ -0,0 +1,179 @@
...
+mainClass.matches(Symbol("com/Main."))
+```
+
+To match only the `println(String)` method overload and not the `println()`
+method.
+
+```scala mdoc
+val printlnString  = SymbolMatcher.exact("scala/Predef.assert(+1).")
+printlnString.matches(Symbol("scala/Predef.assert()."))
+printlnString.matches(Symbol("scala/Predef.assert(+1)."))
```
ShaneDelmore: Did the example here change from assert to println?

olafurpg: Good catch! Will fix in a followup PR, thanks ❤️

PR #877, opened 2018-09-20
Add noUniversalEquality option to DisableSyntax

### LINKED ISSUE #874: Let's add DisableSyntax.noUniversalEquality
Following up from discussions in https://gitter.im/scalacenter/scalafix?at=5ba383391043315947bbdfed

We can add an option `DisableSyntax.noUniversalEquality` that reports an error for all usages of `==`.

gabro:
Closes #874

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
@@ -196,6 +196,12 @@ final class DisableSyntax(config: DisableSyntaxConfig)
               "Default args makes it hard to use methods as functions.",
               m.pos)
           }
+      case Term.ApplyInfix(_, t @ Term.Name("=="), _, _) if config.noUniversalEquality =>
+        Seq(Diagnostic(
+          "noUniversalEquality",
+          "== (universal equality) is disabled",
+          t.pos
+        ))
     }
     val FinalizeMatcher = DisableSyntax.FinalizeMatcher("noFinalize")
     doc.tree.collect(DefaultMatcher.orElse(FinalizeMatcher)).flatten
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntaxConfig.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntaxConfig.scala
@@ -65,6 +65,10 @@ case class DisableSyntaxConfig(
       "Report error when pattern matching in val assignment with non-tuple patterns.")
     noValPatterns: Boolean = false,
     @Description(
+      "Report error on `==` (universal equality)"
+    )
+    noUniversalEquality: Boolean = false,
+    @Description(
       "Report error if the text contents of a source file matches a given regex.")
     @ExampleValue(
       """|[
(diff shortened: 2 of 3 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -196,6 +196,12 @@ final class DisableSyntax(config: DisableSyntaxConfig)
               "Default args makes it hard to use methods as functions.",
               m.pos)
           }
+      case Term.ApplyInfix(_, t @ Term.Name("=="), _, _) if config.noUniversalEquality =>
+        Seq(Diagnostic(
+          "noUniversalEquality",
+          "== (universal equality) is disabled",
```
olafurpg: What do you think about allowing users to customize the message? I suspect in many codebases the convention is to use an alternative like `===` or similar.

gabro: Makes sense. I'll get to it tomorrow (running out now)

olafurpg: Concretely, we can make this message configurable by adding `noUniversalEqualityMessage: String` with the default message "universal equality should be avoided since it is not typesafe".

scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -196,6 +196,12 @@ final class DisableSyntax(config: DisableSyntaxConfig)
               "Default args makes it hard to use methods as functions.",
               m.pos)
           }
+      case Term.ApplyInfix(_, t @ Term.Name("=="), _, _) if config.noUniversalEquality =>
+        Seq(Diagnostic(
+          "noUniversalEquality",
```
olafurpg: Let's use `==` here so that the ID becomes `DisableSyntax.==`, which is consistent with `DisableSyntax.throw` and other settings

scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntax.scala
```diff
@@ -196,6 +196,12 @@ final class DisableSyntax(config: DisableSyntaxConfig)
               "Default args makes it hard to use methods as functions.",
               m.pos)
           }
+      case Term.ApplyInfix(_, t @ Term.Name("=="), _, _) if config.noUniversalEquality =>
```
olafurpg: Let's add a case for `Term.Apply` as well since `1.==(2)` is valid

gabro: Yeah, I've initially excluded it on purpose.

> I've explicitly restricted the rule to infix applications of == as I don't think handling non-infix applications (e.g., 1.==(2)) is worth the (albeit low) effort.

But today is a new day and I've changed my mind.

scalafix-rules/src/main/scala/scalafix/internal/rule/DisableSyntaxConfig.scala
```diff
@@ -64,6 +64,15 @@ case class DisableSyntaxConfig(
...
     noValPatterns: Boolean = false,
+    @Description(
+      "Report error on `==` (universal equality)"
+    )
+    noUniversalEquality: Boolean = false,
+    @Description(
+      "Reporter message for noUniversalEquality"
+    )
+    noUniversalEqualityMessage: String =
+      "Universal equality should be avoided since it's not typesafe",
```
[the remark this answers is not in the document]
gabro: I'm open to suggestions :)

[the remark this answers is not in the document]
olafurpg: `== is unsafe since it allows comparing two unrelated types`

PR #881, opened 2018-09-21
Java 11 compat

markehammons:
Made the changes necessary to get this building on my system. Currently 20 of the tests fail, as opposed to 40 without the fix. I've attached the tests log below. It looks like a lot of the failures have to do with scalameta, but I don't have time at the moment to look into what's going wrong. It may end up that scalameta will have to have fixes applied before everything work.
[tests.txt](https://github.com/scalacenter/scalafix/files/2405289/tests.txt)

### DIFF
--- a/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
+++ b/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
@@ -18,6 +18,7 @@ object ClasspathOps {
 
   def bootClasspath: Option[Classpath] = sys.props.collectFirst {
     case (k, v) if k.endsWith(".boot.class.path") => Classpath(v)
+    case (k, v) if k == "java.class.path" => Classpath(v)
   }
 
   val devNull = new PrintStream(new OutputStream {
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@ noPublish
 // force javac to fork by setting javaHome to get error messages during compilation,
 // see https://github.com/sbt/zinc/issues/520
 def inferJavaHome() =
-  Some(file(System.getProperty("java.home")).getParentFile)
+  Some(file(System.getProperty("java.home")))
 
 lazy val interfaces = project
   .in(file("scalafix-interfaces"))
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,16 @@ noPublish
 
 // force javac to fork by setting javaHome to get error messages during compilation,
 // see https://github.com/sbt/zinc/issues/520
-def inferJavaHome() =
-  Some(file(System.getProperty("java.home")))
+def inferJavaHome() = {
+  val jHome = file(System.getProperty("java.home"))
+  if((jHome / "bin" / "javac").exists()) {
+    Some(jHome)
+  } else if((jHome.getParentFile / "bin" / "javac").exists()) {
+    Some(jHome.getParentFile)
+  } else {
+    None
+  }
+}
 
 lazy val interfaces = project
   .in(file("scalafix-interfaces"))
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
@@ -164,8 +164,8 @@ object MainOps {
           DiffUtils.unifiedDiff(
             input.syntax + "-ondisk",
             input.syntax + "-semanticdb",
-            input.text.lines.toList,
-            doc.text.lines.toList,
+            input.text.linesIterator.toList,
+            doc.text.linesIterator.toList,
             3
           )
         }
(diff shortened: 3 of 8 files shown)


scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
```diff
@@ -18,6 +18,7 @@ object ClasspathOps {
 
   def bootClasspath: Option[Classpath] = sys.props.collectFirst {
     case (k, v) if k.endsWith(".boot.class.path") => Classpath(v)
+    case (k, v) if k == "java.class.path" => Classpath(v)
```
olafurpg: On my machine this returns the  path to the sbt jar
```scala
System.getProperty("java.class.path"): "/usr/local/Cellar/sbt/1.1.6/libexec/bin/sbt-launch.jar"
```
The expected output for `bootClasspath` is `rt.jar` and friends

```scala
bootClasspath: Some(
  Classpath(
    List(
      AbsolutePath(
        /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/resources.jar
      ),
      AbsolutePath(/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/rt.jar),
      AbsolutePath(
        /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/sunrsasign.jar
      ),
      AbsolutePath(
        /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/jsse.jar
      ),
      AbsolutePath(/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/jce.jar),
      AbsolutePath(
        /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/charsets.jar
      ),
      AbsolutePath(/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/lib/jfr.jar),
      AbsolutePath(/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre/classes)
    )
  )
)
```

olafurpg: rt.jar is read from disk using `java.util.jar.JarFile` here

https://github.com/scalameta/scalameta/blob/7cdcba551547d72a4ac7ec3063e86ef2c861d574/semanticdb/metacp/src/main/scala/scala/meta/internal/classpath/ClasspathIndex.scala#L92

We need this to look up information like "what is the signature of java.lang.String`. It seems the Scala compiler handled the removal of `sun.boot.class.path` by simply requiring applications to explicitly provide the boot classpath. It seems we will need to do the same, which requires updating integrations like sbt-scalafix.

As long as the user includes rt.jar in the --classpath argument then this should not be necessary.

I propose we change this logic to return `.getOrElse(Classpath(Nil))` if `sun.boot.class.path` does not exist. This allows users to explicitly pass in rt.jar if they so wish or risk getting error such as "missing symbol java/lang/String#".

WDYT?

markehammons: @olafurpg I'm not sure this is workable at all.  As I mentioned above, rt.jar simply does not exist in java 9 and above. it's been broken up into .so files and stuff. 

I think what is needed for the runtime classes is something like this: https://docs.oracle.com/javase/9/docs/api/java.compiler-summary.html. But I'm not sure, because I'm not sure what you're doing with the classfiles in the above example.

markehammons: I've done a bit of digging, and to get class file paths like in your example above for the rt classes for java, you do like this: 

https://gist.github.com/markehammons/74b4c3bd9417afad84e19bc327d9cd24

these are ```jrt:/``` paths of the form ```jrt:/modules/java.activation/com/sun/activation/registries/LineTokenizer.class```. I'm pretty sure that you can extract the bytecode from said class using the path ```jrt:/java.activation/com/sun/activation/registries/LineTokenizer.class```. So, there's obviously going to be some specialization required for java 9 support in scalafix. A quick question though, are you doing anything other than trawling the class tree of ```rt.jar``` and pulling the bytecode from those classes? I've provided a sample method for the first, and I almost certainly can make a sample method for the second. If those are the only two things you're doing all I need to know from there is how you'd like those integrated.

markehammons:
oh i thought i had. one second

olafurpg:
Thank you @markehammons I can't promise when Java 11 support lands but learning about the `jrt:/` file system makes me optimistic we're not too far from it. It just requires some effort to write against that API and setup the CI infrastructure to make sure it continues to work 😉

PR #889, opened 2018-09-22
Remove unused on top-level template statements

### LINKED ISSUE #626: RemoveUnusedTerms does not work on class/object

gabro:
Fixes #626 

This is a workaround that tries to match by pos.start instead of full position.

I'm sure there are edge cases I haven't thought of, wdyt @olafurpg?

/cc @joan38

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -75,6 +75,12 @@ class RemoveUnused(config: RemoveUnusedConfig)
           }
         case i: Defn if isUnusedTerm(i.pos) =>
           defnTokensToRemove(i).map(Patch.removeTokens).asPatch.atomic
+        case i @ Defn.Val(_, List(pat), _, _)
+            if isUnusedTerm.exists(p => p.start == pat.pos.start) =>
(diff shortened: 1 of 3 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -53,8 +53,6 @@ class RemoveUnused(config: RemoveUnusedConfig)
         diagnostic.message.startsWith("local") &&
         diagnostic.message.endsWith("is never used")) {
         isUnusedTerm += diagnostic.position
-      } else {
-        () // ignore
```
gabro: this didn't seem to do much, so I've just removed it

joan38:
Looks good to me

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
olafurpg: It's was more for stylistic reasons but I don't feel strongly about it. I don't like having such a large `if else` without a closing `else`

PR #895, opened 2018-09-28
Add "Browse sources" link

olafurpg:
Live demo: https://olafurpg.github.io/scalafix/


website/siteConfig.js
```diff
@@ -2,6 +2,8 @@
 // site configuration options.
 
 const repoUrl = "https://github.com/scalacenter/scalafix";
+const metadocUrl =
+  "https://scalameta.org/metadoc/#/scalafix/scalafix-core/src/main/scala/scalafix/patch/Patch.scala#L19C23-L19C28";
```
gabro: Is linking directly to Patch intentional?

It's a bit arbitrary if so, but I understand it's as good as any other point to get started.

It would actually be nice to link to sources directly from the specific docs sections.

gabro: Also, is metadoc continuously updated? If not, it would be a bit strange.

gabro:
Nice addition! Now I'm second guessing the idea of having it in the navigation bar. It looks a bit confusing and I don't know what to expect by clicking it... 🤔

website/siteConfig.js
olafurpg: > It would actually be nice to link to sources directly from the specific docs sections.

That's a good idea! But

> Also, is metadoc continuously updated? If not, it would be a bit strange.

It's not, but I have a script that builds scalameta master + scalafix master and re-generates the site. It could be automated with a daily Travis cron job but I think the scalafix/scalameta sources will be stable enough in the coming weeks that a manual approach will do fine for now. Right now I care most about getting metadoc into the hands of users so they can try it out and report feedback.

olafurpg:
I agree it's a bit invasive and unclear what the button leads to. I want to give it a try in the navigation bar just to see what people think. It will take a while to see how metadoc fits into the picture but we have to start somewhere. By moving the link elsewhere we risk people not seeing it.

Issue #894, opened 2018-09-26
Performance improvements in EscapeHatch


marcelocenerine:
Opening this ticket as a follow-up of #812 . The goal here is to identify possible hot-spots in the suppression mechanism and eventually address them.

### Profiling
I profiled `EscapeHatch` using jProfiler as well as manually adding instrumentation (aka `System.currentTimeMillis()`) to the main parts of the code:
  - `@SuppressWarnings` parsing
  - `scalafix:ok` parsing
  - `scalafix:on|off` parsing
  - `EscapeHatch.filter`

For this exercise, I used a very basic syntactic rule that just adds a comment to the top of each source file (doesn't perform tree traversal):
```scala
override def fix(implicit doc: SyntacticDocument): Patch = 
  Patch.addLeft(doc.tree, "/** Foo **/\n").atomic
```

... and ran Scalafix against the `spark` (3152 files) and `scala` (1626 files) repos:
`java -Xms2g -Xmx2g -jar scalafix-cli/target/scala-2.12/scalafix.jar --rules Foo --sourceroot <path to repo>`

Before running Scalafix, I disabled the optimizations implemented in #812 so that I could try out tweaks that benefited code with/without suppressions.

The figures shown below were collected after a single run of Scalafix against each repo (no warmup or anything as such). Proper benchmarking can be done at a later stage (maybe as part of a resulting PR).

#### Baseline:

| Repo   | Scalafix (total)   | `EscapeHatch` (total) | `:ok`         | `:on\|:off`   | `@`          | `filter`      | 
| ------ | ------- | ------------- | ------------- | ------------- | ------------ | ------------- |
| spark  | 61332ms | 29001ms (47%) | 10486ms (17%) | 264ms (0.4%)  | 6772ms (11%) | 11395ms (18%) |
| scala  | 32973ms | 14379ms (43%) | 5513ms (17%)  | 131ms (0.4%)  | 3086ms (9%)  | 5573ms (17%)  |

In both repos, the suppression algorithm accounts for almost 50% (!!!) of the total running time. What is common in the most expensive parts of the algorithm is that they traverse the scala.meta AST: `:on|:off` (once), `@SuppressWarnings` (once) and `filter` (twice).  `scala.meta.transversers.Traverser.apply` is by far the main hot spot:

![image](https://user-images.githubusercontent.com/1107367/46048612-2cd4ea00-c122-11e8-975d-566d95ca1700.png)

[...]

PR #898, opened 2018-10-07
Optimizations in escape hatch

marcelocenerine:
This PR implements some optimizations in `EscapeHatch` (see details in https://github.com/scalacenter/scalafix/issues/894):

- short circuit the parsing logic when there is no suppression in a given file (each suppression method  is short circuited individually). This avoids the cost of traversing the scalameta `Tree`.
- Make atomic empty patches be considered empty. This will potentially maximize the cases where parts or the entire `EscapeHatch` is short circuited
- A few other minor improvements

The tables below compare the running time of what is on `master` vs. this PR + https://github.com/scalameta/scalameta/pull/1793. These figures are the median of 5 consecutive runs of Scalafix against the Spark code base (3201 files; 103181 lines). The setup is the same as explained [here](https://github.com/scalacenter/scalafix/issues/894#issue-363820370):
>The rule utilized for this exercise is purposely simple to reduce noise and make `EscapeHatch` stand out. Different rules will give different results.

- Without suppression:

|        | Scalafix (total)   | `EscapeHatch` (total) | `:ok`         | `:on\|:off`   | `@`          | `filter`      | 
| ------ | ------- | ------------- | ------------- | ------------- | ------------ | ------------- |
| Before  | 50354ms | 17454ms (34%) | 10326ms (20%) | 252ms (0.5%)  | 6795ms (13%) | 0ms (0%) |
| After | 26066ms | 61ms (0%)     | 0ms (0%)      | 0ms (0%)      | 18ms (0%)    | 0ms (0%)  |

- With suppression (all flavors put together in every source file):

|        | Scalafix (total)   | `EscapeHatch` (total) | `:ok`         | `:on\|:off`   | `@`          | `filter`      | 
| ------ | ------- | ------------- | ------------- | ------------- | ------------ | ------------- |
| Before   | 55094ms | 27479ms (50%) | 9885ms (18%) | 342ms (0.6%)  | 6470ms (11%) | 10718ms (19%) |
| After | 30598ms | 11456ms (37%) | 6330ms (20%) | 356ms (1%)    | 2364ms (7%)    | 2362ms (7%)  |

I believe that most projects using Scalafix don't use suppressions often (only exceptionally as it should be). For those, the cost of `EscapeHatch` is now zero. Any source file containing suppressions only pay the price of the chosen suppression method.

As highlighted [here](https://github.com/scalacenter/scalafix/issues/894#issue-363820370), other parts of Scalafix got faster due to the improved tree traverser currently [under review](https://github.com/scalameta/scalameta/pull/1793) in Scalameta. Rules will also benefit from that.


scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -71,30 +67,24 @@ class EscapeHatch private (
...
+          PatchInternals.treePatchApply(underlying).exists { tp =>
+            val byGit = diffDisable.isDisabled(tp.tok.pos)
             val byEscape = isDisabledByEscape(name, tp.tok.pos.start)
             byGit || byEscape
           }
-        }
+        if (hasDisabledPatch) EmptyPatch else underlying
 
-        if (hasDisabledPatch) EmptyPatch
-        else loop(name, underlying)
```
marcelocenerine: it shouldn't be necessary to check whether the underlying patches are disabled if it was already verified that the parent (`AtomicPatch`) contains a disabled patch. This should prevent some expensive tree traversals

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -210,45 +219,37 @@ object EscapeHatch {
...
-      mods.exists {
-        case Mod.Annot(Init(Type.Name(SuppressWarnings), _, _)) => true
-        case Mod.Annot(
-            Init(Type.Select(_, Type.Name(SuppressWarnings)), _, _)) =>
-          true
-        case _ => false
-      }
+    private object SuppressWarningsArgs {
+      def unapply(mods: List[Mod]): Option[List[Term]] =
+        mods.collectFirst {
```
marcelocenerine: no need to traverse all elements since the annotation can only appear once

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -132,17 +125,23 @@ object EscapeHatch {
...
+  private case class EscapeOffset(offset: Int) extends AnyVal
 
   private object EscapeOffset {
     implicit val ordering: Ordering[EscapeOffset] = Ordering.by(_.offset)
   }
 
-  def apply(tree: Tree, associatedComments: AssociatedComments): EscapeHatch =
+  def apply(
+      input: Input,
+      tree: => Tree,
```
olafurpg: Note: my experience with by-name is that it's easy to accidentally trigger the value. You can also use `() => Tree` or the `LazyValue[T]` in https://github.com/scalacenter/scalafix/blob/18650f572178ab1fa461846604bd82f9fed996ac/scalafix-core/src/main/scala/scalafix/internal/v1/LazyValue.scala which has the benefit of being a normal type and avoids accidental double computations. I prefer explicit `tree.value` or `tree()` to make it clearer when the function is called.

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -180,17 +184,22 @@ object EscapeHatch {
   }
 
   private object AnnotatedEscapes {
-    private val SuppressWarnings = "SuppressWarnings"
+    private val SuppressWarnings = classOf[SuppressWarnings].getSimpleName
```
olafurpg: Is this necessary? I would avoid runtime reflection when possible even if that means hardcoding a string literal, the name is not going to change in the JDK anytime soon anyways.

scalafix-tests/unit/src/test/scala/scalafix/tests/core/EscapeHatchSuite.scala
```diff
@@ -0,0 +1,151 @@
...
+import scalafix.patch.Patch
+import scalafix.rule.RuleName
+import scalafix.v0.SemanticdbIndex
+import scalafix.v1.SyntacticDocument
+
+import scala.meta.{Source, Tree}
+import scala.meta.contrib.AssociatedComments
+import scala.meta.inputs.Input
+
+class EscapeHatchSuite extends FunSuite {
```
olafurpg: 😍 nice tests

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -95,4 +96,20 @@ class PatchSuite extends SyntacticRuleSuite {
       |object A
       |""".stripMargin
   )
+
+  val tree = Input.String(original).parse[Source].get
+
+  test("Patch.empty") {
+    val empty = Patch.empty
+    assert(empty.isEmpty)
+    assert(empty.atomic.isEmpty)
```
olafurpg: 👍

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -23,13 +24,13 @@ import scalafix.v0._
  * precedence over the former in case there are overlaps.
  * See `AnchoredEscapes` and `AnnotatedEscapes` for more details.
  */
-class EscapeHatch private (
+private[scalafix] class EscapeHatch private (
```
olafurpg: Note: it's generally not necessary to mark things private in `scalafix.internal` unless it helps readability with highlighting the key components from implementation details

marcelocenerine: Reverted

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: fair enough. Reverted

scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
marcelocenerine: done

olafurpg:
We can release scalafix v0.9.1 together with scalameta v4.1.0

PR #907, opened 2018-11-13
Improve regex integration with DisableSyntax 

Jacoby6000:
I think the best way to see what I've changed is to look at the [updated docs](https://github.com/scalacenter/scalafix/blob/64e39aaa13120504dfb4ea927901ef5f4c528358/docs/rules/DisableSyntax.md).  To summarize:

1. Now error messages will underline the whole area of matched regex
2. The relevant capture group can be specified, in case you have complicated regex's which may match on parts which aren't helpful in the error message
3. scalafix errors which came from a regex match have access to the capture groups for substitution in the error messages.

I'm not dead set on any of the naming or formatting I've chosen.  Feel free to have me change them

### DIFF
--- a/docs/rules/DisableSyntax.md
+++ b/docs/rules/DisableSyntax.md
@@ -55,16 +55,17 @@ DisableSyntax.regex = [
 ]
 ```
 
-1. The first way has a simple object providing an `id`, `pattern`, and `message`.
+1. The first way has an object providing an `id`, `pattern`, and `message`.
 2. The second way is just the pattern.  When this is used, the `id` is set equal 
    to the pattern, and a generic message is provided for you.
-3. The third way allows you to specify what capture-group the problematic thing 
-   is in, in case your regex is complicated.
+3. The third way allows you to specify what capture-group the problematic piece 
+   of code is in, in case your regex is complicated and also matches characters 
+   not useful in an error message.
 
 ### Error Messages  
 
-Error messages have access to the capture groups of the regex. Simply use `{$n}` 
-where `n` is the index of the capture group you wish to appear in that part of 
-the message.
+Error messages have access to the capture groups of the regex. To access the 
+capture groups of the regex, use `{$n}` where `n` is the index of the capture 
+group you wish to appear in that part of the message.
 
 You can see this used in the 3rd example.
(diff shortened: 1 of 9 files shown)


Jacoby6000:
My current example is the `magicNumber` thing, and I don't like it.  Just couldn't come up with a generic rule that has more complex regex.

docs/rules/DisableSyntax.md
```diff
@@ -31,3 +31,40 @@ println(
...
+    regex = {
+      pattern = "(?:(?:[,(]\\s)|(?:^\\s*))+(\\d+(\\.\\d+)?)"
+      captureGroup = 1
+    }
+    message = "Numbers ({$1} in this instance) should always have a named parameter attached, or be assigned to a val."
+  }
+]
+```
+
+1. The first way has a simple object providing an `id`, `pattern`, and `message`.
```
olafurpg: s/simple//

docs/rules/DisableSyntax.md
```diff
@@ -31,3 +31,40 @@ println(
...
+2. The second way is just the pattern.  When this is used, the `id` is set equal 
+   to the pattern, and a generic message is provided for you.
+3. The third way allows you to specify what capture-group the problematic thing 
+   is in, in case your regex is complicated.
+
+### Error Messages  
+
+Error messages have access to the capture groups of the regex. Simply use `{$n}`
```
olafurpg: > To access to the capture groups of the regex use `{$n}` ...

scalafix-tests/input/src/main/scala/test/disableSyntax/DisableSyntaxBase.scala
```diff
@@ -52,8 +60,12 @@ case object DisableSyntaxBase {
     def -(other: String): String = s"$value - $other"
   }
 
+  5 // assert: DisableSyntax.magicNumbers
```
olafurpg: Can we assert the contents of the message here?
```scala
 5 /* assert: DisableSyntax.magicNumbers
 ^
 Numbers (5 in this instance) should always ...
*/
```
See end of this section https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#use-diagnostic-to-report-linter-errors

Jacoby6000:
Thanks for the feedback.  I believe I've addressed all the comments

PR #911, opened 2018-11-28
Add support for Java 11

### LINKED ISSUE #880: Scalafix does not work with java 11
When trying to run scalafix on java 11 I get the following error: ```error: Unable to load symbol table```

Having looked into it, it looks like the issue is related to these two files: 

This one seems to be the source of the error message:
https://github.com/scalacenter/scalafix/blob/master/scalafix-testkit/src/main/scala/scalafix/testkit/SemanticRuleSuite.scala

This one seems to be the cause of the error:
https://github.com/scalacenter/scalafix/blob/47e25af12e014017fd7cb4a519ffc51639ea937a/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala

In ClasspathOps.scala, you have some code that looks like this:

```scala
  def bootClasspath: Option[Classpath] = sys.props.collectFirst {
    case (k, v) if k.endsWith(".boot.class.path") => Classpath(v)
  }
```

```sun.boot.class.path``` doesn't exist in java 9+ as you can see in this stack overflow issue: https://stackoverflow.com/questions/50451536/how-to-get-boot-class-path-in-jdk-9-or-later

The solution is listed there as well, you need to look for ```java.class.path```. The fixed code should look something like this: 

```scala
  def bootClasspath: Option[Classpath] = sys.props.collectFirst {
    case (k, v) if k.endsWith(".boot.class.path") => Classpath(v)
    case (k,v) if k == "java.class.path" => Classpath(v)
  }
```

I tried to download and build scalafix to test my theory, but I ran into a number of issues with the build process. It seems like scala.predef is disabled in my builds somehow, but a ton of classes rely on the predef methods, like turning java streams into lists, calling ```.foreach``` on java streams, and other things. In any case, I could not get the project to build so I could test my code.

olafurpg:
This was enabled by the upgrade to Scalameta v4.1.0.

Fixes #880

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -6,4 +6,3 @@
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
 -XX:+CMSClassUnloadingEnabled
--XX:+UseConcMarkSweepGC
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ jobs:
       script: sbt ci-211
     - env: TEST="2.12"
       script: sbt ci-212
+    - jdk: openjdk11
+      env: TEST="2.12"
+      script: sbt ci-212
 
     # Disabled until stable v0.6
     # - env: TEST="mima"
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,13 @@ noPublish
 
 // force javac to fork by setting javaHome to get error messages during compilation,
 // see https://github.com/sbt/zinc/issues/520
-def inferJavaHome() =
-  Some(file(System.getProperty("java.home")).getParentFile)
+def inferJavaHome() = {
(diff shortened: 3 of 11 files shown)


markehammons:
Nice! I'll try it out some time!

glammers1:
Sorry for comment in a merged PR but I have a question.

Should the installation.md be updated?

> Java 8. Scalafix supports only Java 8 at the moment. Running Scalafix on Java 9+ results in error: unable to load symbol table errors. See #880 for updates on adding support for Java 11.

Latest release note:

> Scalafix now runs on Java 11. Our CI additionally runs the full build on Java 11 for every PR to prevent regressions.

I don't know if more work is needed in the `java-11-compat` branch.

PR #923, opened 2018-12-16
Remove Importee.Rename if it does not shadow anything

### LINKED ISSUE #915: error fix:unused import
before
```
import akka.actor.typed.scaladsl.{ ActorContext ⇒ SActorContext }
```
after
```
import akka.actor.typed.scaladsl.{ ActorContext ⇒ _ }
```
source :akka.actor.typed.TypedBenchmarkActors

marcelocenerine:
Fixes #915 

/cc @hepin1989


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -39,55 +40,102 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
       Patch.empty
     } else {
-      doc.tree.collect {
-        case i: Importee if isUnusedImport(importPosition(i)) =>
-          i match {
-            case Importee.Rename(_, to) =>
+      fix(doc.tree, unusedTermsPosition.toSet, unusedImportsPosition.toSet)
+    }
+  }
+
```
marcelocenerine: Algorithm:
- traverses the tree using depth-first traversal
- for each enclosing scope (source, package, object, class, def, block, ...)
  - any used import wildcards AND unused import renames are tracked and passed down to nested scopes following them
  - any other unused import gets removed
  - at the end of each scope
    - any unused import rename (from outer our current scope) whose package is the same as any wildcard import in scope (outer or current) gets unimported
    - any remaining unused import rename belonging exclusively to the current scope (not the outer) gets removed

marcelocenerine:
>I am a bit concerned about the increased complexity from this change

Indeed it adds some complexity to handle the case where the rename and wildcard are in different `import` clauses. If the original ask (#614) was only about unused renames in the same `import` as the wildcard, then your suggestion is good enough.

marcelocenerine:
@olafurpg do you prefer the simpler approach and restrict the fix for the original issue (#614) only to unused renames in the same import as the wildcard?

olafurpg:
I think the simple solution fixes also https://github.com/scalacenter/scalafix/issues/915 

In the diff here https://github.com/scalacenter/scalafix/commit/38db60ce9b7339f4778923e83e405951789ab04f#diff-5a5912bc5da72f0e5da63e0919c8d2d0 for example it removes ` import scala.io.StdIn.{readByte => _}` imports

marcelocenerine:
Hey @olafurpg, many of those test cases were more relevant to the original algorithm. I added most of them back, except the ones related to scoping. The new implementation doesn't care about scopes

dwijnand:
Thanks, @marcelocenerine!

PR #924, opened 2018-12-19
Add beforeStart and afterComplete methods on Rule

rorygraves:
Rules can then implement initialisation and cleanup behaviors before and after
all semantic documents have been seen.


rorygraves:
N.b. SyntaticRules test suite does not currently call these methods - as it appears to use the v0 rules and I'm not sure of the semantics we are trying to maintain with the split.

There is currently no test for this - happy to take suggests on what shape the test should take.

rorygraves:
Pushed formatted version.

rorygraves:
@olafurpg 
I am unable to run ```sbt mimaReportBinaryIssues```
I get unresolved errors on 
```bt.librarymanagement.ResolveException: unresolved dependency: ch.epfl.scala#scalafix-cli_2.12.7;0.5.0: not found```

olafurpg:
@rorygraves thank you for your this contribution! I'll be away until January so I'm afraid I won't be able to review this until then, I'm sorry.

olafurpg:
@rorygraves Do you have any thoughts on where rules should dump their metadata? I'm thinking if we should add `BeforeStart` and `AfterComplete` parameters to `beforeStart/afterComplete` that could potentially have information like an output directory for rules to dump metadata. We can add more information to those context types down the road without breaking binary compatibility.

rorygraves:
I think the parameters are a great idea. 
I don’t have a strong idea where that metadata should be written.

olafurpg:
I think this a great idea but I am afraid I won't have time to thoroughly review the design and tests for this feature addition in the near future as I'm busy with Metals. I don't want to rush feature additions to the core API since I take merged features as a commitment to support them (ideally) forever.

One workaround the lack of this feature is to write another `main` function that calls the Scalafix main function and then `afterComplete` once it return. The `withConfiguration` method can be used instead of `beforeStart`. Alternatively, you can maintain a fork. It would be valuable to get some real-world experience on the design before adding it to the core API.

PR #960, opened 2019-04-26
Upgrade to Scalameta 4.1.9

olafurpg:
This release merged a lot of modules into one so that we no longer have
to depend on micro-libraries like metacp, metap or symtab.


olafurpg:
Ignoring flaky resolution error in AppVeyor

PR #981, opened 2019-09-03
expose all synthetic semantic trees

ingarabr:
The optimized lookup only exposed the last tree for a given position. A position can have multiple synthetic semantic trees. This occurs when we have a function with only an implicit argument.

Given the code:
```
object MyObject {
  trait Value[A] { def value: A }
  implicit val intValue: Value[Int] = new Value[Int] { def value: Int = 1 }
  class Foo[A]() {
    def baz[AA >: A](implicit s: Value[AA]): AA = s.value
  }

  val foo = new Foo[Int]()
  foo.baz
}
```

The tree for the `foo.baz` looks like this:
```
Term.Select(Term.Name("foo"), Term.Name("baz"))
```

When calling `term.synthetic` on the `Term` we're not able to extract the implicit arguments. If we add parentheses, `foo.baz()`, it will be wrapped in a `Term.Apply`, where the implicit is available. When querying the semantic tree from the given position we're ending up with two results where scalafix has optimized the lookup and expects at most one result. 

This change exposes all the semantic trees that are available in the given position. It also depricates the API where we're expecting zero or one result.


scalafix-core/src/main/scala/scalafix/v1/package.scala
```diff
@@ -12,18 +12,32 @@ package object v1 extends Api {
...
+      doc.internal.synthetic(term.pos).headOption
+
+    def synthetics(implicit doc: SemanticDocument): List[SemanticTree] =
       doc.internal.synthetic(term.pos)
   }
   implicit class XtensionTermInfixScalafix(infix: Term.ApplyInfix) {
+    @deprecated("0.9.7", "Use syntheticOperators")
     def syntheticOperator(
         implicit doc: SemanticDocument
     ): Option[SemanticTree] = {
```
oyvindberg: `syntheticOperators.headOption`

PR #1070, opened 2020-03-07
convert empty values to None or Nil in ExplicitResultType rule

### LINKED ISSUE #990: ExplicitResultTypes should convert Option.empty[T] into None
The ExplicitResultTypes rewrite currently produces the following diff
```diff
-   var focusedDocument = Option.empty[AbsolutePath]
+   var focusedDocument: Option[AbsolutePath] = Option.empty[AbsolutePath]
```

I expected the rewrite to produce the following diff

```diff
-   var focusedDocument = Option.empty[AbsolutePath]
+   var focusedDocument: Option[AbsolutePath] = None
```

I would expect the same behavior for `List.empty[Int]` becoming `Nil`.

taisukeoe:
Fixes #990

Thank you all for your hard work to make our migrations / linting easier!

This includes the following value conversion in ExplicitResultType rule:

- from Option.empty[T] to None
- from List.empty[T] to Nil
- from Seq.empty[T] to Nil

While I haven't come up with any problem if Seq.empty[T] value is converted to Nil as long as its type is annotated as Seq[T], please let me know if there is any.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
@@ -119,6 +119,12 @@ final class ExplicitResultTypes(
     case Defn.Def(_, name, _, _, _, _) => name
   }
 
+  def defnBody(defn: Defn): Option[Term] = Option(defn).collect {
+    case Defn.Val(_, _, _, term) => term
+    case Defn.Var(_, _, _, Some(term)) => term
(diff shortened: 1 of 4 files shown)


olafurpg:
Thank you for this contribution! LGTM 👍 I pushed a commit to make sure this special case works as expected even when `Option` does not point to `scala.Option`

olafurpg:
Triggered a new 0.9.12 release with this change 😄

taisukeoe:
Thank you for your reviewing and pointing out the special case!

Issue #985, opened 2019-09-29
Runtime support for Java 11


NicolasRouquette:
As mentioned in the JDK compatiblity notes (https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html), it is important to support JDK 11 since it is an LTS release.

As far as scalafix is concerned, there is at lease an issue in `ClasspathOps` due to casts to `URLClassLoader`, e.g.:

```scala
  def thisClassLoader: URLClassLoader =
    this.getClass.getClassLoader.asInstanceOf[URLClassLoader]
```

This is one of the known migration problems described here:
https://blog.codefx.org/java/java-11-migration-guide/#Casting-To-URL-Class-Loader

There are some possible workarounds:
http://java9.wtf/class-loading/
https://stackoverflow.com/questions/46694600/java-9-compatability-issue-with-classloader-getsystemclassloader

I wonder if it is necessary to have JDK-specific `ClasspathOps` code to handle the differences in JDK8 vs. JDK9+.

RicoGit:
Any progress about this? Docs says that java11 is supported.
https://scalacenter.github.io/scalafix/docs/users/installation.html

olafurpg:
The JDK 11 issues should be fixed in the latest release v0.9.13. Our CI runs tests on JDK 11. Please open a new ticket if the problem persists

PR #1080, opened 2020-03-19
scalafix_cli - jdk11

dos65:
Fix for #985

### DIFF
--- a/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
+++ b/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
@@ -3,6 +3,7 @@ package scalafix.internal.reflect
 import java.io.File
 import java.io.OutputStream
 import java.io.PrintStream
+import java.net.URL
 import java.net.URLClassLoader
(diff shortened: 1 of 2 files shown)


scalafix-reflect/src/main/scala/scalafix/internal/reflect/ClasspathOps.scala
```diff
@@ -29,15 +29,16 @@ object ClasspathOps {
   }
 
   def thisClassLoader: URLClassLoader =
-    this.getClass.getClassLoader.asInstanceOf[URLClassLoader]
+    new URLClassLoader(Array.empty, this.getClass().getClassLoader)
```
olafurpg: We can pass `this.getClass.getClassLoader` to `getURLs` from the method here https://github.com/scalameta/metals/blob/2eb7a239a2e4e379c794ae2d66c6977ac9014bda/mtags/src/main/scala/scala/meta/internal/mtags/ClasspathLoader.scala#L14-L52

That should preserve the behavior on JDK 8 and 9+

dos65: Thanks! Fixed

PR #1081, opened 2020-03-19
Migrate most tests to GH Actions for CI

olafurpg:


olafurpg:
The Windows CI failure looks legitimate, I opened https://github.com/scalameta/scalameta/pull/2007 which should fix it

PR #1090, opened 2020-04-13
Update sbt and coursier

ckipp01:
This pr updates sbt from 1.2.8 to 1.3.9. Tests all seemed to pass locally as well as publishing locally. Since we were using an sbt version prior to 1.3.x, I'm assuming that's why we were explicitly using `coursier-sbt`. I therefore removed that dependency on the plugin.

Supercedes #1085


ckipp01:
Missed some relevant stuff. Looking into it.

ckipp01:
I'll need to spend some more time on this 🤔 ...

I continually get the following
```
error: error while loading Object, Missing dependency 'object scala.native in compiler mirror', required by /Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/rt.jar(java/lang/Object.class)
```
If found some information about how forking can help this, [see here](https://github.com/sbt/sbt/issues/2958#issuecomment-568496384), but then I have issues with it not finding the sematic information it needs in tests.

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -82,7 +82,7 @@ class ScalafixImplSuite extends FunSuite with DiffAssertions {
   }
 
   test("error") {
-    val cl = new URLClassLoader(Array())
+    val cl = new URLClassLoader(Array(), null)
```
ckipp01, line 85: Ahhh, I tried the `fork := true` up above like you did, but didn't catch this one down here.

olafurpg:
Thank you for this contribution! It wasn't at all obvious what was causing those errors. The Scalafix test suite previously relied on both 

* the shape of `this.getClass.getClassLoader` to match exactly the layered classloader from sbt v1.2.x
* the working directory to be the root of the build

The last commit enables forking and fixes the build to use a correct working directory as well as fixes a test that was written in a fragile way.

scalafix-tests/unit/src/test/scala-2.12/scalafix/tests/cli/ScalafixImplSuite.scala
olafurpg: It was also necessary to update `baseDirectory.in(Test) := baseDirectory.in(ThisBuild).value`. That should honestly be the default behavior in sbt, but changing the default would break a lot of builds

PR #1094, opened 2020-04-14
include semanticEnabled settings in installation

taisukeoe:
Although I might miss 0.9.14 release, please let me send a PullRequest to include this change to be included in an installation guide column.
https://github.com/scalacenter/sbt-scalafix/pull/89

It would be nicer to be documented,  although I would think we needn't to hurry it.

### DIFF
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -82,7 +82,7 @@ following settings to `build.sbt`.
  )
 ```
 
-- sbt 1.2 or less
+- all sbt versions
 
 ```diff
  // build.sbt
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -63,10 +63,9 @@ is not enabled for this project. The second error says `RemoveUnused` requires
 the Scala compiler option `-Ywarn-unused`. To fix both problems, add the
 following settings to `build.sbt`. 
 
-- sbt 1.3.x or above
 
 ```diff
- // build.sbt
+ // build.sbt, for sbt 1.3x and newer
  inThisBuild(
    List(
      scalaVersion := "@SCALA212@", // @SCALA211@, or @SCALA213@
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -66,7 +66,7 @@ following settings to `build.sbt`
 ```diff
  // build.sbt
  lazy val myproject = project.settings(
-   scalaVersion := "@SCALA212@", // or @SCALA211@
+   scalaVersion := "@SCALA212@", // @SCALA211@, or @SCALA213@
 +  addCompilerPlugin(scalafixSemanticdb), // enable SemanticDB
    scalacOptions ++= List(
 +    "-Yrangepos",          // required by SemanticDB compiler plugin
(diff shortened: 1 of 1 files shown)


docs/users/installation.md
```diff
@@ -75,6 +75,22 @@ following settings to `build.sbt`
...
+> To enable SemanticDB related settings for all projects in your build.sbt, 
+> use the following setting which is available from Scalafix 0.9.14 and sbt 1.3.x . 
+> 
+> ```diff
+>  // build.sbt
+> inThisBuild(
+>  List(
+>    scalaVersion := "@SCALA212@", // @SCALA211@, or @SCALA213@
+> +  semanticdbEnabled := true,
+> +  semanticdbIncludeInJar := true,
```
olafurpg, line 87: Let's recommend setting this to false by default.

docs/users/installation.md
```diff
@@ -75,6 +75,22 @@ following settings to `build.sbt`
  )
 ```
 
+> To enable SemanticDB related settings for all projects in your build.sbt,
```
olafurpg, line 78: I think it would be good to recommend this as the default installation method and document the other approach for users on sbt <v1.3. What do you think?

docs/users/installation.md
taisukeoe: You're right.  Let's do it.
(I might have something wrong in my local project, which required `semanticdbIncludeInJar := true`  ;) )

docs/users/installation.md
taisukeoe: I'm a bit worrying that ThisBuild settings introduce SemanticDB settings across the projects, some of which may not require SemanticDB but incur overhead in compilation time.

Disclaimer: I have not faced the case the overhead is critical. Do you think it would be enough minor?

taisukeoe: In case ThisBuild setting becomes as default, it would be better to modify an error message to ThisBuild scoped settings. 

https://github.com/scalacenter/sbt-scalafix/blob/ae173f44cf4875e9b1d4a172d9ce7d76e2e7076e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L202-L203

olafurpg: I agree, we can do that separately

taisukeoe: Okay, I can do it.
I'll update this PR with mentioning about this risk as well.

> I'm a bit worrying that ThisBuild settings introduce SemanticDB settings across the projects, some of which may not require SemanticDB but incur overhead in compilation time.

taisukeoe: I've sent a PullRequest to change the error message.
https://github.com/scalacenter/sbt-scalafix/pull/90

taisukeoe:
I've reflected review comments. Would you please review it?

docs/users/installation.md
```diff
@@ -61,12 +61,33 @@ The first error message means the
...
+   scalacOptions ++= List(
++    "-Ywarn-unused-import" // required by `RemoveUnused` rule
+   )
+ )
+```
+
+- all sbt versions
 
 ```diff
  // build.sbt
```
olafurpg, line 88:
```suggestion
```diff
 // build.sbt, for sbt 1.2.x and older
```

taisukeoe: IMO it would be nicer to have *explicitly sbt 1.3.x compatible* sub-project specific setting example for new comers. So I'd just add another example.

taisukeoe:
Thank you for reviewing. I've reflected all comments.

docs/users/installation.md
```diff
@@ -60,13 +60,45 @@ RemoveUnused ...
...
+
+ lazy val myproject = project.settings(
+   scalacOptions += "-Ywarn-unused-import" // required by `RemoveUnused` rule
+ )
+```
+
+```diff
+ /*
+  * build.sbt, for sbt 1.3x and newer
+  * SemanticDB is enabled only for a sub-project.
```
olafurpg, line 88: It's really helpful to separate the "for all projects" and "for one project" examples. It's not obvious for sbt beginners how to do this 👍

PR #1095, opened 2020-04-14
Small update to intro docs

### LINKED ISSUE #1089: "-Ywarn-unused" does not work
When I tried to follow [this documentation](https://scalacenter.github.io/scalafix/docs/users/installation.html) I realized that _-Ywarn-unused_ actually should be _-Ywarn-unused:imports_, at least with Scala 2.13.1.

Should this be changed in the documentation?

ckipp01:
This pr just adds in a small note to remind users that may be new to scalafix that the `-Ywarn-unused` options differs a bit between scala versions (although that info is in the `RemoveUnused` section.

I also updated docusaurus to the latest, built it locally and ensured that things still look good.

closes #1089

### DIFF
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -11,7 +11,7 @@ pull request is tested on both Linux and Windows.
 **Java 8 or Java 11.**
 
 **Scala 2.11, 2.12 or 2.13**: Scalafix works only with the latest version of Scala
-2.11,  Scala 2.12 and Scala 2.13. 
+2.11,  Scala 2.12 and Scala 2.13.
 
 | Scalafix  | Scala Compiler                       | Scalameta   |
 | --------- | ------------------------------------ | ----------- |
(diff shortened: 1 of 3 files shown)


website/i18n/en.json
```diff
@@ -43,6 +43,9 @@
       "rules/DisableSyntax": {
         "title": "DisableSyntax"
       },
+      "rules/ExplicitResultTypes": {
```
ckipp01, line 46: This was just auto-generated since it was missing.

docs/users/installation.md
```diff
@@ -60,8 +60,9 @@ RemoveUnused ...
 The first error message means the
 [SemanticDB](https://scalameta.org/docs/semanticdb/guide.html) compiler plugin
 is not enabled for this project. The second error says `RemoveUnused` requires
-the Scala compiler option `-Ywarn-unused`. To fix both problems, add the
-following settings to `build.sbt`
+the Scala compiler option `-Ywarn-unused` (Keep in mind that this compiler
```
olafurpg, line 63:
```suggestion
the Scala compiler option `-Ywarn-unused-import` (or `-Wunused:imports` in 2.13.x)
```

olafurpg: Let's provide the exact option to use in 2.13. Most users only care about removing unused imports, not other unused expressions. I sometimes regret having merged the rule to remove unused imports with the rule to remove other unused expressions..

ckipp01: I suppose, being explicit is probably better here. Changed this 👍

PR #1105, opened 2020-04-24
Fix #1104

### LINKED ISSUE #1104: Can't get symbol from TextDocument for arg of ApplyInfix which have parens around
```scala
trait SymbolTest {
  def shouldBe(right: Any)
  def arg = 1
  this shouldBe (arg)
}
```
```scala
val arg :: Nil = doc.tree.collect {
  case Term.ApplyInfix(_, Term.Name("shouldBe"), _, arg :: Nil) => arg
}
assertNotEquals(arg.symbol, Symbol.None)
```
assertNotEquals fail

giabao:
Fix #1104

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/util/EagerInMemorySemanticdbIndex.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/util/EagerInMemorySemanticdbIndex.scala
@@ -5,6 +5,7 @@ import scala.meta._
 import scala.meta.internal.symtab._
 import scala.meta.internal.{semanticdb => s}
 import scalafix.internal.v0._
+import scalafix.internal.v1.TreePos
 import scalafix.util.SemanticdbIndex
 import scalafix.v0
 import scalafix.v0._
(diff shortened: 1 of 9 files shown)


ohze:
Given:
```scala
0 trait SymbolTest {
1   def shouldBe(right: Any)
2   def arg = 1
3   this shouldBe (arg)
4 }               ↑   ↑
                 16   21
```
We [use](https://github.com/scalacenter/scalafix/blob/3ab7437/scalafix-core/src/main/scala/scalafix/internal/v1/InternalSemanticDoc.scala#L120) `textDocument.occurrences` produced by `semanticdb-scalac` to get symbols (String, eg `(arg)`) from a Range/ Position (eg, `[3:16..3.21)`)

To workaround for scalameta/scalameta#1083, when [get Position](https://github.com/scalacenter/scalafix/blob/02e606d/scalafix-core/src/main/scala/scalafix/internal/v1/TreePos.scala#L7) for name `(arg)`, instead of returning `[3:16..3.21)`, we return `[3:17..3.20)` and use that Position to [lookup](https://github.com/scalacenter/scalafix/blob/3ab7437/scalafix-core/src/main/scala/scalafix/internal/v1/InternalSemanticDoc.scala#L53) symbols from `occurrences`.

But in `textDocument.occurrences` produced by `semanticdb-scalac`, there is only `SymbolOccurrence(range = [3:16..3.21), symbol = "(arg)")` (have no Occurrence with range == `[3:17..3.20)`). So the lookup will return None.

Code changes in this PR: Only [re-try getting](https://github.com/scalacenter/scalafix/pull/1105/files#diff-66059b546ef1029435a93b743eb0f29eR20) symbol from range `[3:16..3.21)` when workingaround scalameta/scalameta#1083.

ohze:
> Is there a relationship between this change and [scalameta/scalameta#2043](https://github.com/scalameta/scalameta/pull/2043) ?

No. The bug resolved in this PR have no relation to implicit conversion as in scalameta/scalameta#2043.

Given:
```scala
trait AnyShouldWrapper {
  def shouldBe(right: Any)
}
trait SymbolTest {
  implicit def convert(o: Any)(implicit foo: Int): AnyShouldWrapper
  implicit val foo: Int = 1
  def arg = 1
  this shouldBe (arg)
}
```
In the last statement `this shouldBe (arg)`:
+ This PR: Getting symbol for name `(arg)` return Some(symbol), not None.
+ scalameta/scalameta#2043: Getting symbol for `this` return `SymbolTest#this().`, not `SymbolTest#convertToAnyShouldWrapper().`

olafurpg:
Please let me know if you're blocked by this fixed and need a new release. I would prefer to wait a bit longer to cut a new release since it always takes a while to synchronize scalafix-core/sbt-scalafix

PR #1114, opened 2020-05-06
improve RemoveUnused rule

xuwei-k:
I think `importees` can be remove if `forall(_.is[Importee.Unimport])`

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -67,6 +67,9 @@ class RemoveUnused(config: RemoveUnusedConfig)
       Patch.empty
     } else {
       doc.tree.collect {
+        case Importer(_, importees)
+            if importees.forall(_.is[Importee.Unimport]) =>
+          importees.map(Patch.removeImportee).asPatch
         case Importer(_, importees) =>
           val hasUsedWildcard = importees.exists {
             case i: Importee.Wildcard => !isUnusedImport(importPosition(i))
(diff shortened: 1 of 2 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -67,6 +67,9 @@ class RemoveUnused(config: RemoveUnusedConfig)
       Patch.empty
     } else {
       doc.tree.collect {
+        case Importer(_, importees)
+            if importees.forall(_.is[Importee.Unimport]) =>
+          importees.map(Patch.removeImportee).asPatch
```
liancheng, line 72: I wonder is it safe to remove any unimports without an accompanying `_`? E.g.:

```scala
import p.{a, b => _}
import p.{c => c1, d => _}
```

I think it's also safe to remove `b => _` and `d => _` above?

The Scala language spec is a little bit vague about the semantics of unimports (see [here][1]).

> If the target in an import selector is a wildcard, the import selector hides access to the source member. For instance, the import selector 𝑥 => _ “renames” 𝑥 to the wildcard symbol (which is unaccessible as a name in user programs), and thereby effectively prevents unqualified access to 𝑥. *This is useful if there is a final wildcard in the same import selector list, which imports all members not mentioned in previous import selectors.*

The last sentence says "this is useful ...", not "this is _only_ useful ...". It makes me wonder whether a "standalone" unimport without an accompanying `_` is meaningful at all.

A quick test shows that they seem to be meaningless:

```scala
package p {
  object a
}

object X {
  import p.a
  import p.{a => _}

  val x = a
}

object Y {
  import p._
  import p.{a => _}

  val x = a
}
```

The above snippet compiles successfully, which is quite expected. Otherwise, it means that unimports are order-sensitive.

I'm asking this partly because I'm trying to answer this question for handling unimports in the [`OrganizeImports` rule][2].

[1]: https://scala-lang.org/files/archive/spec/2.13/04-basic-declarations-and-definitions.html#import-clauses
[2]: https://github.com/liancheng/scalafix-organize-imports/pull/50/files

bjaglin: That looks like another scenario where the rule could be be more aggressive indeed. @xuwei-k any feedback on this? We could merge this as-is and handle your scenario in another PR @liancheng.

PR #1115, opened 2020-05-07
RemoveUnused: document 2.13.2 -Wconf to keep `-Xfatal-warnings`

bjaglin:
I have successfully used https://github.com/scala/scala/pull/8373 on my build to keep/relax `-Xfatal-warnings`, and delegate unused imports fix/enforcement from scalac to scalafix.

The only impact of `-Wconf:cat=unused:info` on semanticdb-scalac 4.3.10 seems to be the `severity` attribute, so the rule works as-is.
```
  diagnostics {
    range {
      start_line: 2
      start_character: 17
      end_line: 2
      end_character: 23
    }
    severity: INFORMATION
    message: "Unused import"
  }
```


docs/rules/RemoveUnused.md
```diff
@@ -17,8 +17,9 @@ To use this rule:
 - Enable the Scala compiler option `-Ywarn-unused` (or `-Wunused` in 2.13). In
   sbt, this is done with `scalacOptions += "-Ywarn-unused"`.
 - Disable `-Xfatal-warnings` if you have it enabled. This is required so the
-  compiler warnings do not fail the build before running Scalafix.
-  Unfortunately, it's not possible exclude unused imports from -Xfatal-warnings.
+  compiler warnings do not fail the build before running Scalafix. If you are
+  running 2.13.2 or later, you may keep `-Xfatal-warnings` by modifying how
+  specific warnings are handled via `scalacOptions += "-Wconf:cat=unused:info"`.
```
bjaglin, line 22: I also tried
```suggestion
  specific warnings are handled via `scalacOptions += "-Wconf:cat=unused:silent"`.
```

but it does what you can expect: the diagnostic is gone!

olafurpg: Interesting, I suppose that makes sense! The info messages are anyways a good reminder that you have unused imports that can be removed

Issue #1110, opened 2020-05-01
Scala fix fails to run on test


daddykotex:
Hi, thanks for the project, it's truly helpful!

I'm working hard to add this as part of our CI but I can't figure out why I can't run it against our test sources. At first I thought it might be the dependency on scalatest, but it's not. A very dummy repo reproduces:

https://github.com/daddykotex/scalafix-issue-repro

```
sbt 1.3.10
addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.15")
scala := 2.13.1
```

I'm getting a bunch of error when I run `test:scalafix --check`:

```
[info] Compiling 1 Scala source to /Users/david/workspace/scalafix-test/module1/target/scala-2.13/test-classes ...
[warn] there was one deprecation warning (since 2.13.0); re-run with -deprecation for details
[warn] one warning found
java.io.IOException: Scala signature package has wrong version
 expected: 5.0
 found: 5.2 in package.class
	at scala.reflect.internal.pickling.UnPickler$Scan.checkVersion(UnPickler.scala:124)
	at scala.reflect.internal.pickling.UnPickler$Scan.<init>(UnPickler.scala:66)
	at scala.reflect.internal.pickling.UnPickler.unpickle(UnPickler.scala:47)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.unpickleOrParseInnerClasses(ClassfileParser.scala:1176)
  | => mat scala.tools.nsc.symtab.classfile.ClassfileParser.parseClass(ClassfileParser.scala:465)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.$anonfun$parse$2(ClassfileParser.scala:161)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.$anonfun$parse$1(ClassfileParser.scala:146)
	at scala.tools.nsc.symtab.classfile.ClassfileParser.parse(ClassfileParser.scala:129)
	at scala.tools.nsc.symtab.SymbolLoaders$ClassfileLoader.doComplete(SymbolLoaders.scala:329)
	at scala.tools.nsc.symtab.SymbolLoaders$SymbolLoader.complete(SymbolLoaders.scala:230)
	at scala.reflect.internal.Symbols$Symbol.info(Symbols.scala:1539)
	at scala.reflect.internal.SymbolTable.openPackageModule(SymbolTable.scala:355)
[...]

Issue #998, opened 2019-11-30
Make ExplicitResultTypes work with more Scala versions


olafurpg:
Unlike other rules, `ExplicitResultTypes` only works with Scala 2.12.10 at the moment because the rule interfaces directly with the presentation compiler. It would be great if we could add support for Scala 2.13 and 2.11 but doing so would require rethinking how rules are compiled/published/launched.

* should Scalafix rules be cross-built against multiple Scala versions if they want to use the presentation compiler API? This would require changing sbt-scalafix to launch the correct Scalafix version (2.11/2.12/2.13) depending on the project being fixed.
* should Scalafix provide an Java-API on top of the presentation compiler allowing 2.12-compiled rules to work with 2.13/2.11? We use this approach in Metals (the server runs in 2.12, but completions work with 2.11/2.12/2.13).

I'm inclined to do the second approach since I like that Scalafix rules can work regardless of the Scala version being fixed. Regardless of the approach we use, I estimate it will require non-trivial work to implement.

mlachkar:
in 2.13, instead of Serializable, the type found is AnyRef.

  val access: Serializable = new Serializable {
    private val results: List[Int] = List.empty
    protected val results2: List[Int] = List.empty
  }

bjaglin:
@mlachkar not sure I follow, you mean that the rule as it's currently written has some bugs specific to scala 2.13? Is that a case of https://github.com/scalacenter/sbt-scalafix/pull/121#issuecomment-638404680 ?

mlachkar:
When I cross compiled this rule to 2.13, this test broke. This rule is not working for all cases that were supported in 2.12. 
I was not able to fix all rules, and I thought I will do this work in other PRs before v1.

bjaglin:
Ah ok, makes sense! Sorry, I hadn't seen https://github.com/scalacenter/scalafix/tree/master/scalafix-tests/input/src/main/scala/test/explicitResultTypes vs https://github.com/scalacenter/scalafix/tree/master/scalafix-tests/input/src/main/scala-2.12/test/explicitResultTypes.

PR #1116, opened 2020-05-09
align behavior with semanticdb-scalac: last arg wins

bjaglin:
Should fix https://github.com/scalacenter/scalafix/issues/1110, as failure to look up the target root triggers the presentation compiler, with a version that might not match the bytecode injected (see https://github.com/scalacenter/scalafix/issues/998).

The `SemanticDBPlugin` from sbt 1.3.x duplicates the `targetroot` option on `Test` (or any configuration that extends one where scalafix is enabled) via parent configuration delegation.: https://github.com/sbt/sbt/blob/a1b54d027a87654cbc782270872afc1b4c5a8852/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala#L42-L45.

In that case, `semanticdb-scalac` picks the last: https://github.com/scalameta/scalameta/blob/de206a796c7558f0ceb8bf3a557f3296f0ac34de/semanticdb/scalac/library/src/main/scala/scala/meta/internal/semanticdb/scalac/ConfigOps.scala#L105-L106, while scalafix was picking the first.

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -336,7 +336,8 @@ case class Args(
   def semanticdbOption(name: String): Option[String] = {
     val flag = s"-P:semanticdb:$name:"
     scalacOptions
-      .find(_.startsWith(flag))
+      .filter(_.startsWith(flag))
+      .lastOption
       .map(_.stripPrefix(flag))
   }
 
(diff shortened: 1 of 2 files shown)


daddykotex:
Thanks to you, I could include a work around in the meantime:

```
lazy val scalaFixTestWorkAround = Def.settings(
  scalacOptions in Test := {
    val initial = (scalacOptions in Test).value
    val semanticDbOptions = initial.filter(_.contains("-P:semanticdb:"))
    val semanticDbLess = initial.filterNot(_.contains("-P:semanticdb:"))
    semanticDbLess ++ semanticDbOptions.lastOption.toSeq
  }
)
```

Then use as `.settings(scalaFixTestWorkAround)` on the modules you need.

bjaglin:
@daddykotex using the sbt 1.2.x way should also do the trick since the "problem" is in the sbt 1.3.x plugin:
```diff
 lazy val myproject = project.settings(
   scalaVersion := "2.12.11", // 2.11.12, or 2.13.2
+  addCompilerPlugin(scalafixSemanticdb), // enable SemanticDB
   scalacOptions ++= List(
+    "-Yrangepos",          // required by SemanticDB compiler plugin
+    "-Ywarn-unused-import" // required by `RemoveUnused` rule
   )
 )
```
https://scalacenter.github.io/scalafix/docs/users/installation.html#sbt

PR #1118, opened 2020-05-11
Migrate to 2.13

mlachkar:
The missing points that need to be addressed to complete this migration to 2.13: 
- Rule NoAutoTupling: unitAdaptations doesn't work (it should be some missing ScalaOption)
- ExplicitResultTypes: will be handled in another issue #998
- test AnchorPatches: Workes only for 2.12 (contains ExplicitResultTypes rule)
- Doc: I had trouble with doc compilation, and just removed the lines that do not compile, which is not a solution.


project/Dependencies.scala
```diff
@@ -21,7 +21,8 @@ object Dependencies {
   def metacp = "org.scalameta" %% "metacp" % scalametaV
   def semanticdbPluginLibrary = "org.scalameta" % "semanticdb-scalac-core" % scalametaV cross CrossVersion.full
   def scalameta = "org.scalameta" %% "scalameta" % scalametaV
-  def scalatest = "org.scalatest" %% "scalatest" % "3.0.8"
+  def bijectionCore = "com.twitter" %% "bijection-core" % "0.9.7"
+  def scalatest = "org.scalatest" %% "scalatest" % "3.1.2"
```
olafurpg, line 25: Is there a reason this upgrade is needed? The upgrade was reverted in https://github.com/scalacenter/scalafix/pull/1083

mlachkar: There is no real reason for that. I was just updating some libs, removing some todo, and adding the real change on top of that. 
I will remove that.

scalafix-tests/input/src/main/scala/tests/ExplicitResultTypesRefinement.scala
```diff
@@ -27,10 +27,11 @@ object ExplicitResultTypesRefinement {
   def tparam[T <: CharSequence](e: T) = new Serializable {
     val results: List[Int] = List(e.length())
   }
-  val access = new Serializable {
-    private val results: List[Int] = List.empty
-    protected val results2: List[Int] = List.empty
-  }
+  //todo: in 2.13, the type found is AnyRef
```
olafurpg, line 30: It's usually better to open an issue in my experience and link to from here instead of adding a todo comment.

scalafix-tests/unit/src/test/scala/scalafix/tests/rule/MavenFuzzSuite.scala
```diff
@@ -26,7 +26,7 @@ class MavenFuzzSuite extends FunSuite with DiffAssertions {
       sourceJars: Seq[Path],
       tmp: Path
   ): Seq[Path] = {
```
olafurpg, line 28: I think this diff would have fixed the compile error as well
```suggestion
  ): collection.Seq[Path] = {
```

.github/workflows/ci.yml
```diff
@@ -14,27 +14,46 @@ jobs:
...
+    name: JDK11/scala_2.13 tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: adopt@1.11
+      - run: sbt ci-213
+  windows_212:
+    name: Windows/scala_2.12 tests
```
olafurpg, line 41: I think it's OK to drop this job and only test 2.13 in JDK 11

scalafix-tests/input/src/main/scala/tests/ExplicitResultTypesRefinement.scala
mlachkar: I will create all corresponding issues.

ohze:
I think [this line](https://github.com/scalacenter/scalafix/blob/94c5c69/project/ScalafixBuild.scala#L37)
```scala
if (isScala213.value) "-Ywarn-unused:imports"
```
should be `... "-Wunused:imports"` ?

mlachkar:
You're right @ohze. Thanks for noticing that.

PR #1122, opened 2020-05-13
clarify syntax of --tool-classpath

bjaglin:
* I naively assumed a `URLClassLoader` would require URL(s), and it did work for a while (the `file:` prefix is supported in `java.nio.file.Paths.get()`), until my path contained '%' which was not supposed to be escaped
* Seeing that repeating the argument did not work, I had to look up the source code to figure out the usage of a single, `java.ioFile.pathSeparator`-separated argument value


bjaglin:
any feedback on this @olafurpg @mlachkar ?

PR #1150, opened 2020-06-04
expose scalafix 2.13 patch version in scalafix-interfaces

bjaglin:
Follow-up of https://github.com/scalacenter/scalafix/pull/1118

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixImpl.scala
@@ -27,5 +27,7 @@ final class ScalafixImpl extends Scalafix {
     Versions.scala211
   override def scala212(): String =
     Versions.scala212
+  override def scala213(): String =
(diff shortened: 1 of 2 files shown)


bjaglin:
I wonder if these scala methods are actually used/useful, as you need `scalafix-cli` to get an implementation for that interface, and to load that `scalafix-cli`, you must have looked up the scala versions from `scalafix-interfaces.properties`. Is there another scenario I am missing?

olafurpg:
It's a good point. Maybe not 🤔 I feel like the architecture would make more sense if we have a bit of glue code implemented in Java as part of scalafix-interfaces that provides

* coursier resolution/fetch 
* classloading of the correct scalafix-cli

bjaglin:
Agreed, that would be a good first step to push up some stuff currently in sbt-scalafix (and mill-scalafix, etc), as also discussed in https://github.com/scalacenter/sbt-scalafix/pull/121#pullrequestreview-424259117. I'll file an issue for that.

Issue #1151, opened 2020-06-04
Make it easier to load an implementation of scalafix.interfaces.Scalafix


bjaglin:
As discussed in https://github.com/scalacenter/scalafix/pull/1150#issuecomment-638803769, `scalafix-interfaces` should provide (Java) helpers for:
 - coursier resolution/fetch
 - classloading of the correct `scalafix-cli`

The current helper is pretty low-level: https://github.com/scalacenter/scalafix/blob/d4f3b8e2744166115157f5c7cbef308e905332d1/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L71-L83

This will remove the need for custom code in `sbt-scalafix`, `mill-scalafix`, and all other clients.

olafurpg:
FWIW, in Scalafmt we have a `scalafmt-dynamic_2.{11,12,13}` library that is written in Scala and provides an implementation of `scalafix-interfaces` that dynamically loads the correct version of `scalafmt-cli`. Both Metals and IntelliJ use scalafmt-dynamic instead of `scalafix-interfaces`. The benefit of this approach is that scalafmt-interfaces only contains Java interfaces
 and all the logic is implemented in Scala.

bjaglin:
I have a local branch almost ready with a pure Java implementation inside `scalafix-interfaces` (using https://github.com/coursier/interface), which I think is fine. We can discuss once I open the PR!

bjaglin:
Maybe another benefit of aligning the approach is to try to align classloaders so that scalameta can at some point be shared across scalafmt & scalafix somehow?

PR #1152, opened 2020-06-06
scalafix-interfaces: higher-level API for class loading

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/1151
Ref https://github.com/scalacenter/scalafix/issues/998

This ports the module resolution/fetching code currently in `sbt-scalafix` into `scalafix-interfaces`. All JVM clients can now benefit from it, and avoid runtime errors due to different Scala binary versions across classloaders.


scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import coursierapi.Repository
+import org.scalatest.FunSuite
+import scalafix.interfaces.Scalafix
+
+import scala.jdk.CollectionConverters._
+
+/**
+ * Tests in this suite require scalafix-cli & its dependencies to be published (`sbt +publishLocal`) so that
+ * Coursier can fetch them.
```
bjaglin, line 13: This is annoying, but I really wanted to exercise different Scala binary versions between the client & the implementation to confirm the proper isolation. Initially, I had every scala version (as set in sbt) run against every scala version (as previously published, fetched via Coursier), but I couldn't find a nice way to automate the `publishLocal` step across versions, as it cannot be done in a task. https://github.com/sbt/sbt-projectmatrix could help with that, but would require big changes in the build.

I ended up hardcoding a single version as the target no matter what the sbt version is, and picking the default sbt version so when running that default sbt version (2.13.2), https://github.com/scalacenter/scalafix/pull/1152/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR246 makes the step unecessary.

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixInterfacesClassloader.java
```diff
@@ -0,0 +1,28 @@
...
+    public ScalafixInterfacesClassloader(ClassLoader parent) {
+        super(null);
+        this.parent = parent;
+    }
+
+    @Override
+    protected Class<?> findClass(String name) throws ClassNotFoundException {
+        if (name.startsWith("scalafix.interfaces")
+                // include types in scalafix.interfaces.* signatures
+                || name.startsWith("coursierapi")) {
```
bjaglin, line 22: I initially had the new overloads in `ScalafixArguments` implemented within `scalafix-interfaces` as Java8 default methods (leaving `scalafix-cli` untouched), so this was not required. It did feel weird though, so I ended up pushing them down to `scalafix-interfaces`, keeping most of the logic in `scalafix-interfaces`. I am not sure what's best...

scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java
```diff
@@ -58,15 +67,96 @@
...
+        }
+
+        String scalafixVersion = properties.getProperty("scalafixVersion");
+        String scalaVersion = properties.getProperty(scalaVersionKey);
+        if (scalafixVersion == null || scalaVersion == null)
+            throw new ScalafixException("Failed to lookup versions from '" + propertiesPath + "'");
+
+        List<URL> jars = ScalafixCoursier.scalafixCliJars(repositories, scalafixVersion, scalaVersion);
+        ClassLoader parent = new ScalafixInterfacesClassloader(Scalafix.class.getClassLoader());
+        return classloadInstance(new URLClassLoader(jars.stream().toArray(URL[]::new), parent));
```
bjaglin, line 141: can we assume java8 at runtime since that's what CI run tests with?

olafurpg: Correct. It’s ok to target java 8

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -25,10 +31,45 @@
...
+     * @param customURLs                   Extra URLs for classloading and compiling external rules.
+     * @param extraDependenciesCoordinates Extra dependencies for classloading and compiling external rules.
+     *                                     For example "com.github.liancheng::organize-imports:0.3.0".
+     *                                     Artifacts will be resolved against the Scala version in the classloader
+     *                                     of the parent {@link Scalafix} instance and fetched using Coursier.
+     * @throws ScalafixException in case of errors during artifact resolution/fetching.
+     */
+    ScalafixArguments withToolClasspath(
+            List<URL> customURLs,
+            List<String> extraDependenciesCoordinates
```
olafurpg, line 49: Should this be a coursierinterface.Dependency?

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+        return jars;
+    }
+
+    public static List<URL> scalafixCliJars(
+            List<Repository> repositories,
+            String scalafixVersion,
+            String scalaVersion
+    ) throws ScalafixException {
+        Dependency scalafixCli = Dependency.parse(
```
olafurpg, line 45: TIL Dependency.parse 🤩

I have implemented this logic too many times, nice to have it built in!

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixInterfacesClassloader.java
olafurpg: This looks good

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
olafurpg: The metals build contains custom sbt logic to publish local across different Scala versions if you want to take a second stab at this

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
bjaglin: `coursierinterface.Dependency` has the Scala version hardcoded, and since that version is changing ([from 2.12 to the eventually target sources one](https://github.com/scalacenter/sbt-scalafix/pull/121)), I thought it was better to request coordinates and let the implementation append the right suffix.

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
bjaglin: Indeed! The hardest is/was to find a good name for the input ([`dep`](https://github.com/coursier/interface/blob/d55842de4f86ceb82cf332d0ccb416a47a6726dc/interface/src/main/java/coursierapi/Dependency.java#L44) is not very clear). I saw `coordinates` was used in other places of Coursier source code so I went for this... Any suggestion is welcome!

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
bjaglin: By supporting `org:::module`, the client can also easily make use of advanced community rules that are built against a full Scala version, without having to explicitly update the suffix at each patch release.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
bjaglin: Great, thanks for the pointer, I'll look at it - maybe I'll follow-up in a separate PR though.

bjaglin:
@olafurpg  sorry for the amend while you were reviewing. I was simply adding an assertion to introspect/confirm the running Scalafix version within the tool classpath: https://github.com/scalacenter/scalafix/compare/17447d939e33f83e418053cb98f2d4c116b8318b..e1ef4d59

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
```diff
@@ -0,0 +1,86 @@
...
+            |        scalafix.lint.LintSeverity.Error
+            |      )
+            |    )
+            |}""".stripMargin
+        }
+      var maybeDiagnostic: Option[ScalafixDiagnostic] = None
+      val scalafixMainCallback = new ScalafixMainCallback {
+        override def reportDiagnostic(diagnostic: ScalafixDiagnostic): Unit =
+          maybeDiagnostic = Some(diagnostic)
+      }
```
bjaglin, line 58: SAM is behind `-Xexperimental` in 2.11 so I decided to go the verbose way

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
```diff
@@ -0,0 +1,87 @@
...
+class ScalafixSuite extends FunSuite {
+
+  def tmpFile(prefix: String, suffix: String)(content: String): Path = {
+    val path = Files.createTempFile(prefix, suffix)
+    Files.write(path, content.getBytes)
+    path
+  }
+
+  def fetchAndLoad(scalaVersion: String): Unit = {
+    test(s"fetch & load instance for Scala version $scalaVersion", SkipWindows) {
```
bjaglin, line 26: `error: Unknown rule 'file:C:\Users\RUNNER~1\AppData\Local\Temp\CaptureScalaVersion3780416901127757866.scala'` on Windows... I saw that source-compilation was tagged `SkipWindows` in `ToolClasspathSuite` as well, so I didn't look further.

bjaglin:
@joan38 @evis will this help for your respective plugins?

joan38:
@bjaglin yes, thanks for tagging me!
I'll take advantage of it once released.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
olafurpg: Makes sense, good idea

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala
bjaglin: @olafurpg I managed to adapt https://github.com/scalameta/metals/commit/3ab0608f4660793e25cff324d475fd334dadc7bb (which is just gold!), to do exactly what I wanted (modulo [a scala bug](https://github.com/scala/bug/issues/10603) I ran into) https://github.com/scalacenter/scalafix/compare/4a382e5f..a082cd3e :tada:

The overhead of cross-publishing all artifacts before running tests is surprisingly acceptable, so I think this is ready to go.

project/ScalafixBuild.scala
```diff
@@ -311,6 +317,31 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
+    }.value,
+    crossPublishLocalBinTransitive := {
+      val currentState = state.value
+      val ref = thisProjectRef.value
+      val versions = crossScalaVersions.value
+      versions.map {
+        version =>
+          val withScalaVersion = Project
+            .extract(currentState)
+            .appendWithoutSession(
```
bjaglin, line 333: with this (as opposed to `appendWithSession`), we don't observe the effect of `++`, which makes it harder to override `scalaVersion` as it seems that it's done at the project level

bjaglin:
@mlachkar I'll cut 0.9.17 with this if that's OK for you

mlachkar:
@bjaglin yes, please do. Thanks

joan38:
Could you comment on why it's using `coursierapi.Repository` and not `coursier.Repository`?

bjaglin:
@joan38 I can't find any `coursier.Repository` in the classpath of `scalafix-interfaces`, could you point me at the one you are looking at? AFAIK all the Java facade that coursier-interface is exposes its API under `coursierapi`.

joan38:
https://github.com/coursier/coursier/blob/master/modules/core/shared/src/main/scala/coursier/core/Repository.scala

bjaglin:
`coursier-interface` (the pure Java, non-cross-versioned module we want clients to depend on to prevent scala binary version conflicts) abstracts the regular (scala) Coursier API under https://github.com/coursier/interface/blob/1dd201e88ef95e2f4d48d580d9678e541d8a4b2e/build.sbt#L53, so it cannot be used directly.

bjaglin:
>  And how can I convert to a `coursierapi.Repository` from a `coursier.Repository`

You should be able to copy the logic from https://github.com/coursier/interface/blob/master/interface/src/main/scala/coursier/internal/api/ApiHelper.scala#L162-L176.

joan38:
Thanks @bjaglin for the code, it would have be great if coursier was exposing those utils instead of us translating via a copy and past.

bjaglin:
Indeed. From what I understand, those utils would have to be exposed via a third module that would depend on `interface` and a non-shared `core`, so it's not trivial. Maybe you can follow up on the coursier channels to confirm that there is no better way.

joan38:
So I realised this thing is not resolving `scalafix-rules` and that's ok.
But I'm hitting this one:
```
project.fix java.util.ServiceConfigurationError: scalafix.v1.Rule: scalafix.internal.rule.DisableSyntax not a subtype
    java.base/java.util.ServiceLoader.fail(ServiceLoader.java:591)
    java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1238)
    java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1266)
    java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1301)
    java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1386)
    scala.collection.convert.Wrappers$JIteratorWrapper.hasNext(Wrappers.scala:43)
    scala.collection.Iterator.foreach(Iterator.scala:943)
    scala.collection.Iterator.foreach$(Iterator.scala:943)
    scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
    scala.collection.generic.Growable.$plus$plus$eq(Growable.scala:62)
    scala.collection.generic.Growable.$plus$plus$eq$(Growable.scala:53)
    scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:184)
    scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:47)
    scala.collection.TraversableOnce.to(TraversableOnce.scala:348)
    scala.collection.TraversableOnce.to$(TraversableOnce.scala:346)
    scala.collection.AbstractIterator.to(Iterator.scala:1431)
    scala.collection.TraversableOnce.toList(TraversableOnce.scala:332)
    scala.collection.TraversableOnce.toList$(TraversableOnce.scala:332)
    scala.collection.AbstractIterator.toList(Iterator.scala:1431)
    scalafix.internal.v1.Rules$.all(Rules.scala:101)
    scalafix.v1.RuleDecoder$$anon$1.<init>(RuleDecoder.scala:99)
    scalafix.v1.RuleDecoder$.decoder(RuleDecoder.scala:98)
    scalafix.internal.interfaces.ScalafixArgumentsImpl.rulesThatWillRun(ScalafixArgumentsImpl.scala:167)
    com.goyeau.mill.scalafix.ScalafixModule$.fixAction(ScalafixModule.scala:87)
    com.goyeau.mill.scalafix.ScalafixModule.$anonfun$fix$4(ScalafixModule.scala:52)
    mill.api.Result$Success.flatMap(Result.scala:29)
    com.goyeau.mill.scalafix.ScalafixModule.$anonfun$fix$1(ScalafixModule.scala:37)
    mill.define.ApplyerGenerated.$anonfun$zipMap$6(ApplicativeGenerated.scala:15)
    mill.define.Task$MappedDest.evaluate(Task.scala:392)
```
Anyone can help?

bjaglin:
Sure - I'll check the PR and comment there

PR #1155, opened 2020-06-10
sbt scalafixResolvers: coursiersmall -> coursierapi

bjaglin:
Merge only when https://github.com/scalacenter/sbt-scalafix/pull/123 is released.

The `coursier-interface` facade is pretty poor when it comes to resolver helpers.

Issue #1146, opened 2020-06-03
Separate versioning of sbt-scalafix and scalafix-core


mlachkar:
* Separate versioning of sbt-scalafix and scalafix-core similar to sbt-scalafmt and scalafmt-core. Users should declare the Scalafix version in .scalafix.conf so that clients like Metals or maybe IntelliJ can load the correct version of Scalafix without having to inspect project/plugins.sbt

* It will make easier the release of scalafix, which is now depending on sbt-scalafix. 

#fixes #1084

bjaglin:
From my experience with `scalafmt` / `sbt-scalafmt`: it can be confusing for the users if the release cycles are distinct but the versions remains in the same vicinity. I don't have a good solution for that though, as we might to signal 1.0 at the same time for `scalafix-core` and `sbt-scalafix`.

mlachkar:
I agree that it's quite annoying to specify the sbt-scalafix version, and then add a version inside the scalafix.conf from the user point of view, that looks almost the same. Moreover, I wonder if people will update the version inside .scalafix.conf (except if scala steward is able to manage this). 

On the other hand, I think that those two versions should be independent since semantically they are different and they could occur independently. It will make easier the release management. 

I wonder if it's possible to have a default value for scalafix version, provided by the sbt plugin, which will need a commit to evolve, but not an urgent one (avoiding the error when fetching artifacts for really early adopters ^^ ). The scafix-version config, will be used only for the advanced user (metals for example).

So my idea is not to show directly this version conf to avoid the confusion that raises from  "versions remaining almost the same". What do you think ?

olafurpg:
I think it's a good idea to have clearly distinct version numbers for sbt-scalafix and scalafix-core, for example sbt-scalafix:20.0.0 and scalafix-core:1.0.0. It might be a bit weird at first sight but it helps clarify that the two versions are independent. I regret we didn't do that with sbt-scalafmt (it was proposed).

I would require the version field in .scalafix.conf from scalafix-core:1.0.0 going forward. Otherwise I am concerned many users will stay on old scalafix versions using the latest default from scalafix.

One problem I see with this scheme is that sbt-scalafix won't know what version of SemanticDB to install since it's provided by scalafix-interfaces. The options I see to fix this problem are

* keep sbt-scalafix and scalafix-core versions in sync and pay the complexity price for releases
* dynamically fetch and load resources for scalafix-interfaces.jar that matches the version in `.scalafix.conf`
[...]

PR #1157, opened 2020-06-12
delay doc publishing until a release is tagged

bjaglin:
New features/deprecations should not be publicly documented until a corresponding release is cut. `mdoc` is now run as part of the CI workflow to limit the risk of `docs/docusaurusPublishGhpages` failing at release time (would a doc change breaking mdoc be visible only on master now?).

This came up as I realized I should hold https://github.com/scalacenter/scalafix/pull/1155 & https://github.com/scalacenter/scalafix/pull/1160 (granted, it's not a big deal if I don't) after seeing the `SNAPSHOT` version in https://scalacenter.github.io/scalafix/docs/users/installation.html#help. Holding documentation PRs is hard to manage - we would need GitHub milestones to handle them properly?.

Once https://github.com/scalacenter/scalafix/issues/1146 is done, the concept of release will be local, so another strategy will need to be found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 jdk:
 - openjdk8
 stages:
-  - name: test
   - name: release
-    if: (branch = master AND type = push) OR (tag IS present)
+    if: tag IS present
 before_install:
   - if [ "$TRAVIS_BUILD_STAGE_NAME" = 'Release'  ]; then
       curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.9.4;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,9 +47,10 @@ jobs:
       - run: sbt ci-213-windows
         shell: bash
   checks:
-    name: Scalafmt
+    name: Scalafmt & docs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: olafurpg/setup-scala@v7
       - run: ./scalafmt --test
+      - run: sbt docs/mdoc


.travis.yml
```diff
@@ -1,9 +1,8 @@
 jdk:
 - openjdk8
 stages:
-  - name: test
```
bjaglin, line 4: dead conf?

olafurpg: Looks dead

bjaglin:
Thanks for the feedback @olafurpg! I am new to this, so I naively thought this would be better, but it's not that simple indeed.

> it’s desirable to be able to update the documentation after a release based on user feedback

Without going the full way of an (over-engineered) versioned documentation, we could have a `docs` branch that we merge master into or reset to master on each tag/release, on which we could cherry-pick doc changes that apply / are relevant to the latest stable. Any push on that branch (and on that branch only) would trigger doc publishing. I am expecting such changes to be rare and the cost of cherry-picking to be acceptable, but that might be a biased assumption.

However, as sbt-scalafix and scalafix-core drift apart lifecycle/version-wise, this would no longer work unless the sbt-specific pages are migrated to sbt-scalafic repo (with potentially the same strategy), so the website becomes the combination of both docs. Maybe it's slightly off topic and should rather be discussed in https://github.com/scalacenter/scalafix/issues/1146, but do you have some thoughts on this?

olafurpg:
I haven't used a custom branch strategy in any projects so I'm not sure how it would work in practice. I suspect I personally would think it requires too much manual effort to cherry-pick commits into the `docs` branch.

In Metals, we have an `mdoc:since` modifier that allows us to declare in what Metals version a feature got released and mdoc displays a warning if the latest stable version is older than the version where the feature got released.

https://github.com/scalameta/metals/blob/8d0924af4348dcad0a2ae8a228c9d2f1fdcbff55/metals-docs/src/main/scala/docs/VersionedDocsModifier.scala

We haven't used `mdoc:since` much, but I think it's a decent solution to this problem where you want to document new features as early as possible without confusing users with documentation of features that are not available in the stable release.

With all things considered, my recommendation would be to continue updating docs on every merge into master and hold off merging PRs documenting new functionality.

bjaglin:
> With all things considered, my recommendation would be to continue updating docs on every merge into master and hold off merging PRs documenting new functionality.

Makes sense, thanks for sharing your experience!

PR #1160, opened 2020-06-13
document scalafixAll InputKey

bjaglin:
Merge only when https://github.com/scalacenter/sbt-scalafix/pull/126 is released

### DIFF
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -150,7 +150,8 @@ Great! You are all set to use Scalafix with sbt :)
 
 | Name | Type | Description
 | ---- | ---- | -----------
-| `scalafix <args>` | `InputTaskKey[Unit]` | Invoke scalafix command line interface directly. Use tab completion to explore supported arguments or consult [--help](#help).
+| `scalafix <args>` | `InputKey[Unit]` | Invoke scalafix command line interface directly. Use tab completion to explore supported arguments or consult [--help](#help).
+| `scalafixAll <args>` | `InputKey[Unit]` | Invoke `scalafix` across all configurations where scalafix is [enabled](#integration-tests).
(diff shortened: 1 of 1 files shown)


docs/users/installation.md
```diff
@@ -181,34 +183,17 @@ configurations. To enable Scalafix for other configuration like
...
-Optionally, add a command alias to your build to run Scalafix on your entire
-project with the shorthand `fix`
-
-```scala
-// top of build.sbt
-addCommandAlias("fix", "all compile:scalafix test:scalafix")
-```
+Both the `scalafix` & the `scalafixAll` task aggregate like the `compile`
+and `test` tasks. To run Scalafix on all projects for both main and test
+sources you can execute `all scalafix test:scalafix` or simply `scalafixAll`.
```
olafurpg, line 188: This seems like the easiest to explain option and it handles custom integrations like `it:scalafix`
```suggestion
sources you can execute `scalafixAll`.
```

Is there a difference between `all scalafixAll` and only `scalafixAll`?

docs/users/installation.md
```diff
@@ -161,13 +162,14 @@ Great! You are all set to use Scalafix with sbt :)
...
 `myproject/test:scalafix` instead. To run on both main and test sources, execute
-`; myproject/scalafix ; myproject/test:scalafix`
+`myproject/scalafixAll`.
 
 ### Integration tests
 
 By default, the `scalafix` command is enabled for the `Compile` and `Test`
-configurations. To enable Scalafix for other configuration like
-`IntegrationTest`, add the following to your project settings
+configurations, ans `scalafixAll` will run on both of them. To enable
```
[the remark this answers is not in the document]
bjaglin: oops

docs/users/installation.md
bjaglin: indeed :+1: 

> Is there a difference between all scalafixAll and only scalafixAll?

None as in both cases, configurations invocations will run independently/concurrently. I guess the latter allows a single `last scalafixAll` to see an overview of the logs easily. And the former is a command while the latter is a task obviously.  Why do you ask?

bjaglin:
Commit amended with the suggestions :+1:

PR #1163, opened 2020-06-14
respect -Xlint or -Xlint:unused scalac options in RemoveUnused rule

taisukeoe:
This PullRequests suppress RemoveUnused configuration error if scalac options contain `-Xlint` or `-Xlint:unused`, both of which generate and add unused diagnostics message string into SemanticDB.

Note: There might be discussions at the following point.

## -Xlint linting options notation
Currently this PullRequest supports no explicit linting option (just `-Xlint`) or only a linting option per a line (`-Xlint:unused`) for implementation simplicity. IMHO this covers majority of `-Xlint` usage, as long as I check with "-Xlint" GitHub search.

This PR does NOT support multiple linting options together like `-Xlint:unused,constant`, wildcard like `-Xlint:_`, nor unary operator like `-Xlint:-constant,_`.

If we want to support multiple linting options notation or else above, it may need to introduce a simple scalac option parser (or just check in a naive way, like 
 `option.startsWith("-Xlint:") && (option.contains("_") || option.contains("unused")) && !option.contains("-unused")` ?).

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -27,15 +27,17 @@ class RemoveUnused(config: RemoveUnusedConfig)
   override def isRewrite: Boolean = true
 
   private def warnUnusedPrefix = List("-Wunused", "-Ywarn-unused")
+  private def warnUnusedString = List("-Xlint", "-Xlint:unused")
   override def withConfiguration(config: Configuration): Configured[Rule] = {
     val hasWarnUnused = config.scalacOptions.exists(option =>
-      warnUnusedPrefix.exists(prefix => option.startsWith(prefix))
+      warnUnusedPrefix.exists(prefix => option.startsWith(prefix)) ||
+        warnUnusedString.contains(option)
     )
     if (!hasWarnUnused) {
       Configured.error(
         s"""|The Scala compiler option "-Ywarn-unused" is required to use RemoveUnused.
             |To fix this problem, update your build to use at least one Scala compiler
-            |option that starts with -Ywarn-unused or -Wunused (2.13 only)""".stripMargin
+            |option like -Ywarn-unused, -Xlint:unused, or -Wunused (2.13 only).""".stripMargin
       )
     } else {
       config.conf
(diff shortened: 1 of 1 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -27,15 +27,18 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
+      warnUnusedPrefix.exists(prefix =>
+        option.startsWith(prefix) || warnUnusedString.contains(option)
+      )
     )
     if (!hasWarnUnused) {
       Configured.error(
         s"""|The Scala compiler option "-Ywarn-unused" is required to use RemoveUnused.
             |To fix this problem, update your build to use at least one Scala compiler
-            |option that starts with -Ywarn-unused or -Wunused (2.13 only)""".stripMargin
+            |option like -Ywarn-unused, -Xlint:unused, or -Wunused (2.13 only).""".stripMargin
```
taisukeoe, line 41: I didn't add `-Xlint` to suggestions in the error message since `-Xlint` is for multi purposes.

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -27,15 +27,17 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
-      warnUnusedPrefix.exists(prefix => option.startsWith(prefix))
+      warnUnusedPrefix.exists(prefix => option.startsWith(prefix)) ||
+        warnUnusedString.contains(option)
     )
     if (!hasWarnUnused) {
       Configured.error(
         s"""|The Scala compiler option "-Ywarn-unused" is required to use RemoveUnused.
             |To fix this problem, update your build to use at least one Scala compiler
-            |option that starts with -Ywarn-unused or -Wunused (2.13 only)""".stripMargin
+            |option like -Ywarn-unused, -Xlint:unused (2.12.2 or above), or -Wunused (2.13 only)""".stripMargin
```
taisukeoe, line 40: Unused warnings by `-Xlint` or `-Xlint:unused` are effective since Scala 2.12.2.

https://github.com/scala/bug/issues/10270
https://github.com/scala/scala/pull/5402

So suggesting `-Xlint:unused` is not relevant in Scala 2.12.1 or lower while Scalafix still supports 2.11 . To be safer, I would clarify which versions are supported.

PR #1166, opened 2020-06-16
document sbt-scalafix scalafixScalaBinaryVersion

bjaglin:
Merge only when https://github.com/scalacenter/sbt-scalafix/pull/121 is released

Improves https://github.com/scalacenter/scalafix/issues/998

### DIFF
--- a/docs/rules/ExplicitResultTypes.md
+++ b/docs/rules/ExplicitResultTypes.md
@@ -34,13 +34,6 @@ This rule has several known limitations, which are most likely fixable with some
 effort. At the time of this writing, there are no short-term plans to address
 these issues however.
 
-### Limited Scala support
-
-Due to implementation details, this rule works only with @SCALA212@. Scala 2.11
-and 2.13 are not supported for this rule. See
-[#988](https://github.com/scalacenter/scalafix/issues/998) if you want to
-contribute support for other Scala versions.
-
 ### Imports ordering
 
 The rewrite inserts imports at the bottom of the global import list. Users are
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -155,6 +155,8 @@ Great! You are all set to use Scalafix with sbt :)
 | `scalafixConfig` | `SettingKey[Option[File]]` | `.scalafix.conf` file to specify which scalafix rules should run, together with their potential options. Defaults to `.scalafix.conf` in the root directory, if it exists.
 | `scalafixDependencies` | `SettingKey[Seq[ModuleID]]` | Dependencies making [custom rules](#run-custom-rules) available via their simple name. Must be set in `ThisBuild`. Defaults to `Nil`.
 | `scalafixResolvers` | `SettingKey[Seq[Repository]]` | Custom resolvers where `scalafixDependencies` are resolved from. Must be set in `ThisBuild`. Defaults to: Ivy2 local, Maven Central, Sonatype releases & Sonatype snapshots.
+| `scalafixScalaBinaryVersion` | `SettingKey[String]` | Scala binary version used for Scalafix execution. Defaults to 2.12. For advanced rules such as ExplicitResultTypes to work, it must match the binary version defined in the build for compiling sources. Note that `scalafixDependencies` artifacts must be published against that Scala version.
+
 
 ### Main and test sources
 


docs/users/installation.md
```diff
@@ -155,6 +155,8 @@ Great! You are all set to use Scalafix with sbt :)
 | `scalafixConfig` | `SettingKey[Option[File]]` | `.scalafix.conf` file to specify which scalafix rules should run, together with their potential options. Defaults to `.scalafix.conf` in the root directory, if it exists.
 | `scalafixDependencies` | `SettingKey[Seq[ModuleID]]` | Dependencies making [custom rules](#run-custom-rules) available via their simple name. Must be set in `ThisBuild`. Defaults to `Nil`.
 | `scalafixResolvers` | `SettingKey[Seq[Repository]]` | Custom resolvers where `scalafixDependencies` are resolved from. Must be set in `ThisBuild`. Defaults to: Ivy2 local, Maven Central, Sonatype releases & Sonatype snapshots.
+| `scalafixScalaBinaryVersion` | `SettingKey[String]` | Scala binary version used for Scalafix execution. Defaults to 2.12. For advanced rules such as ExplicitResultTypes to work, it must match the binary version defined in the build for compiling sources. Note that `scalafixDependencies` artifacts must be published against that Scala version.
```
bjaglin, line 158: > Note that `scalafixDependencies` artifacts must be published against that Scala version.

That's a bit vague on purpose, as I didn't want to explain which full Scala version would be used in case we have external rules using the presentation compiler.

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -59,9 +59,10 @@ final class ExplicitResultTypes(
       }
     if (config.scalacClasspath.nonEmpty && config.scalaVersion != supportedScalaVersion) {
       Configured.error(
-        s"The ExplicitResultTypes rule only supports the Scala version '$supportedScalaVersion'. " +
-          s"To fix this problem, either remove `ExplicitResultTypes` from .scalafix.conf or change the Scala version " +
-          s"in your build to match exactly '$supportedScalaVersion'."
+        s"The ExplicitResultTypes rule only supports the Scala version '$supportedScalaVersion' for this Scala " +
+          "binary version. To fix this problem, either remove `ExplicitResultTypes` from .scalafix.conf, or adjust " +
+          s"the patch Scala version in your build to match exactly '$supportedScalaVersion' after making sure that " +
+          s"the `scalafixBinaryScalaVersion` setting key matches the Scala binary version of your build."
```
bjaglin, line 65: Not very happy about that as the first sentence can be confusing when binary scala version does not match.... Ideally, we should get `supportedScalaVersion` from `scalafix-cli` based on `config.scalaVersion`, but that's impossible. Any suggestion is welcome!

docs/rules/ExplicitResultTypes.md
```diff
@@ -34,13 +34,6 @@ This rule has several known limitations, which are most likely fixable with some
 effort. At the time of this writing, there are no short-term plans to address
 these issues however.
 
-### Limited Scala support
-
-Due to implementation details, this rule works only with @SCALA212@. Scala 2.11
-and 2.13 are not supported for this rule. See
-[#988](https://github.com/scalacenter/scalafix/issues/998) if you want to
-contribute support for other Scala versions.
-
```
bjaglin, line 43: Should we keep this until the rule itself is fully functionnal with 2.13? Since my intention was not to close https://github.com/scalacenter/scalafix/issues/998 with this, I guess so...

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: After re-reading https://github.com/scalacenter/sbt-scalafix/pull/121#discussion_r435117738, I was wondering if we could simply remove the full version constraint here and focus on the binary version (maybe by cross-testing rules at the patch level to verify compatibility)?

docs/rules/ExplicitResultTypes.md
bjaglin: I guess we can remove it now that https://github.com/scalacenter/scalafix/pull/1167 is going to close #998

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
mlachkar: I don't understand exactly what you mean, but I agree that now this rule could be used for 2.11, 2.12, and 2.13.. 
You would like to test many version of the same 2.13  ??

docs/rules/ExplicitResultTypes.md
mlachkar: Agree. 
I

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: Currently, the rule will fail verbosely if the sources were compiled with a different patch version than the one scalafix-rules was built against.

It's either:
1. a different patch version, in which case the user must update its built version to match the one from scalafix
1. a different binary version, in which case the user must align the `scalafixScalaBinaryVersion`  AND potentially the patch release as in (1). The first sentence is a bit confusing as it will say "The ExplicitResultTypes rule only supports the Scala version 2.12.11" for sources compiled in 2.13.1 for example, which is not true as there is now support for 2.13.2.

That's why I was wondering if we should relax the constraint to only fail on a binary version mismatch, and assume it will work most of the time for patch version mismatch. That's my interpretation of @olafurpg's [comment](https://github.com/scalacenter/sbt-scalafix/pull/121#discussion_r435117738). In that case, it might be wise to run the testkit against input cross-built against many patch versions (and not just one per binary version), although this will significantly increase CI time if we are exhaustive for 2.11, 2.12 & 2.13.

So my comment/suggestion is not related to this PR, but writing the recommendation on failure made me realize that having 2 different cases and actions makes it hard to use.

I am not familiar with the presentation compiler and how it's used in the rule to judge whether it's acceptable to ignore the patch release even for built-in rules, but since `scalafix-rules` is cross-built with a full version, I guess there was a reason?

mlachkar: In my opinion, we should relax this constraint.
 I propose to support 3 scala versions of the compiler for each binary version of Scala, by the corresponding CI-tests.
It will increase ci-tests, but I think it's acceptable.

mlachkar:
What do you think about merging this PR and I will finish/rebase this [PR](https://github.com/scalacenter/scalafix/pull/1174) on top of it ?

bjaglin:
I was holding the merge mostly because of what was discussed in https://github.com/scalacenter/scalafix/pull/1157 (and also to get @olafurpg's feedback), but I agree we can merge this and maybe aim at releasing `0.9.18` in the coming days (together will all other PRs and your follow-up) ? Go ahead and merge!

bjaglin:
I'll rebase this against https://github.com/scalacenter/scalafix/pull/1174 once it's merged

bjaglin:
@mlachkar that's rebased!

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
ckipp01: Just randomly jumping in here 😬, but quick question on the wording here. I actually just came across this while using https://github.com/joan38/mill-scalafix. I hit on this issue the message produced seems to be very _sbt-specific_ while there is no guarantee that the user is using sbt here. I'm just curious if there is a better way to word this and avoid the `setting key`. I'm all for making error messages clear and actionable, but when the build tool is assumed, then it can get really confusing for someone not familiar with the ecosystem.

bjaglin: Hey @ckipp01, thanks for the feedback and sorry for the late follow-up! You are right, this is very confusing, my proposal for this is https://github.com/scalacenter/scalafix/issues/1199, let's follow-up there if you have any idea.

PR #1167, opened 2020-06-17
Add test for 2.11 and 2.13 version of ExplicitResultType

### LINKED ISSUE #998: Make ExplicitResultTypes work with more Scala versions

mlachkar:
closes #998


scalafix-tests/output/src/main/scala-2.13/tests/ExplicitResultTypesBug.scala
```diff
@@ -0,0 +1,56 @@
+package tests
```
bjaglin, line 1: so this one doesn't work on scala 2.11?

scalafix-tests/output/src/main/scala-2.11/test/explicitResultTypes/ExplicitResultTypesInfix.scala
```diff
@@ -0,0 +1,11 @@
+package test.explicitResultTypes
```
bjaglin, line 1: so this one doesn't work yet on scala 2.13?

scalafix-tests/input/src/main/scala/test/explicitResultTypes/Rename.scala
```diff
@@ -3,11 +3,11 @@ rules = ExplicitResultTypes
 */
 package test.explicitResultTypes
 
-import scala.collection.{Seq => SSeq}
+import scala.collection.immutable.{List => LList}
```
bjaglin, line 6: why this change?

scalafix-tests/input/src/main/scala/tests/ExplicitResultTypesRefinement.scala
```diff
@@ -27,11 +27,15 @@ object ExplicitResultTypesRefinement {
...
-  //  }
+  val access = new Serializable {
+    private val results: List[Int] = List.empty
+    protected val results2: List[Int] = List.empty
+  }
+  val product = new Product {
+    override def productArity: Int = ???
+    override def productElement(n: Int): Any = ???
+    override def canEqual(that: Any): Boolean = ???
+  }
```
bjaglin, line 38: maybe you can add a `new Serializable with Product` since the fix seems to be related to that?

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
```diff
@@ -313,9 +313,19 @@ class CompilerTypePrinter(g: ScalafixGlobal, config: ExplicitResultTypesConfig)(
...
-          }
+          val productRootClass = definitions.ProductClass.seq
+            .toSet[Symbol] + definitions.ProductRootClass
+          val serializableClass = definitions.SerializableClass
+          val strippedParents =
+            if (productRootClass
+                .intersect(parents.map(_.typeSymbol).toSet)
+                .nonEmpty && parents
+                .map(_.typeSymbol)
+                .contains(serializableClass)) {
```
bjaglin, line 324: could you extract `val parentsClass = parents.map(_.typeSymbol).toSet` to make the condition more readable?

scalafix-tests/input/src/main/scala/tests/ExplicitResultTypesRefinement.scala
mlachkar: good Idea.

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
mlachkar: yes.

scalafix-tests/output/src/main/scala-2.11/test/explicitResultTypes/ExplicitResultTypesInfix.scala
mlachkar: So this one doesn't find the same types in 2.11 (compared to 2.12 and 2.13).
On 2.12 and 2.13, there is still an output test that looks like that
```
object ExplicitResultTypesInfix {
  val ab: Int \/ String = null.asInstanceOf[Int \/ String]
  def foo: Int \/ String = identity(ab)

  case class :+:[T1, T2]()
  def bar: Double :+: Long = :+:[Double, Long]()
}
```

scalafix-tests/output/src/main/scala-2.13/tests/ExplicitResultTypesBug.scala
mlachkar: I forgot this. I will check

scalafix-tests/input/src/main/scala/test/explicitResultTypes/Rename.scala
mlachkar: scala.Seq is now an alias for scala.collection.immutable.Seq[A] (instead of scala.collection.Seq[A]) in 2.13.
So I would need different output tests. The goal is to test the Rename ... this is why I did this

scalafix-tests/output/src/main/scala-2.11/test/explicitResultTypes/ExplicitResultTypesInfix.scala
bjaglin: makes sense, sorry I ad missed the 3 different outputs!

scalafix-tests/output/src/main/scala/test/explicitResultTypes/Rename.scala
```diff
@@ -1,11 +1,11 @@
 
 package test.explicitResultTypes
 
-import scala.collection.{Seq => SSeq}
+import scala.collection.immutable.{List => LList}
 import java.lang.{Boolean => JBoolean}
 
 object Rename {
-  type Seq = Int
-  def foo(a: Int*): SSeq[Int] = identity(a)
```
bjaglin, line 9: I haven't looked at the git history, but maybe there was something with varargs/identity that this was specifically testing? In that case, maybe it's better to go for:
```
def foo(a: Int*): LList[Int] = identity(a.toList)
```

scalafix-tests/output/src/main/scala/test/explicitResultTypes/Rename.scala
```diff
@@ -1,11 +1,11 @@
...
+  type List = Int
+  def foo(a: Int): LList[Int] = List(a)
```
bjaglin, line 9: I haven't looked at the git history, but maybe there was something with varargs/identity that this was specifically testing? In that case, maybe it's better to go for:
```suggestion
  def foo(a: Int*): LList[Int] = identity(a.toList)
```

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
```diff
@@ -222,6 +225,7 @@ class CompilerTypePrinter(g: ScalafixGlobal, config: ExplicitResultTypesConfig)(
 
   private def isPossibleSyntheticParent(tpe: Type): Boolean = {
     definitions.isPossibleSyntheticParent(tpe.typeSymbol) ||
+    CompilerCompat.serializableClass(g).toSet[Symbol](tpe.typeSymbol) ||
```
bjaglin, line 228: isn't that redundant with the check above?

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
```diff
@@ -313,9 +317,19 @@ class CompilerTypePrinter(g: ScalafixGlobal, config: ExplicitResultTypesConfig)(
...
         case RefinedType(parents, _) =>
           // Remove redundant `Product with Serializable`, if possible.
-          val strippedParents = parents.filterNot { tpe =>
-            definitions.isPossibleSyntheticParent(tpe.typeSymbol)
-          }
+          val productRootClass = definitions.ProductClass.seq
+            .toSet[Symbol] + definitions.ProductRootClass
+          val serializableClass =
+            CompilerCompat.serializableClass(g).toSet[Symbol]
+          val parentSymbol = parents.map(_.typeSymbol).toSet
```
bjaglin, line 324:
```suggestion
          val parentSymbols = parents.map(_.typeSymbol).toSet
```

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
mlachkar: not really... for scala 2.12 and 2.11, this method `definitions.isPossibleSyntheticParent` wasn't checking both Serializable traits.

bjaglin: got it, thanks!

PR #1169, opened 2020-06-18
Fix AnchorPatches test

mlachkar:
Trying to re-enable this test for all scala versions. Right now it fails for 2.13. This test uses RemoveUnused Rule and specifically here will remove imports. This test requires `-Wunused:imports` in 2.13.

After debugging in 2.13, I can see that `scalacOptions` that is set in `TestkitProperties.scala` file is equal to `List(-target:jvm-1.8, -Wunused:imports, -encoding, UTF-8, -feature, -unchecked)`. So it's supposed to work, but it's not. In fact, there are no diagnostic in SemanticDocument of this file, so the rule is not able to remove unused imports. 

* Weirdly removing `scalacOptions += "-Ywarn-unused" ` from `build.sbt` makes the test run, but many other tests will fail like RemoveUnusedStatements (still in 2.13)

I am not able to understand why changing `build.sbt` makes the test passes... I am not sure which scalaOptions are used to compile and get SemanticDocument for this test.

### DIFF
--- a/scalafix-tests/input/src/main/scala/test/AnchorPatches.scala
+++ b/scalafix-tests/input/src/main/scala/test/AnchorPatches.scala
@@ -5,7 +5,7 @@ ExplicitResultTypes.memberVisibility = [Public]
 unsafeShortenNames = true
 ExplicitResultTypes.unsafeShortenNames = true
  */
-package test.escapeHatch
+package test
 
 import scala.collection.mutable
 import scala.collection.immutable // scalafix:ok RemoveUnusedImports
--- a/scalafix-tests/output/src/main/scala/test/AnchorPatches.scala
+++ b/scalafix-tests/output/src/main/scala/test/AnchorPatches.scala
@@ -1,4 +1,4 @@
-package test.escapeHatch
+package test
 
 import scala.collection.immutable // scalafix:ok RemoveUnusedImports
 


bjaglin:
> I am not able to understand why changing build.sbt makes the test passes... I am not sure which scalaOptions are used to compile and get SemanticDocument for this test.

That looks like what I fixed in https://github.com/scalacenter/sbt-scalafix/pull/120/commits/4278a1f4f5e7ff28d086f8431c7e64af6f5d4084#diff-dea2d76d7c0198f6f23505efa21b3471L44. I didn't realize there was a copy of it in https://github.com/scalacenter/scalafix/blob/master/build.sbt#L197-L227, so if you updated it, the right `scalacOptions` should be used.

bjaglin:
https://github.com/scalacenter/sbt-scalafix/pull/125/files would also need to be backported, but I guess it's not really required.

mlachkar:
Thanks, I will check this

mlachkar:
ScalacOptions are correct to compile testInput 
```
sbt:unit> show scalacOptions
[info] * -target:jvm-1.8
[info] * -Wunused:imports
[info] * -encoding
[info] * UTF-8
[info] * -feature
[info] * -unchecked
``` 
I will need in fact this change at least to change scalaVersion for the other PR, but here it's doesn't fix my problem.

mlachkar:
Clearly I don't understand everything about this test. For sure test don't compile TestsInput, which is compiled with the right scalacOptions.
Moving this test outside of escapeHatch makes the test passes for 2.13 and all the other scala versions.

bjaglin:
I'll have a look at that this week!

PR #1173, opened 2020-06-21
Document local rules for sbt builds

bjaglin:
Merge only when https://github.com/scalacenter/sbt-scalafix/pull/121 is released

This is the public documentation for https://github.com/scalacenter/sbt-scalafix/pull/100. It's a bit late as I found https://github.com/scalacenter/sbt-scalafix/pull/121 to be key for any non-2.12 project to use it. Since the original PR was not thoroughly reviewed/advertized, I guess it would be nice to iterate on the implementation itself (that we have been using internally with `ScalafixTestkitPlugin` on a multi-project build), as documentation highlights possible improvements & shortcomings.

I decided to create a separate section as a first step, since integrating it in the current tutorial was quite demanding. Depending on the main target for Scalafix rules development (lib or application authors?), this could become the main tutorial.

### DIFF
--- a/docs/developers/local-rules.md
+++ b/docs/developers/local-rules.md
@@ -1,0 +1,212 @@
+---
+id: local-rules
+title: Local rules
+---
+
+If your custom rules only target a single codebase or repository, you can
+define and develop these rules as part of the same sbt build. This way, you do
+not have to version your rules and publish them to a Maven or Ivy repository.
+
+When implementing rules that are very specific to your domain, this strategy
+can be interesting as it allows rules to be added or extended in the same
+change sets as the domain sources they will be run against.
+
+## Prerequisite
+
+Make sure the sbt plugin, as well as the Scala compiler plugin and options [are
+set up correctly](../users/installation.md#sbt).
+
+Although it is possible to define your rules in a Scala binary version that
+does not match your build, it is highly recommended that you align them via:
+
+```diff
+ // build.sbt
++scalafixScalaBinaryVersion in ThisBuild :=
++  CrossVersion.binaryScalaVersion(scalaVersion.value)
+```
+
+> Note that any potential external Scalafix rule, loaded with the
+> `scalafixDependencies` setting key, must be built and published against the
+> same Scala binary version.
+
+## Single-project sbt build
+
+If you do not have any sub-project in your build, custom rules should be
+defined under the Scalafix Ivy configuration sources directory, which by
+default is `src/scalafix/scala`.
+
+```diff
+ repository
+ ├── build.sbt
+ ├── project
+ │   └── plugins.sbt
+ └── src
+     ├── main
+     │   └── ...
+     ├── test
+     │   └── ...
++    └── scalafix
++       ├── resources/META-INF/services
++       │   └── scalafix.v1.Rule // not mandatory, but allows syntax
++       |                        // `MyRule1` over `class:fix.MyRule1`
++       └── scala/fix
++           ├── MyRule1.scala
++           ├── MyRule2.scala
++           └── ...
+```
+
+```diff
+ // build.sbt
++libraryDependencies +=
++  "ch.epfl.scala" %%
++    "scalafix-core" %
++    _root_.scalafix.sbt.BuildInfo.scalafixVersion %
++    ScalafixConfig
+```
+
+```bash
+$ sbt
+> scalafix MyRule1 MyRule2
+```
+
+## Multi-project sbt build
+
+If your build contains several sub-projects, rules can be defined either:
+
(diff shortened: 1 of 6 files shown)


docs/developers/local-rules.md
```diff
@@ -0,0 +1,196 @@
...
++      sourceDirectories.in(`scalafix-input`, Compile).value,
++    scalafixTestkitInputClasspath :=
++      fullClasspath.in(`scalafix-input`, Compile).value,
++    scalafixTestkitInputScalacOptions :=
++      scalacOptions.in(`scalafix-input`, Compile).value,
++    scalafixTestkitInputScalaVersion :=
++      scalaVersion.in(`scalafix-input`, Compile).value
++  )
++  .dependsOn(`scalafix-input`, `scalafix-rules`)
++  .enablePlugins(ScalafixTestkitPlugin)
```
bjaglin, line 195: diff is nice but hard to copy/paste, so I wonder if it's not better to leave it raw when 99% of the things are new, and the snippet is big

docs/developers/local-rules.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+> Note that any potential external Scalafix rule, loaded with the
+> `scalafixDependencies` setting key, must be built and published against the
+> same Scala binary version.
+
+## Single-project sbt build
+
+If you do not have any sub-project in your build, custom rules should be
+defined under the Scalafix Ivy configuration sources directory, which by
```
bjaglin, line 35: we found this to be a problem with IntelliJ's built-in compiler, as unknown Ivy configurations fallback to `Compile` without any proper eviction. I'll follow up with some notes & recommendations when that's a problem.

docs/developers/local-rules.md
```diff
@@ -0,0 +1,212 @@
...
+
+```diff
+ // build.sbt
+ lazy val service1 = project
++  .settings(
++    libraryDependencies +=
++      "ch.epfl.scala" %%
++        "scalafix-core" %
++        _root_.scalafix.sbt.BuildInfo.scalafixVersion %
++        ScalafixConfig
```
bjaglin, line 122: I guess we could make this automatic, as long as we update https://github.com/scalacenter/sbt-scalafix/blob/60199fdae7bb9c56b9d09969b0889923a6b39108/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L201-L202 to avoid always invalidating the default tool classpath.

docs/developers/local-rules.md
```diff
@@ -0,0 +1,212 @@
...
++├── scalafix // same source structure as the scalafix.g8 project template
++│   ├── input/src/main/scala/fix
++│   │   ├── Test1.scala
++│   │   ├── Test2.scala
++│   │   └── ...
++│   ├── output/src/main/scala/fix
++│   │   ├── Test1.scala
++│   │   ├── Test2.scala
++│   │   └── ...
++│   ├── rules/src
```
bjaglin, line 151: this could be 
```suggestion
+│   ├── rules/scalafix
```
to potentially benefit from automatic addition of `scalafix-core`, but it would drift from the current template and would make the sbt `dependsOn` more awkward (`scalafix->scalafix`), so I decided to stick to it.

docs/developers/local-rules.md
mlachkar: I think we could have the final file here (instead of the diff). 
By the way, I like your comments/questions you add on your own PR, I find this quite useful and helpful for the review.

docs/developers/local-rules.md
mlachkar: I don't get it. It's normal to add scalafix-core to library dependencies?

bjaglin: It would be OK to add scalafix-core to library dependencies *in the ScalafixConfig* configuration, since anything there is meant to be rules, which require types from there.

Another downside is that it would potentially expose classpath collisions in IntelliJ mentioned in https://github.com/scalacenter/scalafix/pull/1173/files#r443266186 to all plugin users, whether or not they use local rules.

docs/developers/local-rules.md
bjaglin: > By the way, I like your comments/questions you add on your own PR, I find this quite useful and helpful for the review.

Thanks! However, sometimes they give context that should be captured in the git history as a source/commit comment instead, so I am trying to always question them since they can be an anti-pattern for long-term maintainability.

bjaglin:
For information, I am going to write an article on my company's blog https://medium.com/teads-engineering to explain in details our usage Scalafix and this. Maybe this is something we can integrate also, either here or in a 1.0 release post?

PR #1174, opened 2020-06-22
Add support for the 3 last versions of scala for each binary version for rules

mlachkar:


scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -19,10 +21,20 @@ final class ExplicitResultTypes(
...
 
-  private def supportedScalaVersion = Properties.versionNumberString
+  private def supportedScalaVersions: Set[String] = {
+    val scalaVersion = Properties.versionNumberString
+    val split = scalaVersion.split('.')
+    val binaryVersion = split.take(2).mkString(".")
+    val compilerVersion = Try(split.last.toInt).toOption
+    val supprtedVersions = compilerVersion
+      .map(int => Set(int - 2, int - 1, int).filter(_ >= 0))
+      .getOrElse(Set())
```
bjaglin, line 31:
```suggestion
    val supportedVersions = List.range(compilerVersion-2, compilerVersion).filter(_>=0)
```

project/ScalafixBuild.scala
```diff
@@ -194,15 +194,27 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         "unit/test" ::
         "docs/run" ::
         "interfaces/doc" ::
+        s"++$scala213V1" ::
```
bjaglin, line 197: instead of hardcoding the previous versions, you can probably generate the list dynamically and loop over it

mlachkar: I have more tests that fails when testing against more scala Version. 
I was thinking about adding a specific test that will test only ExplicitResultType : 
`+unit/testOnly scalafix.tests.rule.RuleSuite -- -z explicitResult `

mlachkar:
If I set scalaVersion for both testsInput and testsOutput to another scalaVersion, it's supposed to verify what we want to test.
The test will look like :
```
set testsInput/scalaVersion := "2.12.8"
set testsOutput/scalaVersion := "2.12.8"
unit/testOnly scalafix.tests.rule.RuleSuite -- -z explicitResult
```
But I noticed that it's not failing if it's not within the 3 versions range. The version check uses as the scala version from TestkitProperties.version, which is not the correct one (?)

bjaglin:
> The version check uses as the scala version from TestkitProperties.version, which is not the correct one (?)

That sounds a lot like the same problem as https://github.com/scalacenter/scalafix/pull/1169#issuecomment-646581997, except that this time it's `scalaVersion` (and not `scalacOptions`) that it picked up from the test runner instead of the compiled sources.

mlachkar:
I will check that.

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -19,10 +21,20 @@ final class ExplicitResultTypes(
 
   def this() = this(ExplicitResultTypesConfig.default, LazyValue.now(None))
 
-  private def supportedScalaVersion = Properties.versionNumberString
+  private def supportedScalaVersions: List[String] = {
```
mlachkar, line 24: I am not completely satisfied of duplicating this logic. Is there a way to retrieve `supportedScalaVersionsForExplicitResultTypeRule` from builldInfo

project/ScalafixBuild.scala
```diff
@@ -212,6 +213,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
     },
+    commands += Command.command("ci-explicit-result-types-rule") { s =>
+      supportedScalaVersionforExplicitResultRule
+        .flatMap {
+          case (k, versions) =>
+            versions.flatMap { v =>
+              List(
+                s"""set ThisBuild/scalaVersion := "$k"""",
+                s"""set  testsInput/scalaVersion := "$v"""",
+                s"""set testsOutput/scalaVersion := "$v\"""",
```
bjaglin, line 224: that's not really required as only the source is used for assertion

project/Dependencies.scala
```diff
@@ -36,4 +42,13 @@ object Dependencies {
...
   )
+
+  private def supportedScalaVersions(scalaVersion: String): List[String] = {
+    val split = scalaVersion.split('.')
+    val binaryVersion = split.take(2).mkString(".")
+    val compilerVersion = Try(split.last.toInt).toOption
+    val supportedVersions =
+      compilerVersion.map(version => List.range(version - 2, version + 1).filter(_ >= 0)).getOrElse(Nil)
+    supportedVersions.map(v => s"$binaryVersion.$v")
+  }
```
bjaglin, line 53: I think this should be called after what it does, not after what it is used for
```suggestion
  private def latestVersions(scalaVersion: String): List[String] = {
    val split = scalaVersion.split('.')
    val binaryVersion = split.take(2).mkString(".")
    val compilerVersion = Try(split.last.toInt).toOption
    val latestPatchVersions =
      compilerVersion.map(version => List.range(version - 2, version + 1).filter(_ >= 0)).getOrElse(Nil)
    latestPatchVersions.map(v => s"$binaryVersion.$v")
  }
```

project/ScalafixBuild.scala
```diff
@@ -212,6 +213,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         s"unit/testOnly -- -l scalafix.internal.tests.utils.SkipWindows" ::
         s
     },
+    commands += Command.command("ci-explicit-result-types-rule") { s =>
```
bjaglin, line 216: Instead of doing a new command that loops through all binary versions, maybe we could patch existing `ci-212` & `ci-213` commands to run an extra `unit/test scalafix.tests.rule.RuleSuite` for the N-1 & N-2 versions (since N is already run as part of the `unit/test`)? I think CI failures there would be clearer to troubleshoot (and everything would be already compiled)

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -57,11 +69,21 @@ final class ExplicitResultTypes(
           )
         }
       }
-    if (config.scalacClasspath.nonEmpty && config.scalaVersion != supportedScalaVersion) {
+    if (config.scalacClasspath.nonEmpty && !supportedScalaVersions.contains(
+        config.scalaVersion
+      )) {
+      val shortScalaVersion =
```
bjaglin, line 75:
```suggestion
      val inputBinaryScalaVersion =
```

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -57,11 +69,21 @@ final class ExplicitResultTypes(
...
+        config.scalaVersion.split(".").take(2).mkString(".")
+      val shortRuleScalaVersion =
```
bjaglin, line 77:
```suggestion
      val runtimeBinaryScalaVersion =
```

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -57,11 +69,21 @@ final class ExplicitResultTypes(
...
-        s"The ExplicitResultTypes rule only supports the Scala version '$supportedScalaVersion'. " +
-          s"To fix this problem, either remove `ExplicitResultTypes` from .scalafix.conf or change the Scala version " +
-          s"in your build to match exactly '$supportedScalaVersion'."
+        s"Your scalaVersion is ${config.scalaVersion} but the ExplicitResultTypes rule is compiled with ${shortRuleScalaVersion}" +
+          s"and only supports these Scala versions '$supportedScalaVersions'. " +
+          (if (shortScalaVersion == shortRuleScalaVersion)
+             s"To fix this problem, either remove `ExplicitResultTypes` from .scalafix.conf or change the Scala version to be within '$supportedScalaVersions'"
+           else
+             s"To fix this problem, either remove `ExplicitResultTypes` from .scalafix.conf or " +
+               s"specify a `scalafixScalaBinaryVersion` key in your build.sbt equal to  $shortScalaVersion")
```
bjaglin, line 86: For clarity, I think this message could be different depending on whether the binary version matches or not.

#### inputBinaryScalaVersion == runtimeBinaryScalaVersion
> The ExplicitResultTypes rule only supports the exact Scala versions '$supportedScalaVersion' for this binary version.  To fix this problem, either remove `ExplicitResultTypes` from `.scalafix.conf` or change the Scala version of your build to match one of the supported versions.

($supportedScalaVersion should list the full versions supported for the binary version $inputBinaryScalaVersion)

#### inputBinaryScalaVersion != runtimeBinaryScalaVersion
>The ExplicitResultTypes rule needs to run with the same Scala binary version as the one used to compile target sources ($inputBinaryScalaVersion). To fix this problem, either remove `ExplicitResultTypes` from `.scalafix.conf` or make sure the `scalafixScalaBinaryVersion` setting key matches  $inputBinaryScalaVersion.

project/ScalafixBuild.scala
```diff
@@ -67,6 +67,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         "scalameta" -> scalametaV,
         scalaVersion,
         "supportedScalaVersions" -> supportedScalaVersions,
+        "supportedScalaVersionsForExplicitResultTypeRule" -> supportedScalaVersionforExplicitResultRule,
```
bjaglin, line 70: this is the build info of core, I don't think it should know anything about built-in rules

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: definitely! you should define a specific buildInfo for the rules project in build.sbt - look for stuff with the `buildInfo` prefix there to see how it works (keys + plugin activation)

project/Dependencies.scala
```diff
@@ -11,6 +13,10 @@ object Dependencies {
   def coursierV = "2.0.0-RC5-6"
   def coursierInterfaceV = "0.0.22"
   val currentScalaVersion = scala213
+  // we support 3 last binary versions of each scala Version, except for 2.11 because semanticdb-scalac_2.11.10 is missing.
+  val supportedScalaVersionforExplicitResultRule =
```
bjaglin, line 17: I think this should not be specific to a single rule - we could test all of them (`unit/testOnly scalafix.tests.rule.RuleSuite`) against that set of versions (which will only be useful for those that use the presentation compiler, but at least we won't miss cross-testing it when a new one is introduced).

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
mlachkar: ok

project/ScalafixBuild.scala
bjaglin: > you mean set testsOutput/scalaVersion := "$v is not required.

yes, that specific line is not required, but the 2 others are indeed!

mlachkar:
Thanks for your review and comments @bjaglin. I do find the new version way better.

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -18,11 +18,14 @@ final class ExplicitResultTypes(
...
+  val supportedScalaVersions = BuildInfo.supportedScalaVersionforRules
+  val runtimeScalaVersion = BuildInfo.scalaVersion
+
+  private def toBinaryVersion(v: String) = v.split('.').take(2).mkString(".")
 
-  private def supportedScalaVersion = Properties.versionNumberString
   override def description: String =
     "Inserts type annotations for inferred public members. " +
-      s"Only compatible with Scala $supportedScalaVersion."
+      s"Only compatible with Scala $supportedScalaVersions."
```
bjaglin, line 28: does the `toString` representation of that map reads well? maybe we should only show the values (and maybe mkstring them) ?
```suggestion
      s"Only compatible with Scala ${supportedScalaVersions.values.flatten}."
```

project/ScalafixBuild.scala
```diff
@@ -190,21 +190,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         s
     },
     commands += Command.command("ci-213") { s =>
-      s"++$scala213" ::
+      s"""set ThisBuild/scalaVersion := "$scala213"""" ::
```
bjaglin, line 193: wasn't the previous `++` syntax working?

project/ScalafixBuild.scala
```diff
@@ -342,4 +342,16 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       }
     }
   )
+  private def testRulesFor(scalaVersion: String, state: State): State = {
+    supportedScalaVersionforRules(scalaVersion)
+      .filter(_ != scalaVersion)
+      .flatMap { v =>
+        List(
+          s"""set testsInput/scalaVersion := "$v"""",
+          s"""set testsOutput/scalaVersion := "$v"""",
```
bjaglin, line 351: as discussed in the other thread, this does not need to be (re)-compiled, as only sources are used for assertions
```suggestion
```

project/ScalafixBuild.scala
```diff
@@ -342,4 +342,16 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
+          s"unit/testOnly scalafix.tests.rule.RuleSuite"
```
bjaglin, line 352: I am not familiar with the syntax, but can that `RuleSuite` be excluded from the `unit/test` above? (to avoid the current version being tested twice)

project/ScalafixBuild.scala
```diff
@@ -342,4 +342,16 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       }
     }
   )
+  private def testRulesFor(scalaVersion: String, state: State): State = {
+    supportedScalaVersionforRules(scalaVersion)
+      .filter(_ != scalaVersion)
+      .flatMap { v =>
+        List(
+          s"""set testsInput/scalaVersion := "$v"""",
```
bjaglin, line 350: looking at the logs, it's not clear what version is set 
> 2020-07-01T13:07:50.4524207Z [info] Defining testsInput / scalaVersion
2020-07-01T13:07:50.4525011Z [info] The new value will be used by testsInput / Compile / compileBinaryFileInputs, testsInput / Compile / compileSourceFileInputs and 34 others.
2020-07-01T13:07:50.4525764Z [info] 	Run `last` for details.

so maybe we can make it explicit for troubleshooting?

```suggestion
          s"""set testsInput/scalaVersion := "$v"""",
          "show testsInput/scalaVersion",
```

build.sbt
```diff
@@ -71,13 +71,17 @@ lazy val rules = project
   .in(file("scalafix-rules"))
   .settings(
     moduleName := "scalafix-rules",
+    buildInfoKeys ++= Seq[BuildInfoKey](
+      "supportedScalaVersionforRules" -> supportedScalaVersionforRules
```
bjaglin, line 75: since this is local to the rules module, I don't think the suffix is required for the key
```suggestion
      "supportedScalaVersions" -> supportedScalaVersionforRules
```

build.sbt
```diff
@@ -71,13 +71,17 @@ lazy val rules = project
...
+    ),
     description := "Built-in Scalafix rules",
     libraryDependencies ++= List(
       "org.scalameta" % "semanticdb-scalac-core" % scalametaV cross CrossVersion.full,
       collectionCompat
     )
   )
   .dependsOn(core)
+  .enablePlugins(BuildInfoPlugin)
```
bjaglin, line 84: did you confirm the file was generated in a package that wouldn't potentially conflict with other build infos depending on the classpath loading?

project/ScalafixBuild.scala
mlachkar: you're right .. forgot to remove it.

project/ScalafixBuild.scala
mlachkar: once you do `++`, I cannot set specific scalaVersion for any project.

project/ScalafixBuild.scala
mlachkar: The current version is not tested twice anymore, since I filter `.filter(_ != scalaVersion) `

build.sbt
mlachkar: No, and actually I dont know how to verify that. 
In fact reflect dependsOn core and rules, which both contains BuildInfo. I will change buildInfo object name .

project/ScalafixBuild.scala
bjaglin: isn't the current version tested in the `unit/test` that exists on master?

mlachkar: the current version is tested in fact in unit/test,
then we set a different version for testsInput that is different from the current version. 

Currently, we don't test at the same place ruleSuite for all supportedVersions..

bjaglin: my bad I misread the filter indeed!

bjaglin: maybe a clearer implementation would be to rename `latestVersions` to `previousVersions`, to exclude the current version in the implementation

build.sbt
```diff
@@ -71,13 +71,18 @@ lazy val rules = project
   .in(file("scalafix-rules"))
   .settings(
     moduleName := "scalafix-rules",
+    buildInfoKeys ++= Seq[BuildInfoKey](
+      "supportedScalaVersions" -> supportedScalaVersionforRules
+    ),
```
bjaglin, line 76:
```suggestion
    buildInfoKeys ++= Seq[BuildInfoKey](
      "supportedScalaVersions" -> scalaVersion.value :: testedPreviousScalaVersions(scalaBinaryVersion.value)
    ),
```

project/Dependencies.scala
```diff
@@ -11,6 +13,10 @@ object Dependencies {
   def coursierV = "2.0.0-RC5-6"
   def coursierInterfaceV = "0.0.22"
   val currentScalaVersion = scala213
+  // we support 3 last binary versions of each scala Version, except for 2.11 because semanticdb-scalac_2.11.10 is missing.
+  val supportedScalaVersionforRules =
+    List(scala213, scala212).map(version => version -> latestVersions(version)).toMap ++
+      Map(scala211 -> List(scala211))
```
bjaglin, line 19:
```suggestion
  val testedPreviousScalaVersions =
    List(scala213, scala212).map(version => version -> previousVersions(version)).toMap 
```

project/Dependencies.scala
```diff
@@ -37,4 +43,13 @@ object Dependencies {
...
   )
+
+  private def latestVersions(scalaVersion: String): List[String] = {
+    val split = scalaVersion.split('.')
+    val binaryVersion = split.take(2).mkString(".")
+    val compilerVersion = Try(split.last.toInt).toOption
+    val latestPatchVersions =
+      compilerVersion.map(version => List.range(version - 2, version + 1).filter(_ >= 0)).getOrElse(Nil)
+    latestPatchVersions.map(v => s"$binaryVersion.$v")
+  }
```
bjaglin, line 54:
```suggestion
  private def previousVersions(scalaVersion: String): List[String] = {
    val split = scalaVersion.split('.')
    val binaryVersion = split.take(2).mkString(".")
    val compilerVersion = Try(split.last.toInt).toOption
    val previousPatchVersions =
      compilerVersion.map(version => List.range(version - 2, version).filter(_ >= 0)).getOrElse(Nil)
    previousPatchVersions.map(v => s"$binaryVersion.$v")
  }
```

project/ScalafixBuild.scala
```diff
@@ -342,4 +342,16 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       }
     }
   )
+  private def testRulesFor(scalaVersion: String, state: State): State = {
+    supportedScalaVersionforRules(scalaVersion)
+      .filter(_ != scalaVersion)
```
bjaglin, line 347:
```suggestion
  private def testRulesAgainstPreviousScalaVersions(scalaVersion: String, state: State): State = {
    testedPreviousScalaVersions(scalaVersion)
```

project/ScalafixBuild.scala
```diff
@@ -190,21 +190,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         s
     },
     commands += Command.command("ci-213") { s =>
-      s"++$scala213" ::
+      s"""set ThisBuild/scalaVersion := "$scala213"""" ::
         "unit/test" ::
         "docs/run" ::
         "interfaces/doc" ::
-        s
+        testRulesFor(scala213, s)
```
bjaglin, line 197:
```suggestion
        testRulesForPreviousScalaVersions(scala213, s)
```

project/ScalafixBuild.scala
```diff
@@ -190,21 +190,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
     },
     commands += Command.command("ci-212") { s =>
-      s"++$scala212" ::
+      s"""set ThisBuild/scalaVersion := "$scala212"""" ::
         "unit/test" ::
-        s
+        testRulesFor(scala212, s)
```
bjaglin, line 202:
```suggestion
        testRulesForPreviousScalaVersions(scala212, s)
```

project/ScalafixBuild.scala
```diff
@@ -190,21 +190,21 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
     },
     commands += Command.command("ci-211") { s =>
-      s"++$scala211" ::
+      s"""set ThisBuild/scalaVersion := "$scala211"""" ::
         "unit/test" ::
-        s
+        testRulesFor(scala211, s)
```
bjaglin, line 207:
```suggestion
        testRulesForPreviousScalaVersions(scala211, s)
```

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -18,11 +18,14 @@ final class ExplicitResultTypes(
 ) extends SemanticRule("ExplicitResultTypes") {
 
   def this() = this(ExplicitResultTypesConfig.default, LazyValue.now(None))
+  val supportedScalaVersions = RulesBuildInfo.supportedScalaVersions
+  val runtimeScalaVersion = RulesBuildInfo.scalaVersion
```
bjaglin, line 22: As far as I remember, I suggested the name `runtimeScalaVersion` because you were using `scala.util.Properties.versionNumberString` in an earlier commit of this PR. Now that it's using `RulesBuildInfo`, I think the right naming would be
```suggestion
  val compilerScalaVersion = RulesBuildInfo.scalaVersion
```

In some rare cases, I guess the runtime can differ from that. But from my understanding anyway, this is actually closer to what we want: compare the presentation compiler version brought by scalameta (using the exact same scalaVersion as `rules`) against the version of scalac used for the input.

Am I getting this right @olafurpg ?

bjaglin: Anyway, this is just naming so it can be tweaked later (maybe in the same PR as the previous set of cosmetic suggestions I had), I just wanted to make sure we were doing/testing the right thing here.

dwijnand: > against the version of scalac used for the input

JFYI, there's also `scala.tools.nsc.Properties.versionNumberString` which is the compiler's version, but it's highly unlikely to vary compared to the standard library version.

bjaglin: > it's highly unlikely to vary compared to the standard library version.

I guess that can only happen if the scala version of scalafix-core & scalameta drift, which is rarely the case (https://github.com/scalacenter/scalafix/pull/1179/files bumps both for example)
```
$ cs resolve -t ch.epfl.scala::scalafix-rules:0.9.17
  Result:
└─ ch.epfl.scala:scalafix-rules_2.13:0.9.17
   ├─ ch.epfl.scala:scalafix-core_2.13:0.9.17
   │  ├─ com.geirsson:metaconfig-typesafe-config_2.13:0.9.10
   │  │  ├─ com.geirsson:metaconfig-core_2.13:0.9.10
   │  │  │  ├─ com.lihaoyi:pprint_2.13:0.5.9
   │  │  │  │  ├─ com.lihaoyi:fansi_2.13:0.2.9
   │  │  │  │  │  └─ com.lihaoyi:sourcecode_2.13:0.2.1
   │  │  │  │  └─ com.lihaoyi:sourcecode_2.13:0.2.1
   │  │  │  ├─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │  │  │  ├─ org.scala-lang.modules:scala-collection-compat_2.13:2.1.2 -> 2.1.6
   │  │  │  │  └─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │  │  │  └─ org.typelevel:paiges-core_2.13:0.3.0
   │  │  │     └─ org.scala-lang:scala-library:2.13.0 -> 2.13.2
   │  │  ├─ com.typesafe:config:1.2.1
   │  │  └─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │  ├─ com.googlecode.java-diff-utils:diffutils:1.3.0
   │  ├─ org.scala-lang:scala-library:2.13.2
   │  ├─ org.scala-lang.modules:scala-collection-compat_2.13:2.1.6
   │  │  └─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │  └─ org.scalameta:scalameta_2.13:4.3.14
   │     ├─ org.scala-lang:scala-library:2.13.2
   │     ├─ org.scala-lang:scalap:2.13.2
   │     │  └─ org.scala-lang:scala-compiler:2.13.2
   │     │     ├─ net.java.dev.jna:jna:5.3.1
   │     │     ├─ org.jline:jline:3.14.1
   │     │     ├─ org.scala-lang:scala-library:2.13.2
   │     │     └─ org.scala-lang:scala-reflect:2.13.2
   │     │        └─ org.scala-lang:scala-library:2.13.2
   │     └─ org.scalameta:parsers_2.13:4.3.14
   │        ├─ org.scala-lang:scala-library:2.13.2
   │        └─ org.scalameta:trees_2.13:4.3.14
   │           ├─ com.thesamet.scalapb:scalapb-runtime_2.13:0.10.3
   │           │  ├─ com.google.protobuf:protobuf-java:3.11.4
   │           │  ├─ com.lihaoyi:fastparse_2.13:2.3.0
   │           │  │  ├─ com.lihaoyi:geny_2.13:0.6.0
   │           │  │  └─ com.lihaoyi:sourcecode_2.13:0.2.1
   │           │  ├─ com.thesamet.scalapb:lenses_2.13:0.10.3
   │           │  │  ├─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │           │  │  └─ org.scala-lang.modules:scala-collection-compat_2.13:2.1.6
   │           │  │     └─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │           │  ├─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │           │  └─ org.scala-lang.modules:scala-collection-compat_2.13:2.1.6
   │           │     └─ org.scala-lang:scala-library:2.13.1 -> 2.13.2
   │           ├─ org.scala-lang:scala-library:2.13.2
   │           ├─ org.scalameta:common_2.13:4.3.14
   │           │  ├─ com.lihaoyi:sourcecode_2.13:0.2.1
   │           │  └─ org.scala-lang:scala-library:2.13.2
   │           └─ org.scalameta:fastparse_2.13:1.0.1
   │              ├─ com.lihaoyi:sourcecode_2.13:0.1.7 -> 0.2.1 (possible incompatibility)
   │              ├─ org.scala-lang:scala-library:2.13.0 -> 2.13.2
   │              └─ org.scalameta:fastparse-utils_2.13:1.0.1
   │                 ├─ com.lihaoyi:sourcecode_2.13:0.1.7 -> 0.2.1 (possible incompatibility)
   │                 └─ org.scala-lang:scala-library:2.13.0 -> 2.13.2
   ├─ org.scala-lang:scala-library:2.13.2
```

dwijnand: It's impossible in sbt without disabling things as it has this `OverrideScalaMediator` (https://github.com/sbt/sbt/pull/2634) which keeps scala-compiler/scala-library/scala-reflect in sync.  And then there's are also, in the compiler, various sanity checks on the classes in the library, so it crashes early if you mix major versions for example.

project/ScalafixBuild.scala
```diff
@@ -342,4 +342,19 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       }
     }
   )
+  private def testRulesAgainstPreviousScalaVersions(
+      scalaVersion: String,
+      state: State
+  ): State = {
+    testedPreviousScalaVersions(scalaVersion)
+      .filter(_ != scalaVersion)
```
bjaglin, line 350: I think this is now redundant (if the `previousVersions` is updated!)
```suggestion
      .filter(_ != scalaVersion)
```

project/Dependencies.scala
```diff
@@ -37,4 +42,13 @@ object Dependencies {
     "com.chuusai" %% "shapeless" % "2.3.3",
     scalacheck
   )
+
+  private def previousVersions(scalaVersion: String): List[String] = {
+    val split = scalaVersion.split('.')
+    val binaryVersion = split.take(2).mkString(".")
+    val compilerVersion = Try(split.last.toInt).toOption
+    val previousPatchVersions =
+      compilerVersion.map(version => List.range(version - 2, version + 1).filter(_ >= 0)).getOrElse(Nil)
```
bjaglin, line 51:
```suggestion
      compilerVersion.map(version => List.range(version - 2, version).filter(_ >= 0)).getOrElse(Nil)
```

PR #1179, opened 2020-06-26
Upgrade to Scala 2.13.3

olafurpg:


project/Dependencies.scala
```diff
@@ -2,12 +2,12 @@ import sbt._
 /* scalafmt: { maxColumn = 120 }*/
 
 object Dependencies {
-  val scalametaV = "4.3.15"
+  val scalametaV = "4.3.16"
```
bjaglin, line 5: That was fast 👍

bjaglin, line 5: That was fast 👍

olafurpg: Not so fast to merge 😅

PR #1182, opened 2020-06-27
Github action: Cache Ivy, Sbt, Coursier

giabao:


olafurpg:
Thank you for reporting! Can you elaborate on the motivation for this change? I'm personally on the fence about enabling CI cache for GitHub Actions because my experience is that the speedup gains from the cache haven't been significant enough to outweigh the downsides of enabling cache (occasional need to clear cache)

bjaglin:
I agree with @olafurpg - the reasonnable build time (<10min) and the volume of PRs does not really justify the (hidden) cost of a cache.

tgodzik:
Btw. there is a coursier cache action that does it all by default and we've also never had to clear a github actions cache.

Issue #1198, opened 2020-07-03
fix ci-213-windows


bjaglin:
The GitHub action [started failing on master](https://github.com/scalacenter/scalafix/actions?query=workflow%3ACI+branch%3Amaster) with https://github.com/scalacenter/scalafix/actions/runs/155292713.

Initial investigation by @mlachkar and myself shows that it's a race condition or a combination of factors, as reverting back to Scala 2.13.2 (no matter what the scalameta version is) OR removing other GitHub actions workflows OR adding `println` OR reverting https://github.com/scalacenter/scalafix/pull/1174 make the problem disappear. We have not been able to reproduce the problem locally.

The debug logs added seem to show that the test failure comes from a [double-"crash" of the presentation compiler wrapper resulting in an empty patch](https://github.com/scalacenter/scalafix/blob/v0.9.18/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L91-L104). Adding some logs is sometimes enough to make the problem disappear, although [we observed a "recovered" crash](https://github.com/scalacenter/scalafix/runs/834295929):
```
2020-07-03T12:14:50.0692483Z first attempt failed, retrying...
2020-07-03T12:14:50.1985464Z [info] - test/explicitResultTypes/LocalImplicits.scala (227 milliseconds)
```

bjaglin:
It might be worth digging into https://github.com/scalacenter/scalafix/commit/ba74db38dde7fcdd4971b56b10ec94cb6fef7092 to get some background.

mlachkar:
Good idea.

bjaglin:
@olafurpg maybe you have some idea/context/pointers?

olafurpg:
What was the motivation to close the revert PR? https://github.com/scalacenter/scalafix/pull/1191

The diff in #1181 is a bit problematic because it uses the presentation compiler directly to lookup types of local symbols. Before that change, ExplicitResultTypes rule only inserted types for public members, which I think is a good constraint. The `skipLocalImplicits` configuration option should be removed, I think it was unused before #1181

bjaglin:
[...]

PR #1209, opened 2020-07-10
Remove support for ExplicitResultTypes.skipLocalImplicits option

bjaglin:
Fix https://github.com/scalacenter/scalafix/issues/1198

Reverts https://github.com/scalacenter/scalafix/pull/1181

Rationale: https://github.com/scalacenter/scalafix/issues/1198#issuecomment-656605904


bjaglin:
Unfortunately @giabao we'll have to revert your PR as the implementation of skipLocalImplicits is not safe.

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
```diff
@@ -45,13 +45,11 @@ class CompilerTypePrinter(g: ScalafixGlobal, config: ExplicitResultTypesConfig)(
...
   ): Option[v1.Patch] = {
     val gpos = unit.position(pos.start)
-    val tpe = GlobalProxy.typedTreeAt(g, gpos)
-    val inverseSemanticdbSymbol =
-      if (sym.isLocal) tpe.symbol
-      else
-        g.inverseSemanticdbSymbols(sym.value)
-          .find(s => g.semanticdbSymbol(s) == sym.value)
-          .getOrElse(g.NoSymbol)
+    GlobalProxy.typedTreeAt(g, gpos)
```
bjaglin, line 48: @olafurpg is this called just for its side effect https://github.com/scalacenter/scalafix/commit/ad8a4bb481d0a425b01b09a9dadee145c97d3599? That goes far beyong my understanding, but when it was added in https://github.com/scalacenter/scalafix/commit/162db399bbf5584632e617bf7c3135a600dc06c8#diff-0771fc97e5f15a43af8af2a9e65cbc62R137, there was no explicit side effect so I wonder.

olafurpg: Correct. This method performs “positioned type checking”, which loads up all the necessary symbols from the classpath. The return value is helpful in other situations like in metals where we use this method to implement parameter hints or completions

PR #1215, opened 2020-07-16
Ondemand compilation doesn't take scalacOptions

mlachkar:


mlachkar:
linked to this https://github.com/scalameta/scalameta/pull/2094

bjaglin:
Just landed on this and trying to understand the status: `scalacOptions` is actually passed but there is a bug in InteractiveSemanticdb for 2.13, right?

mlachkar:
So if we pass the scalacOptions, it fixes the onDemand compilaton for 2.11 and 2.12, but not 2.13

bjaglin:
Oh ok. I guess that's still something good to take in then (by removing tests for 2.13), and follow up 2.13 in another PR?

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -376,11 +376,15 @@ case class Args(
...
-      settings.classpath.value = validatedClasspath.syntax
-      val reporter = new ConsoleReporter(settings)
-      LazyValue.fromUnsafe(() => new Global(settings, reporter))
+      LazyValue.fromUnsafe(() => {
+        val ss = new Settings()
+        val command = new CompilerCommand(scalacOptions, ss)
+        val settings = command.settings
+        settings.YpresentationAnyThread.value = true
+        settings.classpath.value = validatedClasspath.syntax
+        val reporter = new StoreReporter()
```
bjaglin, line 385: out of curiosity: is this block evaluated lazily on purpose to capture some state that it not available eagerly?

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
```diff
@@ -257,4 +258,54 @@ class CliSemanticSuite extends BaseCliSuite {
     )
   }
 
+  test("on-demand with scalacOptions") {
+    // Todo: Ondemand compilation doesn't take scalacOptions or the reporter doesn't gather all warns in 2.13
+    // see https://github.com/scalacenter/scalafix/pull/1215
+    if (ScalaVersions.isScala213) true
```
bjaglin, line 264: should we open a separate issue to track that?

bjaglin (approved):
:+1: thanks for making this incremental!

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: So the creation of Global was lazy, I just put the entire block as lazy. 
Usually, if the semanticdb file is present, global is never instantiated. 
I don't think there is any state that is captured later..

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
mlachkar: ok

PR #1217, opened 2020-07-19
add --triggered flag to load overlay section in .scalafix.conf

taisukeoe:
This is a follow up PullRequest of https://github.com/scalacenter/sbt-scalafix/pull/148 .

As discussed at the previous PullRequest, we find it would be better to support on-compile specific scalafix configuration in CLI, rather than sbt-scalafix.

So that I would add `--on-compile` flag and support onCompile section in .scalafix.conf .

## sbt-scalafix follow up
With this PullRequest, sbt-scalafix can pass `--on-compile` as a scalafix InputTask argument. So scalafix running on-compile will respect `onCompile` configuration in `.scalafix.conf`.

And I have an extra idea:

  What do you think about adding `scalafixArgumentOnCompile: SettingKey[String]` (or `SettingKey[Seq[String]]`) key to sbt-scalafix?
  With `scalafixArgumentOnCompile := "--check"`, users can switch their rules to *lint only* in their every scalafix run on-compile, which might be useful.


scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -242,9 +247,28 @@ case class Args(
...
   }
 
+  // In running with onCompile flag, look for settings in onCompile block first, and fallback to standard settings.
+  //
+  // Please watch out that its merge scope is shallow. For example,
+  //   DisableSyntax.noVars = true
+  // will be overridden by
+  //   onCompile = {
+  //     DisableSyntax.noThrows = true
+  //   }
```
taisukeoe, line 257: I'm wondering what is the best... 🤔 
Or should we allow to override `DisableSyntax.noThrows` only,
 and keep `DisableSyntax.noVars = true` or other DisableSyntax settings?

Your feedback is much appreciated.

bjaglin:
Thanks for following up @taisukeoe - it's vacation time here, so expect some delays for the review

taisukeoe:
@bjaglin Thank you for your response.
No problem, please enjoy your vacation 👍

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
taisukeoe: I thought merging-shallow strategy is simpler and clearer, since users can understand what is going to run without merging Confs one by one in their mind.

However, I'd imagine several concrete use cases -- for example, users just want to turn `OrganizeImports.removeUnused` into `false` on compile. With merging-shallow strategy, they might face severe conf duplications especially in `OrganizeImports.groups` (or it can be generalized as `Conf.Lst`s), which are not pleasant.

So I have changed my mind and implementation with 342f74d . 
Now Confs are merged deeply, so that `onCompile` Conf.Obj can work just as a patch against original.

I'd still want to hear your opinions about this.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -81,6 +81,11 @@ case class Args(
         "configured in .scalafix.conf or via --rules"
     )
     syntactic: Boolean = false,
+    @Description(
+      "Run only rules defined in onCompile section in .scalafix.conf," +
+        "or rules explicitly passed via --rules"
```
bjaglin, line 86: Maybe this is a good place to mention the fact that the section only overlays the default stuff?
```suggestion
      "Overlay the default rules & rule settings in .scalafix.conf with the `onCompile` section"
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -242,9 +247,21 @@ case class Args(
     RuleDecoder.decoder(ruleDecoderSettings.withConfig(scalafixConfig))
   }
 
+  // With a --on-compile flag, looking for settings in onCompile block first, and fallback to standard settings.
+  def preProcessedConf(base: Conf): Conf =
```
bjaglin, line 251: if we use the `overlay` term, this could be more precise
```suggestion
  def maybeOnCompileOverlaidConf(base: Conf): Conf =
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -242,9 +247,21 @@ case class Args(
...
+    if (onCompile) {
+      val confOnCompile = ConfGet.getKey(base, "onCompile" :: Nil)
+      confOnCompile.fold(base)(
+        ConfOps.merge(ScalafixConfOps.drop(base, "onCompile"), _)
+      )
```
bjaglin, line 256: while you are at it, can't you extract the entire get/fold/merge logic to `ScalafixConfOps` so we better understand the intent here?

bjaglin (commented):
Big apologies for getting back so late to this PR! Thanks for following up after our long discussion in https://github.com/scalacenter/sbt-scalafix/pull/148, let's get this (and the sbt-scalafix companion to come) in before the next release!

I have a few implementation cosmetic comments, but mostly I am wondering if `onCompile` is not too specific as a build tool might invoke this in a different context than compile. From your use-case, I feel like you want a specific set of rules and/or rule settings when scalafix is triggered "in the background" (which can be "sneaky" for the user). Would `--non-interactive` / `nonInteractive` be a more generic name for that?

taisukeoe:
@bjaglin Thank you for your reviewing!
Indeed `on-compile` would be too specific, and it would be nicer to be relevant name for a build tool or an editor.

While I'd agree that `--non-interactive` can be applied to more general use case than `--on-compile`, I'm wondering how we can make it more straight forward name. (In general, it would be nicer to rephrase `non-XXX` to `YYY`).

Considering the nature that it adds a hook on compile, what do you think about `--hooked` or `--triggered`?

I'm happy to apply suggested changes to this branch, once we fix the term.
I'm planning to do this stuff, let's say this weekend. Is it OK for your release schedule?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainMode.java
```diff
@@ -29,4 +29,8 @@
      */
     AUTO_SUPPRESS_LINTER_ERRORS,
 
+    /**
+     * Enable onCompile mode and let scalafix use onCompile configuration in .scalafix.conf.
+     */
+    ON_COMPILE
```
bjaglin, line 35:
```suggestion
    /**
     * Use when the client triggers the run as a side effect of something else, as opposed to an explicit, interactive invocation. Write fixed contents in-place, with a custom configuration if it exists.
     */
    IN_PLACE_TRIGGERED
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -106,7 +106,7 @@ ScalafixArguments withToolClasspath(
     ScalafixArguments withConfig(Optional<Path> config);
 
     /**
-     * @param mode The mode to run via --check or --stdout or --auto-suppress-linter-errors
+     * @param mode The mode to run via --check or --stdout or --auto-suppress-linter-errors or --on-compile
```
bjaglin, line 109:
```suggestion
     * @param mode The mode to run via --check or --stdout or --auto-suppress-linter-errors or --triggered
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -81,6 +81,11 @@ case class Args(
         "configured in .scalafix.conf or via --rules"
     )
     syntactic: Boolean = false,
+    @Description(
+      "Run only rules defined in onCompile section in .scalafix.conf," +
+        "or rules explicitly passed via --rules"
+    )
+    onCompile: Boolean = false,
```
bjaglin, line 88:
```suggestion
    @Description(
      "Overlay the default rules & rule settings in .scalafix.conf with the `triggered` section"
    )
    triggered: Boolean = false,
```

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSyntacticSuite.scala
```diff
@@ -175,6 +175,34 @@ class CliSyntacticSuite extends BaseCliSuite {
...
 
+  check(
+    name = "--on-compile is respected",
+    originalLayout = s"""|
+                         |/.scalafix.conf
+                         |rules = [
+                         |  RemoveUnused
+                         |]
+                         |
+                         |onCompile.rules = [ProcedureSyntax]
```
bjaglin, line 186:
```suggestion
                         |triggered.rules = [ProcedureSyntax]
```

bjaglin (commented):
> While I'd agree that `--non-interactive` can be applied to more general use case than `--on-compile`, I'm wondering how we can make it more straight forward name. (In general, it would be nicer to rephrase `non-XXX` to `YYY`).
> 
> Considering the nature that it adds a hook on compile, what do you think about `--hooked` or `--triggered`?

I would say `triggered` is the best proposal so far - I went through the PR and added suggestions in a few places to see how it looks. @mlachkar your opinion is welcome on this as I guess this can be useful for the Metals integration.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
taisukeoe: `Overlay` seems to be a nicer word 👍

taisukeoe:
I've reflected review comments with `triggered`.
I'm still fine to change the term `triggered` if better proposal comes later.

bjaglin (approved):
:+1: LTGM!

@mlachkar any objection/opinion?

> What do you think about adding scalafixArgumentOnCompile: SettingKey[String] (or SettingKey[Seq[String]]) key to sbt-scalafix?
With scalafixArgumentOnCompile := "--check", users can switch their rules to lint only in their every scalafix run on-compile, which might be useful.

For that, we could have a `scalafixCheckOnCompile` setting key (mutually exclusive with `scalafixOnCompile`), which would request a `CHECK_TRIGGERED` mode execution I guess? Having an arbitrary list of arguments means that we wouldn't fully control the mode, so it can be confusing to know which configuration (overlay) is used, no?

taisukeoe:
> For that, we could have a scalafixCheckOnCompile setting key (mutually exclusive with scalafixOnCompile), which would request a CHECK_TRIGGERED mode execution I guess? 

Since we can just use both of CHECK and IN_PLACE_TRIGGERED at once, I wouldn't think we need to introduce new CHECK_TRIGGERED mode. I also worry that exclusiveness between different sbt key seems to be tricky.  

> Having an arbitrary list of arguments means that we wouldn't fully control the mode, so it can be confusing to know which configuration (overlay) is used, no?

I see your point.
If we have another mode to overlay config than `triggered` in future, it will be confusing.

bjaglin:
@olafurpg any objection?

PR #1218, opened 2020-07-30
Enable scalafix plugin

mlachkar:


mlachkar:
I have issues with semanticdb file not generated for tests anymore.

bjaglin:
> I have issues with semanticdb file not generated for tests anymore.

I think this is because you added the semanticdb compiler plugin via the sbt 1.3.x way (`SemanticDBPlugin`), which has a different output directory than the raw `scalac` way, used in https://github.com/scalacenter/scalafix/blob/dad392101f0106508cf5bc08ae01d5456dacace9/project/ScalafixBuild.scala#L83-L92.

mlachkar:
To fix the issue, I just didn't enable scalafix-semanticDb in the entire build. I don't know if it's the correct way to fix the issue I was experimenting with. 
I don't know if we want or not to enable scalafix rules. In my opinion, it's nice to have a number of rules applied automatically (organizeImport, and ExplicitResultType)

bjaglin:
> To fix the issue, I just didn't enable scalafix-semanticDb in the entire build. I don't know if it's the correct way to fix the issue I was experimenting with.

I think we should keep it everywhere - but you need to stick to one way to enable the plugin, not two as I mentioned above (probably the sbt 1.3 way documented in https://scalacenter.github.io/scalafix/docs/users/installation.html#sbt since we are running 1.3+)

bjaglin:
As for whether we should enable a few rules on the project itself: yes I think it's always good to eat your own dog food!

mlachkar:
A weird error is now happening with mdoc. 
One of the snippet of semantic-tree.md fails when running `docs/run`
```scala mdoc:fail
val x: Int = ""
```
```
[info] error: /Users/meriamlachkar/scalacenter/scalafix/docs/developers/semantic-tree.md:164:1: Expected compile errors but program compiled successfully without errors
[info] val x: Int = ""
[info] ^^^^^^^^^^^^^^^
```

And of course, it also fails if we remove mdoc`:fails`. I just removed `scala mdoc:fail` from this snippet for now

mlachkar:
@bjaglin I would like another review on this (since it required a rebase every time.) Thanks.
I will create another PR/issue to fix warnings and errors on documentation

bjaglin (approved):
Sure, let's get that in, I can open a PR to address https://github.com/scalacenter/scalafix/pull/1218#issuecomment-682527441 if you want?

mlachkar:
Now, it only uses the new way of semanticdb (sbt 1.3+). I removed all "org.scalameta" % "semanticdb-scalac".
So I think this comment has been addressed

mlachkar:
It's left only as Library dependency for Rules, since it's used to read semanticdb files.

Issue #1154, opened 2020-06-08
scalafix-interfaces: expose rewrites as patches


bjaglin:
Current situation as of Scalafix 0.9.17:
- https://github.com/scalacenter/scalafix/blob/v0.9.17/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixDiagnostic.java only exposes lint patches
- Rewrites diffs (as generated by java-diff-utils) can be observed by inspecting stdout with the `ScalafixMainMode.STDOUT` mode, but are not really actionnable

Goal:
- `scalafix-interfaces` clients should be able to decide when, how and which rewrites are applied

Client use cases:
- implement local refactoring actions in an IDE via Scalafix rules (https://github.com/scalameta/metals-feature-requests/issues/18?)
- attach actionable diffs to updated lines within the code review UI (using https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-a-suggested-change through https://github.com/reviewdog/reviewdog for example)

Possible implementations:
1. change the behavior of `ScalafixMainMode.STDOUT` to produce an actionnable diff using a format like [unified diff](https://en.wikipedia.org/wiki/Diff#Unified_format)
1. introduce `ScalafixMainMode.STDOUT_UNIFIED_FORMAT` to expose an actionnable diff using a format like [unified diff](https://en.wikipedia.org/wiki/Diff#Unified_format)
1. add a `ScalafixMainCallback.reportRewrite(ScalafixRewrite rewrite)` with `ScalafixRewrite` being something similar to https://github.com/java-diff-utils/java-diff-utils/blob/java-diff-utils-parent-4.7/java-diff-utils/src/main/java/com/github/difflib/patch/Patch.java, which gets called for all `ScalafixMainMode`s.

Limitations:
- Since (Scalafix) patches are applied sequentially, the information of which rule requested the rewrite is tricky to expose, and depends on application of "previous" rules rewrites, so clients willing to segregate rewrites by rules would have to run scalafix once per rule.

olafurpg:
This is a good idea! I'm 👍 for approaches 2 and 3. Approach 2 is already kind of implemented with the --check flag
```sh
❯ scalafix -r ProcedureSyntax --check project/Dependencies.scala
--- /Users/lgeirsson/dev/scalafix/project/Dependencies.scala
+++ <expected fix>
@@ -2,7 +2,7 @@
[...]

PR #1223, opened 2020-08-03
Add methods to scalafix api to be able to retrieve patches

mlachkar:
I added a ScalafixResult that contains:
* Patches 
* Diagnostics
* Error

Applying patches still consist on writing the file in the desired Path if there is a diff. 
You can also apply selective patches if needed. The idea is to enable another [project](https://contributors.scala-lang.org/t/the-scala-2-to-scala-3-typer-and-implicit-resolver/4446) that needs this feature.


scalafix-tests/unit/src/test/scala/scalafix/tests/rule/MavenFuzzSuite.scala
```diff
@@ -158,7 +160,7 @@ class MavenFuzzSuite extends AnyFunSuite with DiffAssertions {
         .withPaths(paths.asJava)
         .withRules(List(rule).asJava)
         .withClasspath(classfiles.asJava)
-      // .withMode(ScalafixMainMode.CHECK)
+        .withMode(ScalafixMainMode.CHECK)
```
adpi2, line 163: What does this mean?

adpi2 (changes requested):
This looks very good to me!

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixResult.java
```diff
@@ -0,0 +1,34 @@
...
+    ScalafixError[] getError();
+
+    /**
+     * @return Optional<String> if the run is not successful and an error message is provided.
+     */
+    Optional<String> getMessageError();
+
+    /**
+     * G
+     * @return ScalafixOutput[] for each file we store diagnostics and patches
```
bjaglin, line 26: clarify if files that have no diagnostics nor patches (=empty patch) will show up here

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return the output file after running scalafix if no error
+     */
+    Optional<String> getOutputFileFixed();
+
+
+    /**
+     *
+     * @return scalafixPatches
+     */
```
bjaglin, line 55:
```suggestion
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
```diff
@@ -0,0 +1,9 @@
+package scalafix.interfaces;
+
+public interface ScalafixPatch {
```
bjaglin, line 3: do we need to expose a patch as single unit? it's coupled to a file and usually use together with others, so instead I would take a kind or an id as parameter on the `apply` overloads (since I guess the only purpose of getting them is to filter before applying them?)

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return list of diagnostics
+     */
+    ScalafixDiagnostic[] getDiagnostics();
+
+    /**
+     *
+     * @return if the fixed file is different from the input file, this method return non empty unified diff
+     */
+    Optional<String> getUnifiedDiff();
```
bjaglin, line 42: :+1: great for https://github.com/scalacenter/scalafix/issues/1154

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixResult.java
```diff
@@ -0,0 +1,34 @@
+package scalafix.interfaces;
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public interface ScalafixResult {
```
bjaglin, line 6: I was not expecting to see side-effect methods on a `Result` object, which seems like the end of the process. I can't really find the right term for it, but I was thinking about something along
```suggestion
public interface ScalafixEvaluation {
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -205,4 +205,14 @@ ScalafixArguments withToolClasspath(
...
+    /**
+     *
+     * Compared to {@link #run()}, it doesn't write the diff to the requested file,
+     * but instead return the list of patches and diagnostics, which can be written to file using
+     * {@link ScalafixResult#writeResult()}}.
+     * withCallback doesn't have any effect, all diagnostics are directly stored in ScalafixResult
+     *
+     * @return ScalafixResult
+     */
+    ScalafixResult runAndReturnResult();
```
bjaglin, line 217: To echo the comment below about `ScalafixResult`, this could be something like
```suggestion
    ScalafixEvalution evaluate();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -205,4 +205,14 @@ ScalafixArguments withToolClasspath(
      */
     ScalafixError[] run();
 
+    /**
+     *
+     * Compared to {@link #run()}, it doesn't write the diff to the requested file,
+     * but instead return the list of patches and diagnostics, which can be written to file using
+     * {@link ScalafixResult#writeResult()}}.
+     * withCallback doesn't have any effect, all diagnostics are directly stored in ScalafixResult
```
bjaglin, line 213: can we fail or feed both rather than ignoring?

bjaglin (commented):
I finally found some time to look at this promising PR! I only focused on the API declaration & tests without looking at the implementation to keep a naive consumer standpoint as much as possible.

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixOutputtImpl.scala
```diff
@@ -0,0 +1,145 @@
...
+    val exitStatus = fixed.toTry
+      .flatMap(MainOps.appyDiff(args, file = originalPath, _))
+      .getOrElse(ExitStatus.UnexpectedError)
+    ScalafixErrorImpl.fromScala(exitStatus).toSeq.toArray
+  }
+
+  override def getOutputFixedWithSelectivePatches(
+      selectedPatches: Array[ScalafixPatch]
+  ): Optional[String] = {
+    val ids = selectedPatches.toList.map(_.getId())
```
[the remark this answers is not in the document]
sjrd: Could we use the identity of the `ScalafixPatch` itself as the ID? This code would like the following:
```scala
val selectedPatchesSet = new java.util.IdentityHashMap[ScalafixPatch, Unit](selectedPatches.length)
for (patch <- selectedPatches)
  selectedPatchesSet.put(patch, ())
val filteredPatches = patches.filter(selectedPatchesSet.containsKey(_))
```
That way, we wouldn't have to expose `getId()` at all in the interface.

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixOutputtImpl.scala
```diff
@@ -0,0 +1,145 @@
+package scalafix.internal.interfaces
```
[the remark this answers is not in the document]
mlachkar: thanks

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -76,7 +85,66 @@ object MainOps {
...
+                      messages
+                    )(args, ctx, index)
+                  } else {
+                    val diff = DiffUtils.unifiedDiff(
+                      file.toString(),
+                      "<expected fix>",
+                      input.text.linesIterator.toList,
+                      fixed.linesIterator.toList,
+                      3
+                    )
```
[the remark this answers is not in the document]
mlachkar: done

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -76,7 +85,66 @@ object MainOps {
...
+                      ExitStatus.Ok,
+                      patches,
+                      messages
+                    )(args, ctx, index)
+                  }
+              }
+              .getOrElse(
+                ScalafixOutputtImpl
+                  .from(file, ExitStatus.UnexpectedError, "failure")(args)
+              )
```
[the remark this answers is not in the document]
mlachkar: Thanks, agree with you

scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
```diff
@@ -37,36 +37,59 @@ object PatchInternals {
...
       doc: v1.SyntacticDocument,
       suppress: Boolean
-  ): (String, List[RuleDiagnostic]) = {
+  ): (
+      String,
+      List[v0.Patch],
+      v0.RuleCtx,
+      Option[v0.SemanticdbIndex],
+      List[RuleDiagnostic]
+  ) = {
```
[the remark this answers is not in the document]
mlachkar: Agree

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: good name

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixResult.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     *
+     * @return Boolean true if there are no error on any file scalafix run on it.
+     */
+    Boolean isSuccessful();
```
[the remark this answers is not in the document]
mlachkar: you're right !

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+import scalafix.testkit.DiffAssertions
+import scalafix.tests.core.Classpaths
+import scalafix.tests.util.ScalaVersions
+import scalafix.v1.SemanticRule
+
+import scala.meta.io.AbsolutePath
+import collection.JavaConverters._
+
+/**
+ * Tests in this suite require scalafix-cli & its dependencies to be cross-published so that Coursier can fetch them.
```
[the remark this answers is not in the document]
mlachkar: good idea. I will instantiate ScalafixArgumentsImpl directly

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixResult.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * G
+     * @return ScalafixOutput[] for each file we store diagnostics and patches
+     * */
+    ScalafixOutput[] getScalafixOutputs();
+
+
+    ScalafixError[] writeResult();
+
+    ScalafixError[] writeResult(Path path);
```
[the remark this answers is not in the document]
mlachkar: ```ScalafixError[] writeResult(Path path);``` this one is not required. 
The difference is that write result, will apply changes to all files in ScalafixResult, and apply patches in scalafixOutput is linked to only one file

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: So if the name is evaluate, it make sense that the Mode or callback are not called ?

mlachkar:
The interfaces look like that 
```
interface ScalafixArguments {
  ScalafixEvaluation evaluate();
}

interface ScalafixEvaluation {
   ScalafixOutput[] getScalafixOutputs(); // for each file
   ScalafixError[] writeResult();
}
interface ScalafixOutput {
  Path path();
  Rules[] rules;
  ScalafixPatch[] patches();
  ScalafixDiagnostic[] getDiagnostics();
  Optional<String> getUnifiedDiff();

// write to file the result of patches
 void applyAllPatches();
 void applySelectivePatches();

// and equivalent function that returns the result of applying patches with writing the diff.
 Optional<String> evaluateAllPatches();
 Optional<String> evaluateSelectivePatches();
 }
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
```diff
@@ -0,0 +1,9 @@
...
+    /**
+     * Can be RemoveGlobalImport, RemoveImportee, AddGlobalImport, AddGlobalSymbol, ReplaceSymbol
```
bjaglin, line 5: it looks like an enum

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @return boolean true if there are no error on any file scalafix run on it.
+     */
+    boolean isSuccessful();
+
+    /**
+     * If not successful, the error is stored in this method.
+     * @returnOptional<ScalafixError>
+     */
+    ScalafixError[] getError();
```
bjaglin, line 16:
```suggestion
    ScalafixError[] getErrors();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,35 @@
...
+     * @return ScalafixOutput[] for each file we store the scalafix evaluation: If the evaluation is successful,
+     * we store the list of patches, diagnostics, the corresponding unified diff, otherwise we store the error resulted from the evaluation
+     * */
+    ScalafixOutput[] getScalafixOutputs();
+
+    /**
+     *
+     * @return if no error, write result to file configured in {@link ScalafixArguments()}
+     */
+    ScalafixError[] writeResult();
```
bjaglin, line 34: align name with ScalafixOutput ?
```suggestion
    ScalafixError[] apply();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+
+    ScalafixPatch[] getPatches();
+
+    /**
+     *
+     * @return the output content after evaluating scalafix if no error. The original file will stay unchanged.
+     */
+    Optional<String> getOutputFixed();
```
bjaglin, line 46: `output` was not really clear to me since it's the name of the class itself, while I think here it means the content of the path after the fix is applied. Also, the fix does not really show that this is a dry-run version of `applyPatches`. What about
```
interface ScalafixTargetEvaluation {
  Path getTarget(); 
  Optional<String> previewPatches();
```

(we should stick to one wording for file/target/path to designate what scalafix runs on by the way - not sure what's the best or what's already used)

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
mlachkar: you are right, but it's not corresponding to all possible cases in Patch sealed trait.
Creating an Enum may add too much constraint for this field, especially that the only use case that I see right now is applying only patches relate to `Import`

My opinion is that String is enough for now, but maybe later we can add an enum.

bjaglin: > the only use case that I see right now is applying only patches relate to Import

Can you elaborate the use-case? maybe we don't need to expose the path class hierarchy as-is, but just what we need - I am concerned we are leaking internals here.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
bjaglin: discussed offline: this PR is driven by a need to try to apply some patches until files reach a certain state (via `previewSelectivePatches`), but we don't know beforehand which patch will be tried, so we need to be able to address them individually as we try and iterate.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -205,4 +205,12 @@ ScalafixArguments withToolClasspath(
      */
     ScalafixError[] run();
 
+    /**
+     * Compared to {@link #run()}, it evaluates scalafix on each file and returns the list of patches and diagnostics,
+     * which can be written to file using {@link ScalafixEvaluation#apply()}}.
+     * withMainCallback it's not called, all diagnostics are directly stored in ScalafixEvaluation
+     *
+     * @return ScalafixEvaluation
+     */
```
bjaglin, line 214:
```suggestion
    /**
     * Similar to {@link #run()}, but without any side effects on the source files. Via the returned {@link ScalafixEvaluation},
     * for each file, diagnostics can be inspected, and patches can be previewed and applied.  
     * <p>
     * Incompatible with {@link #withMainCallback()} and {@link #withMode()}.
     */
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,32 @@
...
+
+public interface ScalafixEvaluation {
+    /**
+     * @return boolean true if scalafix has successfully been evaluated on all files configured.
+     */
+    boolean isSuccessful();
+
+    /**
+     * @return List of scalafixErrors resulted from scalafix evaluation
+     */
```
bjaglin, line 13: I don't think this doc adds anything that the signature doesn't tell?
```suggestion
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,32 @@
...
+    ScalafixError[] getErrors();
+
+    /**
+     * @return A message error is provided if the run is not successful
+     */
+    Optional<String> getMessageError();
```
bjaglin, line 19: how does this relate to the `getErrors` ?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * @return ScalafixOutput[] for each file we store the scalafix evaluation: If the evaluation is successful,
+     * we store the list of patches, diagnostics and the corresponding unified diff,
+     * otherwise we store the errors resulted from the evaluation
+     * */
+    ScalafixOutput[] getScalafixOutputs();
+
+    /**
+     * @return If no error, write result to file configured in {@link ScalafixArguments()}
+     */
```
bjaglin, line 30:
```suggestion
    /**
     * Applies all patches for all files
     */
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -0,0 +1,32 @@
...
+    ScalafixError[] getErrors();
+
+    /**
+     * @return A message error is provided if the run is not successful
+     */
+    Optional<String> getMessageError();
+
+    /**
+     * @return ScalafixOutput[] for each file we store the scalafix evaluation: If the evaluation is successful,
```
bjaglin, line 22: is the type really needed in javadoc `@return` comments?
```suggestion
     * @return for each file we store the scalafix evaluation: If the evaluation is successful,
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @return original path of the file to be fixed
+     */
+    Path getPath();
+
+    /**
+     *
+     * @return corresponding list of rules that has been evaluated
+     */
+    ScalafixRule[] getRules();
```
bjaglin, line 17:
```suggestion
    ScalafixRule[] getEvaluatedRules();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public interface ScalafixOutput {
+
+    /**
+     * @return original path of the file to be fixed
+     */
+    Path getPath();
```
bjaglin, line 11:
```suggestion
    Path getEvaluatedPath();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+    Optional<String> previewPatches();
+
+    /**
+     *
+     * @return apply all patches and write the result to file.
+     */
+    ScalafixError[] applyPatches();
+
+    /**
+     * @param patches the patches should belong to this scalafixOutput's patches. If not, no patch will be applied to the file.
```
bjaglin, line 54: what do you mean by "if not"?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
+package scalafix.interfaces;
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public interface ScalafixOutput {
```
bjaglin, line 6: Same comment as before, I find it weird to have an `Output` class (a very passive name) from where we can trigger side effects. I am not sure my suggestion is better though
```suggestion
public interface ScalafixFileEvaluation {
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    ScalafixError[] applyPatches();
+
+    /**
+     * @param patches the patches should belong to this scalafixOutput's patches. If not, no patch will be applied to the file.
+     *
+     * @return It returns the new content file after applying the desired patches if no error.
+     * The original file will stay unchanged.
+     */
+    Optional<String> previewSelectivePatches(ScalafixPatch[] patches);
```
bjaglin, line 59: can't this be an overload?
```suggestion
    Optional<String> previewPatches(ScalafixPatch[] patches);
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**applySelectivePatches
+     *
+     * @return apply selected patches to file
+     */
+    ScalafixError[] applySelectivePatches(ScalafixPatch[] patches);
```
bjaglin, line 65:
```suggestion
    ScalafixError[] applyPatches(ScalafixPatch[] patches);
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @param patches the patches should belong to this scalafixOutput's patches. If not, no patch will be applied to the file.
+     *
+     * @return It returns the new content file after applying the desired patches if no error.
+     * The original file will stay unchanged.
+     */
+    Optional<String> previewSelectivePatches(ScalafixPatch[] patches);
+
+    /**applySelectivePatches
```
bjaglin, line 61:
```suggestion
    /**
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
mlachkar: good idea. I forgot we can do this

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
mlachkar: If we apply a rule that looks like explicitResultType: When a type is added, sometimes an import is also added. 
We would like to be able to apply all import patches first. 
I understand your concern. We can maybe expose 3 kinds: ReplaceSymbol, AddImport, RemoveImport? 
I can also remove `kind` until it's clearer what we would need, and add this in a second PR later.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
mlachkar: It's a global message error that can be empty. Errors are a concatenation of code errors that result from a merge function.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
mlachkar: no, type is not necessary in `return` java doc

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixOutput.java
mlachkar: At least evaluation is consistent with the naming we decide for ScalafixEvaluation. So I prefer your naming

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
bjaglin: >  We would like to be able to apply all import patches first.

I am not very familiar with the patch internals, but doesn't https://scalacenter.github.io/scalafix/docs/developers/patch.html#atomic allow to group patches that should be applied "together" - maybe we need to expose patches as group of atomic patches?

bjaglin: >  I can also remove kind until it's clearer what we would need, and add this in a second PR later.

That would probably be better as there is already a lot of surface in the new APIs. `ScalafixPatch` would then become an empty trait for now, highlighting the fact that it's used for its identify (to iterate and converge on a proper state), rather than for its attributes.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
bjaglin: I think we should fail if `evaluate` is called on a `ScalafixArguments` that was explicitly set a mode or a callback, since they have no effect.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -205,4 +205,11 @@ ScalafixArguments withToolClasspath(
      */
     ScalafixError[] run();
 
+    /**
+     * Similar to {@link #run()}, but without any side effects on the source files. Via the returned {@link ScalafixEvaluation},
+     * for each file, diagnostics can be inspected, and patches can be previewed and applied.
+     * <p>
+     * Incompatible with {@link #withMainCallback} and {@link #withMode}.
```
bjaglin, line 212: can this be tested? I don't see a check in the implementation

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+public interface ScalafixFileEvaluation {
+
+    /**
+     * @return original path of the file scalafix has evaluated on
+     */
+    Path getEvaluatedPath();
```
bjaglin, line 11: to highlight that it's a file (and not a directory that can have been passed via `ScalafixArguments#withPaths`)
```suggestion
    Path getEvaluatedFile();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -0,0 +1,65 @@
...
+     * The original file will stay unchanged.
+     */
+    Optional<String> previewPatches(ScalafixPatch[] patches);
+
+    /**
+     *
+     * @param patches the patches should belong to this scalafixFileEvaluation's patches.
+     *                If you apply patches that result from another ScalafixFileEvaluation,
+     *                no patch will be applied to the file.
+     * @return apply selected patches to file
```
bjaglin, line 62: that's not what's returned
```suggestion
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -0,0 +1,65 @@
...
+    ScalafixError[] getErrors();
+
+    ScalafixDiagnostic[] getDiagnostics();
+
+    /**
+     *
+     * @return if the fixed file is different from the input file and the evaluation
+     * is successful this method return non empty unified diff
+     */
+    Optional<String> getUnifiedDiff();
```
bjaglin, line 34:
```suggestion
    /**
     *
     * @return if the fixed file is different from the input file and the evaluation
     * is successful this method return non empty unified diff. The original file will stay unchanged.
     */
    Optional<String> previewPatchesAsUnifiedDiff();
```

bjaglin (commented):
Just a few last comments on the API, but overall LGTM. I haven't looked at the implementation, but this can/will evolve as you use it I guess, so please don't wait for me.

I am still a bit concerned about the opaque `ScalafixPatch` object that is exposed to be able to apply/preview a random subset of patches, but I don't have a better idea to tackle this use-case.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: Evaluate should not fail from my point of view. It's a new way to interact with scalafix, that requires fewer arguments.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: Yes, I will add test for this.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
mlachkar: thanks.

scalafix-tests/unit/src/test/scala/scalafix/tests/rule/MavenFuzzSuite.scala
mlachkar: there are 3 modes. this one only checks and doesn't write to file.

mlachkar:
Thank you for all reviews. I will merge it now.

PR #1224, opened 2020-08-10
Fix ScalafixImplSuite test and add a test for configuration path

mlachkar:
Right now, even if you specify a config file = None, scalafix still reads the default config path. 
The goal is to document this behavior through a test

### DIFF
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
@@ -9,9 +9,13 @@ import java.nio.file.Files
 import java.nio.file.Paths
 import java.util.Collections
 import java.util.Optional
+
 import coursier._
 import org.scalatest.funsuite.AnyFunSuite
+
 import scala.collection.JavaConverters._
+import buildinfo.RulesBuildInfo
+
 import scala.meta.io.AbsolutePath
 import scala.meta.io.Classpath
 import scalafix.Versions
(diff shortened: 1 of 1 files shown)


mlachkar:
This test was running only for 2.12. It tests the interface implementations.
Now that we cross-compile example-rule, it's possible to run this test for the 3 scala versions.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -97,6 +99,47 @@ class ScalafixImplSuite extends AnyFunSuite with DiffAssertions {
...
+      """.stripMargin,
+        charset
+      )
+    val args = api
+      .newArguments()
+      .withConfig(Optional.empty())
+      .withWorkingDirectory(cwd.toNIO)
+    args.validate()
+    assert(
+      args.rulesThatWillRun().asScala.toList.map(_.toString) == List(
```
bjaglin, line 124: do you really need to fallback on a `toString` for assertion? (same comment for the second assertion)

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -97,6 +99,47 @@ class ScalafixImplSuite extends AnyFunSuite with DiffAssertions {
     assert(e.isPresent)
     assert(e.get().getMessage.contains("-Ywarn-unused"))
   }
+  test("configure scalafix path") {
```
bjaglin, line 102: i don't understand this description - isn't the test more about
```suggestion
  test("rulesThatWillRun") {
```

bjaglin (approved):
:+1: I have added a few minor suggestions

PR #1225, opened 2020-08-12
Add scala version requierment to ExplicitResultTypes doc page.

mlachkar:


bjaglin (approved):
:+1: maybe the "programmatic" description pulled in https://scalacenter.github.io/scalafix/docs/rules/overview.html could also automatically be added on rules doc pages?

PR #1251, opened 2020-10-05
Add a message error to scalafixFileEvaluation

mlachkar:
* Modify isSuccessful of ScalafixEvaluation: If a lintError is found, isSuccessful is false.
A LintError is produced when scalafix has run successfully and there are diagnostics with severity Error.
The issue is the default severity is Error, and rules don't usually modify that.
I think it's misleading to report that scalafix has failed, when there are diagnostics.

* Fix the reporting of StaleSemanticDB error. We were reporting unexpected Error instead of StaleSemanticdbError


scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -104,6 +104,7 @@ object MainOps {
         val files = getFilesFrom(args)
         val fileEvaluations = {
           files.map { file =>
+            println(s"file = ${file}")
```
bjaglin, line 107:
```suggestion
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -22,6 +22,8 @@
      */
     boolean isSuccessful();
 
+    Optional<String> getMessageError();
```
bjaglin, line 25: could you add some javadoc? does "error" include lint errors?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,7 +4,7 @@
 
 public interface ScalafixEvaluation {
     /**
-     * @return boolean true if scalafix has successfully been evaluated on all files configured and no ScalafixError has been found
+     * @return boolean true if scalafix has successfully been evaluated on all files configured with no error except LintErrors.
```
bjaglin, line 7: should the `ScalafixFileEvaluation#isSuccessful` implementation & doc be updated as well?

bjaglin (commented):
Makes sense! A few comments regarding the API changes.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
mlachkar: ok. It includes lintErrors.

tgodzik (commented):
Looks good! Just one question: is `scalafix-interfaces/.DS_Store` added on purpose? Not sure what that file does.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
bjaglin: sorry if I wasn't clear but I was talking about `getMessageError` - I think the `getErrors` are clear enough given their signature so the javadoc there is redundant.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -150,6 +150,38 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
+    Files.write(main, staleCode.getBytes(StandardCharsets.UTF_8))
+    val code2 = FileIO.slurp(AbsolutePath(main), StandardCharsets.UTF_8)
+
+    val evaluation2 = args.evaluate()
+    assert(evaluation2.getErrors.head.toString == "StaleSemanticdbError")
+    assert(evaluation2.getMessageError == Optional.empty())
+    assert(!evaluation2.isSuccessful)
+    val fileEval = evaluation2.getFileEvaluations.head
+    assert(fileEval.getErrors.head.toString == "StaleSemanticdbError")
+    assert(fileEval.getMessageError.get.startsWith("Stale SemanticDB"))
```
bjaglin, line 183: :+1: can we cover the other `isSuccessful` while we are at it?
```suggestion
    val evaluation2 = args.evaluate()
    assert(evaluation2.getErrors.head.toString == "StaleSemanticdbError")
    assert(evaluation2.getMessageError == Optional.empty())
    assert(!evaluation2.isSuccessful)
    val fileEval = evaluation2.getFileEvaluations.head
    assert(fileEval.getErrors.head.toString == "StaleSemanticdbError")
    assert(fileEval.getMessageError.get.startsWith("Stale SemanticDB"))
    assert(!fileEval.isSuccessful)
```

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixEvaluationImpl.scala
```diff
@@ -16,11 +16,11 @@ final case class ScalafixEvaluationImpl(
     fileEvaluations: Seq[ScalafixFileEvaluationImpl]
 ) extends ScalafixEvaluation {
 
-  override def isSuccessful: Boolean = error.isOk
-
   override def getErrors: Array[ScalafixError] = {
     ScalafixErrorImpl.fromScala(error)
   }
+  override def isSuccessful: Boolean =
+    getErrors.toList.filter(_ != ScalafixError.LinterError).isEmpty
```
bjaglin, line 23: i think the behavior change of both `isSuccessful` is lacking a test case to prevent regressions

mlachkar: ok, good idea

mlachkar:
LinterErrors were only added in the end, when scalafixFileEvaluations did contain nonEmpty diagnostics (through adjustExitCode). 
ScalafixEvaluation is a new api, that is not meant to be used as a cli, and only thought code calls. LinterErrors are already represented as a ScalafixDiagnostics. 
The last commit removes the adjustExitCode, so we don't need anymore to filter LinterErrors. 

I did a choice to concatenate all scalafixFileEval error messages if scalafix has run successfully partly. Just a suggestion

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,19 +4,25 @@
...
      */
     boolean isSuccessful();
 
     ScalafixError[] getErrors();
 
+    /**
+     *
+     * @return a more detailed error message that can be
+     *  - an explanation of why scalafix failed running at all
+     *  - a concatenation of all {@Link ScalafixFileEvaluation} message Errors
```
bjaglin, line 17: Cosmetic, but I think you need tags for this list of bullet points to appear properly in Javadoc like https://javadoc.io/doc/ch.epfl.scala/scalafix-interfaces/0.9.21/scalafix/interfaces/ScalafixFileEvaluation.html. Have a look at the rest of the Javadoc, it's very clean, so I think we should try to keep the same standard.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -18,10 +18,21 @@
...
      *
-     * @return boolean true if there is no error when scalafix has been evaluated on the file
+     * @return boolean true if there is no error except LintError when scalafix has been evaluated on the file
      */
     boolean isSuccessful();
 
+    /**
+     *
+     * @return a more detailed error in case of scalafix ParseError, StaleSemanticdbError,
+     * MissingSemanticdbError or an UnexpectedError. LinterErrors are found in {@link ScalafixDiagnostic}
```
bjaglin, line 28: I am not in front of an IDE, but looking at the code, it looks like the error message is only defined for `UnexpectedError`.

https://github.com/scalacenter/scalafix/blob/e31e5c946a14bb3839ef5c087b660953a3485db8/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala#L122-L125

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,19 +4,25 @@
...
-     * @return boolean true if scalafix has successfully been evaluated on all files configured and no ScalafixError has been found
+     * @return boolean true if scalafix has successfully been evaluated on all files configured with no error.
      */
     boolean isSuccessful();
 
     ScalafixError[] getErrors();
 
+    /**
+     *
+     * @return a more detailed error message that can be
```
bjaglin, line 15: more than what?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,19 +4,25 @@
 
 public interface ScalafixEvaluation {
     /**
-     * @return boolean true if scalafix has successfully been evaluated on all files configured and no ScalafixError has been found
+     * @return boolean true if scalafix has successfully been evaluated on all files configured with no error.
      */
     boolean isSuccessful();
 
     ScalafixError[] getErrors();
```
bjaglin, line 11: missing javadoc since not all values of the enum might be returned?

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixFileEvaluationImpl.scala
```diff
@@ -63,7 +63,9 @@ final case class ScalafixFileEvaluationImpl(
   override def getErrors(): Array[ScalafixError] =
     ScalafixErrorImpl.fromScala(error)
 
-  override def isSuccessful: Boolean = error.isOk
```
bjaglin, line 66: is error (the exit status) still used then?

bjaglin (commented):
I am afraid I am still having a hard time figuring out the behavior by looking at the docs. I think the problem comes from the re-use of the enum `ScalafixError`.

This API is brand new, maybe it's still time to deprecate `getErrorMessage` (which concatenates errors in a way that makes it hard to consume for a non-human) & `getErrors` which have a similar signature in `ScalafixEvaluation` and `ScalafixFileEvaluation`, yet different possible values.

What about replacing them with `ScalafixEvaluationError[] getEvaluationErrors()`, `ScalafixEvaluationError` being an interface with a `getType` returning an enum with only "operational errors", and `getMessage`? Same thing for `ScalafixFileEvaluationError`, with its own embedded enum. It sounds like a lot of changes, but it would avoid a lot of ambiguous documentation and testing.

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
mlachkar: the error , wchich is only the type of the Enum

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
mlachkar: true, but this has been fixed in this PR. 
```
case Failure(exception) =>
   .from(file,  ExitStatus.from(exception), exception.getMessage)
...

  def from(throwable: Throwable): (ExitStatus) =
    throwable match {
      case _: ParseException =>
        ExitStatus.ParseError
      case _: SemanticDocument.Error.MissingSemanticdb =>
        ExitStatus.MissingSemanticdbError
      case _: StaleSemanticDB =>
        ExitStatus.StaleSemanticdbError
      case NonFatal(_) =>
        ExitStatus.UnexpectedError
    }
````

**Comment by mlachkar** [comment]
I agree. It will be easier to add new `ScalafixErrors`.  I will rework this. Thanks

File: scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java [hunk 11]
```diff
@@ -18,11 +19,22 @@
...
+     * @return boolean true if there is no error except LintError when scalafix has been evaluated on the file
      */
     boolean isSuccessful();
 
-    ScalafixError[] getErrors();
+    /**
+     *
+     * @return a more detailed error in case of scalafix.
+     */
+    Optional<String> getErrorMessage();
```
bjaglin, line 30: in case of what? I think this adds confusion rather than thelp the user
```suggestion
    Optional<String> getErrorMessage();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -18,11 +19,22 @@
 
     /**
      *
-     * @return boolean true if there is no error when scalafix has been evaluated on the file
+     * @return boolean true if there is no error except LintError when scalafix has been evaluated on the file
```
bjaglin, line 22: Now that `LintError` is no longer something that `getError` can return, I think it's better not to mention it
```suggestion
     * @return boolean true if scalafix was successfully evaluated on the file
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,19 +4,31 @@
 
 public interface ScalafixEvaluation {
     /**
-     * @return boolean true if scalafix has successfully been evaluated on all files configured and no ScalafixError has been found
+     * @return boolean true if scalafix has successfully been evaluated on all files configured with no error.
```
bjaglin, line 7: From what I see in the implementation, this won't be set to error if we have a stale semanticdb for a file, is that correct?
```suggestion
     * @return boolean true if scalafix ran on at least one file; there might be errors in file evaluation though
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEvaluation.java
```diff
@@ -4,19 +4,31 @@
...
+    @Deprecated ScalafixError[] getErrors();
+    /**
+     * @deprecated  replaced by {@link #getErrorMessage()}
+     */
+    @Deprecated Optional<String> getMessageError();
+
+    /**
+     * @return a more detailed error message that can be
+     */
+    Optional<String> getErrorMessage();
```
bjaglin, line 25: I think the signature talks for itself now
```suggestion
    Optional<String> getErrorMessage();
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixFileEvaluation.java
```diff
@@ -1,5 +1,6 @@
 package scalafix.interfaces;
 
+import javax.swing.text.html.Option;
```
bjaglin, line 3:
```suggestion
```

scalafix-interfaces/src/main/java/scalafix/interfaces/EvaluationError.java
```diff
@@ -0,0 +1,16 @@
+package scalafix.interfaces;
+
+public enum EvaluationError {
```
bjaglin, line 3: It's a bit verbose, but shouldn't these new enums follow the current pattern and be prefixed by `Scalafix`  for consistency?

bjaglin (commented):
Much clearer with new enums! A last round of comments about the behavior and a clearer javadoc, and I guess we are good to go!

bjaglin:
(rebase needed)

.scalafix.conf
```diff
@@ -9,7 +9,7 @@ ExplicitResultTypes {
 OrganizeImports {
  groupedImports = Explode
  expandRelative = true
- removeUnused = true # done already by RemoveUnused rule
```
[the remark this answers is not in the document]
mlachkar: true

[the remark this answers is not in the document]
mlachkar: but we don't have RemoveUnused rule anymore.

scalafix-interfaces/src/main/java/scalafix/interfaces/EvaluationError.java
mlachkar: yes. ok

mlachkar:
Thanks for your review. I addressed all your comments.

bjaglin (approved):
:+1:

PR #1252, opened 2020-10-05
fix pretty print for backquoted symbols

mlachkar:
Example:
```
  implicit val `b-b`: Int = 12
  val crazy1 = implicitly[Int]
```
the synthetics of this piece of code is `List(*(b-b))`, and I would expect ```List(*(`b-b`))```

Maybe there is another way to get correctly the toString, and this is just a proposition to get more feedback.
Thanks

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
+++ b/scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
@@ -16,7 +16,7 @@ final class Symbol private (val value: String) {
   def isGlobal: Boolean = value.isGlobal
   def isLocal: Boolean = value.isLocal
   def owner: Symbol = Symbol(value.owner)
-  def displayName: String = value.desc.name.value
+  def displayName: String = encode(value.desc.name.value)
   def info(implicit doc: Symtab): Option[SymbolInformation] =
     doc.info(this)
   def normalized: Symbol = SymbolOps.normalize(this)
@@ -38,6 +38,18 @@ final class Symbol private (val value: String) {
       case _ => false
     })
   override def hashCode(): Int = value.##
+
+  private def encode(value: String): String = {
+    if (value == "") {
+      "``"
+    } else {
+      val (start, parts) = (value.head, value.tail)
+      val isStartOk = Character.isJavaIdentifierStart(start)
+      val isPartsOk = parts.forall(Character.isJavaIdentifierPart)
+      if (isStartOk && isPartsOk) value
+      else "`" + value + "`"
+    }
+  }
 }
 
 object Symbol {
--- a/scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
+++ b/scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
@@ -16,7 +16,7 @@ final class Symbol private (val value: String) {
   def isGlobal: Boolean = value.isGlobal
   def isLocal: Boolean = value.isLocal
   def owner: Symbol = Symbol(value.owner)
-  def displayName: String = encode(value.desc.name.value)
+  def displayName: String = value.desc.name.value
   def info(implicit doc: Symtab): Option[SymbolInformation] =
     doc.info(this)
   def normalized: Symbol = SymbolOps.normalize(this)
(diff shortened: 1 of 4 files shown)


scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
```diff
@@ -38,6 +38,18 @@ final class Symbol private (val value: String) {
       case _ => false
     })
   override def hashCode(): Int = value.##
+
+  private def encode(value: String): String = {
+    if (value == "") {
+      "``"
```
bjaglin, line 44: is that even a valid identifier?

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
```diff
@@ -38,6 +38,18 @@ final class Symbol private (val value: String) {
...
+    } else {
+      val (start, parts) = (value.head, value.tail)
+      val isStartOk = Character.isJavaIdentifierStart(start)
+      val isPartsOk = parts.forall(Character.isJavaIdentifierPart)
+      if (isStartOk && isPartsOk) value
+      else "`" + value + "`"
```
bjaglin, line 50:
```suggestion
      else s"`$value`"
```

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
```diff
@@ -16,7 +16,7 @@ final class Symbol private (val value: String) {
   def isGlobal: Boolean = value.isGlobal
   def isLocal: Boolean = value.isLocal
   def owner: Symbol = Symbol(value.owner)
-  def displayName: String = value.desc.name.value
+  def displayName: String = encode(value.desc.name.value)
```
bjaglin, line 19: maybe I am off but could https://gitter.im/scalacenter/scalafix?at=5f72c62cefefe620f2569259 be used?

bjaglin (commented):
Nice!

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
olafurpg: If you're inserting `Symbol.displayName()` into source code then it should be wrapped inside `scala.meta.Term.Name(...)`. This won't handle the case however when you need a leading space before the `:` colon in `def * : Int = 42`, the `ExplicitResultTypes` rule should have special logic for that to insert a the space if `!Character.isJavaIdentifierPart(lastCharacter))`.

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
```diff
@@ -38,6 +38,18 @@ final class Symbol private (val value: String) {
       case _ => false
     })
   override def hashCode(): Int = value.##
+
+  private def encode(value: String): String = {
```
olafurpg, line 42: This method doesn't handle some cases that `scala.meta.Term.Name` handles, for example keywords
```
@ scala.meta.Term.Name("for").toString
res4: String = "`for`"
```

The motivation to call it "displayName" instead of "name" was to make it emphasize that this name is not a valid `scala.meta.Name`, although I admit this fact is not obvious when I think of it now.

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
mlachkar: The motivation for modifying displayName is that is used in SemanticTree.toString. I am using it to insert valid synthetics into code source. It works except for Symbol names that are backquoted. 
Do you recommend to modify instead directly pprint for symbol to this :
```
  def pretty(sym: Symbol): Doc =
    Doc.text(scala.meta.Name(sym.displayName).toString())
``` 
Instead of  
```
  def pretty(sym: Symbol): Doc =
    Doc.text(ym.displayName)
```

Thanks

scalafix-core/src/main/scala/scalafix/v1/Symbol.scala
mlachkar: Trying a new approach.

PR #1253, opened 2020-10-07
Activate scalafix on unit project

mlachkar:


mlachkar:
Just adding scalafix Plugin to another project.

PR #1254, opened 2020-10-08
Handle secondary constructor in Procedure syntax

### LINKED ISSUE #1235: ProcedureSyntax for secondary constructor
The `ProcedureSyntax` rule does not handle the case of secondary constructors.

I would expect the following piece of code
```scala
class Bar(a: String) {
  def this() {
    this("bar")
  }
}
```

to be rewritten into
```scala
class Bar(a: String) {
  def this() = {
    this("bar")
  }
}
```

The following custom rule does the job for secondary constructors only: [ohze/scala-rewrites/`fix.scala213.ConstructorProcedureSyntax`](https://github.com/ohze/scala-rewrites/tree/dotty/#fixscala213constructorproceduresyntax).

mlachkar:
* code inspired from [ConstructorProcedureSyntax](https://github.com/ohze/scala-rewrites/blob/dotty/rewrites/src/main/scala/fix/scala213/ConstructorProcedureSyntax.scala)
* We don't remove braces in case of 1 statement

fixes #1235

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
@@ -1,6 +1,9 @@
 package scalafix.internal.rule
 
 import scala.meta._
+import scala.meta.tokens.Token
+import scala.meta.tokens.Token.Equals
+import scala.meta.tokens.Token.RightParen
 
 import scalafix.util.Trivia
 import scalafix.v1._
@@ -9,6 +12,7 @@ class ProcedureSyntax extends SyntacticRule("ProcedureSyntax") {
 
   override def description: String =
     "Replaces deprecated procedure syntax with explicit ': Unit ='"
+
   override def isRewrite: Boolean = true
 
   override def fix(implicit doc: SyntacticDocument): Patch = {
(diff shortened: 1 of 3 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -23,6 +27,21 @@ class ProcedureSyntax extends SyntacticRule("ProcedureSyntax") {
           toAdd <- doc.tokenList.leading(bodyStart).find(!_.is[Trivia])
         } yield Patch.addRight(toAdd, s": Unit =").atomic
         fixed.getOrElse(Patch.empty)
+
+      /** @see [[https://github.com/ohze/scala-rewrites/blob/dotty/rewrites/src/main/scala/fix/scala213/ConstructorProcedureSyntax.scala ConstructorProcedureSyntax.scala]] */
```
bjaglin, line 31: @giabao it seems that https://github.com/ohze/scala-rewrites is not licensed, can you advise since this PR is basically inlining code from there (with the test) ?

mlachkar:
I am not completely satisfied with closing this PR. I think this addition is really interesting for this scalafix Rule. I can maybe rewrite it without looking to @giabao code and avoid inling his code?

bjaglin:
> I think this addition is really interesting for this scalafix Rule

I completely agree. Maybe my concern about licensing was overthinking. It's probably something you should agree on within the Scala Center, so I leave it up to you!

PR #1255, opened 2020-10-08
Fix warn messages in doc

### LINKED ISSUE #1190: Stop using deprecated APIs in documentation
See warnings in https://travis-ci.org/github/scalacenter/scalafix/builds/703864130. It looks like most of them come from https://github.com/scalacenter/scalafix/pull/981/files#diff-fb3f807a020266da1788e6b103fbc190R15 (which seems to have reverse the description & since parameters).

Maybe we could then enable `-Xfatal-warnings` to keep the doc up to date?

mlachkar:
* use the new api for SemanticTree
* add @unchecked for non exhaustive pattern matching (in doc)

 fixes #1190

### DIFF
--- a/docs/developers/semantic-type.md
+++ b/docs/developers/semantic-type.md
@@ -49,7 +49,7 @@ Use `MethodSignature.returnType` to get the types of vals.
 
 ```scala mdoc
 def getType(symbol: Symbol): SemanticType =
-  (symbol.info.get.signature: @unchecked) match {
+  symbol.info.get.signature match {
     case MethodSignature(_, _, returnType) =>
       returnType
   }
@@ -167,7 +167,7 @@ def simpleDealias(tpe: SemanticType): SemanticType = {
       case _ =>
         symbol
     }
-  (tpe: @unchecked) match {
+  tpe match {
     case TypeRef(prefix, symbol, typeArguments) =>
       TypeRef(prefix, dealiasSymbol(symbol), typeArguments.map(simpleDealias))
   }
--- a/docs/developers/semantic-type.md
+++ b/docs/developers/semantic-type.md
@@ -49,7 +49,7 @@ Use `MethodSignature.returnType` to get the types of vals.
 
 ```scala mdoc
 def getType(symbol: Symbol): SemanticType =
-  symbol.info.get.signature match {
+  (symbol.info.get.signature: @unchecked) match {
     case MethodSignature(_, _, returnType) =>
       returnType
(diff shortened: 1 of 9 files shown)


bjaglin (commented):
Nice, fixes https://github.com/scalacenter/scalafix/issues/1190 ?

mlachkar:
yes, it will fix  #1190

docs/developers/semantic-type.md
```diff
@@ -49,7 +49,7 @@ Use `MethodSignature.returnType` to get the types of vals.
 
 ```scala mdoc
 def getType(symbol: Symbol): SemanticType =
-  symbol.info.get.signature match {
+  (symbol.info.get.signature: @unchecked) match {
```
bjaglin, line 52: maybe it's clearer with a `case _` ?

bjaglin (commented):
:+1: what about using `-Xfatal-warnings` so that docs remain up to date as deprecations are added?

mlachkar (commented):
I am not a big fun of -Xfatal-warnings on documentation. I hesitated between adding a `@unchecked` or finding the correct scalacOptions to remove warning on exhaustive pattern match. I found on google the `@unchecked` first

docs/developers/semantic-type.md
bjaglin: That would make the doc harder to read indeed... Does removing the `--unchecked` from `scalacOption` for the doc project silence these exceptions? If not, there is https://github.com/scala/scala/pull/8373 that could be used as well to avoid repeating this everywhere (which should be usable since the doc is built using 2.13?).

bjaglin:
> I am not a big fun of -Xfatal-warnings on documentation

why?

mlachkar:
> > I am not a big fun of -Xfatal-warnings on documentation
> 
> why?
I mean in documentation having non exhaustive pattern matching is ok, it makes the doc easier to read. I am not against -Xfatal-warnings in general, just for this case. 
I agree it's nice to have for depreciate api. 
I am also worried about the confusion `@unchecked` brings.

docs/developers/semantic-type.md
mlachkar: removing -unchecked doesn't silence these exceptions

mlachkar:
I fixed last warning and I added xfatal-wanings , and also used -Wconf to silence pattern match errors.

bjaglin (approved):

Issue #1108, opened 2020-04-27
Towards Scalafix v1.0


olafurpg:
Scalafix v0.9.x has been quite stable for a long time now so I think it would be nice to start preparing a v1.0.0 release. I would like to keep the current API mostly unchanged in v1.x so the upgrade should be smooth for most users. This issue is just to track what changes I think are worth getting done before we release v1.x

* cross-build Scalafix for 2.13
* change sbt-scalafix to run `scalafix_2.13` by default instead of `scalafix_2.12`
* We may want an optional mode in sbt-scalafix to use `scalafix_2.12` on 2.12 sources and `scalafix_2.11` on 2.11 sources, which would unblock `ExplicitResultType` usage on more Scala versions. However, this would mean that custom Scalafix rules would also need to be cross-published.
* separate versioning of sbt-scalafix and scalafix-core similar to sbt-scalafmt and scalafmt-core. Users should declare the Scalafix version in `.scalafix.conf` so that clients like Metals or maybe IntelliJ can load the correct version of Scalafix without having to inspect `project/plugins.sbt`
* change `ScalafixArguments` from an interface to abstract class
* remove the long-deprecated `scalafix.v0` package

taisukeoe:
@olafurpg 
I'm excited to see v1.0.0 is approaching!

> We may want an optional mode in sbt-scalafix to use scalafix_2.12 on 2.12 sources and scalafix_2.11 on 2.11 sources

Just for my curiosity, but hope it's a relevant question for everyone:

Is there any reason to give up `Scalafix provide an Java-API on top of the presentation compiler` option, described in #998 issue?
(I'm looking over #998 and trying to find time to solve it.)

olafurpg:
@taisukeoe you may want to sync with @mlachkar on #998 

I believe a good starting point regardless of approach would be to add 2.13 support for scalafix-core.

I'm still open to go the Java-API approach. My biggest concern with that direction is that it will be difficult to come up with generic methods since the compiler provides a lot of functionality. If we decide to only implement a small API for ExplicitResultTypes and not expand the scope of that Java API then maybe it will be OK

bjaglin:
As a rule author, I can assume that rules built against `scalafix-core` 1.x (and probably earlier, since I am not sure there was ever any breaking change there?) will be forward-compatible with any `scalafix-interfaces` 1.x.

[...]

PR #1259, opened 2020-10-19
check the whole caret position in the testkit

gontard:
This pull request ensures that the testkit checks the whole caret (`^^^^`) position range. Before only the range start was tested.

### DIFF
--- a/scalafix-testkit/src/main/scala/scalafix/internal/testkit/AssertDelta.scala
+++ b/scalafix-testkit/src/main/scala/scalafix/internal/testkit/AssertDelta.scala
@@ -25,9 +25,7 @@ case class AssertDelta(
     sameKey(assert.key) &&
       (assert.caretPosition match {
         case Some(carPos) =>
-          (carPos.start == lintDiagnostic.position.start) &&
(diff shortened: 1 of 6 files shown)


bjaglin (commented):
Thanks for the PR! That's a breaking change for testkit users as the second commit shows, but a welcome one in my opinion, as I always expected it worked like this. WDYT @olafurpg @mlachkar ?

Can you update the "doc" as well https://github.com/scalacenter/scalafix/blob/e31e5c946a14bb3839ef5c087b660953a3485db8/scalafix-testkit/src/main/scala/scalafix/internal/testkit/CommentAssertion.scala#L26?

olafurpg (approved):
LGTM 👍 Would be good to mention this in the v1.0 release notes as a breaking change.

bjaglin (approved):

bjaglin:
@olafurpg I have added the 1.0.0 milestone on this PR and added [a bullet point in the outstanding tasks for the 1.0.0 release](https://github.com/scalacenter/scalafix/issues/1108). Can you think of any other breaking change we should document so we can tag it as well?

PR #1260, opened 2020-10-22
add a new method to Patch Api

mlachkar:
- Patch.around could be useful when show implicit conversions

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
@@ -165,7 +165,7 @@ object PatchInternals {
     builder.result()
   }
 
-  private def getPatchUnits(patch: Patch): Seq[Patch] = {
+  private[scalafix] def getPatchUnits(patch: Patch): Seq[Patch] = {
     val builder = Seq.newBuilder[Patch]
     foreachPatchUnit(patch) {
       case _: LintPatch => ()
@@ -185,7 +185,7 @@ object PatchInternals {
     patch.isEmpty || hasLintMessage && onlyLint
   }
 
-  def foreach(patch: Patch)(f: Patch => Unit): Unit = {
+  private def foreach(patch: Patch)(f: Patch => Unit): Unit = {
     def loop(patch: Patch): Unit = patch match {
       case Concat(a, b) =>
         loop(a)
@@ -201,7 +201,9 @@ object PatchInternals {
   }
 
   // don't decompose Atomic Patch
-  def foreachPatchUnit(patch: Patch)(f: Patch => Unit): Unit = {
+  private[scalafix] def foreachPatchUnit(
+      patch: Patch
+  )(f: Patch => Unit): Unit = {
     def loop(patch: Patch): Unit = patch match {
       case Concat(a, b) =>
         loop(a)
--- a/scalafix-core/src/main/scala/scalafix/patch/Patch.scala
+++ b/scalafix-core/src/main/scala/scalafix/patch/Patch.scala
@@ -110,6 +110,14 @@ object Patch {
   def addLeft(tree: Tree, toAdd: String): Patch =
     tree.tokens.headOption.fold(Patch.empty)(addLeft(_, toAdd))
 
+  def addAround(tok: Token, left: String, right: String): Patch =
+    Add(tok, left, right)
+
+  def addAround(tree: Tree, left: String, right: String): Patch = {
+    (tree.tokens.headOption.fold(Patch.empty)(addLeft(_, left)) +
+      tree.tokens.lastOption.fold(Patch.empty)(addRight(_, right))).atomic
+  }
+
   /**
    * Remove named imports for this symbol.
    *
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
@@ -8,8 +8,12 @@ import scala.meta._
 import scala.meta.tokens.Token.Ident
 
 import org.scalatest.funsuite.AnyFunSuiteLike
+import scalafix.internal.patch.PatchInternals
 import scalafix.internal.tests.utils.SkipWindows
 import scalafix.patch.Patch
+import scalafix.patch.Patch.internal.Add
+import scalafix.patch.Patch.internal.AtomicPatch
+import scalafix.patch.Patch.internal.Concat
 import scalafix.testkit.AbstractSyntacticRuleSuite
 import scalafix.v0.Rule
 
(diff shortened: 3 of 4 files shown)


scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
```diff
@@ -154,7 +154,7 @@ object PatchInternals {
       .mkString
   }
 
-  private def underlying(patch: Patch): Seq[Patch] = {
+  def underlying(patch: Patch): Seq[Patch] = {
```
adpi2, line 157: Why removing this `private` modifier?

adpi2 (approved):
I guess you want to use this patch for handling things like:
```diff
def pretty[A](a: A)(implicit ev: A => Pretty): String =
-  a.print
+   ev(a).print
```

scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
mlachkar: to use it in test and show that it will create two patches

adpi2: Then I suggest you use a package private modifier:

```suggestion
  private[scalafix] def underlying(patch: Patch): Seq[Patch] = {
```

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -110,6 +110,18 @@ object Patch {
   def addLeft(tree: Tree, toAdd: String): Patch =
     tree.tokens.headOption.fold(Patch.empty)(addLeft(_, toAdd))
 
+  def addAround(tok: Token, left: String, right: String): Patch =
+    Add(tok, left, right)
+
+  // if the tree contains more than one token, this patch is not atomic
```
bjaglin, line 116: That looks like something we want in the Scaladoc
```suggestion
  /** If the tree contains more than one token, this patch is not atomic. */
```

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -135,4 +136,31 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
+      |
+      |object A
+      |*/""".stripMargin
+  )
+
+  test("Patch.around") {
+    val multiplePatches = Patch.addAround(tree, "/*", "*/")
+    assert(!multiplePatches.isEmpty)
+    assert(PatchInternals.underlying(multiplePatches).length == 2)
+    // atomic doesn't do
```
bjaglin, line 161:
```suggestion
    // patch can't be made atomic
```

mlachkar:
waiting for this PR: https://github.com/scalacenter/scalafix/pull/1261 => the goal is to make Patch.around atomic

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -135,4 +140,37 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
+      |
+      |object A
+      |""".stripMargin,
+    """/*import a.b
+      |
+      |object A
+      |*/""".stripMargin
+  )
+
+  test("Patch.around") {
```
bjaglin, line 161:
```suggestion
  test("Patch.addAround") {
```

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -135,4 +140,37 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
+    val atomicPatch = PatchInternals.getPatchUnits(patch)
+    assert(atomicPatch.length == 1)
+    atomicPatch.head match {
+      case AtomicPatch(Concat(patch1, patch2)) =>
+        assert(patch1.isInstanceOf[Add])
+        assert(patch2.isInstanceOf[Add])
+      case _ =>
+        throw new Exception(
+          s"Test `Patch.around` failed. $atomicPatch is not AtomicPatch(Concat(patch1, patch2))"
+        )
```
bjaglin, line 173:
```suggestion
        fail(s"$atomicPatch is not of expected type AtomicPatch(Concat(_, _))")
```

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -6,10 +6,15 @@ import java.nio.file.Paths
 
 import scala.meta._
 import scala.meta.tokens.Token.Ident
+import scala.meta.tokens.Token.classifiable
```
bjaglin, line 9: :thinking: is that used? I guess so because we have scalafix guarding against unused imports?

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
```diff
@@ -110,6 +110,14 @@ object Patch {
   def addLeft(tree: Tree, toAdd: String): Patch =
     tree.tokens.headOption.fold(Patch.empty)(addLeft(_, toAdd))
 
+  def addAround(tok: Token, left: String, right: String): Patch =
+    Add(tok, left, right)
+
+  def addAround(tree: Tree, left: String, right: String): Patch = {
+    (tree.tokens.headOption.fold(Patch.empty)(addLeft(_, left)) +
+      tree.tokens.lastOption.fold(Patch.empty)(addRight(_, right))).atomic
```
bjaglin, line 118: maybe it's worth testing and/or creating a special case for single-token trees?

bjaglin (approved):

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
mlachkar: no, we don't apply scalafix on this project. But It worth adding it I think. Next  PR ^^

scalafix-core/src/main/scala/scalafix/patch/Patch.scala
mlachkar: It's really hard to create a tree with one token. I was not able to do that.

bjaglin (approved):

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
mlachkar: Actually we apply scalafix on this project. It's an implicit import, and I guess it's hard to see that it's not used. Even Intellij doesn't see it's not used, or maybe it's used, but it's not a necessary import

PR #1261, opened 2020-10-26
Respect the intent of rules authors when returning atomic patches

mlachkar:
In scalafixFileEvaluation, there is a possibility to apply only some patches.
The current issue, is that we were returning the low level patches, by recursively decomposing even Atomic patches.

If the author rule has created an atomic patch that contains two Add patches, we should not decompose it,
and therefore `previewPatches()` can only be called with "atomic" patches

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
@@ -42,7 +42,7 @@ import scalafix.internal.diff.DiffUtils
 import scalafix.internal.interfaces.ScalafixEvaluationImpl
 import scalafix.internal.interfaces.ScalafixFileEvaluationImpl
 import scalafix.internal.patch.PatchInternals
-import scalafix.internal.patch.PatchInternals.patchApply
+import scalafix.internal.patch.PatchInternals.tokenPatchApply
 import scalafix.lint.LintSeverity
 import scalafix.lint.RuleDiagnostic
 import scalafix.v0
@@ -374,7 +374,7 @@ object MainOps {
       ctx: RuleCtx,
       index: Option[v0.SemanticdbIndex]
   ): Option[String] =
-    Try(patchApply(ctx, index, patches)).toOption
+    Try(tokenPatchApply(ctx, index, patches)).toOption
 
   def applyPatches(
       args: ValidatedArgs,
@@ -385,7 +385,7 @@ object MainOps {
   ): Try[ExitStatus] = {
     val input = args.input(file)
     for {
-      fixed <- Try(patchApply(ctx, index, patches))
+      fixed <- Try(tokenPatchApply(ctx, index, patches))
       if (fixed != input.text)
       toFix = args.pathReplace(file).toNIO
       _ <- Try(Files.createDirectories(toFix.getParent))
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
@@ -42,7 +42,7 @@ import scalafix.internal.diff.DiffUtils
 import scalafix.internal.interfaces.ScalafixEvaluationImpl
 import scalafix.internal.interfaces.ScalafixFileEvaluationImpl
 import scalafix.internal.patch.PatchInternals
-import scalafix.internal.patch.PatchInternals.tokenPatchApply
+import scalafix.internal.patch.PatchInternals.patchApply
 import scalafix.lint.LintSeverity
 import scalafix.lint.RuleDiagnostic
 import scalafix.v0
@@ -374,7 +374,7 @@ object MainOps {
       ctx: RuleCtx,
       index: Option[v0.SemanticdbIndex]
   ): Option[String] =
-    Try(tokenPatchApply(ctx, index, patches)).toOption
+    Try(patchApply(ctx, index, patches)).toOption
 
   def applyPatches(
       args: ValidatedArgs,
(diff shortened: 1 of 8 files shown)


scalafix-tests/unit/src/main/scala/scalafix/test/FqnRule.scala
```diff
@@ -50,3 +51,17 @@ class SyntacticRuleV1 extends v1.SyntacticRule("SyntacticRuleV1") {
     Patch.addRight(doc.tree, "\nobject SyntacticRuleV1")
   }
 }
+
+class CommentFileRule1 extends v1.SyntacticRule("CommentFileRule1") {
```
bjaglin, line 55:
```suggestion
class CommentFileNonAtomic extends v1.SyntacticRule("CommentFileNonAtomic") {
```

scalafix-tests/unit/src/main/scala/scalafix/test/FqnRule.scala
```diff
@@ -50,3 +51,17 @@ class SyntacticRuleV1 extends v1.SyntacticRule("SyntacticRuleV1") {
...
+  override def fix(implicit doc: SyntacticDocument): Patch = {
+    Patch.addLeft(doc.tree, "/*") +
+      Patch.addRight(doc.tree, "*/")
+  }
+}
+
+class CommentFileRule2 extends v1.SyntacticRule("CommentFileRule2") {
```
bjaglin, line 62:
```suggestion
class CommentFileAtomic extends v1.SyntacticRule("CommentFileAtomic") {
```

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -214,6 +213,74 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
+        List(
+          "CommentFileRule1"
+        ).asJava
+      )
+      .withSourceroot(src)
+
+    val fileEvaluation1 = run1.evaluate().getFileEvaluations.head
+    val patches1 = fileEvaluation1.getPatches
+    assert(patches1.length == 2)
+    val obtained1 = fileEvaluation1.previewPatches().get
```
bjaglin, line 228: no assertion?

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -214,6 +213,74 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
     assert(contentAfterRun == fileEvaluation.previewPatches().get)
   }
 
+  test("ScalafixArguments.evaluate retrieve only atomic patches") {
```
bjaglin, line 216: I think tests would be more readable (and avoid 1/2 suffixes in variables) if you had more tests with less assertions

scalafix-core/src/main/scala/scalafix/v0/Rule.scala
```diff
@@ -83,7 +83,7 @@ abstract class Rule(ruleName: RuleName) { self =>
     // This overload of apply if purely for convenience
     // Use `applyAndLint` to iterate over Diagnostic without printing to the console
     val result =
-      PatchInternals(patches, ctx, semanticOption)
+      PatchInternals(patches, ctx, semanticOption, suppress = false)
```
bjaglin, line 86: are changes in this file related to the other one above?

scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
```diff
@@ -191,6 +200,20 @@ object PatchInternals {
     loop(patch)
   }
 
+  // don't decompose Atomic Patch
+  def foreachPatchUnit(patch: Patch)(f: Patch => Unit): Unit = {
```
bjaglin, line 204: this is a very generic signature, yet it's used only for accumulation, maybe the 2 new methods could be merged?

bjaglin (commented):
Even with `?w=1` the diff is really hard to follow, and I feel like many changes are unnecessary or could be extracted to no-op refactoring commits. Is there any chance you could amend it to try to produce a minimal diff? Thanks!

scalafix-core/src/main/scala/scalafix/v0/Rule.scala
mlachkar: no. you are right

scalafix-core/src/main/scala/scalafix/internal/patch/PatchInternals.scala
mlachkar: I use it twice

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -374,7 +374,7 @@ object MainOps {
       ctx: RuleCtx,
       index: Option[v0.SemanticdbIndex]
   ): Option[String] =
-    Try(tokenPatchApply(ctx, index, patches)).toOption
+    Try(patchApply(ctx, index, patches)).toOption
```
mlachkar, line 377: this is in fact a rename.

mlachkar (commented):
yes, I will amend to reduce the diff.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
mlachkar: I didi an assertion 
`assertNoDiff(obtained1, obtained2)`

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
mlachkar: ok

mlachkar:
I removed all my small refactoring (renames, and removing `return` keyword). 
The only change is about not using `underlying` that loop inside an `Atomic Patch`.

bjaglin (approved):
thanks for taking the time to amend, much clearer :+1:

PR #1272, opened 2020-11-03
Kill and make sure that the presentation compiler is interrupted befo…

### LINKED ISSUE #1229: Scalafix plugin uses a lot of memory in the sbt project
I added the scalafix plugin to https://github.com/sbt/sbt and the build became noticeably sluggish. It soon became clear the memory utilization of sbt had skyrocketed after adopting the plugin. I took a heap dump after running scalafix and the heap was dominated by scalafix classes: 
Screen Shot 2020-08-24 at 9 49 18 AM
In particular notice that instance of `ScalafixGlobal` are taking up 725MB and that there are 34 `ScalfixGlobalThreads` running (which I also verified with jstack).

mlachkar:
…re starting a new thread

fixes #1229

### DIFF
--- a/scalafix-rules/src/main/scala/scala/tools/nsc/interactive/ScalafixGlobalProxy.scala
+++ b/scalafix-rules/src/main/scala/scala/tools/nsc/interactive/ScalafixGlobalProxy.scala
@@ -2,6 +2,7 @@ package scala.tools.nsc.interactive
 
 import java.util.logging.Level
 
+import scala.tools.nsc.util.WorkScheduler
 import scala.util.control.NonFatal
 
 import scala.meta.internal.pc.ScalafixGlobal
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
Nice job tracking this down! Manual thread management is tricky 😞

bjaglin:
:clap: Maybe we can add a test that exercises this code path many times to cause a OutOfMemory in CI if that ever regresses?

mlachkar:
for CI, we fork each test. it's hard to see the OOM.

scalafix-rules/src/main/scala/scala/tools/nsc/interactive/ScalafixGlobalProxy.scala
```diff
@@ -40,6 +41,7 @@ trait ScalafixGlobalProxy { this: ScalafixGlobal =>
           )
       }
     }
+    this.scheduler = new WorkScheduler
```
ckipp01, line 44: So I guess I don't fully understand this part. What exactly does create and assigning a new WorkScheduler do? Mainly how does it help this situation?

PR #1273, opened 2020-11-03
Use directly the presentation compiler thread

### LINKED ISSUE #1229: Scalafix plugin uses a lot of memory in the sbt project

mlachkar:
fixes #1229


mlachkar:
The other solution has been validated. We keep using a new thread, but we make sure to kill the presentation compiler first

PR #1275, opened 2020-11-04
revert unnecessary breaking change in testkit

bjaglin:


mlachkar (approved):

mlachkar:
My change was not that clean. I can fix tests here if you want.

build.sbt
```diff
@@ -219,12 +219,8 @@ lazy val unit = project
...
       put(
         "inputClasspath",
-        fullClasspath.in(testsInput, Compile).value.map(_.data)
-      )
-
-      put(
-        "semanticdbRootDirectory",
-        Seq(semanticdbTargetRoot.in(testsInput, Compile).value)
+        fullClasspath.in(testsInput, Compile).value.map(_.data) :+
+          semanticdbTargetRoot.in(testsInput, Compile).value
```
bjaglin, line 223: similar strategy as in https://github.com/scalacenter/sbt-scalafix/commit/cd9efb1b1e16f178da6be72a4cc621524e8a301a

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
```diff
@@ -170,7 +170,7 @@ class CliSemanticSuite extends BaseCliSuite {
   checkSemantic(
     name = "-P:semanticdb:targetroot",
     args = {
-      val (targetroot :: classDir :: Nil, jars) =
+      val (_ :: targetroot :: Nil, jars) =
```
bjaglin, line 173: not new to this PR, but the fact that order matters here makes the test a bit fragile - I think it's acceptable as we don't put any ordering constraint for (external) testkit client, this is just a shortcut for fixturing.

PR #1276, opened 2020-11-06
ExplicitResultTypes path-dependent types import bugfix & tweaks

bjaglin:
Best reviewed by commits (as each of them should be green test-wise): the first one prevents compilation error when running the rule against a codebase using https://github.com/tpolecat/doobie (see commit message and comments in test), while the second improve reference shortening for path-dependent types.

### DIFF
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
@@ -396,8 +396,19 @@ class ScalafixGlobal(
             Iterator(lookupSymbol(name), lookupSymbol(name.otherName))
           results.flatten.filter(_ != LookupNotFound).toList match {
             case Nil =>
-              missingImports(name) = short
-              true
+              // Missing imports must be addressable via the dot operator
+              // syntax (as type projection is not allowed in imports).
+              // https://lptk.github.io/programming/2019/09/13/type-projection.html
+              if (
+                sym.isStaticMember || // Java static
+                sym.owner.ownerChain.forall { s =>
+                  // ensure the symbol can be referenced in a static manner, without any instance
+                  s.isPackageClass || s.isPackageObjectClass || s.isModule
+                }
+              ) {
+                missingImports(name) = short
+                true
+              } else false
             case lookup =>
               lookup.forall(_.symbol.isKindaTheSameAs(sym))
           }
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
@@ -238,7 +238,21 @@ class ScalafixGlobal(
             // "scala.Seq[T]" even when it's needed.
             loop(ThisType(sym.owner), name)
           } else if (sym.hasPackageFlag || sym.isPackageObjectOrClass) {
-            if (history.tryShortenName(name)) NoPrefix
+            val dotSyntaxFriendlyName = name.map { name0 =>
+              if (name0.symbol.isStatic) name0
+              else {
+                // Use the prefix rather than the real owner to maximize the
+                // chances of shortening the reference: when `name` is directly
+                // nested in a non-statically addressable type (class or trait),
+                // its original owner is that type (requiring a type projection
+                // to reference it) while the prefix is its concrete owner value
+                // (for which the dot syntax works).
+                // https://docs.scala-lang.org/tour/inner-classes.html
+                // https://danielwestheide.com/blog/the-neophytes-guide-to-scala-part-13-path-dependent-types/
+                ShortName(name0.symbol.cloneSymbol(sym))
+              }
+            }
+            if (history.tryShortenName(dotSyntaxFriendlyName)) NoPrefix
             else tpe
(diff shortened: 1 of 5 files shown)


scalafix-tests/output/src/main/scala/test/explicitResultTypes/ImportAbstractOwner.scala
```diff
@@ -0,0 +1,22 @@
+package test.explicitResultTypes
+
+import test.explicitResultTypes.Package.{ Foo, Nested }
```
bjaglin, line 3: I initially explored going for
```suggestion
import test.explicitResultTypes.Trait.{ Foo, Nested }

  val fooRef: Trait#Foo = foo
```
but the implementation looked more complicated, and less generic

mlachkar (approved):

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
```diff
@@ -238,7 +238,13 @@ class ScalafixGlobal(
             // "scala.Seq[T]" even when it's needed.
             loop(ThisType(sym.owner), name)
           } else if (sym.hasPackageFlag || sym.isPackageObjectOrClass) {
-            if (history.tryShortenName(name)) NoPrefix
+            val importableName = name.map(_.symbol) match {
```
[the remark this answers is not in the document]
bjaglin: Sure - actually while trying to explain these lines I realized I misunderstood the root problem - I'll push a completely revised commit history shortly!

bjaglin:
@mlachkar I just force-pushed after iterating a few hours, tell me what you think! The compiler and the type system are pretty much unknown to me, so I might be missing something or use the wrong terminology, but at least it's definitely clearer in my head now, and the diff is (I think) easier to follow.

scalafix-tests/output/src/main/scala/tests/ExplicitResultTypesImports.scala
```diff
@@ -18,8 +18,7 @@ object ExplicitResultTypesImports {
 
   val timezone: ju.TimeZone = null.asInstanceOf[java.util.TimeZone]
 
-  // TODO: Is this desirable behavior?
-  val inner: Searching.SearchResult = null.asInstanceOf[scala.collection.Searching.SearchResult]
+  val inner: SearchResult = null.asInstanceOf[scala.collection.Searching.SearchResult]
```
olafurpg, line 21: The primary motivation for the old behavior comes from the desire to insert `Term.Name` and `import scala.meta.Term` instead of `import scala.meta.Term.Name`. Scalameta isn't the best motivating example, but I feel like this pattern of nesting generic names inside a class is not only limited to Scalameta. Any thoughts on how to support both use-cases? I'm open for various routes

* default to importing FQN with option to flip default
* ability to manually include/exclude individual prefixes from nesting

bjaglin: Thanks for feedback, I didn't realize the current behavior was an active choice. It makes sense with your example, I am fine keeping it, this was just a side effect I got when iterating on my own test. I have amended the last commit to keep/document the current behavior and align it with path-dependent types, see https://github.com/scalacenter/scalafix/pull/1276/commits/fffba0667b7e72767e58f360e8e245b0999fb851.

bjaglin:
Build failures are related to https://gitter.im/sbt/sbt?at=5fa7ee2fb86f6407043af431

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
```diff
@@ -500,6 +525,9 @@ class ScalafixGlobal(
     def isJavaModule: Boolean =
       sym.isJava && sym.isModule
 
+    def isObject: Boolean =
+      sym.isModule && !sym.hasPackageFlag && !sym.isPackageObject
```
bjaglin, line 529: maybe there is something simpler?

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
```diff
@@ -396,8 +410,19 @@ class ScalafixGlobal(
...
-              true
+              // Missing imports must be addressable via the dot operator
+              // syntax (as type projection is not allowed in imports).
+              // https://lptk.github.io/programming/2019/09/13/type-projection.html
+              if (
+                sym.isStaticMember || // Java static
+                sym.owner.ownerChain.forall { s =>
+                  // ensure the symbol can be referenced in a static manner, without any instance
+                  s.isPackageClass || s.isPackageObjectClass || s.isModule
+                }
```
bjaglin, line 421: This has the same semantics as `isStatic`, but `isStatic` yields false negatives when fed with the special case `ShortName(name0.symbol.cloneSymbol(sym))` above, as (I believe) not all flags are properly set since the symbol reattached to the name does not come from the same compiler codepath.

olafurpg (approved):
LGTM 👍 It's tricky to navigate the compiler APIs, nice work!

bjaglin:
> It's tricky to navigate the compiler APIs, nice work!

Thanks. Indeed, I didn't realize all the work you put into that rule until now! :sweat_smile:

PR #1278, opened 2020-11-11
allow building on JDK 15

SethTisue:
I don't know anything about this flag, but JDK 15 doesn't accept it.

A possible alternate approach here would be to add `-XX:+IgnoreUnrecognizedVMOptions`.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -3,5 +3,4 @@
 -Xmx2G
 -XX:ReservedCodeCacheSize=1024m
 -XX:+TieredCompilation
--XX:+CMSClassUnloadingEnabled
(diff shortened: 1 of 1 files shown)


bjaglin (approved):
LGTM as we use Java8+ anyway on CI for the build. See https://github.com/scalacenter/scalafix/blob/25e2e79c9279f47d49cfa4f4e824a48d3d46a067/.jvmopts#L8.

bjaglin:
From my understanding based on https://github.com/scalacenter/scalafix/pull/1279, there is no point trying to run CI on JDK15 (to prevent this to regress) until 2.13.4 is out.

PR #1282, opened 2020-11-13
Update setup-scala githubAction

mlachkar:
Warn from setup-scala: The `set-env` command is deprecated and will be disabled on November 16th


bjaglin (approved):
:+1:

PR #1291, opened 2020-11-26
Upgrade to Scala 2.13.4

mlachkar:
Similar to https://github.com/scalacenter/scalafix/pull/1179


mlachkar:
@bjaglin I need help on the mima ci issue. I don't understand why it's still try to download `/semanticdb-scalac-core_2.13.4/4.3.24`. 

Thanks

bjaglin:
> @bjaglin I need help on the mima ci issue. I don't understand why it's still try to download `/semanticdb-scalac-core_2.13.4/4.3.24`.
> 
> Thanks

@mlachkar you need to update https://github.com/scalacenter/scalafix/blob/4c2eb94b9bf5cda08263cfca42466f0c881a29fc/project/ScalafixBuild.scala#L271

mlachkar:
I think I don't fully understand the` PreviousScalaVersion Map`. It could be empty, it would give the same result.

bjaglin:
> I think I don't fully understand the` PreviousScalaVersion Map`. It could be empty, it would give the same result.

Indeed, my suggestion didn't make sense. I'll have a look at that today.

bjaglin:
It looks like a regression in MiMa introduced in https://github.com/lightbend/mima/pull/508 - see last commit details. I'll follow up there, for the time being I think it's easier to downgrade.

mlachkar:
If it's ok for you, we can merge and then release ?

bjaglin:
> If it's ok for you, we can merge and then release ?

sure of course!

daddykotex:
Thanks for the hard work 👍

PR #1292, opened 2020-11-30
Update scalafix plugin to 0.9.24

mlachkar:


project/ScalafixBuild.scala
```diff
@@ -266,11 +266,6 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     )
   )
 
-  private val PreviousScalaVersion = Map(
-    "2.12.13" -> "2.12.12",
-    "2.13.4" -> "2.13.3"
```
bjaglin, line 271: Right, this needs to be bumped or targeting an unknown version once an artifact of the current version is available (independantly of the actual change here related to sbt-scalafix). Maybe you could keep an empty list to facilitate the next transition (so we don't need to lookup how that was done before?). Anyway, not a big deal!

bjaglin (approved):
:+1:

project/ScalafixBuild.scala
mlachkar: ok

PR #1293, opened 2020-11-30
Fix Patch.replaceSymbols bug that affects ScalaTest rule.

olafurpg:
Previously, the ScalaTest autofix rule didn't properly handle cases with
renamed imports due to a bug in Scalafix. This commit fixes that bug so
that the ScalaTest rule works correctly as expected.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
@@ -3,6 +3,7 @@ package scalafix.internal.patch
 import scala.meta._
 import scala.meta.internal.trees._
 
+import scalafix.internal.util.SymbolOps
 import scalafix.internal.util.SymbolOps.Root
 import scalafix.internal.util.SymbolOps.SignatureName
 import scalafix.patch.Patch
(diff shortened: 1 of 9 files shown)


mlachkar (approved):

mosesn:
@mlachkar I'm not familiar with the scalafix OSS process, what else would this patch need to be merged in?  We'd love to take advantage of the improvement :) (we think we can start cross-building for Scala 3 after this lands!)

mlachkar:
Of course, I can merge it.

mosesn:
@mlachkar great, thank you!  What is the typical release cadence for scalafix?  I'm going to try to figure out how we can build a snapshot locally (unless you publish nightlies?) but it would be cool to have this released too :)

bjaglin:
@mosesn until scalafix 0.9.25 is out, you can use `0.9.24+4-ae8741e5-SNAPSHOT` available on Sonatype OSS snapshots.

If you use sbt-scalafix, this should do it:
#### build.sbt
```diff
+ThisBuild / scalafixResolvers += coursierapi.MavenRepository.of("https://oss.sonatype.org/content/repositories/snapshots"),
```

#### project/plugins.sbt
```diff
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.24")
+resolvers += Resolver.sonatypeRepo("snapshots")
+libraryDependencies += "ch.epfl.scala" % "scalafix-interfaces" % "0.9.24+4-ae8741e5-SNAPSHOT"
```

mosesn:
An update from my side!  I finally got some time to look at this, and I was able to use that snapshot version.  I'm trying it out as we speak, thanks!!  Unfortunately, running a scalafix task across our entire codebase takes all day.

mosesn:
I tried it out, and I still have a problem.  I'm not very familiar with scalafix, can someone who is more familiar with it help me figure out how to solve my problem?  I'll file a ticket to explain.

The short version is that if it's possible to do a rename that causes a name collision, which can make code uncompileable.

Issue #1299, opened 2020-12-04
0.9.24: loss of meaningful braces


rossabaker:
In http4s, since upgrading to 0.9.24, we're seeing a loss of braces that were preserved in 0.9.23.  Input:

```scala
    final def doFetchR2[B](req: Request[G], f: Response[G] => G[B])(implicit G: Sync[G]): G[B] =
      fetch[B](req) { f }
```

Diff:

```scala
     final def doFetchR2[B](req: Request[G], f: Response[G] => G[B])(implicit G: Sync[G]): G[B] =
-      run(req).use[G, B] f
+      run(req).use[G, B] { f }
```

It's also affecting more complex expressions, like partial function literals.  Input:

```scala
  def doClientFetchR5[F[_]: Sync, A](client: Client[F], req: Request[F], f: Response[F] => F[A]): F[A] = {
    client.fetch[A](req) { case res@Response(_, _, _, _, _) => f(res) }
  }
```

Diff:

```scala
   def doClientFetchR5[F[_]: Sync, A](client: Client[F], req: Request[F], f: Response[F] => F[A]): F[A] = {
-    client.run(req).use[F, A] case res@Response(_, _, _, _, _) => f(res)
+    client.run(req).use[F, A] { case res@Response(_, _, _, _, _) => f(res) }
```

I don't have a minimization yet, but it can be observed from `sbt ++2.12.12 scalafixTests/test` on https://github.com/http4s/http4s/pull/3943/commits/b4f0ca5632b534b74576ce21ca3e2cb79b657c1a.

Could this be related to the scalameta-4.4.0 upgrade?

bjaglin:
Thanks for the report. I have reproduced the problem with scalafix 0.9.24, as well as with master that now includes scalameta-4.4.2. I'll try to find some time in the coming days to extract a repro from https://github.com/http4s/http4s/tree/main/scalafix/rules/src/main/scala/fix to file a ticket in scalameta as it does seem related to the bump.

rossabaker:
I haven't connected it to a scalameta bump, but our tests are passing again with 0.9.25.

Thanks!

PR #1327, opened 2021-02-04
Use directly the presentation compiler thread

mlachkar:
I think it could be nice to remove this unnecessary complexity.


bjaglin (commented):
That would be nice indeed! Doesn't https://github.com/scalacenter/scalafix/pull/1272 showed that handling failures and restarting the compiler was sometimes necessary though (otherwise there would have been no leak?). I ran into into https://github.com/scalameta/metals/pull/2476/files#r570479181 and I have the same concerns basically.

Maybe you could smoke-test this against https://github.com/scalacenter/scalafix/issues/1229#issuecomment-680087256 and other large open source projects that use scalafix with `ExplicitResultTypes` (via a github search)? I had written instructions on how to use a scalafix-interfaces snapshot with sbt-scalafix in https://github.com/scalacenter/scalafix/issues/1299#issuecomment-745187397.

mlachkar:
We would not have a leak if we use directly the the presentation compiler #1229, I profiled both solutions. 
In metals issue, they first considered removing the hijackPresentationCompiler, before using the same fix than in Scalafix. 
I do think that the code we have that do that was a copy from metals. 
The only change compared to presentation compiler thread is in this https://github.com/scalacenter/scalafix/blob/master/scalafix-rules/src/main/scala/scala/tools/nsc/interactive/ScalafixGlobalThread.scala#L46-L62

I agree that we need to do some testing to make sure it doesn't have any impact. I will add those tests in this Pull request

mlachkar:
I tried the 0.9.26 and this version on Zio code base, and the profile of cpu/memory is the same: what I was expecting.


`0.9.26` : run in `[success] Total time: 86 s (01:26), completed 29 Mar 2021, 11:27:26`


image



This PR: `[success] Total time: 82 s (01:22), completed 29 Mar 2021, 11:31:26`

image

I tried both with the same conditions: I compiled first the entire project, then I run scalafixAll.
I don't think we can conclude that this PR performs better, but we can conclude that the performance should be similar. 
What do you think @bjaglin ?

project/Mima.scala
```diff
@@ -17,7 +17,17 @@ object Mima {
...
+      ProblemFilters.exclude[MissingTypesProblem]("scala.meta.internal.pc.ScalafixGlobal"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.internal.pc.ScalafixGlobal.presentationCompilerThread"),
+      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.ScalafixGlobalProxy"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.internal.pc.ScalafixGlobal.hijackPresentationCompilerThread"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.internal.pc.ScalafixGlobal.isHijacked"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("scala.meta.internal.pc.ScalafixGlobal.metalsAsk"),
+      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.ScalafixGlobalThread"),
+      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.ScalafixGlobalThread$"),
+      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.InterruptException"),
+      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.InterruptException$")
```
bjaglin, line 30: Maybe classes that are outside the scalafix namespace just to access internals should be fully excluded?
```suggestion
      ProblemFilters.exclude[MissingTypesProblem]("scala.meta.internal.*"),
      ProblemFilters.exclude[MissingClassProblem]("scala.tools.nsc.interactive.*")
```

mlachkar: better, in fact.

Issue #1333, opened 2021-02-18
sbt-scalafix: drop support for sbt 0.13


bjaglin:
It is probably a good time to drop support for sbt 0.13 - sbt 1.0 was released 3,5 years ago.

Benefits/TODO:
- use scala 2.12 features
- get rid of the [1.0/0.13 compatibility layer](https://github.com/scalacenter/sbt-scalafix/tree/master/src/main/scala-sbt-1.0/sbt/internal/sbtscalafix)
- use slash notation (in code and in docs)
- restructure scripted to run most (all?) of them against different sbt 1.x versions, like https://github.com/thesamet/sbt-protoc/pull/196

bjaglin:
Considering that it most likely won't be particularly harder to ship https://github.com/scalacenter/scalafix/issues/1146 with sbt 0.13 support, I think it would be nice to hold this just a little longer, so that clients stuck to sbt 0.13 can still benefit from future versions of scalafix itself. Once that's done, then I think it's really safe to drop support and leverage newer sbt features.

mlachkar:
We can maybe close this ticket now ?

bjaglin:
Why? The work hasn't started.

mlachkar:
ok. I thought we dropped support of sbt 0.13 in sbt-scalafix, but I am wrong. https://github.com/scalacenter/sbt-scalafix/pull/219
We dropped support of 0.13 for testkit only.

PR #1342, opened 2021-03-01
use slash notation in the build

mlachkar:


build.sbt
```diff
@@ -1,20 +1,16 @@
 import Dependencies._
-inThisBuild(
```
bjaglin, line 2: is this deprecated??? I liked it better :thinking:

bjaglin (commented):
:+1: did you use https://eed3si9n.com/syntactic-scalafix-rule-for-unified-slash-syntax? ~It looks like you are missing `project/*.scala`~

build.sbt
mlachkar: No, it's not deprecated. It looks a little like the old syntax, so I proposed to change it, but let's keep it, since I cannot change the template scalafix.g8 to use it only.

bjaglin: I like the idea of grouping build-level settings together, just like you do for a project

bjaglin (approved):
:+1: 

related to, but does not close https://github.com/scalacenter/scalafix/issues/1333 - we need sbt-scalafix to drop 0.13 to update use unified syntax in docs for example

Issue #1264, opened 2020-10-29
Remove Unused depends on -Xmaxwarns on Scala 2.13


melezov:
The Remove Unused Scalafix rule seems to be attached to the `-Xmaxwarns` scala compiler option.

The documentation of the `-Xmaxwarns` does not seem to reflect that processing will change, only the presentation of the warnings:

    -Xmaxwarns <n> Maximum warnings to print

If the `-Xmaxwarns` is not set, it defaults to 100.
It means that if the unused imports/code is not within these "printed" warnings, it will not have been processed on Scala 2.13

This is a regression from Scala 2.12.x, where this setting did not seem to affect the scalafix rule execution.

Specific versions I'm referring to: Scala 2.13.3/2.12.12

mlachkar:
In fact the rule RemoveUnusued depends on this parameter, but I don't know what we can do to change this behavior. 
We can document this setting that impact removeUnusued and all rules that depends on Warn messages. 
@bjaglin 
1 - Could we modify the default value on the project where the scalafix plugin is enabled , or just during the run of scalafix, as we do for `-Xfatal-warning` 
2- Report as warning in `withConfiguration` for RemoveUnusued that this value could be increased ?

bjaglin:
1 - That looks like a good idea indeed to leverage https://github.com/scalacenter/sbt-scalafix/pull/196, although I am not sure what's a good value?
2 - This looks like something https://github.com/scalacenter/scalafix/issues/1199 could signal. Since we have the `scalacOptions`, we know the value of `-Xmaxwarns` (implicit default or explicit), so provided all warnings are reported as diagnostics in semanticdb, we could even fail the build when we detect that we have exactly as many diagnostics as the maximum number of warns, since there is a good chance we are missing out some.

mlachkar:
Actually after thinking more seriously about this, semanticdb should not depend on this to record warnings. I m working on a fix on scalemeta side. Those parameters are supposed to change the number of warning or errors printed.

mlachkar:
linked to https://github.com/scalameta/scalameta/pull/2402

mlachkar:
RemoveUnused doesn't depend on Maxwarn anymore!
Thanks for reporting this issue

PR #1345, opened 2021-03-02
Fix RemoveUnusedPatternVars

### LINKED ISSUE #604: ignore generated (expanded macros) code
#559

### LINKED ISSUE #1343: Remove unused doesn't remove unused pattern vars when inside a method
To reproduce the issue:
```
  def test(): Unit = {
    example match {
      case AB(a, b) => println("Not used, wrong")
    }
  }
```
The regex used to catch the diagnostic message is too specific to values

vincenzobaz:
This PR aims at fixing #1343 by using a more general regexp and adding several tests.

closes #1343

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -145,6 +145,7 @@ lazy val testsInput = project
     scalacOptions += warnAdaptedArgs.value, // For NoAutoTupling
     scalacOptions += warnUnusedImports.value, // For RemoveUnused
     scalacOptions += "-Ywarn-unused", // For RemoveUnusedTerms
+    scalacOptions ++= maxwarns.value, // Increase the maximum warnings to print
     logLevel := Level.Error, // avoid flood of compiler warnings
     libraryDependencies += bijectionCore,
     testsInputOutputSetting,
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -39,9 +39,14 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       crossVersion := CrossVersion.full
     )
     lazy val isScala213 = Def.setting { scalaVersion.value.startsWith("2.13") }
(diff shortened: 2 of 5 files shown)


mlachkar (approved):

mlachkar:
@vincenzobaz can you check why the tests are failing .. I find that strange the renameSuite is failing. Can you also run bin/scalafmt. Thanks

mlachkar:
I am having issues understanding the issue with this PR. The change should not affect the other tests, but it's the case. 
When debugging, the list of diagnostics is empty in the failing tests. 
In my tests, the semanticdb files were produce during compilation time, not by interactive.global. 
I am quite blocked by this bug. @bjaglin what do you think we should do ?

mlachkar:
same code, but the tests are passing https://github.com/scalacenter/scalafix/pull/1355

mlachkar:
So after further debugging: in fact when adding these new case tests, we don't have warnings anymore on some files...

mlachkar:
I have just removed a number of tests that were added by this pull request.
Something in those tests was making the compiler not producing anymore "unused warning" for other files.

bjaglin:
>  Something in those tests was making the compiler not producing anymore "unused warning" for other files.

Could it be a limit in the compiler output? https://docs.scala-lang.org/overviews/compiler-options/index.html
>  -Xmaxwarns ARG
>
>   Maximum warnings to print
>   Default: 100

mlachkar:
I think it should be that! completly forgot about this :s

mlachkar:
It also means that this rule fixes maximum 100 warns by default ^^

bjaglin:
> It also means that this rule fixes maximum 100 warns by default (in each run)^^

https://github.com/scalacenter/scalafix/issues/1264

We could mention it in the docs at least?

PR #1347, opened 2021-03-03
fix instructions to use nightlies

bjaglin:
`sbt-scalafix` is versioned independently (at least for SNAPSHOTs) and we don't have its latest SNAPSHOT version here, so it's `scalafix-interfaces` the user should override to get the latest `scalafix-core` & `scalafix-rules`.

### DIFF
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -451,19 +451,15 @@ Scalafix is supported in other build tools via externally maintained plugins:
 ## SNAPSHOT
 
 Our CI publishes a snapshot release to Sonatype on every merge into master. Each
-snapshot release has a unique version number, jars don't get overwritten. To
-find the latest snapshot version number, go to
-<https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-core_2.12/>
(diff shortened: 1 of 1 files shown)


mlachkar (approved):

mlachkar (commented):
I have just tested this documentation, it works perfectly. Maybe we can add a link to where we can find those snapshots (this was the most difficult part) https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-interfaces/

bjaglin:
> Maybe we can add a link to where we can find those snapshots (this was the most difficult part) https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalafix-interfaces/

Well, I actually removed the link towards the Sonatype OSS listing, since the latest available version is advertised in the generated doc thanks to `@NIGHTLY_VERSION@`, and I think the most common use-case is to use the latest available version. Advanced users trying to bisect a regression would probably know where to look for the artifact anyway.

mlachkar:
I didn't know about the doc being generating the right snapshot. Yes I think it's enough

PR #1348, opened 2021-03-04
Close global after compute ondemand semanticdb files

mlachkar:
The state before the fix. It possible to reproduce by running multiple time the test on-demand with scalacOptions
The presentation compiler is never closed.

image

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
@@ -12,6 +12,7 @@ import java.nio.file.attribute.BasicFileAttributes
 
 import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
+import scala.tools.nsc.interactive.Global
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
@@ -37,6 +38,7 @@ import org.typelevel.paiges.{Doc => D}
 import scalafix.Versions
 import scalafix.cli.ExitStatus
 import scalafix.interfaces.ScalafixEvaluation
+import scalafix.internal.compat.CompilerCompat.XtensionGlobal
 import scalafix.internal.config.PrintStreamReporter
 import scalafix.internal.diff.DiffUtils
 import scalafix.internal.interfaces.ScalafixEvaluationImpl
@@ -132,6 +134,7 @@ object MainOps {
           }
           // Then afterComplete for each rule
           args.rules.rules.foreach(_.afterComplete())
+          shutdownCompiler(args.global)
 
           ScalafixEvaluationImpl.from(fileEvaluations, ExitStatus.Ok)
         } else
@@ -267,7 +270,6 @@ object MainOps {
           )
         } catch {
           case NonFatal(_) =>
-            args.global.restart()
             TextDocument.defaultInstance
         }
       g.unitOfFile.clear()
@@ -347,6 +349,17 @@ object MainOps {
     } else {
       args.rules.syntacticPatch(doc, args.args.autoSuppressLinterErrors)
     }
+  }
+
+  private def shutdownCompiler(global: LazyValue[Option[Global]]): Unit = {
+    global.foreach(_.foreach(g => {
+      try {
+        g.askShutdown()
+        g.closeCompat()
+      } catch {
+        case NonFatal(_) =>
+      }
+    }))
   }
 
   def previewPatches(
@@ -440,6 +453,7 @@ object MainOps {
     }
 
     args.rules.rules.foreach(_.afterComplete())
+    shutdownCompiler(args.global)
 
     adjustExitCode(args, exit, files)
   }
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -383,7 +383,7 @@ case class Args(
         settings.YpresentationAnyThread.value = true
         settings.classpath.value = validatedClasspath.syntax
         val reporter = new StoreReporter()
-        new Global(settings, reporter)
+        new Global(settings, reporter, "Scalafix")
       })
     }
 


scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -267,7 +269,6 @@ object MainOps {
           )
         } catch {
           case NonFatal(_) =>
-            args.global.restart()
```
bjaglin, line 270: is this related to the change?

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -347,6 +348,18 @@ object MainOps {
     } else {
       args.rules.syntacticPatch(doc, args.args.autoSuppressLinterErrors)
     }
+    // close global if evaluated
+    cleanGlobalRessources(args.global)
```
bjaglin, line 352: should that be in a finally block instead?

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -347,6 +348,18 @@ object MainOps {
...
+    results
+  }
+
+  private def cleanGlobalRessources(global: LazyValue[Option[Global]]): Unit = {
+    if (global.isEvaluated)
+      global.value.foreach { g =>
+        g.askShutdown()
+        g.closeCompat()
+      }
```
bjaglin, line 361: is that the prefered way to shut down the compiler? https://github.com/scalacenter/scalafix/blob/f6bff33bf13512a9e73c83e448eb5456de1b5db2/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L38-L47 is a bit different

bjaglin (commented):
good catch! some questions/comments as I am not very familiar with `Global`

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
mlachkar: no, but I don't understand why we do that. 
even if the compiler crash, the global don't crash. We can still close it after this.

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
mlachkar: ok

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
mlachkar: this what happens if we let this line
```try {
          InteractiveSemanticdb.toTextDocument(
            g,
            doc.input.text,
            doc.internal.input.syntax,
            10000,
            Nil
          )
          throw new Exception("exception")
        } catch {
          case NonFatal(_) =>
            println("exception catched")
            args.global.restart()
            TextDocument.defaultInstance
        }
        ```
image

Comment (mlachkar) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala:270]: it's never closed

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 2 above)
Comment (mlachkar) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala:352]: yes. but where?

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 3 above)
Comment (mlachkar) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala:361]: those two methods are not supposed to throw exceptions. but why not

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 2 above)
Comment (bjaglin) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala]: I don't understand your question - here?

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 3 above)
Comment (bjaglin) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala]: I didn't look at the git history, but I am guessing this was done for a reason? or to be defensive, which is not a bad thing

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 1 above)
Comment (bjaglin) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala:270]: I see, I just realized`restart()` was just dropping the reference https://github.com/scalacenter/scalafix/blob/e31e5c946a14bb3839ef5c087b660953a3485db8/scalafix-core/src/main/scala/scalafix/internal/v1/LazyValue.scala#L17-L19 :+1:

**Review by bjaglin — COMMENTED** [review]
Maybe a larger question but do we want to keep supporting ondemand generation of semanticdb files in 1.0? This will require significant effort to support it with Scala 3 sources, and as far as I know, that feature is undocumented.

What are the use-cases behind it? Was that done to cover shortcomings of build tool integration? It looks like @olafurpg added it while working on `ExplicitResultTypes` but the git history is not very clear.

**Comment by mlachkar** [comment]
It's useful if the semanticdb files are not present, especially if it's used through the API when we don't depend on compile task.
We may open another issue to discuss this. I agree it's confusing not to know if Scalafix is compiling files, or relying on existing semanticdb files, using interactive global (which up to the latest release, was not giving the same diagnostics for example)

**Review by bjaglin — APPROVED** [review]
> It's useful if the semanticdb files are not present, especially if it's used through the API when we don't depend on compile task.
> We may open another issue to discuss this.

:+1: 

LGTM (moving the call to `shutdownCompiler` in a finally block would be a bonus I guess)

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala (hunk 2 above)
Comment (mlachkar) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala]:
```
try {
 val tree = LazyValue.later { () =>
      args.parse(input).get: Tree
    }
    val doc = SyntacticDocument(input, tree, args.diffDisable, args.config)
    val results = if (args.rules.isSemantic) {
      val relpath = file.toRelative(args.sourceroot)
      val sdoc = SemanticDocument.fromPath(
        doc,
        relpath,
        args.classLoader,
        args.symtab,
        () => compileWithGlobal(args, doc)
      )
      val result =
        args.rules.semanticPatch(sdoc, args.args.autoSuppressLinterErrors)
      assertFreshSemanticDB(
        input,
        file,
        result.fixed,
        sdoc.internal.textDocument
      )
      result
    } else {
      args.rules.syntacticPatch(doc, args.args.autoSuppressLinterErrors)
    }
    results
  } finally {
      // close global if evaluated
    shutdownCompiler(args.global)
    }
```

Comment (mlachkar) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala]: something like this ?

Comment (bjaglin) [reply scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala]: hard to tell without proper indentation and highlighting but yes :+1:  (the `results` val can be removed)

**Comment by mlachkar** [comment]
try finally will be usefull since wr throw a potentiel exception with the Assert. Thank you for the advise

File: scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala [hunk 4]
```diff
@@ -321,32 +322,44 @@ object MainOps {
...
+          input,
+          file,
+          result.fixed,
+          sdoc.internal.textDocument
+        )
+        result
+      } else {
+        args.rules.syntacticPatch(doc, args.args.autoSuppressLinterErrors)
+      }
+    } finally shutdownCompiler(args.global)
```
bjaglin, line 352: actually I guess the finally block could be scoped to where it's actually used - around the tunk calling `compileWithGlobal`?

mlachkar: You are right, it's again on the wrong place .. we need it elsewhere, where the function is called. we don't want to open a new gloabl for each file.

olafurpg (approved):
Nice fix!

olafurpg:
The library API could support an enum mode to determine how to load SemanticDB files. I agree it can sometimes be confusing when the fallback triggers.

bjaglin (approved):

PR #1354, opened 2021-03-26
Trampoline Concat patches.

rpiaggio:
Avoids stack overflows when processing large `Concat` structures by using trampolining.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
@@ -4,6 +4,7 @@ import scala.annotation.tailrec
 import scala.collection.immutable.TreeMap
 import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
+import scala.util.control.TailCalls._
 
 import scala.meta._
 import scala.meta.contrib._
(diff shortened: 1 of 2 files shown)


olafurpg (approved):
LGTM 👍 Thank you for this contribution! I have never used the `TailRec` class so this was interesting to learn about. wdyt @mlachkar @bjaglin

mlachkar:
Thanks for the PR. I am curious, have you ever experienced StackOverflow in this method and with which rule?
This method is called for a patch per rule per file. So usually for one file, with the rules I know, usually, we don't exceed 100 underlying patches, which should be around 100 recursive calls. 
On the other hand, in fact, nothing prevents this method from going StackOverflow.

rpiaggio:
Thank you!

> have you ever experienced StackOverflow in this method and with which rule?

Yes, I experienced it when developing a custom rule that removes (and then regenerates) a large chunk of code. The resulting patch includes a `Patch.removeTokens` which results in a large `Concat` structure. This would overflow in the `loop` recursion within the `foreach` in `PatchInternals.scala`.

bjaglin (approved):
:+1: great usage of `TailRec` (which I had no clue had made its way to the standard library)

PR #1356, opened 2021-03-29
Don't produce each time a RichCompilationUnit

mlachkar:
The bug doesn't seem to exist anymore.
This call was increasing the memory usage of ExpicitResultType


mlachkar:
The allocation rate is lower with this change. Moreover this call has been removed in `metals` [here](https://github.com/scalameta/metals/blob/4748fcdb4a016b1b63295eaa1049f7e4ec2523e2/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImportsProvider.scala#L26) and `scala3-migrate`. 
I do believe that the impact will be higher on projects that are just adopting ExplicitResultType Rule, since we call this method as many times as we find missing explicit result types. 


**With 0.9.26**
image



**With this PR**
image

bjaglin (approved):
Cool! Do we need `GlobalProxy` at all then?

mlachkar:
It's still a private method in a package (the method typeTreeAt). we just need it to modify the package hierarchy

PR #1373, opened 2021-04-16
Allow configuring the dialect in Scalafix 

mlachkar:
Scala3 dialect is a superset of Scala213 dialect: 
```  
implicit val Scala3 = Scala213
    .withAllowAndTypes(true)
    // there 3 different ways to specify vargs, some will be removed in future Scala 3 versions
    .withAllowAtForExtractorVarargs(true) // both @ and : work currently for Scala 3
    .withAllowColonForExtractorVarargs(true) // both @ and : work currently for Scala 3
    .withAllowPostfixStarVarargSplices(true)
    .withAllowEnums(true)
    .withAllowImplicitByNameParameters(true)
    .withAllowInlineMods(true)
    .withAllowLiteralTypes(true)
    .withAllowMethodTypes(false)
    .withAllowOrTypes(true) // Scala 3: `val a: A | B`
    .withAllowTrailingCommas(true)
    .withAllowTraitParameters(true)
    .withAllowTypeLambdas(true)
    .withAllowViewBounds(false) // View bounds have been removed in Dotty
    .withAllowXmlLiterals(false) // Scala 3: parser doesn't support xml
    .withAllowGivenUsing(true)
    .withAllowExtensionMethods(true)
    .withAllowOpenClass(true)
    .withAllowSpliceAndQuote(true)
    .withAllowToplevelStatements(true)
    .withAllowOpaqueTypes(true)
    .withAllowExportClause(true)
    .withAllowCommaSeparatedExtend(true)
    .withAllowEndMarker(true)
    .withAllowInterpolationDolarQuoteEscape(true)
    .withAllowSignificantIndentation(true)
    .withAllowQuestionMarkPlaceholder(true)
    .withAllowTypeParamUnderscore(false)
    .withAllowByNameRepeatedParameters(true)
    .withAllowLazyValAbstractValues(true)
    .withAllowUpperCasePatternVarBinding(true)
    .withAllowDerives(true)
    .withAllowTypeInBlock(true)
    .withAllowPolymorphicFunctions(true)
    .withAllowMatchAsOperator(true)
    .withAllowTypeMatch(true)
    .withAllowInfixMods(true)
    .withAllowSymbolLiterals(false)
    .withAllowDependentFunctionTypes(true)
    .withAllowAllTypedPatterns(true)
    .withAllowAsForImportRename(true)
    .withAllowStarWildcardImport(true)
    ```

    Maybe we don't need a new parameter to choose which dialect.

### CODE REVIEW

File: scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala [hunk 1]
```diff
@@ -13,16 +13,10 @@ case class ParserConfig(
     inlineKeyword: Boolean = false
 ) {
 
-  private val dialect = scala.meta.dialects.Scala213.copy(
```
mlachkar, line 16: I don't see the point of writing our own definition of dialect. this one is not correct.
  implicit val Scala213 = Scala212
    .withAllowImplicitByNameParameters(true)
    .withAllowLiteralTypes(true)
    .withAllowNumericLiteralUnderscoreSeparators(true)
    .withAllowTryWithAnyExpr(true)

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
```diff
@@ -13,16 +13,10 @@ case class ParserConfig(
...
-    allowLiteralTypes = literalTypes,
```
mlachkar, line 17: already true in 2.12

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
```diff
@@ -13,16 +13,10 @@ case class ParserConfig(
...
-    allowTrailingCommas = trailingCommas,
```
mlachkar, line 18: already true in 2.12

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
```diff
@@ -13,16 +13,10 @@ case class ParserConfig(
...
-    allowInlineMods = inlineKeyword,
```
mlachkar, line 19: we set allowInlineMods to false, and it's true in scala3

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
```diff
@@ -13,16 +13,10 @@ case class ParserConfig(
...
-    allowInlineIdents = !inlineKeyword
```
mlachkar, line 20: allowInlineIdent is true since 2.10

tgodzik:
Hmm... this is true in most cases and I was also thinking about maybe doing this, but it might not be best to use it by default. Especially later on after some syntax will be deprecated then we will for sure need a new dialect. And some things are already unsupported for example:
```
  .withAllowViewBounds(false) // View bounds have been removed in Dotty
```

I would rather add another option to `.scalafix.conf`.

scalafix-core/src/main/scala/scala/meta/internal/scalafix/ScalafixScalametaHacks.scala
```diff
@@ -8,8 +8,8 @@ import scala.meta.internal.trees.Origin
 
 object ScalafixScalametaHacks {
   def dialect(language: String): Dialect =
-    if (language == "Scala") dialects.Scala212
-    else if (language.isEmpty) dialects.Scala212
+    if (language == "Scala") dialects.Scala3
```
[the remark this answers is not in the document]
bjaglin: Sounds right, it's under `internal` anyway

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
bjaglin: I am not sure either - does it make sense to keep `ParserConfig` at all?

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
```diff
@@ -38,14 +38,12 @@ trait ScalafixMetaconfigReaders {
     import ScalafixConfig.{DefaultDialect => Default}
     import scala.meta.dialects._
     ReaderUtil.oneOf[Dialect](
```
bjaglin, line 40: where is this `ConfDecoder`  used?

bjaglin (commented):
> Scala3 dialect is a superset of Scala213 dialect:

I guess this is a wrong assumption, so scalafix needs to know whether it's dealing with 2.x or 3 sources.


> I would rather add another option to .scalafix.conf.

I initially thought we should follow the scalafmt way (optional `runner.dialect` key), but scalafix is a bit different, because for semantic rules, the scala version is already mandatory, as a CLI/invocation argument.


So instead of adding a configuration key, we could make that argument mandatory also for syntactic rules. For build tool clients, it's an acceptable breaking change, but for CLI users like https://eed3si9n.com/syntactic-scalafix-rule-for-unified-slash-syntax (maybe a bad example as in this case we look at the extension to infer the dialect), it's quite annoying.

What about keeping the scala version optional for syntactic rules, but when missing try to pars with scala2.x dialect and fall back to scala3 on error (once per invocation no matter how many files we have)? Is there a case where both parsers would succeed but produce different trees?

tgodzik:
> What about keeping the scala version optional for syntactic rules, but when missing try to pars with scala2.x dialect and fall back to scala3 on error (once per invocation no matter how many files we have)? Is there a case where both parsers would succeed but produce different trees?

There are a couple, especially related to optional braces, but mostly if a code was weirdly formatted to start with.

```scala
object A{
  def b =
   12
   23
}
```

I think in case of Scala 3, 12 and 23 will belong to b, while for Scala 2 only 12 will.

tgodzik:
If possible I would try to declare the Scala version explicitly, but if it turns out to be more problematic for different clients, we could try with the fallback, but maybe issue a warning?

bjaglin:
Any thought on this correctness vs simplicity dilemma @olafurpg?

olafurpg:
I would lean towards requiring users to explicitly declare that the Scala 3 dialect should be used. As Tomasz pointed out, there are subtle risks from enabling full Scala 3 syntax by default. I would avoid heuristics and attempting an automatic fallback.  One thing we could do is to customize syntax error messages to recommend enabling the Scala 3 dialect to fix the problem.

scalafix-core/src/main/scala/scalafix/internal/config/ParserConfig.scala
olafurpg: IIRC, the motivation for the custom dialect was to allow as much as possible of Scala 3 syntax without introducing regressions for Scala 2 users. Some Scala 3 syntax can be enabled safely while other syntax like significant indent may change the parsed structure of Scala 2 programs.

mlachkar: Ah, thanks! I understand better! Clever idea.

mlachkar: Ah, thanks! I understand better! Clever idea.

mlachkar:
in scalafix.conf, we can already provide a `Dialect`. In this commit, I just make sure that it's used to parse files

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
```diff
@@ -34,33 +37,6 @@ object ScalafixConfig {
...
-    allowViewBounds = true,
-    // Are `with` intersection types supported by this dialect?
-    allowWithTypes = true,
-    // Are XML literals supported by this dialect?
-    // We plan to deprecate XML literal syntax, and some dialects
-    // might go ahead and drop support completely.
-    allowXmlLiterals = true
-  )
-
+  val DefaultDialect = scala.meta.dialects.Scala213
```
bjaglin, line 40: Following up on the previous discussion [above](https://github.com/scalacenter/scalafix/pull/1373/files#r614766745): this is a breaking change since currently some scala3 syntax can be parsed without tweaking the default. That's what we want but we should document that.

build.sbt
```diff
@@ -63,6 +62,7 @@ lazy val core = project
   .settings(
     moduleName := "scalafix-core",
     buildInfoSettingsForCore,
+    scalacOptions ++= syntheticsOn.value,
```
bjaglin, line 65: instead of adding it to every scala2 project, can't you keep it in `ThisBuild` and remove it for the test-input? that would reduce the noise

project/ScalafixBuild.scala
```diff
@@ -108,6 +115,12 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       "unit/test:runMain scalafix.tests.util.SaveExpect" ::
         s
     },
+    // just to launch first test in scala 3
+    commands += Command.command("ci-3") { s =>
+      s"""set testsInput/scalaVersion := "$scala3"""" ::
+        s"""set testsOutput/scalaVersion := "$scala3"""" ::
+        "unit/testOnly scalafix.tests.rule.RuleSuite" :: s
+    },
```
bjaglin, line 123: :+1:

scalafix-core/src/main/scala/scalafix/config/CustomMessage.scala
```diff
@@ -13,7 +13,9 @@ class CustomMessage[T](
 
 object CustomMessage {
   implicit val SymbolDecoder: ConfDecoder[CustomMessage[Symbol.Global]] =
-    decoder[Symbol.Global](field = "symbol")
+    decoder[Symbol.Global](field = "symbol")(
+      ScalafixMetaconfigReaders.symbolGlobalReader
+    )
```
bjaglin, line 18: is that a no-op?

scalafix-core/src/main/scala/scalafix/v1/SyntacticDocument.scala
```diff
@@ -27,19 +29,11 @@ final class SyntacticDocument private[scalafix] (
 }
 
 object SyntacticDocument {
-  def fromInput(input: Input): SyntacticDocument = {
```
bjaglin, line 30: deprecate instead of removing?

scalafix-tests/output/src/main/scala-3/test/HelloWorld.scala
```diff
@@ -0,0 +1,6 @@
+package test
+
+object HelloWorld:
+
+  @main def hello = println("Hello, world!")
+// This comment is added by Scalafix
```
bjaglin, line 6: :tada:

bjaglin (commented):
Getting there :+1: 

> in scalafix.conf, we can already provide a Dialect. In this commit, I just make sure that it's used to parse files.

Ah great if that's already available. Could you add a unit test to demonstrate this?

>  This commit dosn't add the possibilit to configure the dialect through the CLI.

For semantic rules, the scala version is mandatory in the CLI args - it should have priority since it means class files exist with that version, so if it does not match with the dialect in the configuration, should we proceed with that one but issue a warning?

scalafix-core/src/main/scala/scalafix/v1/SyntacticDocument.scala
mlachkar: ok

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
mlachkar: I can keep the same if needed. I just need to check which one was actually used.

scalafix-core/src/main/scala/scalafix/config/CustomMessage.scala
mlachkar: yes

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
mlachkar: so this default is based on 212 and it was only used in old rules (none of the ones we have in scalafix).
the real dialect that was used is this one, where all the additional configuration is already part of scala213.
```
  private val dialect = scala.meta.dialects.Scala213.copy(
    allowLiteralTypes = true,
    allowTrailingCommas = true,
    allowInlineMods = flase, 
    allowInlineIdents = true 
  )
  ```
  I don't know if it's useful to allow scala 3 syntax in scala213, but we could allow some of it.

**Comment by mlachkar** [comment]
Still missing the CLI:
We can deduce the dialect using the scalaVersion if present, which would override the dialect configured in .scalafix.conf.
or we can add the argument `dialect` in cli, and take it into account.

I don't know what's the best solution.

File: scalafix-core/src/main/scala/scalafix/config/CustomMessage.scala (hunk 11 above)
Comment (mlachkar) [reply scalafix-core/src/main/scala/scalafix/config/CustomMessage.scala:18]: the goal was just to know which implicit is used from ScalafixMetaconfigReaders

File: scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala [hunk 14]
```diff
@@ -121,6 +123,11 @@ case class Args(
       "The Scala compiler version that was used to compile this project."
     )
     scalaVersion: String = scala.util.Properties.versionNumberString,
+    @Description(
+      "default Scala213" +
+        "Possibilities: Scala3, Scala213"
+    )
```
mlachkar, line 129: Need to be improved

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
```diff
@@ -41,11 +41,8 @@ trait ScalafixMetaconfigReaders {
       Default,
       Scala211,
       Scala212,
-      Sbt0137,
-      Sbt1,
-      Dotty,
-      Paradise211,
-      Paradise212
+      Scala213,
+      Scala3
```
bjaglin, line 45: what happened to `Sbt1` ? I guess it could be useful if someone has a sbt-type file with a different extension than `.sbt` (very far-fetched)

scalafix-core/src/main/scala/scalafix/v1/SyntacticDocument.scala
```diff
@@ -27,19 +29,15 @@ final class SyntacticDocument private[scalafix] (
 }
 
 object SyntacticDocument {
+  @deprecated("use fromInput(input: Input, dialect: Dialect) instead", "0.9.27")
```
bjaglin, line 32:
```suggestion
  @deprecated("use fromInput(input: Input, dialect: Dialect) instead", "0.9.28")
```

build.sbt
```diff
@@ -1,14 +1,15 @@
 import Dependencies._
+import sbt.Keys.scalacOptions
 
 inThisBuild(
   List(
     onLoadMessage := s"Welcome to scalafix ${version.value}",
     scalaVersion := scala213,
     crossScalaVersions := List(scala213, scala212, scala211),
     fork := true,
+    scalacOptions ++= syntheticsOn.value,
```
bjaglin, line 10: I guess you can now revert this change and get rid of `syntheticsOn` since the scala3 customization is only applied where scala3 cross-building is enabled.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -121,6 +123,11 @@ case class Args(
       "The Scala compiler version that was used to compile this project."
     )
     scalaVersion: String = scala.util.Properties.versionNumberString,
+    @Description(
+      "default Scala213" +
```
bjaglin, line 127:
```suggestion
      "Defaults to Scala213 for `.scala` files and Sbt1 for `.sbt` files" +
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -121,6 +123,11 @@ case class Args(
...
+        "Possibilities: Scala3, Scala213"
```
bjaglin, line 128: doesn't metaconfig generate this out of the box?

project/ScalafixBuild.scala
```diff
@@ -108,6 +115,12 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       "unit/test:runMain scalafix.tests.util.SaveExpect" ::
         s
     },
+    // just to launch first test in scala 3
```
bjaglin, line 118: this is unlikely to be updated when we have more than one test
```suggestion
```

bjaglin (approved):
A few minor comments, great work!

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: Ah, I don't know

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixMetaconfigReaders.scala
mlachkar: even the old code don't really take `sbt` dialect into account.
sbt dialect is always 213 (which is the default) with allowTopLevelItems. 
and now sbtDialct is `val DefaultSbtDialect: Dialect = scala.meta.dialects.Sbt1`

  private val sbtDialect = dialect.copy(allowToplevelTerms = true)

  def dialectForFile(path: String): Dialect =
    if (path.endsWith(".sbt")) sbtDialect
    else dialect

So users are supposed to configure either scala213 or scala3 in dialect. Do you think we should restrict to thode two values? 
  implicit lazy val dialectReader: ConfDecoder[Dialect] = {
    import ScalafixConfig.{DefaultDialect => Default}
    import scala.meta.dialects._
    ReaderUtil.oneOf[Dialect](
      Scala213,
      Scala3,
      //Scala211
      //Scala212
    )
  }

bjaglin: > Do you think we should restrict to those two values?

If we do, then it would be `Scala2` and `Scala3`, no?

bjaglin: > By the way, I thought sbt1 is just sbt with a higher version 1.x.x
> Isn't that the case?

It is, the format and binary backward compatibility is frozen in 1.x

bjaglin: > even the old code don't really take sbt dialect into account.
> sbt dialect is always 213 (which is the default) with allowTopLevelItems.

Interesting! I think we should rather use the dialect defined in scalameta than defning our own variant here, no?
https://github.com/scalameta/scalameta/blob/6c0a370cc278d473588f973c436d8378a691d2b1/scalameta/dialects/shared/src/main/scala/scala/meta/dialects/package.scala#L70-L72

mlachkar: yes! It's what I did. I Like the idea of having only scala2 and Scala3 dialect !!

PR #1374, opened 2021-04-16
Update the build

mlachkar:


build.sbt
```diff
@@ -183,40 +150,38 @@ lazy val testkit = project
...
         .dependsOn(
           testsInput / Compile / compile,
-          testsOutput / Compile / compile
+          testsOutput / Compile / compile,
+          testsShared / Compile / compile
         )
         .value
     },
     Test / resourceGenerators += Def.task {
       // copy-pasted code from ScalafixTestkitPlugin to avoid cyclic dependencies between build and sbt-scalafix.
```
mlachkar, line 177: I don't understand why we use java.properties intead of BuildInfo here ?

bjaglin: Indeed, maybe here it would be possible to use `BuildInfo` to generate a `TestkitProperties` instance and load it from the test runner. But as a regular testkit user (rule author, using sbt or not), being forced to generate code does not seem very nice. So I guess the rationale for using java properies here is that supporting java properties is a key feature, and we might as well eat our own dog food and test it here?

bjaglin (commented):
Thanks for the cleanup!  Apart from the changes in `Dependencies.scala` which I found harder to read, LGTM. Just to confirmed: this is only pure refactoring and does not impact anything in the resulting artifacts right?

project/Dependencies.scala
```diff
@@ -4,39 +4,51 @@ import scala.util.{Properties, Try}
...
+  val coursierInterfaceV = "1.0.3"
+  val scalacheckV = "1.15.2"
+  val collectionCompatV = "2.4.3"
+  val java8CompatV = "0.9.0"
+  val nailgunV = "0.9.1"
+  val logbackV = "1.2.3"
+  val commontTextV = "1.9"
 
-  var testClasspath: String = "empty"
+  def jgit: ModuleID = "org.eclipse.jgit" % "org.eclipse.jgit" % jgitV
```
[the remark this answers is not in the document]
mlachkar: One of them had the type explicitly, so I did it for all of them. I agree it's better without

project/ScalafixBuild.scala
```diff
@@ -155,8 +161,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
           "supportedScalaVersions" -> supportedScalaVersions,
           "coursierVersion" -> coursierV
         )
-      ),
-      tut / fork := true
```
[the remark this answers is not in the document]
mlachkar: tut was before they moved to mdoc I think. 
I don't think we use it anymore

build.sbt
```diff
@@ -113,57 +92,45 @@ lazy val cli = project
...
-      warnUnusedImports.value,
-      "-Xlint"
-    ),
-    testsInputOutputSetting,
-    coverageEnabled := false,
-    libraryDependencies += bijectionCore
+    noPublishAndNoMima,
+    scalacOptions -= warnUnusedImports.value,
+    libraryDependencies ++= testsDeps,
+    coverageEnabled := false
```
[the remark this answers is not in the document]
mlachkar: I don't know. I just removed `-= -Xlint` because Xlint is definitely not present in the scalacOptions here.

build.sbt
mlachkar: ok

build.sbt
```diff
@@ -265,29 +230,31 @@ lazy val unit = project
...
+
+// force javac to fork by setting javaHome to get error messages during compilation,
+// see https://github.com/sbt/zinc/issues/520
+def inferJavaHome() = {
+  val home = file(sys.props("java.home"))
+  val actualHome =
+    if (System.getProperty("java.version").startsWith("1.8")) home.getParentFile
+    else home
+  Some(actualHome)
+}
```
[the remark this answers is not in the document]
mlachkar: ok

project/Dependencies.scala
```diff
@@ -4,39 +4,51 @@ import scala.util.{Properties, Try}
...
-  def scala210 = "2.10.7"
   def scala211 = "2.11.12"
   def scala212 = "2.12.13"
   def scala213 = "2.13.5"
-  def coursierV = "2.0.0-RC5-6"
-  def coursierInterfaceV = "1.0.3"
-  val currentScalaVersion = scala213
   // we support 3 last binary versions of scala212 and scala213
-  val testedPreviousScalaVersions =
+  def testedPreviousScalaVersions: Map[String, List[String]] =
```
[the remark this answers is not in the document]
mlachkar: no reason. I don't there is a reason either for:
```
 def scala211 = "2.11.12"
  def scala212 = "2.12.13"
  def scala213 = "2.13.5"
```

project/ScalafixBuild.scala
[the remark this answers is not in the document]
mlachkar: and I will remove the other tut/fork

build.sbt
```diff
@@ -62,12 +52,11 @@ lazy val core = project
   .in(file("scalafix-core"))
   .settings(
     moduleName := "scalafix-core",
-    buildInfoSettings,
+    buildInfoSettingsForCore,
     libraryDependencies ++= List(
       scalameta,
       googleDiff,
-      "com.geirsson" %% "metaconfig-typesafe-config" % metaconfigV,
-      "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided,
```
mlachkar, line 70: this lib is removed. but I am not sure of this choice..

build.sbt
```diff
@@ -97,11 +80,7 @@ lazy val reflect = project
   .in(file("scalafix-reflect"))
   .settings(
     moduleName := "scalafix-reflect",
-    isFullCrossVersion,
-    libraryDependencies ++= Seq(
-      "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-      "org.scala-lang" % "scala-reflect" % scalaVersion.value
```
mlachkar, line 103: Those two also are removed. Not sure if it's better to hide the fact that we depend on the compiler.

mlachkar:
Yes, it's just refactoring. It's not supposed to impact anything. 
The change I am not sure about is: When I remove some libdependencies (scala-reflect and scala-conpiler) from core and reflect module. 

I also removed now all the old website configuration.

build.sbt
bjaglin: In that case, since scalafix-reflect is full cross-version we wouldn't want to end up with wrong patch version brought transitively, it would be interesting to keep them. although https://github.com/scalacenter/scalafix/blob/48b201b89fa11edb7496dd923e188cad1b490184/build.sbt#L89 effectively ensure that the patch versions are aligned...

bjaglin (commented):
Regarding the removal of libs: I think it would better to keep the dependencies explicit, rather than relying on what's brough transitively. Actually, what about using https://github.com/cb372/sbt-explicit-dependencies to enforce this rule? The classpath with binary/full version is complex, so I think it would help making sure we declare what's needed.

mlachkar:
ok. let's keep the, and maybe add also the explicit lib plugin. thanks

mlachkar:
I m reverting the last change using the sbt-plugin sbt-explicit-dependencies! the result of the command is not always correct.

project/Dependencies.scala
```diff
@@ -4,40 +4,59 @@ import scala.util.{Properties, Try}
...
+  val commonText = "org.apache.commons" % "commons-text" % commontTextV
+  val coursier = "io.get-coursier" %% "coursier" % coursierV
+  val coursierInterfaces = "io.get-coursier" % "interface" % coursierInterfaceV
+  val googleDiff = "com.googlecode.java-diff-utils" % "diffutils" % googleDiffV
+  val jgit = "org.eclipse.jgit" % "org.eclipse.jgit" % jgitV
+  val metaconfigCore = "com.geirsson" %% "metaconfig-core" % metaconfigV
+  val metaconfigTypesageConfig = "com.geirsson" %% "metaconfig-typesafe-config" % metaconfigV
+  val nailgunServer = "com.martiansoftware" % "nailgun-server" % nailgunV
+  val scalaXml = "org.scala-lang.modules" %% "scala-xml" % scalaXmlV
+  val scalacheck = "org.scalacheck" %% "scalacheck" % scalacheckV
```
bjaglin, line 40: just looking at it on github, this looks unused?

project/ScalafixBuild.scala
```diff
@@ -38,12 +29,23 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
-    lazy val isScala212 = Def.setting { scalaVersion.value.startsWith("2.12") }
+    lazy val isScala213 = Def.setting {
+      scalaVersion.value.startsWith("2.13")
+    }
+    lazy val isScala212 = Def.setting {
+      scalaVersion.value.startsWith("2.12")
+    }
+    lazy val isScala211 = Def.setting {
+      scalaVersion.value.startsWith("2.11")
+    }
```
bjaglin, line 40: unused?

build.sbt
```diff
@@ -265,27 +254,26 @@ lazy val unit = project
...
-    noPublish,
     moduleName := "scalafix-docs",
     scalaVersion := scala213,
     scalacOptions += "-Wconf:msg='match may not be exhaustive':s", // silence exhaustive pattern matching warning for documentation
     scalacOptions += "-Xfatal-warnings",
     mdoc := (Compile / run).evaluated,
     crossScalaVersions := List(scala213),
     libraryDependencies ++= List(
-      "com.geirsson" %% "metaconfig-docs" % metaconfigV
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value,
```
bjaglin, line 271: is the compiler used to generate docs?

bjaglin (approved):
> I m reverting the last change using the sbt-plugin sbt-explicit-dependencies! the result of the command is not always correct.

OK, thanks for trying! I guess that's what caused the updated for scalameta/scala-compiler/scala-reflect/etc? I wonder if it wouldn't be better to leave them as they were, since I don't think there was a specific problem with them? 

In any case, thanks for the cleanup!

build.sbt
mlachkar: it looks like

PR #1375, opened 2021-04-20
add triggered config description to docs

taisukeoe:
ref: https://github.com/scalacenter/scalafix/pull/1217 https://github.com/scalacenter/sbt-scalafix/pull/169

This is a follow up PR to introduce `--triggered` args and `triggered` configuration prefix to run scalafix on compilation with different config.

### DIFF
--- a/docs/users/configuration.md
+++ b/docs/users/configuration.md
@@ -34,8 +34,15 @@ DisableSyntax.noFinalize = true
 
 ## Triggered configuration
 
-Overlay rules and rule specific configuration for scalafix on compilation, which is enabled by `scalafixOnCompile := true`.
-Every configuration can be overlaid by `triggered` prefix. With the following configuration, explicit scalafix invocation will run with `[ DisableSyntax, RemoveUnused ]` rules, while scalafix on compilation will run with `[ DisableSyntax ]` rule.
+When Scalafix is triggered automatically (`scalafixOnCompile := true`), the
+values for keys prefixed with `triggered` takes precedence over the default
+configuration - both for rules and rule-specific configuration. You can think
+of `triggered` as a optional overlay, on top of the default configuration.
+  
+For example, with the following configuration, explicit scalafix invocation
+will run `DisableSyntax` & `RemoveUnused`, while triggered invocations
+will only run `DisableSyntax`. In both cases, the `noFinalize` configuration
+is enabled.
 
 ```scala
 // .scalafix.conf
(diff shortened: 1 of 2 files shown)


docs/users/configuration.md
```diff
@@ -31,3 +31,22 @@ rules = [
 ]
 DisableSyntax.noFinalize = true
 ```
+
+## Triggered configuration
+
+Overlay rules and rule specific configuration for scalafix on compilation, which is enabled by `scalafixOnCompile := true`.
+Every configuration can be overlaid by `triggered` prefix. With the following configuration, explicit scalafix invocation will run with `[ DisableSyntax, RemoveUnused ]` rules, while scalafix on compilation will run with `[ DisableSyntax ]` rule.
```
bjaglin, line 38: I tried to rephrase to highlight the "triggered" concept/naming that we came up with in the original PR.

```suggestion
When Scalafix is triggered automatically (`scalafixOnCompile := true`), the
values for keys prefixed with `triggered` takes precedence over the default
configuration - both for rules and rule-specific configuration. You can think
of `triggered` as a optional overlay, on top of the default configuration.
  
For example, with the following configuration, explicit scalafix invocation
will run `DisableSyntax` & `RemoveUnused`, while triggered invocations
will only run `DisableSyntax`. In both cases, the `noFinalize` configuration
is enabled.
```

bjaglin (commented):
Thanks for the PR! Please see my suggestions above. @mlachkar do you have some feedback? you are probably less biased as you didn't follow the previous discussions.

taisukeoe:
Thank you for reviewing and suggestion! I have reflected both 👍

docs/users/installation.md
```diff
@@ -231,19 +231,22 @@ feature with care as it has several shortcomings, for example:
...
+   after a simple `test`. 
+   To make such invocations less intrusive, you can change the rules and rules
+   configuration used in that case by defining in `.scalafix.conf`
+   [custom values for them](configuration.md#triggered-configuration).
 1. If you run many semantic rules by default, the last one(s) to run might see
    stale information and fail the invocation, which needs to be re-run manually.
    This is [not specific to `scalafixOnCompile`](https://github.com/scalacenter/scalafix/issues/1204),
    but the problem becomes much more visible with it.
-1. To keep the overhad minimal, `scalafixCaching` is automatically enabled when
+1. To keep the overhead minimal, `scalafixCaching` is automatically enabled when
```
bjaglin, line 242: :+1:

bjaglin (approved):

mlachkar (approved):

mlachkar:
The documentation is super clear! thanks for the PR.

Issue #1236, opened 2020-09-10
Add a method in scalafixApi that takes a semanticdbDocument


mlachkar:
For metals, we would like to be able to organize import on file that it's not saved yet. We can through metals generates an interactive semantic db file, that is not persisted. We would like to pass this textDocument as a parameter to the `evaluate method`

bjaglin:
I am wondering if it wouldn't be easier that metals save a temporary file? I haven't looked at the code, but I assume supporting in-memory SemanticDB info would require quite a few changes.

tgodzik:
@bjaglin That might be another way to go around it. We could maybe have a separate source root inside `.metals` for a scenario like that.

tgodzik:
I started looking into using temporary files, but that would require both temporary file within a proper sourceroot hierarchy and a semanticdb file to accompany it. That essentially would mean keeping a copy of the workspace somewhere, which seems like it might not be the perfect solution.

However, since scalafix already can compile semanticdb files if none exist I wonder if we could do two things:
1. Add an ability to run scalafix on a text provided via API instead of reading it from the filesystem
2. Compile semanticdb in case md5 is too old, which would help with anything that was not yet saved to the filesystem

In case proper semanticdb file exists, everything would work as usual.  2 should be simple to do in the existing code, from what I checked, while 1 might require a bit of consideration.

What do you all think ?

bjaglin:
I had a look at the code, and (1)+(2) does look doable indeed. One thing I wonder though, in cases where (2) would kick-in, is whether the current lookup order in https://github.com/scalacenter/scalafix/blob/master/scalafix-core/src/main/scala/scalafix/internal/v1/InternalSemanticDoc.scala (first a global symbol table loaded from the potentially stale class files, with a fallback on a potentially "fresh" on-demand `TextDocument`) would still make sense? I don't know if that corner case could effectively cause issues with `OrganizeImports`, but I guess it could in theory.

olafurpg:
One way to pass a TextDocument is as an array of bytes. Encoding/decoding is cheap. With such an API we could also pass in the symbol table so that the client is responsible for keeping it up to date

mlachkar:
For me solution (2) is not that simple: since we only provide a path to scalafix, if the file is not saved, the content read will be the old one, not the unsaved one. The only case where we get Stale semantic db file, is when the file is saved but the semanticdb is not yet replace by the newst one.

tgodzik:
[...]

PR #1376, opened 2021-04-21
Don't generate semanticdb files if there are missing 

mlachkar:
- instead we report an error of semanticdb files missing

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -13,10 +13,6 @@ import java.util.regex.Pattern
 import java.util.regex.PatternSyntaxException
 
 import scala.annotation.StaticAnnotation
-import scala.tools.nsc.CompilerCommand
-import scala.tools.nsc.Settings
-import scala.tools.nsc.interactive.Global
-import scala.tools.nsc.reporters.StoreReporter
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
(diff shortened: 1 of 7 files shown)


tgodzik:
In this case, would it be possible to be able to provide semanticdb via `Array[Byte]` in the API? Not neccessarily in this PR and I know we discussed it previously, but this would be amazing to have as it would enable us to make it work in a bunch of places.

mlachkar:
We need to think and design this solution. Metals uses scalafix to fix only one file, by the api takes multiple files and in this case how do we differentiate between semanticdb files.
Maybe we need a new api that fixes only one file and in this case we can provide the semantic file.

tgodzik:
> We need to think and design this solution. Metals uses scalafix to fix only one file, by the api takes multiple files and in this case how do we differentiate between semanticdb files.
> Maybe we need a new api that fixes only one file and in this case we can provide the semantic file.

Maybe we could have an additional Map[String, Array[Byte] ? Or Path instead of string, I am not sure exactly what is currently used :sweat_smile:

bjaglin (approved):
Limiting the usage of the compiler to advanced semantic rules (for the presentation compiler) and for scalafix-reflect to compile source rule is probably a step in the right direction.

As already discussed in https://github.com/scalacenter/scalafix/pull/1348#pullrequestreview-604155677, this ondemand, transparent compilation causes a lot of confusion when the compiler kicks in in 2.12 (default loaded by most clients) against 2.13 classfiles (we had a few gitter questions / github tickets about that), so it's doing more harm than good in my opinion. When it works, it's so silent that the user might not realize there is something wrong with its semanticdb setup and could end up blaming scalafix for the very long run time.

bjaglin:
> > We need to think and design this solution. Metals uses scalafix to fix only one file, by the api takes multiple files and in this case how do we differentiate between semanticdb files.
> > Maybe we need a new api that fixes only one file and in this case we can provide the semantic file.
> 
> Maybe we could have an additional Map[String, Array[Byte] ? Or Path instead of string, I am not sure exactly what is currently used sweat_smile

Makes sense, maybe let's follow up on https://github.com/scalacenter/scalafix/issues/1236 for this?

PR #1377, opened 2021-04-22
Remove partly the deprecated api

mlachkar:
We remove here
- LegacyRule
- LegacyRuleCtx. // replaced by `RuleCtxImpl`. I had to modify it to keep the same behavior that we had with LegacyRuleCtx. Is it really Legacy ? 
- LegacySemanticRule
- LegacySyntacticRule
- v0.Rule

Those changes have an impact on:
- AbstractSyntacticRuleSuite.scala: It was still using the old api v0.Rule. This abstract class is not used externally on any github project. We may remove the the deprecated class SyntacticRuleSuite also
- Some test rules: Some internal test rules were using the old api. I fixed that to use the new Rule api.
- PatchInternals: One method was still using the LegacyRuleCtx. I fixed it to use the RuleCtxImpl
- RuleCtxImpl: I change its argument to take doc instead of tree, to make the compute of tree lazy, as it was in LegacyRuleCtx
- RuleDecoderOps.scala: toRule method was classloading both v0.Rule and v1.Rule. Now it will throw an exception if it's v0.Rule

**After finishing this change, I figure out that v0.Rule wasn't deprecated, and there is on github about ten rules that use this API. What do we want to do with the old api? What's considered as an old api? 
All feedbacks are welcomed!**


olafurpg:
I think it's OK to remove the `v0` at this point if it's causing problems. The point of namespacing the API under v0 was to make it possible to support the old and new APIs in parallel. At the time the v1 package was introduced, there was a non-trivial number of public and private rules that were gaining traction and I didn't want to entirely kill that momentum.

olafurpg:
I admit the v0/v1 separation makes the codebase difficult to understand. It took many iterations to get to the current design of Scalafix and every incremental improvement was pretty painful.

mlachkar:
If I understand correctly: Not everything has a v0 and v1 versions? 
For example, LintCategory is only under v0.
There is also some datastructure marked as Legacy, but there are still used like `LegacyRuleCtx`.


For the public API, we can maybe mark as deprecated all the things we want to delete definitely, help users move their rules to use the new api, and delete it.

Changing the dialect makes me realize that there are too many places where it's used/called, and that we can maybe reduce this complexity by deleting the old API and fixing internally the part where we still used the old api.

bjaglin:
> After finishing this change, I figure out that v0.Rule wasn't deprecated, and there is on github about ten rules that use this API. What do we want to do with the old api? What's considered as an old api?

I am not familiar with the differences, but what about a scalafix rule? ;)

mlachkar:
why not!

PR #1388, opened 2021-05-04
Make dialect not an option

mlachkar:
I think having a default value is better than None.

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -56,10 +56,6 @@ case class Args(
     @ExtraName("remainingArgs")
     @ExtraName("f")
     files: List[AbsolutePath] = Nil,
-    @Description(
-      "default Scala2" +
-        "Possibilities: Scala2, Scala3"
-    )
     dialect: Dialect = ScalafixConfig.Scala2,
     @Description(
       "File path to a .scalafix.conf configuration file. " +
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -60,7 +60,7 @@ case class Args(
       "default Scala2" +
         "Possibilities: Scala2, Scala3"
     )
-    dialect: Option[Dialect] = None,
+    dialect: Dialect = ScalafixConfig.Scala2,
     @Description(
       "File path to a .scalafix.conf configuration file. " +
         "Defaults to .scalafix.conf in the current working directory, if any."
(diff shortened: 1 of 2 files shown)


scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -60,7 +60,7 @@ case class Args(
       "default Scala2" +
```
bjaglin, line 60: I guess this is no longer useful since the default shows up in the help message?

bjaglin (approved):
I thought `None` was useful to let the Sbt1/Scala switch based on the file extension happen after, but it looks like it happens anyway, so LGTM. We could add in the doc of the dialect that it is the dialect for parsing *`.scala` files*. If there is one day a `Sbt2` dialect, we could add a `sbt-dialect` option to control it, as it's currently harcoded.

mlachkar:
I think we will have issues if the code base has both dialects... scalafmt has a system of overrides for paths:
```
fileOverride {
  "glob:**/scala-3*/**" {
    runner.dialect = scala3
    rewrite.scala3.convertToNewSyntax = true
    rewrite.rules = [RedundantBraces]
    rewrite.scala3.removeOptionalBraces = yes
  }
}
```
I don't if we need to deal with this case right now or not ?

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: ok

bjaglin:
> I think we will have issues if the code base has both dialects...

I assume https://github.com/scalacenter/sbt-scalafix/blob/4ae45a12921d44f08f6627832aa851e4dc14172e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L660 handles that - sources that do not match the `scalaVersion` (and tus the dialect) won't be passed.

I wonder if that `scala-xx` convention also applies to other build tools though, in which case https://github.com/scalacenter/scalafix/blob/bf0df227c9b72fd6a9466f3a6fe11a8ce5f14efd/scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala#L21-L23 could be a bit smarter?

mlachkar:
> I assume https://github.com/scalacenter/sbt-scalafix/blob/4ae45a12921d44f08f6627832aa851e4dc14172e/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L660 handles that - sources that do not match the `scalaVersion` (and tus the dialect) won't be passed.

I think so too. 

I was more thinking about the case where you have two subproject or modules in your project definition, and one is in scala 2 and the other in scala 3. Scalafix configuration file is not per module, we can only have one.

bjaglin:
> I was more thinking about the case where you have two subproject or modules in your project definition, and one is in scala 2 and the other in scala 3. Scalafix configuration file is not per module, we can only have one.

In that case there is a cross-build, but you can only have one scalaVersion at the time per project, so Zinc/Scalafix only see the corresponding sources.

PR #1389, opened 2021-05-04
Configure correctly semanticdb targetroot depending on the dialect

mlachkar:
it makes me release that sourceroot need also to be configured differently if the dialect Scala3.

There are two commits in this PR. The second is a proposition to make the targetroot configurable through the api

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -349,7 +349,7 @@ case class Args(
   def validatedClasspath: Classpath = {
     val targetrootClasspath = semanticdbTargetroot
       .map(_.toString())
-      .orElse(targetrootFromScalacOptions())
+      .orElse(semanticdbOption("targetroot", Some("-semanticdb-target")))
       .map(option => Classpath(option))
       .getOrElse(Classpath(Nil))
     val baseClasspath =
@@ -367,28 +367,30 @@ case class Args(
   def classLoader: ClassLoader =
     ClasspathOps.toOrphanClassLoader(validatedClasspath)
 
-  def semanticdbOption(name: String): Option[String] = {
-    val flag = s"-P:semanticdb:$name:"
-    scalacOptions
-      .filter(_.startsWith(flag))
-      .lastOption
-      .map(_.stripPrefix(flag))
-  }
-
-  def targetrootFromScalacOptions(): Option[String] =
-    dialect match {
-      case ScalafixConfig.Scala2 => semanticdbOption("targetroot")
-      case ScalafixConfig.Scala3 =>
+  private def semanticdbOption(
+      settingInScala2: String,
+      settingInScala3Opt: Option[String]
+  ): Option[String] = {
+    if (scalaVersion.isEmpty || scalaVersion.startsWith("2")) {
+      val flag = s"-P:semanticdb:$settingInScala2:"
+      scalacOptions
+        .filter(_.startsWith(flag))
+        .lastOption
+        .map(_.stripPrefix(flag))
+    } else {
+      settingInScala3Opt.flatMap { settingInScala3 =>
         scalacOptions
           .sliding(2)
           .collectFirst {
-            case h :: last :: Nil if h == "-semanticdb-target" => last
+            case h :: last :: Nil if h == settingInScala3 => last
           }
+      }
     }
+  }
 
   def semanticdbFilterMatcher: FilterMatcher = {
-    val include = semanticdbOption("include")
-    val exclude = semanticdbOption("exclude")
+    val include = semanticdbOption("include", None)
+    val exclude = semanticdbOption("exclude", None)
     (include, exclude) match {
       case (None, None) => FilterMatcher.matchEverything
       case (Some(in), None) => FilterMatcher.include(in)
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -112,9 +112,9 @@ case class Args(
     sourceroot: Option[AbsolutePath] = None,
     @Description(
       "Absolute path passed to semanticdb with -P:semanticdb:targetroot:<path>. " +
-        "Used to locate semanticdb files. Default, Scalafix will try to locate semanticdb files under the classpath"
+        "Used to locate semanticdb files. By default, Scalafix will try to locate semanticdb files in the classpath"
     )
-    targetroot: Option[AbsolutePath] = None,
+    semanticdbTargetroot: Option[AbsolutePath] = None,
     @Description(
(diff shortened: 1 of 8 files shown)


scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -109,6 +110,11 @@ case class Args(
         "sourceroot. Defaults to current working directory if not provided."
     )
     sourceroot: Option[AbsolutePath] = None,
+    @Description(
+      "Absolute path passed to semanticdb with -P:semanticdb:targetroot:<path>. " +
+        "Used to locate semanticdb files. Default, Scalafix will try to locate semanticdb files under the classpath"
```
bjaglin, line 115:
```suggestion
        "Used to locate semanticdb files. By default, Scalafix will try to locate semanticdb files in the classpath"
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -109,6 +110,11 @@ case class Args(
...
+    )
+    targetroot: Option[AbsolutePath] = None,
```
bjaglin, line 117: targetroot makes sense as a scalac compiler option, but here it's an input, so I think it should be more explicit
```suggestion
    semanticdbRoot: Option[AbsolutePath] = None,
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -144,6 +144,12 @@ ScalafixArguments withToolClasspath(
      */
     ScalafixArguments withSourceroot(Path path);
 
+    /**
+     * @param path The SemanticDB targetroot path passed via --targetroot. Must match <code>path</code>
+     *             in <code>-Xplugin:semanticdb:targetroot:{path}</path></code> if used.
+     */
+    ScalafixArguments withTargetroot(Path path);
```
bjaglin, line 151: see comment about naming

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -367,6 +375,18 @@ case class Args(
...
 
+  def semanticTargetRoot(): Option[String] = {
+    dialect match {
+      case ScalafixConfig.Scala2 => semanticdbOption("targetroot")
+      case ScalafixConfig.Scala3 =>
+        val indexOpt = scalacOptions.zipWithIndex
+          .filter { case (setting, _) => setting == "-semanticdb-target" }
+          .lastOption
+          .map(_._2)
+        indexOpt.flatMap(i => Option(scalacOptions(i + 1)))
```
bjaglin, line 386: can't you use `sliding` to make this more readable and less error-prone?

scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
```diff
@@ -37,5 +37,6 @@ object ScalafixConfig {
     decoder(default)
 
   val Scala2: Dialect = scala.meta.dialects.Scala213
+  val Scala3: Dialect = scala.meta.dialects.Scala3
```
bjaglin, line 40: :+1:

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: or semanticTargetroot?
since we also have sourceroot which also specific to semanticdb options.

bjaglin: Apologies, I didn't see that one! Interestingly, its default does not seem to be infered by the scalacOptions. I guess for symmetry, it makes sense to keep `targetRoot` then, even though I find both options a bit hard to understand for someone not familiar with scalameta. Resolving this

mlachkar: I think it's better to name semanticdbTargetroot. I always had difficulties to understand sourceroot too.

mlachkar: and I agree, it's weird we don't configure semanticdbSourceroot from scalacOptions.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -367,6 +375,17 @@ case class Args(
       .map(_.stripPrefix(flag))
   }
 
+  def targetrootFromScalacOptions(): Option[String] =
+    dialect match {
```
bjaglin, line 379: dialect should be used for parsing source only, I think we should switch on the `scalaVersion` instead

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -367,6 +375,17 @@ case class Args(
...
+      case ScalafixConfig.Scala2 => semanticdbOption("targetroot")
```
bjaglin, line 380: `semanticdbOption()` is used in other places - are the other usages valid in scala3? maybe the scala2/scala3 different mappings should be abstracted inside it?

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: ok

mlachkar: I agree with you on this, but the scalaVersion can be empty and it's a little annoying.

mlachkar: actually if we have the scalaVersion, the dialect is easy to deduce.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: you would prefer something like that ? 
  def semanticdbOption2(settingInScala2: String, settingInScala3: String): Option[String] = {
    if (scalaVersion.isEmpty || scalaVersion.startsWith("2")) {
      val flag = s"-P:semanticdb:$settingInScala2:"
      scalacOptions
        .filter(_.startsWith(flag))
        .lastOption
        .map(_.stripPrefix(flag))
    } else {
      scalacOptions
        .sliding(2)
        .collectFirst {
          case h :: last :: Nil if h == settingInScala3 => last
        }
    }
  }

bjaglin: yes! that would force any client of `semanticdbOption` to specify the scala2/scala3 flags (or maybe there is a pattern to derive one from the other?)

mlachkar: Not really!

bjaglin (approved):

PR #1392, opened 2021-05-07
remove dialect and use scalaVersion

mlachkar:


scalafix-tests/unit/src/test/scala/scalafix/tests/cli/CliSemanticSuite.scala
```diff
@@ -246,29 +246,29 @@ class CliSemanticSuite extends BaseCliSuite {
...
-      "--dialect",
-      "unknown"
-    ),
-    expectedExit = ExitStatus.CommandLineError,
-    outputAssert = { out =>
-      assert(out.contains("Scala3"))
-      assert(out.contains("Scala2"))
-    }
-  )
+//  checkSemantic(
```
bjaglin, line 249: remove?

scalafix-tests/unit/src/test/scala/scalafix/tests/config/ScalaVersionSuite.scala
```diff
@@ -0,0 +1,12 @@
+package scalafix.tests.config
+
+import scalafix.internal.config.ScalaVersion
+
+class ScalaVersionSuite extends munit.FunSuite {
+
+  test("parser Scala2 ") {
+    val scala2 = ScalaVersion.from("2")
```
bjaglin, line 8: it looks like you added support for many more patterns (I am guessing the RC suffix is gracefully dropped for example), it would be nice to document/test them.

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
```diff
@@ -0,0 +1,85 @@
...
+      case BasicVersion(major, minor, patch) =>
+        MajorVersion.from(major.toLong).flatMap{ major =>
+          Success(Patch(major, minor.toInt, patch.toInt))
+        }
+      case PartialVersion(major, minor) =>
+        MajorVersion.from(major.toLong).flatMap{ major =>
+          Success(Minor(major, minor.toInt))
+        }
+      case MajorPattern(major) => MajorVersion.from(major.toLong).flatMap(major => Success(Major(major)))
+      case _ => Failure(new Exception(s"$s not a valid ScalaVersion."))
```
bjaglin, line 82:
```suggestion
      case _ => Failure(new Exception(s"$s not a valid Scala version"))
```

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  private val longPattern = """\d{1,19}"""
+  private val BasicVersion = raw"""($longPattern)\.($longPattern)\.($longPattern)""".r
+  private val MajorPattern = raw"""($longPattern)""".r
+  private val PartialVersion = raw"""($longPattern)\.($longPattern)""".r
+
+  def from(s: String): Try[ScalaVersion] = {
+    val version = s.split("-").head
+    version match {
+      case "Scala2" | "scala2" => Success(scala2) // not sure
```
bjaglin, line 71: not sure about supporting this syntax? Given the documentation of the flag, it seems a bit strange to me that the user would provide something else than `x[.x[.x]]`. Maybe we can strip the potential heading `[S|s]cala` before parsing?

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
```diff
@@ -0,0 +1,85 @@
...
+   object MajorVersion {
+     def from(int: Long): Try[MajorVersion] =
+       int match {
+         case 2 => Success(Scala2)
+         case 3 => Success(Scala3)
+         case _ => Failure(new Exception(s"Major Scala version can be either 2 or 3 not $int"))
+       }
+   }
+
+  private val longPattern = """\d{1,19}"""
```
bjaglin, line 63: I am guessing we can stick to 2? also you capture ints here, not longs

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
```diff
@@ -0,0 +1,85 @@
...
+  private val BasicVersion = raw"""($longPattern)\.($longPattern)\.($longPattern)""".r
```
bjaglin, line 64:
```suggestion
  private val FullVersion = raw"""($longPattern)\.($longPattern)\.($longPattern)""".r
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -129,7 +127,7 @@ case class Args(
     @Description(
       "The Scala compiler version that was used to compile this project."
```
bjaglin, line 128: Since we are extending the use of this flag for syntactic rules, I think we should relax a bit the constraint of "prior compilation". Also, the reference to `project` is a bit ambiguous, considering the other flags, I think `files` would be more consistent.
```suggestion
      "The Scala compiler version that the provided files are targeting, or the actual version that was used to compile them when a classpath is provided."
```

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -371,7 +368,7 @@ case class Args(
       settingInScala2: String,
       settingInScala3Opt: Option[String]
   ): Option[String] = {
-    if (scalaVersion.isEmpty || scalaVersion.startsWith("2")) {
+    if (scalaVersion.value.isEmpty || scalaVersion.value.startsWith("2")) {
```
bjaglin, line 371: add helper on `ScalaVersion`? Or `Ordering`?

bjaglin (approved):
LGTM on using `scalaVersion` to control the internal dialect passed to scalameta for parsing. A few docs & implementation details comments.

As discussed offline, for the record:
- this flag is not mandatory at the moment - it's only required for rules using the presentation compiler (`ExplicitResultTypes` amongst the built-in rules) to ensure compatibility at the class/pickle level
- it would be redundant and confusing for the user to ask for another flag just for the dialect
- sbt-scalafix [only passes it for semantic rules at the moment](https://github.com/scalacenter/sbt-scalafix/blob/33ef9de1210e95de73363f358c105ee14b5558c1/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L432), but it's trivial to update it
- the current default is the runtime Scala version of Scalafix, so effectively 2.12 or 2.13, which does not change the default behavior for parsing sources 
- :+1: for supporting a major version only (single int), to cover the use case of dialect selection via the CLI (or potentially in the configuration file itself) with as little provided information as possible (`--scala-version 3`).

scalafix-tests/unit/src/test/scala/scalafix/tests/config/ScalaVersionSuite.scala
mlachkar: I agree.

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
mlachkar: I think we can just not support this. I was

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
mlachkar: you're right

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
mlachkar: yes definitely. It was just to make everything compiles

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
```diff
@@ -200,8 +194,16 @@ final case class ScalafixArgumentsImpl(args: Args = Args.default)
...
-  override def withScalaVersion(version: String): ScalafixArguments =
-    copy(args = args.copy(scalaVersion = version))
+  override def withScalaVersion(version: String): ScalafixArguments = {
+    ScalaVersion
+      .from(version) match {
+      case Success(value) => copy(args = args.copy(scalaVersion = value))
+      case Failure(exception) =>
+        throw new ScalafixMainArgsException(
+          s"Failed when parsing Scala Version because: ${exception.getMessage}"
+        )
```
bjaglin, line 204: why not use the constructor that wraps an exception? The `because` is implied then.
```suggestion
        throw new ScalafixMainArgsException("Failed to parse Scala version", exception)
```

scalafix-core/src/main/scala/scalafix/v1/SyntacticDocument.scala
```diff
@@ -29,15 +30,27 @@ final class SyntacticDocument private[scalafix] (
...
   }
-  def fromInput(input: Input, dialect: Dialect): SyntacticDocument = {
+  @deprecated(
+    "use fromInput(input: Input, scalaVersion: ScalaVersion) instead",
+    "0.9.28"
+  )
+  def fromInput(inout: Input, dialect: Dialect): SyntacticDocument = {
+    val tree = parse(inout, dialect).get: Tree
+    fromTree(tree)
+  }
```
bjaglin, line 47: you can remove this since it was never published, no?

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
```diff
@@ -169,6 +164,7 @@ ScalafixArguments withToolClasspath(
      *                this must match the binary version available in the classloader of
      *                this instance, as requested/provided in the static factory methods
      *                of {@link Scalafix}.
+     * @throws ScalafixException In case of an error parsing the scalaVersion.
```
bjaglin, line 167: the javadoc could be updated in the same way the CLI doc was updated

bjaglin (approved):

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: ok

scalafix-core/src/main/scala/scalafix/v1/SyntacticDocument.scala
mlachkar: This method was there too!

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixArguments.java
mlachkar: the cli sentence is finally not that clear for me!
```
    @Description(
      "The Scala compiler version that the provided files are targeting, " +
        "or the actual version that was used to compile them when a classpath is provided."
    )
```
what we want is 
- the full scalaverion that was used to compile the code source
- in the absence of the fullVersion, we require either the scala binary version of the code source, or the major scala Version

mlachkar: we need at least the major, to know with which dialect to parse the code source, but some rules may require more..

mlachkar: ok!

bjaglin:
FTR, I just ran into https://github.com/scalameta/scalameta/issues/2323, so I think whatever the outcome of that discussion is, it's good that we hide the dialect and infer it based on `scalaVersion` (and potentially `scalacOptions`).

mlachkar:
I agree with you!

PR #1176, opened 2020-06-23
make it possible to use testkit with ScalaTest 3.2.x

bjaglin:
Fix https://github.com/scalacenter/scalafix/issues/1172.

I ran MiMa locally against 0.9.17 and ~[will open a separate PR to re-introduce it](https://gitter.im/scalacenter/scalafix?at=5ef22c8e6c06cd1bf44d5c9d)~ https://github.com/scalacenter/scalafix/pull/1177.

* Deprecate existing `*RuleSuite` as they cause runtime failure when used in a classpath where ScalaTest 3.0.x gets evicted by 3.2.x
* testkit clients must now mix-in themselves an implementation of `TestRegistration` (as this trait moved between 3.0.x & 3.2.x) like `[Any]FunSuiteLike`
* Run https://github.com/scalatest/autofix/tree/master/3.1.x on `unit`

PR #1395, opened 2021-05-12
Add a migration rule to help migration to 0.9.28

mlachkar:
I find it quite verbose to add all of this in order to keep migration rules.
Maybe we can just create a gist and don't add it at all to the repo?


migration/rules/src/main/scala/fix/v0_9_28.scala
```diff
@@ -0,0 +1,22 @@
...
+class v0_9_28 extends SemanticRule("v0_9_28") {
+  override def fix(implicit doc: SemanticDocument): Patch =
+    doc.tree.collect {
+      case t @ Importee.Name(Name("SemanticRuleSuite")) =>
+        Patch.removeImportee(t) +
+          Patch.addGlobalImport(
+            importer"scalafix.testkit.AbstractSemanticRuleSuite"
+          )
+      case t @ Init(_, _, _) =>
+        if (t.toString() == "SemanticRuleSuite()")
```
bjaglin, line 16: use semanticdb information through [`SymbolMatcher`](https://scalacenter.github.io/scalafix/docs/developers/symbol-matcher.html) rather than relying on representation (the empty parameters list being optional, this will yield false negatives).

migration/rules/src/main/scala/fix/v0_9_28.scala
```diff
@@ -0,0 +1,22 @@
...
+          Patch.addGlobalImport(importer"org.scalatest.FunSuiteLike") +
+            Patch.replaceTree(t, "AbstractSemanticRuleSuite") +
+            Patch.addRight(t, " with FunSuiteLike")
```
bjaglin, line 19: why not
```suggestion
            Patch.replaceTree(t, "AbstractSemanticRuleSuite with FunSuiteLike")
```

migration/rules/src/main/scala/fix/v0_9_28.scala
```diff
@@ -0,0 +1,22 @@
+package fix
+
+import scala.meta._
+
+import scalafix.v1._
+
+class v0_9_28 extends SemanticRule("v0_9_28") {
```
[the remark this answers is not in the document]
bjaglin: might be worth adding a link to https://github.com/scalacenter/scalafix/pull/1176 for the record

migration/rules/src/main/scala/fix/v0_9_28.scala
bjaglin: Also, I guess it does not matter since chances that `AbstractSemanticRuleSuite` was mixing-in other traits are very low, but the rule would erase the existing ones - to be correct, you could have prettyprinted a patched version of `t` instead.

migration/rules/src/main/scala/fix/v0_9_28.scala
```diff
@@ -0,0 +1,22 @@
...
+  override def fix(implicit doc: SemanticDocument): Patch =
+    doc.tree.collect {
+      case t @ Importee.Name(Name("SemanticRuleSuite")) =>
+        Patch.removeImportee(t) +
+          Patch.addGlobalImport(
+            importer"scalafix.testkit.AbstractSemanticRuleSuite"
+          )
+      case t @ Init(_, _, _) =>
+        if (t.toString() == "SemanticRuleSuite()")
+          Patch.addGlobalImport(importer"org.scalatest.FunSuiteLike") +
```
bjaglin, line 17: I just had a double-check on the mess that ScalaTest introduced in a minor release:
- `org.scalatest.FunSuiteLike` is available for ScalaTest < 3.2.0
- `org.scalatest.funsuite.AnyFunSuite` is available for ScalaTest >= 3.1.0

We currently build against 3.0.8, so the first is the right choice, and users of the testkit which have evicted 3.0.8 with 3.2.x would have runtime failures on `SemanticRuleSuite` anyway, so this migration does not apply to them.

We will easily be able to move to `org.scalatest.funsuite.AnyFunSuite` via another migration (at least where Steward runs) for a further version where we bump to 3.2.0 (or something else if we switch to munit).

bjaglin (commented):
I agree it does not belong here:
- the additions in `build.sbt` shows that the new modules are completely independent of the rest of the build
- despite generic names, the input/output modules are coupled to testkit 0.9.27, so if we ever have to create a new migration for, say, 0.9.42, we will need other modules

We could follow the example of https://github.com/scalatest/autofix: create a separate repository `scalacenter/scalafix-migrations`, with a first sbt project `testkit_0_9_28` at the root. I think that would be better than a gist (can a gist be on an organization anyway?), even if there is no CI on it.

bjaglin:
> We could follow the example of https://github.com/scalatest/autofix: create a separate repository scalacenter/scalafix-migrations, with a first sbt project testkit_0_9_28 at the root. I think that would be better than a gist (can a gist be on an organization anyway?), even if there is no CI on it.

I just realized that there are already migrations for rule authors in https://github.com/scalacenter/scalafix.g8/tree/main/scalafix. I think that new one could belong there. Maybe we could rework the layout to follow the autofix "one build per rule" layout?

mlachkar:
https://github.com/scalacenter/scalafix.g8/pull/20

PR #1406, opened 2021-05-16
testkit: extract targetroot from scalacOptions

bjaglin:
Second take at https://github.com/scalacenter/scalafix/pull/1275 leveraging https://github.com/scalacenter/scalafix/pull/1389 (with the addition of the first preliminary commit, allowing the targetroot to be repeated in the CLI when dealing with a complex classpath)

Unlocks https://github.com/scalacenter/sbt-scalafix/pull/219

### DIFF
--- a/project/Mima.scala
+++ b/project/Mima.scala
@@ -34,7 +34,7 @@ object Mima {
       ProblemFilters.exclude[Problem]("scalafix.testkit.SemanticRuleSuite.*"),
       ProblemFilters.exclude[MissingClassProblem]("scalafix.testkit.SyntacticRuleSuite$"),
       ProblemFilters.exclude[Problem]("scalafix.testkit.SyntacticRuleSuite"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.interfaces.ScalafixArguments.withSemanticdbTargetroot")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.interfaces.ScalafixArguments.withSemanticdbTargetroots")
     )
   }
 }
--- a/build.sbt
+++ b/build.sbt
@@ -212,8 +212,7 @@ lazy val unit = project
       }
       put(
         "inputClasspath",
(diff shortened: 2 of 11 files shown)


project/Mima.scala
```diff
@@ -34,7 +34,7 @@ object Mima {
       ProblemFilters.exclude[Problem]("scalafix.testkit.SemanticRuleSuite.*"),
       ProblemFilters.exclude[MissingClassProblem]("scalafix.testkit.SyntacticRuleSuite$"),
       ProblemFilters.exclude[Problem]("scalafix.testkit.SyntacticRuleSuite"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("scalafix.interfaces.ScalafixArguments.withSemanticdbTargetroot")
```
bjaglin, line 37: unreleased

mlachkar (approved):
I didn't think about having multiple targetroot directories. 
I do prefer this approach for specifying semanticdbTargetRoot.

bjaglin:
> I didn't think about having multiple targetroot directories.

Me neither at first, and I didn't expect this PR to take me there (since it's about inferring it from `scalacOptions` in testkit, meaning that it can't be repeated) but I think it demonstrates how/why having a separate CLI argument makes sense: specifically when the input  is so complex that it was compiled in several `scalac` invokations. Of course, in that case we would like to provide several "sets" of `scalacOptions`, but that's another story.

We might want to expose `semanticdbTargetroots` as a property in testkit to [handle such use-cases](https://github.com/liancheng/scalafix-organize-imports/pull/179/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R33-R34), but I thought we could live without it for now and revisit later. I think it's fair for the testkit to follow what the CLI has been doing until now anyway.

mlachkar:
Ok, let's do that like this, but I think if we start using `withTargetroots` in the sbt-plugin, we should do the same in the testkit, even if we retrieve the targetroot from the scalacOptions for now.

PR #1409, opened 2021-05-17
support parsing scala 2.13.6 / 2.12.14 -Xsource:3 dialect

bjaglin:
https://github.com/scalameta/scalameta/issues/2323

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
@@ -5,6 +5,7 @@ import scala.util.Success
 import scala.util.Try
 
 import scala.meta.Dialect
+import scala.meta.dialects
 
 import scalafix.internal.config.ScalaVersion._
 
@@ -16,10 +17,13 @@ sealed trait ScalaVersion {
   val major: MajorVersion
   val minor: Option[Int]
   val patch: Option[Int]
-  val dialect: Dialect = major match {
-    case ScalaVersion.Major.Scala2 => scala.meta.dialects.Scala213
-    case ScalaVersion.Major.Scala3 => scala.meta.dialects.Scala3
-  }
+
+  def dialect(sourceScalaVersion: Option[ScalaVersion]): Dialect =
+    (major, sourceScalaVersion.map(_.major)) match {
+      case (Major.Scala3, _) => dialects.Scala3
+      case (Major.Scala2, Some(Major.Scala3)) => dialects.Scala213Source3
+      case (Major.Scala2, _) => dialects.Scala213
+    }
 
   def isScala2: Boolean = major == Major.Scala2
   def isScala3: Boolean = major == Major.Scala3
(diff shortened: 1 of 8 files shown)


scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -55,17 +56,13 @@ object RuleTest {
...
       val rules = decoder.read(rulesConf).get.withConfiguration(config).get
       (rules, sdoc)
     }
 
     new RuleTest(test, run)
   }
-
-  def getDialectFromScalaV(scalaV: String): Dialect =
-    if (scalaV.startsWith("3")) ScalafixConfig.Scala3
-    else ScalafixConfig.Scala2
```
bjaglin, line 70: I tried to centralize dialect handling to make sure the testkit is not drifting. Turning the entire `testsInput` to `-Xsource:3` is a big task, so unfortunately, the testkit codepath is not tested with the new flag.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -345,6 +345,33 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
+    val cwd = StringFS
+      .string2dir(
+        s"""|/src/Scala2Source3.scala
+          |open class Scala2Source3""".stripMargin,
+        charset
+      )
+      .toNIO
+    val src = cwd.resolve("src")
+    val args = api
+      .withScalaVersion("2.13.6")
```
bjaglin, line 358: This value is not mandatory, but I left it there since what we expect clients to send - I feel like it's enough/better to test for the presence of `-Xsource:` only, so that this PR works in the future with the 2.12.x release that will eventually allow it as well.

scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
```diff
@@ -16,10 +17,13 @@ sealed trait ScalaVersion {
...
   val patch: Option[Int]
-  val dialect: Dialect = major match {
-    case ScalaVersion.Major.Scala2 => scala.meta.dialects.Scala213
-    case ScalaVersion.Major.Scala3 => scala.meta.dialects.Scala3
-  }
+
+  def dialect(sourceScalaVersion: Option[ScalaVersion]): Dialect =
+    (major, sourceScalaVersion.map(_.major)) match {
+      case (Major.Scala3, _) => dialects.Scala3
+      case (Major.Scala2, Some(Major.Scala3)) => dialects.Scala213Source3
```
bjaglin, line 24: that should handle the upcoming 2.12.14 release with `-Xsource:3` support

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
mlachkar: I think it's ok

mlachkar (approved):

bjaglin:
I'll open a PR on sbt-scalafix shortly to pass `scalacOptions` even for syntactic rules

PR #1411, opened 2021-05-20
Create a scala3 variable for documentation

mlachkar:


bjaglin (commented):
I didn't create a templated variable because the support for semanticdb is built-in in the compiler (so we don't need a new scalafix/sbt-scalafix when a new 3.x release comes out).

mlachkar:
ok

PR #1415, opened 2021-05-27
bump scalameta to fix regression for replaceTree(Template, t.syntax)

bjaglin:
https://github.com/scalameta/scalameta/issues/2364

### DIFF
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
@@ -5,6 +5,7 @@ import java.nio.file.Files
 import java.nio.file.Paths
 
 import scala.meta._
+import scala.meta.internal.prettyprinters.TreeSyntax
 import scala.meta.tokens.Token.Ident
 
 import org.scalatest.funsuite.AnyFunSuiteLike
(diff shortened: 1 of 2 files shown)


scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -178,4 +179,46 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
+      }.asPatch
+      assert(patch.nonEmpty)
+      patch
+    }
+  }
+
+  val program: Source =
+    """import a.b
+      |
+      |class Foo(private val q: String = "") extends Bar with Qux {
```
bjaglin, line 198: `class Foo extends Bar` was enough to reproduce the problem but I thought I might as well write a longer program to prevent similar "symmetry" regressions for other trees, but it's of course by no mean exhaustive... I wonder if there is a test input somewhere in scalameta that makes use of all tree types?

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -178,4 +179,47 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
+      |  foo[Bar](null)
+      |  new R()
+      |  object O {}
+      |}""".stripMargin.parse[Source].get
+
+  program
+    .collect { case t => t }
+    .filterNot(_.pos.text == "")
+    .groupBy(_.productPrefix)
+    .map(_._2.head)
```
bjaglin, line 210: `distinctBy` was introduced in scala 2.13 :cry:

PR #1408, opened 2021-05-17
scala 2.13.6

bjaglin:
Also fix the MiMa regression https://github.com/scalacenter/scalafix/pull/1370#issuecomment-820461009

PR #1416, opened 2021-05-28
Update Scala 2.12 to 2.12.14

tgodzik:


project/Dependencies.scala
```diff
@@ -24,7 +24,7 @@ object Dependencies {
   val metaconfigV = "0.9.11"
   val nailgunV = "0.9.1"
   val scalaXmlV = "1.3.0"
-  val scalametaV = "4.4.18"
```
bjaglin, line 27: sorry I just merged a PR that bumped this to `.19` on main

project/Dependencies.scala
```diff
@@ -5,7 +5,7 @@ import scala.util.Try
 
 object Dependencies {
   val scala211 = "2.11.12"
-  val scala212 = "2.12.13"
+  val scala212 = "2.12.14"
```
bjaglin, line 8: you'll need something like https://github.com/scalacenter/scalafix/pull/1408/files#diff-cf70a614681abb9e33115a49e0b4a52865bdfcdb24c4c0df4af6546d92f88f5fR163 for MiMa

bjaglin: and probably also https://github.com/scalacenter/scalafix/pull/1408/files#diff-8fe42aaf3e5d1a3f8804efa64a6206f5af6a826dea8f6ce51ad0ce8105188858R2 since `ExplicitResultTypes` will complain that `2.12.14` is not supported....

tgodzik: Scalafix runs on 2.13, so we should be good!

bjaglin (approved):
Thanks! I'll tag a release soon (maybe I'll wait for https://github.com/scalameta/metaconfig/pull/131).

Issue #1199, opened 2020-07-03
Expose scala versions mismatch as typed errors in SemanticRule#withConfiguration


bjaglin:
Advanced rules such as `ExplicitResultTypes` or [`ExplicitNonNullaryApply`](https://github.com/scala/scala-rewrites/blob/v0.1.1/rewrites/src/main/scala/fix/scala213/ExplicitNonNullaryApply.scala) using the presentation compiler via `ScalafixGlobal` require the build to match `scalafixScalaBinaryVersion`, and potentially the full Scala release used by Scalafix.

Rules themselves should not suggest actions to fix a mismatch (as rules might not be run from sbt for example). That should be up to the rule runner, which should be able to intercept rich information on execution failure.

Implementation: I am not sure the signature of https://www.javadoc.io/static/ch.epfl.scala/scalafix-core_2.12/0.9.18/scalafix/v1/SemanticRule.html#withConfiguration(config:scalafix.v1.Configuration):metaconfig.Configured[scalafix.v1.Rule] would allow that, so we might need an overload.

PR #1433, opened 2021-06-16
Procedure syntax

mlachkar:

### DIFF
--- a/docs/rules/ProcedureSyntax.md
+++ b/docs/rules/ProcedureSyntax.md
@@ -23,3 +23,6 @@ trait A { def doSomething }
 def main(args: Seq[String]): Unit = { println("Hello world!") }
 trait A { def doSomething: Unit }
 ```
+
+This syntax no longer exists in Scala 3 and therefore this rule only makes sense in Scala 2.
+You need to remove `ProcedureSyntax` from `.scalafix.conf` for Scala 3 projects.
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
@@ -5,13 +5,24 @@ import scala.meta.tokens.Token
 import scala.meta.tokens.Token.Equals
 import scala.meta.tokens.Token.RightParen
 
+import metaconfig.Configured
 import scalafix.util.Trivia
 import scalafix.v1._
 
 class ProcedureSyntax extends SyntacticRule("ProcedureSyntax") {
 
   override def description: String =
-    "Replaces deprecated procedure syntax with explicit ': Unit ='"
+    "Replaces deprecated procedure syntax with explicit ': Unit ='" +
+      "This rule is specific to scala 2, since procedure syntax is not supported in Scala 3"
+
+  override def withConfiguration(config: Configuration): Configured[Rule] = {
+    if (config.scalaVersion.startsWith("3"))
+      Configured.error(
+        "This rule is specific to Scala 2, since procedure syntax is not supported in Scala 3. " +
+          "To fix this error, remove ProcedureSyntax from .scalafix.conf"
+      )
+    else Configured.ok(this)
+  }
 
   override def isRewrite: Boolean = true
 
--- a/scalafix-tests/input/src/main/scala-2/test/procedureSyntax/ConstructorProcedureSyntax.scala
+++ b/scalafix-tests/input/src/main/scala-2/test/procedureSyntax/ConstructorProcedureSyntax.scala
@@ -1,7 +1,7 @@
 /*
 rules = ProcedureSyntax
  */
-package test
+package test.procedureSyntax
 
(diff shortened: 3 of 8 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -11,7 +11,8 @@ import scalafix.v1._
 class ProcedureSyntax extends SyntacticRule("ProcedureSyntax") {
 
   override def description: String =
-    "Replaces deprecated procedure syntax with explicit ': Unit ='"
+    "Replaces deprecated procedure syntax with explicit ': Unit ='" +
+      "This rule is specific to scala 2, since procedure syntax is not supported in Scala 3"
```
bjaglin, line 15: I guess it goes back to the question of whether unsupported rules should be silently ignored or trigger a failure, but I would have expected this to come with a check like https://github.com/scalacenter/scalafix/blob/933b9403e269cad1133d65696821a5d387de9c86/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L49-L93. In that case it's not a big deal since code not respecting the procedure syntax wouldn't compile anyway, but there might be other cases where we prefer a verbose failure instead of false negatives.

docs/rules/ProcedureSyntax.md
```diff
@@ -23,3 +23,6 @@ trait A { def doSomething }
 def main(args: Seq[String]): Unit = { println("Hello world!") }
 trait A { def doSomething: Unit }
 ```
+
+This feature no longer exists in Scala 3 and therefore this rule won't
```
bjaglin, line 27:
```suggestion
This syntax no longer exists in Scala 3 and therefore this rule won't 
```

scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
mlachkar: Yes in fact in this case, there is no way to have procedure syntax in scala 3 code, so this rule wouldn't rewrite scala 3 code projects. 
But should we do that in other cases: for example in DisableSyntax, some options don't make sense anymore since  some syntax is not allowed anymore in scala 3. Should we warn about some options to say, that they won't have any effects ? 

- In this case, I agree we can throw an error this rule is not useful in Scala 3

scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -1,17 +1,28 @@
...
+
+  override def withConfiguration(config: Configuration): Configured[Rule] = {
+    if (config.scalaVersion.startsWith("3"))
+      Configured.error(
+        """|This rule is specific to scala 2, since procedure syntax is not supported in Scala 3
+          |To fix this error, remove ProcedureSyntax from .scalafix.conf""".stripMargin
+      )
```
bjaglin, line 23: other rules don't seem to use multi-line error messages, not sure if that can be a problem for some integrations?

scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
```diff
@@ -1,17 +1,28 @@
...
 
   override def description: String =
-    "Replaces deprecated procedure syntax with explicit ': Unit ='"
+    "Replaces deprecated procedure syntax with explicit ': Unit ='" +
+      "This rule is specific to scala 2, since procedure syntax is not supported in Scala 3"
+
+  override def withConfiguration(config: Configuration): Configured[Rule] = {
+    if (config.scalaVersion.startsWith("3"))
+      Configured.error(
+        """|This rule is specific to scala 2, since procedure syntax is not supported in Scala 3
```
bjaglin, line 21:
```suggestion
        """|This rule is specific to Scala 2, since procedure syntax is not supported in Scala 3
```

scalafix-rules/src/main/scala/scalafix/internal/rule/ProcedureSyntax.scala
bjaglin: Maybe now is a good time to design/implement https://github.com/scalacenter/scalafix/issues/1199? I feel like it could address the recurring questions we have with built-in rules (as discussed in https://github.com/scalacenter/scalafix/pull/1394), and that community rules will probably have.

docs/rules/ProcedureSyntax.md
```diff
@@ -23,3 +23,6 @@ trait A { def doSomething }
 def main(args: Seq[String]): Unit = { println("Hello world!") }
 trait A { def doSomething: Unit }
 ```
+
+This syntax no longer exists in Scala 3 and therefore this rule won't 
+have any effect on Scala 3 projects.
```
bjaglin, line 28: this is no longer true with the latest changes

bjaglin (approved):

PR #1436, opened 2021-06-18
WIP - expose requirement for rules

mlachkar:
This implementation is just to start the discussion about #1199 #1394
- So with this implementation we allow rules authors to expose their constraints
- But, the concern raised in  #1199, about enabling the rule runner (sbt, mill, a tool ..) to propose a fix,  is not addressed. Those runners interact with scalafix through the Java interface Scalafx Arguments which can only return a unit or a `ScalafixError` which is just a string. 
Let's start a discussion here to see what we want and what the issue we want to adress.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/v1/InputRequirements.scala
+++ b/scalafix-core/src/main/scala/scalafix/v1/InputRequirements.scala
@@ -1,0 +1,22 @@
+package scalafix.v1
+
+trait InputRequirements {
+  def scalaVersionsRequired: Seq[InputRequirements.ScalaVersion]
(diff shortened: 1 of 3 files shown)


bjaglin:
> Those runners interact with scalafix through the Java interface Scalafx Arguments which can only return a unit or a `ScalafixError` which is just a string.

Good point, I didn't consider the Java layer in the suggestion drafted in https://github.com/scalacenter/scalafix/issues/1199. What about adding methods in https://github.com/scalacenter/scalafix/blob/main/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixRule.java?

PR #1439, opened 2021-06-25
Small refactoring of explicitResultTypes

mlachkar:


scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -92,35 +92,32 @@ final class ExplicitResultTypes(
     }
   }
 
-  override def fix(implicit ctx: SemanticDocument): Patch = {
+  override def fix(implicit ctx: SemanticDocument): Patch =
     try unsafeFix()
     catch {
-      case _: CompilerException =>
-        shutdownCompiler()
```
mlachkar, line 99: If the compiler crashes because usually it cannot type a part of the AST, it will crash again even after a retry. 
Actually we do that on the file level, which means if one explicit type makes the compiler crash, we don't add any types in the entire file. Should we continue to do that ?

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -92,35 +92,32 @@ final class ExplicitResultTypes(
...
+          case t @ Defn.Var(mods, Pat.Var(name) :: Nil, None, Some(body))
+              if isRuleCandidate(t, name, mods, body) =>
+            fixDefinition(t, body, types)
+
+          case t @ Defn.Def(mods, name, _, _, None, body)
+              if isRuleCandidate(t, name, mods, body) =>
+            fixDefinition(t, body, types)
+        }.asPatch
+      case None => Patch.empty
+    }
```
mlachkar, line 120: old code vs new code:
 - in old code, we will always traverse the entire Tree of the file even if we were not able to instantiate the global, which can only happen if the classpath is wrong
 - new code: we always instantiate the global, even if there no element (def, or var or val) in the Tree. But in the other hand we remove an indirection where we create  a `TypePrinter ` just to wrap the case where ScalafixGlobal is None.

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: > The compiler crashes sometimes because it cannot type a part of the AST. If it crashes once, it will crash again even after a retry.

Is that the only case? https://github.com/scalacenter/scalafix/commit/ba74db38dde7fcdd4971b56b10ec94cb6fef7092 is not very verbose, I am not sure... In any case, if you reproduced the issue, it would be good to document the cases where the compiler crashes and how it behaves via tests.

> Actually we do that on the file level, which means if one explicit type makes the compiler crash, we don't add any types to the entire file. Should we continue to do that ?

Catching the exception at a more granular seems like a good idea indeed. Maybe this wasn't done for performance reason because some files would crash the compiler over and over again?

scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: I am not familiar with `global`, but https://github.com/scalacenter/scalafix/commit/c6014cebbbbea44b196d53991822c9418a469a6f did introduce this for something else than an incorrect classpath it seems

mlachkar:
Hello @olafurpg ! Could you have a look on this refactoring. As @bjaglin reported, this PR https://github.com/scalacenter/scalafix/commit/c6014cebbbbea44b196d53991822c9418a469a6f introduced what I am changing again. 
My goal is to reduce the number of abstractions, to make it easier to understand. What do you think ? Thanks !

mlachkar:
I will merge this change! I think it's good to simplify the code for explicit result types!

PR #1444, opened 2021-07-09
add anonymous function parameter support & enhance patvar RemoveUnused

taisukeoe:
Hello. This is a RemoveUnused rule enhancement Pull Request.

More specifically, this Pull Request introduces:
- an anonymous function parameter support to RemoveUnused, with a new `params` config of RemoveUnusedConfig.
- unused pattern variable support enhancement to RemoveUnused, which newly supports:
   - Bind (As) patterns and Typed patterns, not only Variable patterns
   - case clauses in Partial-Function and Try-Catch

Of note, I intentionally skip ForYield unused pattern variables support, since their unused warnings seem to have a lot of false positives (at least in Scala 2.12.14).

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -5,8 +5,8 @@ import scala.collection.mutable
 import scala.meta._
 
 import metaconfig.Configured
-import scalafix.v1._
 import scalafix.util.Trivia
+import scalafix.v1._
 
 class RemoveUnused(config: RemoveUnusedConfig)
     extends SemanticRule(
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -6,6 +6,7 @@ import scala.meta._
 
 import metaconfig.Configured
 import scalafix.v1._
+import scalafix.util.Trivia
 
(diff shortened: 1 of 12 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -146,4 +172,17 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
+      tree.tokens.takeWhile(tk => tk.is[Token.LeftParen] || tk.is[Trivia]).size
+    val rightTokenCount = tree.tokens
+      .takeRightWhile(tk => tk.is[Token.RightParen] || tk.is[Trivia])
+      .size
+    tree.pos match {
+      case Position.Range(input, start, end) =>
+        Position.Range(input, start + leftTokenCount, end - rightTokenCount)
+      case other => other
+    }
+  }
```
taisukeoe, line 187: This is a workaround to match a Position with unused diagnostics, if a patvar is embraces by `()`,  like `(s: String)`.
I'd appreciate it if  your can advice me a smarter way.

mlachkar: Looks good to me!

mlachkar (approved):

mlachkar:
Thanks for your contribution.

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
taisukeoe: Thank you for your confirmation!

PR #1434, opened 2021-06-17
Remove unused

mlachkar:
The error in 2.12 is not easy to understand. If the test file is moved outside the package I have created, the test passes. (this time the issue is not around maxNuberofWarn that I increased. The semanticdb file doesn't contain any diagnostics.
@tgodzik have you ever seen this kind of issue ?

PR #1448, opened 2021-07-15
Unused function params follow-up

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1444 (and https://github.com/scalacenter/scalafix/pull/1434 for the second commit)

Note that the current rule implementation does not generate invalid code, it just misses some rewrites.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -156,6 +156,12 @@ class RemoveUnused(config: RemoveUnusedConfig)
         case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
             if isUnusedParam(posExclParens(param)) =>
           Patch.replaceTree(name, "_")
+        case Term.Function(params, _) =>
+          params.collect {
+            case param @ Term.Param(_, name, _, _)
+                if isUnusedParam(param.pos) =>
+              Patch.replaceTree(name, "_")
+          }.asPatch
       }.asPatch
     }
   }
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -150,11 +150,20 @@ class RemoveUnused(config: RemoveUnusedConfig)
           patchPatVarsIn(cases)
         case Term.Try(_, cases, _) =>
           patchPatVarsIn(cases)
-        case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
-            if isUnusedParam(name.pos) =>
+        case Term.Function(List(param @ Term.Param(_, name, None, _)), _)
+            if isUnusedParam(
+              // diagnostic does not include the implicit prefix (supported only
+              // on 1-arity function without explicit type) captured in param,
+              // so use name instead
+              name.pos
+            ) =>
+          // drop the entire param so that the implicit prefix (if it exists) is removed
           Patch.replaceTree(param, "_")
-        case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
-            if isUnusedParam(posExclParens(param)) =>
+        case Term.Function(List(param @ Term.Param(_, name, Some(_), _)), _)
+            if isUnusedParam(
+              // diagnostic does not include the wrapping parens captured in single param
+              posExclParens(param)
+            ) =>
           Patch.replaceTree(name, "_")
         case Term.Function(params, _) =>
           params.collect {
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -153,6 +153,9 @@ class RemoveUnused(config: RemoveUnusedConfig)
         case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
             if isUnusedParam(name.pos) =>
           Patch.replaceTree(param, "_")
+        case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
+            if isUnusedParam(posExclParens(param)) =>
+          Patch.replaceTree(name, "_")
       }.asPatch
     }
   }
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnusedConfig.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnusedConfig.scala
@@ -16,7 +16,9 @@ case class RemoveUnusedConfig(
       "Remove unused pattern match variables (compatible with Scala 2.12 and 2.13)"
     )
     patternvars: Boolean = true,
-    @Description("Remove unused parameters")
+    @Description(
+      "Remove unused function parameters (compatible with Scala 2.12 and 2.13)"
+    )
     params: Boolean = true
 )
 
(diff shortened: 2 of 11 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnusedConfig.scala
```diff
@@ -16,7 +16,9 @@ case class RemoveUnusedConfig(
       "Remove unused pattern match variables (compatible with Scala 2.12 and 2.13)"
     )
     patternvars: Boolean = true,
-    @Description("Remove unused parameters")
+    @Description(
+      "Remove unused function parameters (compatible with Scala 2.12 and 2.13)"
```
bjaglin, line 20: Unused method parameters are reported by the compiler but not handled by the rule, so I decided to make this explicit in the doc.

Since `def fn(_: Long)` is invalid syntax, we would have to change the signature, which is not safe (or would be limited to private methods and require quite a bit of logic for patching call sites). Therefore I don't anticipate the rule to ever support them, but if we ever do, I guess we can add a `methodparams` config flag (in other word, I think it's acceptable to assign `params` for function params only).

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -153,6 +153,15 @@ class RemoveUnused(config: RemoveUnusedConfig)
         case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
             if isUnusedParam(name.pos) =>
           Patch.replaceTree(param, "_")
```
bjaglin, line 155: the fact that the predicate is on the name and the rewrite on the full param seems to be very specific to a 1-arity function with an implicit modifier, since in all other cases we want to match on the full param and replace only the name

bjaglin: @taisukeoe can you comment on this since that's the test case you added? I had a look at https://www.scala-lang.org/files/archive/spec/2.13/06-expressions.html#anonymous-functions to try to understand how params can be implicit (and whether it's possible in a >1-arity function), but I am still confused.

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -153,6 +153,15 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
+        case Term.Function(List(param @ Term.Param(_, name, _, _)), _)
+            if isUnusedParam(posExclParens(param)) =>
```
bjaglin, line 157: Wrapping parens are captured in `param` only when there is a single param. I couldn't find a way to use parens around a param with `implicit` modifier, so I think that's the only place where `posExclParens` is needed...

mlachkar:
Thanks for this follow-up. Everything looks ok. 
@taisukeoe could you also do a review. Thanks

mlachkar (approved):

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnusedConfig.scala
taisukeoe: Thank you very much for clarifying this.
I'd agree with that this RemoveUnused rule should not support unused explicit method parameters because of unsafety (that's why I have not added) and be nicer to be explicit about it.

One thing I was wondering is about unused implicit method parameters as an opt-in config.
Of course they are also not 100% safe to remove (because call sites can pass the argument explicitly), but still useful in most cases, and easy to fix even if it breaks the code.

```scala
/* O is not used */
def f[T](t: List[T])(implicit O: Ordering[T]): List[T] =  t

// It's safe to remove O
f(List("a"))

// It's not safe to remove O, although this is not so often.
f(List(1))(Ordering.Int)
```

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
taisukeoe: Thank you very much for elaboration. 
While current predicate works only when a param without explicit type, regardless of implicit type, it clearly misses a param with explicit type, because of the following behaviour. 

When `def g(x: String => Unit): Unit = ???` is given,

|matches with unused `s` diagnostics pos?|`g{s => ???}`|`g{implicit s => ???}`|`g{s: String => ???}`|
|:--|:--|:--|:--|
|param.name.pos|〇|〇|×|
|param.pos|〇|×|〇|

>  how params can be implicit (and whether it's possible in a >1-arity function)

As you pointed, 6.2.3 Anonymous Functions says:

> A named parameter of an anonymous function may be optionally preceded by an implicit modifier. In that case the parameter is labeled implicit; however the parameter section itself does not count as an implicit parameter section. Hence, arguments to anonymous functions always have to be given explicitly.

Although It's not very clearly said, `A named parameter of an anonymous function` might mean that it is allowed only for 1-arity function, and it matches what we experimented.

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -153,6 +153,15 @@ class RemoveUnused(config: RemoveUnusedConfig)
...
+          Patch.replaceTree(name, "_")
+        case Term.Function(params, _) =>
+          params.collect {
+            case param @ Term.Param(_, name, _, _)
+                if isUnusedParam(param.pos) =>
+              Patch.replaceTree(name, "_")
+          }.asPatch
```
taisukeoe, line 164: As you leave comments and we discussed in this PR, I'm a bit worrying that the reasons why each condition is chosen are not very clear (from the code, at least):

-  isUnusedParam(name.pos) or isUnusedParam(param.pos) or isUnusedParam(posExclParen(param))
- replaceTree(name, "_") or replaceTree(param, "_")
- match against 1-arity or n-arity function

Rather, it might be better to consolidate, something like:

```suggestion
        case Term.Function(params, _) =>
          params.collect {
            case param @ Term.Param(_, name, declType, _)
                if isUnusedParam(name.pos) || isUnusedParam(
                  posExclParens(param)
                ) =>
              Patch.replaceTree(if (declType.isDefined) name else param, "_")
          }.asPatch
```

What do you think?

taisukeoe: This will traverse Function trees a bit more.
Or, adding this kind of if-guard might be more user-friendly?

```scala
case Term.Function(params, _) if config.params =>
```

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnusedConfig.scala
github-brice-jaglin: > I'd like to just note that I was wondering about unused implicit method parameters as an opt-in config.

Then we agree: a potential new opt-in flag could be added later, but we keep this current one for functions only.

scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
github-brice-jaglin: I agree the code is complex, but I think refactoring everything into one `case` with a nested branch does not make the 3 different use cases and their gotchas explicit. I added 1968ecfc096e07ecf9264e83ccefd70d9d7ab28b to document why the last `case` is not enough.

taisukeoe (approved):
LGTM, thank you very much for your following ups.

PR #1452, opened 2021-07-15
Make NoAutoTupling fixes the insertion of unit

mlachkar:
- and refactoring tests

I don't know if want to insert `()` when we have this kind of warnings (that are only reported by `-deprecation`)
```
adaptation of an empty argument list by inserting () is deprecated: this is unlikely to be what you want
        signature: Some.apply[A](value: A): Some[A]
  given arguments: <none>
 after adaptation: Some((): Unit)
 ```
 
 The test was called NoUnitInsertion. Not clear for me if we want that or not. In the other hand, even if it's not a behavior we want, noAutoTupling insert (), if the user has added -deprecation in his options. It wasn't working for 2.13 since the warning message changed from `Adaptation ...` to `adaptation ...` for unit insertions.
 
 - other point: For this rule, we don't have `withConfiguration`: should we do the same than in removeUnused ?

### DIFF
--- a/docs/rules/NoAutoTupling.md
+++ b/docs/rules/NoAutoTupling.md
@@ -4,27 +4,23 @@ id: NoAutoTupling
 title: NoAutoTupling
 ---
 
-> ⚠️ This rule does not work with Scala 2.13 since `-Yno-adapter-args` has been
-> removed.
-
 Adds explicit tuples around argument lists where auto-tupling is occurring.
 
 To use this rule:
 
-- enable `-Ywarn-adapted-args` (note, `-Yno-adapted-args` will fail compilation,
-  which prevents scalafix from running)
-- disable `-Xfatal-warnings`. Unfortunately, the Scala compiler does not support
-  finer grained control over the severity level per message kind. See
-  [scalameta/scalameta#924](https://github.com/scalameta/scalameta/issues/924)
-  for a possible workaround in the near future.
+- enable `-Ywarn-adapted-args` for Scala 2.11 and 2.12 (note, `-Yno-adapted-args` will fail compilation,
+  which prevents scalafix from running). For Scala 2.13, use instead `-Xlint:adapted-args`.
+- enable also `-deprecation` to get warnings on insertions of `Unit`.
 
 ```scala
 // before
 def someMethod(t: (Int, String)) = ...
 someMethod(1, "something")
+val c: Option[Unit] = Some()
 // after
 def someMethod(t: (Int, String)) = ...
 someMethod((1, "something"))
+val c: Option[Unit] = Some(())
 ```
 
 Auto-tupling is a feature that can lead to unexpected results, making code to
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: scalafix-tests/output/src/main/scala-2/test/noAutoTupling/NoUnitInsertion.scala [hunk 1]
```diff
@@ -0,0 +1,24 @@
+package test.noAutoTupling
+
+class NoUnitInsertion {
```
bjaglin, line 3: so if I understand well:
* this was removed by mistake in https://github.com/scalacenter/scalafix/commit/d5d23dbf37208533254067ddf1cea3dca6904b2b#diff-fd705cc22acdf8f3dc8262c143dd9ff3a6afcdb914b8dc0cd02479258fcdbe3e
* testkit seeing an input without an output is not triggering a failure as it's a valid use-case for linters (`assert` syntax in input file comments)
* The TODOs introduced in https://github.com/scalacenter/scalafix/pull/1118 in `NoAutoTupling` was a subset of this

docs/rules/NoAutoTupling.md
```diff
@@ -4,27 +4,23 @@ id: NoAutoTupling
...
-
 Adds explicit tuples around argument lists where auto-tupling is occurring.
 
 To use this rule:
 
-- enable `-Ywarn-adapted-args` (note, `-Yno-adapted-args` will fail compilation,
-  which prevents scalafix from running)
-- disable `-Xfatal-warnings`. Unfortunately, the Scala compiler does not support
-  finer grained control over the severity level per message kind. See
-  [scalameta/scalameta#924](https://github.com/scalameta/scalameta/issues/924)
```
bjaglin, line 18: Did you remove that following https://github.com/scalacenter/sbt-scalafix/pull/196? It's specific to sbt-scalafix, but I guess it's not worth mentioning "just" for the other clients, so I am inclined to say that we should ignore mentions of `-Xfatal-warnings` considering sbt-scalafix is the main client.

bjaglin (approved):
:+1: Looking at https://github.com/scalacenter/scalafix/pull/352, it does look like it's something we want(ed)! Regarding `withConfiguration`, I guess it would be good to add it to guide usage of the rules as it's getting tricky to set up the right `scalacOptions` indeed!

docs/rules/NoAutoTupling.md
mlachkar: yes and I forgot about the other clients. I think the other client should mention it (or are already menting it)

scalafix-tests/output/src/main/scala-2/test/noAutoTupling/NoUnitInsertion.scala
mlachkar: - I didn't know why this test wasn't running. I deleted it by mistake! 
- While grouping the tests together for each rule, I remembered that a small thing was missing to make this rule working. Maybe I will go throw my previous Todo and see what I can fix.

PR #1493, opened 2021-11-02
support scala 2.13.7

bjaglin:


.scalafix.conf
```diff
@@ -1,5 +1,5 @@
 rules = [
-  ExplicitResultTypes,
+#  ExplicitResultTypes,
```
bjaglin, line 2:
```
[error] (rules / scalafixAll) scalafix.sbt.InvalidArgument: The ExplicitResultTypes rule only supports the exact Scala versions 'List(2.13.6, 2.13.4, 2.13.5)' for this binary version. To fix this problem, either remove ExplicitResultTypes from .scalafix.conf or change the Scala version of your build to match one of the supported versions.
```

mlachkar: make sense. we have to that everytime we upgrade the version!

mlachkar:
I think we can comment the failing test and we will try to fix later. It doesn't seem to be too bad. thank for starting this!

bjaglin:
There are 2 tests failing, [that were not failing with the Scalameta 4.4.29 update](https://github.com/scalacenter/scalafix/pull/1492)

1/
https://github.com/scalacenter/scalafix/blob/3b219fa3e53e3a06fb7a3e5890be845fb07ae3fb/scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/BetterRscCompat.scala#L196
```
###########> Diff       <###########

=======
=> Diff
=======
--- obtained
+++ expected
@@ -194,3 +194,3 @@
∙
-    val poc = new PkgObjClass
+    val poc: PkgObjClass = new PkgObjClass
   }
         
[info] - test/explicitResultTypes/BetterRscCompat.scala *** FAILED *** (592 milliseconds)

```

That seems related to https://github.com/scalameta/scalameta/pull/2502/files#diff-65b44c93ebcf284cd66432972b21ee5671b9ea3c1de809c400f7c8bc16b8041fR109-R115

2/
https://github.com/scalacenter/scalafix/blob/f41dec8fc14c8c96b84f563ece3bd895304e8287/scalafix-tests/shared/src/main/scala/test/TypeToTreeInput.scala#L40
```
[info] PrettyTypeSuite:
[info] - Defn.Trait - Functor *** FAILED *** (1 millisecond)
[info]   --- obtained
[info]   +++ expected
[info]   
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   trait Functor[C[`_`]]
[info]            
[info]   =======
[info]   => Diff
[info]   =======
[info]   --- obtained
[info]   +++ expected
[info]   @@ -1,1 +1,1 @@
[info]   -trait Functor[C[`_`]]
[info]   +trait Functor[C[_]] (PrettyTypeSuite.scala:58)
```

Not sure about this one... Any insight @tgodzik?

SethTisue:
> Not sure about this one...

might have to do with https://github.com/scala/scala/pull/9712 ...?

som-snytt:
`class C[_]` is disallowed, not `trait F[C[_]]`.

tgodzik:
> Not sure about this one... Any insight @tgodzik?

Looks like `_` is now treated as a name instead of a wildcard. Not sure where it stems from, but I could try to investigate next week. Is writing `Functor[C[`_`]]` needed currently though? I don't think we would every write [_] for the explicit results types?

As for 1) issue, in scalameta that was only an issue with tests, it did not cause any problems when using semanticdb as a normal compiler plugin. It might be the same case here.

som-snytt:
I'll refresh my memory about the change in handling wildcard. One progression was allowing multiple wildcards. I linked another change in behavior.

tOverney:
Does this bug means that scalafix will never support 2.13.7 and we're waiting for 2.13.8 ?

tgodzik:
> Does this bug means that scalafix will never support 2.13.7 and we're waiting for 2.13.8 ?

We could also comment out the test for now and merge as is.

bjaglin:
> We could also comment out the test for now and merge as is.

Agreed. I am commenting them to unblock the release.

tOverney:
Thank you for the release!! :)

PR #1496, opened 2021-11-11
Add rewriting rule RedundantSyntax.finalObject

bplommer:


bjaglin (commented):
Thanks for your PR!

It's definitely useful and generic enough to be a built-in check IMHO, but I am not sure this really belongs to `DisableSyntax` as something redundant is not a bad smell/feature like the other flags - the rule description reads "Reports an error for disabled features such as var or XML literals.".

Also, since it's redundant, this would really benefit from being a rewrite rule. I believe there are other redundant tokens that could be rewritten. What about starting a `RedundantTokens` syntactic rewriting rule to group them?

scalafix-tests/input/src/main/scala/test/redundantTokens/FinalObject.scala
```diff
@@ -0,0 +1,11 @@
+/*
+rules = RedundantTokens
+RedundantTokens.finalObject = true
```
bjaglin, line 3: some negative tests here where the final keyword is used on other symbols would be nice against regressions

scalafix-rules/src/main/scala/scalafix/internal/rule/RedundantTokens.scala
```diff
@@ -0,0 +1,33 @@
+package scalafix.internal.rule
+
+import scala.meta._
+
+import metaconfig.Configured
+import scalafix.util.TokenList
+import scalafix.v1._
+
+class RedundantTokens(config: RedundantTokensConfig)
+    extends SemanticRule("RedundantTokens") {
```
bjaglin, line 10: did you make it semantic for further iterations? I think the name of the rule itself calls for a syntactic one
```suggestion
    extends SyntacticRule("RedundantTokens") {
```

bjaglin (commented):
Thanks for the fast turnaround! Could you add some basic docs in /docs/rules/RedundantTokens.md? Thanks!

bjaglin:
Related: https://github.com/scala/bug/issues/11094

bplommer:
> Thanks for the fast turnaround! Could you add some basic docs in /docs/rules/RedundantTokens.md? Thanks!

Will do! (I'm trying to refrain from opening an editor on a Saturday though so the turnaround won't be as quick...)

bplommer:
> Related: https://github.com/scala/bug/issues/11094

Interesting - does that mean we should deactivate this rule for members of non-final traits/classes on Scala 2.12 and earlier, and/or have that behaviour configurable?

scalafix-rules/src/main/scala/scalafix/internal/rule/RedundantTokens.scala
bplommer: Yeah, I was thinking of things like removal of `override` for implementation of abstract members (though I wouldn't want to enable that by default) - should we have two separate rules, a semantic one and a syntactic one?

bjaglin: Having 2 rules sounds a bit extreme. On the other hand, starting with a syntactic one and upgrading to a semantic later would be a breaking change for clients that do not use a build tool integration. I would say this is still acceptable.

I can think of a few other useful **syntactic** rewrites (removing unnecessary string interpolators, empty bodies, etc). The question then is whether such rewrites wouldn't be better (only) in scalafmt. I do see potential in syntactic rules, but mostly for community ones actually.

Any opinion @olafurpg @mlachkar ?

bjaglin:
> > Related: [scala/bug#11094](https://github.com/scala/bug/issues/11094)
> 
> Interesting - does that mean we should deactivate this rule for members of non-final traits/classes on Scala 2.12 and earlier, and/or have that behaviour configurable?

Since it's on by default and may have unexpected yet limited side effects on the bytecode, I'd say that a reference to that ticket in the docs would be enough.

scalafix-rules/src/main/scala/scalafix/internal/rule/RedundantTokens.scala
olafurpg: I would lean towards making most functionality available through syntactic rules as possible. I don’t think it’s extreme to split a rule between its syntactic parts and semantic. Worst case, users add two rules to their config file. However, this makes sense as long as the syntactic rules are useful on their own.

mlachkar: I think we should start with a syntactic rule, and we can decide later if you change it to a semantic one, or create another new one. I do like to group rules when they offer the same functionality: for example Remove Unused can remove 4 kind of warnings, and it quite possible to built in new capacities without requiring from the user to add new rules.

bplommer: Shall we call this one `RedundantSyntax` rather than `RedundantTokens` then?

mlachkar: Good idea!

bjaglin: :+1:

bjaglin:
Hi @bplommer - will you have some time to follow-up and document this? Thanks!

bplommer:
Sorry, I forgot all about this! I'll try to finish it off soon.

bjaglin (approved):
Let's hold the merge until the last moment so docs don't advertise the new rule before it's released.

PR #1519, opened 2022-01-11
support scala 2.13.8

bjaglin:


project/Dependencies.scala
```diff
@@ -28,7 +28,7 @@ object Dependencies {
   val nailgunV = "0.9.1"
   val scalaXmlV = "2.0.1"
   val scalaXml211V = "1.3.0" // scala-xml stops publishing for scala 2.11
-  val scalametaV = "4.4.31"
+  val scalametaV = "4.4.32"
```
bjaglin, line 31: https://github.com/scalameta/scalameta/pull/2519

.scalafix.conf
```diff
@@ -1,5 +1,5 @@
 rules = [
-  ExplicitResultTypes,
+#  ExplicitResultTypes,
```
bjaglin, line 2: temporarily disable that rule: despite [forcing the latest version of scalameta](https://github.com/scalacenter/scalafix/blob/b3752b387349c0ce07840a6394647a118fd4ed21/build.sbt#L12) for the run of scalafix in this repo, sbt-scalafix runs with the current version of scalafix which [does not whitelist 2.13.8 yet](https://github.com/scalacenter/scalafix/blob/8e64923d1a22a4dc741f278a4fc0339c4fca1e92/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L31)

project/ScalafixBuild.scala
```diff
@@ -177,6 +177,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   )
 
   private val PreviousScalaVersion: Map[String, String] = Map(
+    "2.13.8" -> "2.13.7"
```
bjaglin, line 180: help MiMa find the previous artifact (there is so and won't be any 2.13.8 artifact for the current version of scalafix modules)

PR #1526, opened 2022-01-20
add spaceteams/scalafix-rules to list of projects that cross-publish

alexanderkogan:
spaceteams/scalafix-rules was already created with cross-publishing capability, so there is no pull request per se, that added it.


bjaglin:
Thanks for your contribution!

Since your rules have been cross-published from day 1, I don't think it's relevant to add a link here since this was supposed to help people *migrate*. You might want to advertize your rules in https://github.com/scalacenter/scalafix/blob/main/docs/rules/community-rules.md though.

On another note, I see that this page would require a big overhaul to explain/document the Scala 3 strategy :thinking: cc @mlachkar

mlachkar:
Yes I agree!  I will have a look!

Issue #1538, opened 2022-02-09
Missing versions in docs


bjaglin:
https://scalacenter.github.io/scalafix/docs/users/installation.html

> Start by installing the sbt 1.x plugin in project/plugins.sbt
> ```
> // project/plugins.sbt
> addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.0.0")

https://github.com/scalacenter/scalafix/blob/0dab22c5b2f75e808d4f217cf5275de80067e054/docs/users/installation.html#L68

bjaglin:
It looks like it is broken since https://github.com/scalacenter/scalafix/commit/0dab22c5b2f75e808d4f217cf5275de80067e054

bjaglin:
The same thing happened once in the past with https://github.com/scalacenter/scalafix/commit/cfe848a1deddb4d29d52b94f758cc93026c8e1a3 (https://github.com/scalacenter/scalafix/pull/1513), so it looks like it's related to the build environment (Github Actions?) rather than the build itself.

```
$ git log -S'0.0.0' scalacenter/gh-pages
commit 0dab22c5b2f75e808d4f217cf5275de80067e054
Author: Docusaurus bot <mdoc@docusaurus>
Date:   Fri Jan 21 20:51:40 2022 +0000

    Deploy website

    Deploy website version based on 6c21e2662b0bec87e1495d14f20a775a10663ff9

commit e245382680fda4b83d9510c16d84ddc287c14dd8
Author: Docusaurus bot <mdoc@docusaurus>
Date:   Tue Jan 11 17:30:00 2022 +0000

    Deploy website

    Deploy website version based on 0a202ab467e67228a368f0b7bd4372b70d1c8086

commit cfe848a1deddb4d29d52b94f758cc93026c8e1a3
Author: Docusaurus bot <mdoc@docusaurus>
Date:   Mon Dec 20 08:05:56 2021 +0000

    Deploy website

    Deploy website version based on 2518d5ba6fa7ce2d8bda9ac7c71d1402253e58cc
```

PR #1544, opened 2022-02-10
attempt at fixing versions in user-facing docs

bjaglin:
Since [versions](https://github.com/scalacenter/scalafix/blob/6c21e2662b0bec87e1495d14f20a775a10663ff9/project/ScalafixBuild.scala#L96-L98) used in docs (and in metadata available for clients) come from https://github.com/sbt/sbt-dynver, I don't understand how this ever worked with a shallow clone, so this might help with https://github.com/scalacenter/scalafix/issues/1538

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - uses: olafurpg/setup-scala@v13
       - name: Publish ${{ github.ref }}
         run: sbt docs/docusaurusPublishGhpages


.github/workflows/deploy-website.yml
```diff
@@ -8,8 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
bjaglin, line 12: https://github.com/actions/checkout
>    \# Number of commits to fetch. 0 indicates all history for all branches and tags.
>    \# Default: 1

mlachkar: I don't understand why this number is wrong ...but let's try!

bjaglin:
It looks like it worked: https://github.com/scalacenter/scalafix/commit/c151ae682b4f9c1525674dccdc025e81be91f3d9! I have no idea why it worked without this before though... Maybe to be continued.

.github/workflows/deploy-website.yml
bjaglin: For the record the release job gets proper versions because it does an explicit unshallow https://github.com/scalacenter/scalafix/blob/6d991eb0bcaf4afd635c3c1657af59e946548ea0/.github/workflows/release.yml#L13

I guess either approach is correct. The unshallow is maybe a bit more efficient but I like that `fetch-depth` is a flag on the checkout as it's more apparent.

PR #1546, opened 2022-02-14
force usage of shaded TPrint

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1530
Should unblock https://github.com/scalacenter/scalafix/pull/1531

Injecting instances of `pprint.TPrint` to customize the docs no longer has effect as of https://github.com/scalacenter/scalafix/pull/1530, which was a silent breaking change. This commit exposes the breaking change so that rules built against the latest scalafix-core either drop usage of `TPrint` or switch to the shaded version which is used internally by metaconfig.

### DIFF
--- a/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
+++ b/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
@@ -42,11 +42,23 @@ public class ScalafixCoursier {
             String scalafixVersion,
             String scalaVersion
     ) throws ScalafixException {
-        Dependency scalafixCli = Dependency.parse(
+        List<Dependency> dependencies = new ArrayList<Dependency>();
+        dependencies.add(
+            Dependency.parse(
                 "ch.epfl.scala:::scalafix-cli:" + scalafixVersion,
                 ScalaVersion.of(scalaVersion)
+            ).withConfiguration("runtime")
         );
-        return fetch(repositories, Collections.singletonList(scalafixCli), ResolutionParams.create());
+        // Coursier does not seem to fetch runtime dependencies transitively, despite what Maven dictates
+        // https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope
+        // so to be able to retrieve the runtime dependencies of scalafix-core, we need an explicit reference
+        dependencies.add(
+            Dependency.parse(
+                "ch.epfl.scala::scalafix-core:" + scalafixVersion,
+                ScalaVersion.of(scalaVersion)
+            ).withConfiguration("runtime")
+        );
+        return fetch(repositories, dependencies, ResolutionParams.create());
     }
 
     public static List<URL> toolClasspath(
--- a/scalafix-rules/src/main/scala-2.12+/scalafix/internal/rule/TPrintImplicits.scala
+++ b/scalafix-rules/src/main/scala-2.12+/scalafix/internal/rule/TPrintImplicits.scala
@@ -1,0 +1,14 @@
+package scalafix.internal.rule
+
+import java.util.regex.Pattern
+
+import metaconfig.pprint._
+import scalafix.config.CustomMessage
+
+class TPrintImplicits {
+  implicit val tprintPattern: TPrint[List[CustomMessage[Pattern]]] =
+    new TPrint[List[CustomMessage[Pattern]]] {
+      def render(implicit tpc: TPrintColors): fansi.Str =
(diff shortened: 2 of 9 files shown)


scalafix-testkit/src/main/scala/scalafix/testkit/TestkitPath.scala
```diff
@@ -25,14 +25,6 @@ final class TestkitPath(
     val testPath: RelativePath,
     val semanticdbPath: RelativePath
 ) {
-  override def toString: String = {
-    val map = Map(
-      "input" -> input,
-      "testPath" -> testPath,
-      "semanticdbPath" -> semanticdbPath
-    )
-    pprint.PPrinter.BlackWhite.tokenize(map).mkString
```
bjaglin, line 34: Only `pprint.TPrint` has been shaded into metaconfig. I felt like other usages of `pprint` here and below were not worth adding an explicit compile-time dependency to `pprint`

scalafix-rules/src/main/scala-2.12+/scalafix/internal/rule/TPrintImplicits.scala
```diff
@@ -0,0 +1,13 @@
+package scalafix.internal.rule
+
+import java.util.regex.Pattern
+
+import metaconfig.pprint._
+import scalafix.config.CustomMessage
+
+class TPrintImplicits {
+  implicit val tprintPattern: TPrint[List[CustomMessage[Pattern]]] =
```
bjaglin, line 9: This implicit has actually no effect (neither before or after this PR) as the signature of the param it was targeting changed in https://github.com/scalacenter/scalafix/pull/907/files#diff-fa445c2f97a1bfb229caa4cd2fdc97cb9c243ddac155698a132bcbdb22970910R91. I will follow-up with a PR

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
```diff
@@ -45,7 +45,7 @@
         Dependency scalafixCli = Dependency.parse(
                 "ch.epfl.scala:::scalafix-cli:" + scalafixVersion,
                 ScalaVersion.of(scalaVersion)
-        );
+        ).withConfiguration("runtime"); // https://github.com/coursier/coursier/issues/552#issuecomment-310133532
```
bjaglin, line 48:
```
$ cs resolve ch.epfl.scala::scalafix-core:0.9.34+24-e65c6590-SNAPSHOT | grep pprint
com.geirsson:metaconfig-pprint_2.13:0.10.0:default
```

```
$ cs resolve ch.epfl.scala::scalafix-core:0.9.34+24-e65c6590-SNAPSHOT:runtime | grep pprint
com.geirsson:metaconfig-pprint_2.13:0.10.0:default
com.lihaoyi:pprint_2.13:0.6.6:default
```

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
```diff
@@ -42,11 +42,23 @@
...
-        Dependency scalafixCli = Dependency.parse(
+        List<Dependency> dependencies = new ArrayList<Dependency>();
+        dependencies.add(
+            Dependency.parse(
                 "ch.epfl.scala:::scalafix-cli:" + scalafixVersion,
                 ScalaVersion.of(scalaVersion)
+            ).withConfiguration("runtime")
         );
-        return fetch(repositories, Collections.singletonList(scalafixCli), ResolutionParams.create());
+        // Coursier does not seem to fetch runtime dependencies transitively, despite what Maven dictates
```
bjaglin, line 52: I will follow up in coursier/coursier with a repro

mlachkar (approved):

scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
bjaglin: Actually, it works as expected when using sonatype:
```
$ cs resolve -r sonatype:snapshots -r central --no-default   ch.epfl.scala:scalafix-cli_2.13.8:0.9.34+25-b98e220d-SNAPSHOT:runtime | grep com.lihaoyi:pprint
com.lihaoyi:pprint_2.13:0.6.6:runtime
```

There is however something fishy with ivy2local: https://github.com/coursier/coursier/issues/2370.

So the code added there is probably needed only for tests to pass, but it does not hurt anyway.

PR #1556, opened 2022-02-25
Scalameta 4.5.0 (was 4.4.35)

SethTisue:
In the Scala 2 community build, I noticed that 1) this upgrade hasn't happened here yet, and 2) it requires a small source change (because the deprecated `allowMethodTypes` was removed).

And then also I'm seeing failing tests downstream in simulacrum-scalafix, at https://scala-ci.typesafe.com/job/scala-2.13.x-jdk11-integrate-community-build/3535/artifact/logs/simulacrum-scalafix-build.log :

```
[simulacrum-scalafix] =======
[simulacrum-scalafix] => Diff
[simulacrum-scalafix] =======
[simulacrum-scalafix] --- obtained
[simulacrum-scalafix] +++ expected
[simulacrum-scalafix] @@ -24,9 +24,9 @@
[simulacrum-scalafix] âˆ™
[simulacrum-scalafix] -  val fk1 = Î»[FuncK[Option, List]](_.toList)
[simulacrum-scalafix] +  val fk1 = new FuncK[Option, List] { def apply[A$](a$: Option[A$]): List[A$] = a$.toList }
[simulacrum-scalafix]    val fk2 = new FuncK[Option, List] { def apply[A$](a$: Option[A$]): List[A$] = Nil }
[simulacrum-scalafix] -  val fk3 = Î»[List ~> Option](_.headOption)
[simulacrum-scalafix] +  val fk3 = new (List ~> Option) { def apply[A$](a$: List[A$]): Option[A$] = a$.headOption }
[simulacrum-scalafix]    val fk4 = new (List ~> Option) { def apply[A$](xs: List[A$]): Option[A$] = xs.lastOption }
[simulacrum-scalafix]    val fk5 = new (({ type Î»[Î±$] = Foo[Î±$, Int] })#Î» ~> Option) { def apply[A$](a$: Foo[A$, Int]): Option[A$] = None }
[simulacrum-scalafix] -  val fk6 = Î»[FuncK[List, Î»[x => List[List[x]]]]](List(_))
[simulacrum-scalafix] -  val fk7 = Î»[List ~> Option](_.headOption.map(identity))
[simulacrum-scalafix] +  val fk6 = new FuncK[List, ({ type Î»[x] = List[List[x]] })#Î»] { def apply[A$](a$: List[A$]): List[List[A$]] = List(a$) }
[simulacrum-scalafix] +  val fk7 = new (List ~> Option) { def apply[A$](a$: List[A$]): Option[A$] = a$.headOption.map(identity) }
[simulacrum-scalafix]    val fk8 = new (List ~> Option) { def apply[A$](list: List[A$]): Option[A$] = {
```

I don't know what to make of this. Regression in scalameta? Something in scalafix that needs to be adjusted for the new scalameta version? Or is it just something that the simulacrum-scalafix people (@sh0hei maybe?) need to adjust in their test suite?

Properly investigating isn't really something I can take on myself.

Regardless, opening this PR to get the ball rolling. (It isn't easy for the simulacrum-scalafix folks to investigate unless they have a scalafix release to use.)

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -29,7 +29,7 @@ object Dependencies {
   val nailgunV = "0.9.1"
   val scalaXmlV = "2.0.1"
   val scalaXml211V = "1.3.0" // scala-xml stops publishing for scala 2.11
-  val scalametaV = "4.4.35"
+  val scalametaV = "4.5.0"
   val scalatestV = "3.0.8" // don't bump, to avoid forcing breaking changes on clients via eviction
 
   val bijectionCore = "com.twitter" %% "bijection-core" % bijectionCoreV
--- a/scalafix-core/src/main/scala/scalafix/internal/util/DenotationOps.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/util/DenotationOps.scala
@@ -7,7 +7,7 @@ import scalafix.v0._
 
 object DenotationOps {
   val defaultDialect: Dialect =
-    dialects.Scala212.copy(allowMethodTypes = true, allowTypeLambdas = true)
+    dialects.Scala212.copy(allowTypeLambdas = true)
 
   def resultType(
       symbol: Symbol,


bjaglin:
I'll merge this and open a PR on simulacrum-scalafix with the resulting nightly to follow-up on this regression.

bjaglin:
I confirmed the breaking change, identified the [root cause](https://github.com/scalameta/scalameta/pull/2601) and updated the rule code in https://github.com/typelevel/simulacrum-scalafix/pull/194/commits/04f557f9b1b903da2751d7cdb7e7709edfa91ffe. However, this brings questions about compatibility.

### Rule authors

Should we signal this change to rule authors dealing with `Term.Placeholder` through a 0.10.0 release? scalafix 0.9.0 came out with scalameta 4.0.0, so I assume there has been other breaking changes in scalameta along the way. Still, I think this one is "breaking enough" to justify this.

### Running rules built against newer scalafix versions with older scalafix versions

Rules relying on `Term.AnonymousFunction` from scalameta 4.5.0 will fail at runtime when executed with older scalafix versions. That's because scalafix-core is already loaded when external rules are injected in a separate classloader, so their version of scalafix-core (and thus scalameta) is ignored. **scalafix-interfaces forces scalafix-cli's scalameta version, there is no eviction possible here.**

https://github.com/scalacenter/scalafix/blob/8bd700fd5bd46d471d78f65c6a36c628fd57bca5/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java#L69-L76

This is not new (rules matching on Scala3-specific trees such as `Term.Given` need a recent scalafix runner for example), but it's more annoying for existing rules that don't use new feature but just have to adapt to the new parsing output like https://github.com/typelevel/simulacrum-scalafix/pull/194/commits/04f557f9b1b903da2751d7cdb7e7709edfa91ffe.

To mitigate this problem, I suggest we provide an actionable error message when encountering linking exceptions ("a more recent versions of scalafix is needed to run that rule"). This will improve the scalafix user experience if/when new trees are added in scalameta in the future.

### Running rules built against older scalafix versions with newer scalafix versions

Conversely, running rules built against older versions of scalafix might miss some trees (silently I believe?) when ran on the upcoming version of scalafix since they will be exposed to unexpected `Term.AnonymousFunction`.

To mitigate that, we could issue a warning when detecting that external rules are built against a different minor version of scalafix: "running external scalafix rules built for an earlier version - correctness is not guaranteed so downgrade the scalafix version or ask the rule author to release a new version". This would be very verbose though, and seems to be very specific to https://github.com/scalameta/scalameta/pull/2601 so I would suggest not to do anything.

@mlachkar @olafurpg @tgodzik what's your take on all this?

tgodzik:
> To mitigate this problem, I suggest we provide an actionable error message when encountering linking exceptions ("a more recent versions of scalafix is needed to run that rule"). This will improve the scalafix user experience if/when new trees are added in scalameta in the future.

Could we have a requirement in the rules for the minimal Scalafix version? I would even fail in that situation if possible.

> Conversely, running rules built against older versions of scalafix might miss some trees (silently I believe?) when ran on the upcoming version of scalafix since they will be exposed to unexpected Term.AnonymousFunction (see the test failure that triggered this discussion).

It should miss silently in most cases unless any rule depends on a very specific path of tree nodes. I think most rules traverse tree until they find what is needed, so they would skip unknown nodes. We could issue a warning in this case for sure.

SethTisue:
> The problem is that this warning would then be issued for any community rule,

I think we should bite the bullet on that.

Issue #1555, opened 2022-02-25
`RemoveUnused` rewrites nested pattern binders to syntactically incorrect patterns


shonfeder:
We've just started using scalafix in our project, and it's already been very helpful! Thank you for the great work :)

We have found that `scalafix RemoveUnused` consistently produces syntactically invalid rewrites for nested pattern binders. Here's a MWE:

Given

```scala
object Example {
  def f(v: (Int, (Boolean, String))): Int = v match {
    case t @ (i, v @ (_, _)) => i
  }
}
```

scalafix rewrites to:

```scala
object Example {
  def f(v: (Int, (Boolean, String))): Int = v match {
    case  (_, _)) => i
  }
}
```

Which of course leads to a compiler error:

```
'=>' expected but ')' found.
[error]     case  (_, _)) => i
[error]                 ^
[error] one error found
```

Our setup:

```
scalaVersion := "2.12.15"
addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.34")
sbt.version=1.6.1
```

bjaglin:
Thanks for the clear report @shonfeder! I have just opened https://github.com/scalacenter/scalafix/pull/1557.

PR #1557, opened 2022-02-27
RemoveUnused: support nested Pat.Bind

bjaglin:
Fixes https://github.com/scalacenter/scalafix/issues/1555
Follows https://github.com/scalacenter/scalafix/pull/1444


scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
```diff
@@ -106,9 +106,8 @@ class RemoveUnused(config: RemoveUnusedConfig)
             Patch.replaceTree(v, "_")
           case t @ Pat.Typed(v: Pat.Var, _) if checkUnusedPatTree(t) =>
             Patch.replaceTree(v, "_")
-          case b: Pat.Bind if checkUnusedPatTree(b) =>
-            val removableTokens = b.tokens.dropRightWhile(!_.is[Token.At])
```
bjaglin, line 110: this was incorrectly picking up the nested bind when attempting to strip the LHS of the outer one

mlachkar (approved):

PR #1565, opened 2022-03-12
define & apply binary compatibility strategy

bjaglin:
In the last few weeks, 2 dependency bumps broke backward compatibility of nightlies
- metaconfig 0.10.0, at the [source](https://github.com/scalacenter/scalafix/pull/1546) & [binary](https://github.com/scalacenter/scalafix/pull/1530#issuecomment-1061174301) level because of [breaking changes in pprint](https://github.com/com-lihaoyi/PPrint/pull/72), despite [attempts to mitigate the problem](https://github.com/scalameta/metaconfig/pull/154)
- scalameta 4.5.0, at the [functional](https://github.com/typelevel/simulacrum-scalafix/pull/194) level, by [introducing a new node in the AST](https://github.com/scalameta/scalameta/pull/2601) ([detected thanks to the Scala 2 community build](https://github.com/scalacenter/scalafix/pull/1556))

Since these changes haven't made it to a release yet, that seems like a good moment to either
1. clarify compatibility guarantees for rule authors ahead of a minor bump for the upcoming release, or
1. revert the bumps for now and
   1. pin dependencies, or
   1. contribute upstream to mitigate the impact

This PR takes the path of (1).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,9 @@ jobs:
       - uses: olafurpg/setup-scala@v13
       - uses: olafurpg/setup-gpg@v3
       - run: git fetch --unshallow
+      - name: Check that major or minor was bumped upon compatibility breakage
+        if: startsWith(github.ref, 'refs/tags/v')
+        run: sbt versionCheck
       - name: Publish ${{ github.ref }}
         run: sbt ci-release
         env:
--- a/docs/developers/api.md
+++ b/docs/developers/api.md
@@ -4,9 +4,26 @@ title: API Overview
 sidebar_label: Overview
 ---
 
+## Compatibility considerations
+
+[`scalafix-core`](https://index.scala-lang.org/scalacenter/scalafix/scalafix-core)
+makes the packages below available to built-in and external rules.
+
+Its X.Y.Z version scheme follows [Early SemVer](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#early-semver-and-sbt-version-policy)
+for binary compatiblity and aims at keeping source compatibility across
+versions.
+- Running a rule built against an older X or 0.Y version of Scalafix may cause
+  either runtime errors, or false positive/negative tree selection. In that
+  case, a warning is issued when fetching the rules(s) artifact(s).
+- Running a rule built against a more recent X.Y or 0.Y.Z version of Scalafix
+  is not possible. In that case, a `ScalafixException` is raised when fetching
+  the rules(s) artifact(s).
+
+## Packages
+
 The Scalafix public API documentation is composed of several packages.
 
-## Scalafix v1
+### Scalafix v1
 
 Latest Scaladoc:
 [v@VERSION@](https://static.javadoc.io/ch.epfl.scala/scalafix-core_2.12/@VERSION@/scalafix/v1/index.html)
@@ -37,15 +54,7 @@ structures include:
   information such as trees/tokens and semantic information such as
   symbols/types/synthetics.
 
-## Scalafix v0
-
-Latest Scaladoc:
-[v@VERSION@](https://static.javadoc.io/ch.epfl.scala/scalafix-core_2.12/@VERSION@/scalafix/v0/index.html)
-
-This is a legacy API that exists to smoothen the migration for existing Scalafix
-v0.5 rules. If you are writing a new Scalafix rule, please use the v1 API.
-
-## Scalameta Trees
+### Scalameta Trees
 
 Latest Scaladoc:
 [v@SCALAMETA@](https://static.javadoc.io/org.scalameta/trees_2.12/@SCALAMETA@/scala/meta/index.html)
@@ -71,7 +80,7 @@ include:
   position". Statement position is the
 - `Defn`: the supertype for all definitions
 
-## Scalameta Tokens
+### Scalameta Tokens
 
 Latest Scaladoc:
 [v@SCALAMETA@](https://static.javadoc.io/org.scalameta/tokens_2.12/@SCALAMETA@/scala/meta/tokens/Token.html)
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -5,7 +5,8 @@ import sbt.nio.Keys._
 import sbt.plugins.JvmPlugin
 import com.typesafe.tools.mima.plugin.MimaPlugin.autoImport._
 import sbtbuildinfo.BuildInfoKey
-import sbtbuildinfo.BuildInfoPlugin.autoImport.{BuildInfoKey, _}
+import sbtbuildinfo.BuildInfoPlugin.autoImport._
+import sbtversionpolicy.SbtVersionPolicyPlugin.autoImport._
 import com.typesafe.sbt.sbtghpages.GhpagesKeys
 import sbt.librarymanagement.ivy.IvyDependencyResolution
 import sbt.plugins.IvyPlugin
@@ -177,6 +178,13 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   )
 
   private val PreviousScalaVersion: Map[String, String] = Map(
+  )
+
+  override def buildSettings: Seq[Setting[_]] = List(
+    versionPolicyIgnoredInternalDependencyVersions :=
+      Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r),
+    versionScheme := Some("early-semver"),
+    versionPolicyIntention := Compatibility.None // TODO: harden after 0.10.0
   )
 
   override def projectSettings: Seq[Def.Setting[_]] = List(
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
@@ -24,6 +24,10 @@ import scalafix.Versions
 import scalafix.cli.ExitStatus
 import scalafix.interfaces._
 import scalafix.internal.config.ScalaVersion
+import scalafix.internal.util.Compatibility
+import scalafix.internal.util.Compatibility.Compatible
+import scalafix.internal.util.Compatibility.Incompatible
+import scalafix.internal.util.Compatibility.Unknown
 import scalafix.internal.v1.Args
 import scalafix.internal.v1.MainOps
 import scalafix.internal.v1.Rules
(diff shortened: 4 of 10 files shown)


scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
```diff
@@ -76,7 +80,48 @@ final case class ScalafixArgumentsImpl(args: Args = Args.default)
...
+        // the easiest option would be to run several resolutions in isolation, which seems like a massive
+        // cost for just issuing a warning.
+        Compatibility.pvp(dependency.getVersion, Versions.nightly) match {
+          case Incompatible =>
+            throw new ScalafixException(
+              s"Scalafix version ${Versions.scalaVersion} cannot load the registered external rules, " +
+                s"please upgrade to ${dependency.getVersion} or later"
+            )
+          case Unknown =>
+            args.out.println(
```
bjaglin, line 109: sbt-scalafix sets the printstream [very late](https://github.com/scalacenter/sbt-scalafix/blob/0a4dfc6a32d67b701235876fe5658f04c9c9687a/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L390), so this gets dumped on plain stdout. It would be possible to wire a fallback sbt logger from the beginning.

bjaglin:
I'd really like your view on this @olafurpg (and I cannot assign you), thanks!

docs/developers/api.md
```diff
@@ -4,9 +4,24 @@ title: API Overview
 sidebar_label: Overview
 ---
 
+## Compatibility considerations
+
+[`scalafix-core`](https://index.scala-lang.org/scalacenter/scalafix/scalafix-core) makes the packages
+below available to built-in and external rules.
+
+Its A.B.C scheme follows [PVP](https://pvp.haskell.org/) for syntactic and semantic binary compatiblity
```
julienrf, line 12: I find the wording a bit confusing. It seems that you talk about “semantic binary compatibility” and “syntactic binary compatibility”. Although I guess you really talk about the binary compatibility of the API for syntactic and semantic rules?

```suggestion
Its A.B.C version scheme follows [PVP](https://pvp.haskell.org/) for binary compatiblity
```

docs/developers/api.md
```diff
@@ -4,9 +4,24 @@ title: API Overview
...
+and aims at keeping source compatibility across versions.
+- A or B (the major version) is increased whenever running rules built against the upcoming version
+  with an environment frozen at the current version may cause either runtime errors, or false positive
+  or  negative tree selection. A warning is issued at runtime in that case.
```
julienrf, line 16: Here too, I find the wording hard to follow.

```suggestion
- A.B (the major version) is increased whenever a binary incompatibility is introduced. Running a rule built with an older version of scalafix may cause either runtime errors, or false positive
  or  negative tree selection. A warning is issued at runtime in that case.
```

julienrf (commented):
Thank you @bjaglin for adding this!

I suggest also declaring the versioning scheme in your build definition (in the projects that are published, and for which you do perform binary compatibility checks):

~~~ scala
versionScheme := Some("pvp")
~~~

That may help build tools (at least sbt) to correctly resolve conflicting versions of scalafix artifacts.

docs/developers/api.md
bjaglin: I agree this was confusing, particularly in the context of scalafix where we have syntactic and semantic rules - I'll amend with your wording.

For the record, what I meant was that a major version will be bumped even though no linkage error may happen, but just because the behavior/semantics changed. Binary compatibility, as described in https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html for example, seems to focus on link guarantees, so I wanted to highlight the silent functional changes as a factor.

docs/developers/api.md
bjaglin: Thanks for your suggestion. I am adding "major" (older **major** version), and updating the sentence below to follow the same pattern "Running a rule xxxx",

project/ScalafixBuild.scala
```diff
@@ -179,6 +180,11 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   private val PreviousScalaVersion: Map[String, String] = Map(
   )
 
+  override def buildSettings: Seq[Setting[_]] = List(
+    versionScheme := Some("pvp"),
+    versionPolicyIntention := Compatibility.None // TODO: harden after 0.10.0
```
bjaglin, line 185: I verified that with `Compatibility.BinaryCompatible`, we get
```
[error] Incompatibilities with dependencies of scalafix-core:0.9.34
[error]   com.geirsson:metaconfig-core_2.13: incompatible version change from 0.9.15 to 0.10.0 (compatibility: package versioning policy)
[error]   com.geirsson:metaconfig-typesafe-config_2.13: incompatible version change from 0.9.15 to 0.10.0 (compatibility: package versioning policy)
[error]   com.lihaoyi:fansi_2.13: incompatible version change from 0.2.14 to 0.3.0 (compatibility: package versioning policy)
[error]   com.lihaoyi:pprint_2.13: missing dependency
[error]   org.scalameta:common_2.13: incompatible version change from 4.4.32 to 4.5.0 (compatibility: package versioning policy)
[error]   org.scalameta:parsers_2.13: incompatible version change from 4.4.32 to 4.5.0 (compatibility: package versioning policy)
[error]   org.scalameta:scalameta_2.13: incompatible version change from 4.4.32 to 4.5.0 (compatibility: package versioning policy)
[error]   org.scalameta:trees_2.13: incompatible version change from 4.4.32 to 4.5.0 (compatibility: package versioning policy)
[error] Dependencies of module scalafix-core:0.9.34+56-5a4c18d3+20220319-2243-SNAPSHOT break the intended compatibility guarantees 'Compatibility.BinaryCompatible' (see messages above). You have to relax your compatibility intention by changing the value of versionPolicyIntention.
```

.github/workflows/release.yml
```diff
@@ -11,6 +11,9 @@ jobs:
       - uses: olafurpg/setup-scala@v13
       - uses: olafurpg/setup-gpg@v3
       - run: git fetch --unshallow
+      - name: Check that major was bumped if bin compat was broken
+        if: startsWith(github.ref, 'refs/tags/v')
```
bjaglin, line 15: don't block publishing of snapshots

bjaglin:
@julienrf thanks for your feedback. I have just added sbt-version-policy in a second commit, your review is very much welcome on that one too.

project/ScalafixBuild.scala
bjaglin: opened https://github.com/scalacenter/scalafix/pull/1568/commits/79259b475cf4278a0440b9a2ebaa5b5e3bc0e3c6 as draft for a reminder

docs/developers/api.md
julienrf: > a major version will be bumped even though no linkage error may happen, but just because the behavior/semantics changed

In such a case, would it be possible to not change the semantics of the existing methods but to add new methods instead?

julienrf:
Unfortunately, sbt-version-policy expects `early-semver` instead of `pvp` as a versioning scheme. It’s behavior is undefined for `pvp`, so it’s better to not use it in that case (or consider using `early-semver` instead of `pvp`).

docs/developers/api.md
bjaglin: That's what we aim for, but we can control dependencies that are [part of the API](https://scalacenter.github.io/scalafix/docs/developers/api.html#scalameta-trees): https://github.com/scalameta/scalameta/pull/2601 is what resulted in a (silent) breaking change, identified by the [community build](https://github.com/scalacenter/scalafix/pull/1556).

bjaglin:
> Unfortunately, sbt-version-policy expects `early-semver` instead of `pvp` as a versioning scheme. It’s behavior is undefined for `pvp`, so it’s better to not use it in that case

I had a quick look at the code, and it looks like the early-semver assumption is only in `versionCheck`, is that right? `versionPolicyCheck` seems to be fine. Following up in https://github.com/scalacenter/sbt-version-policy/issues/123.

> or consider using `early-semver` instead of `pvp`

I guess I could also go for that if the fix above is not trivial. I had picked pvp over early-semver mostly of out laziness when (re)implementing https://github.com/bjaglin/scalafix/commit/5a4c18d3c47e3e2fecb83b39433a34a63a3dcaa6#diff-d96ec3acf34c2b31f8eefeb8bdd342bb1365fb47091c75b69298b7b49c928aafR14 (since I couldn't find a versioning lib cross-published for 2.11).

julienrf:
> it looks like the early-semver assumption is only in `versionCheck`, is that right? `versionPolicyCheck` seems to be fine

That might be the case, but using sbt-version-policy with anything else than early-semver is undertested. I recommend using early-semver, but I am willing to re-consider supporting other versioning schemes in sbt-version-policy (since you are not the only one asking for it).

bjaglin:
@julienrf I amended to advertize/honor `early-semver`. One argument to stay on PVP would be that scalameta (the main dependency) is on PVP and gets updated quite often. After scalafix reaches 1.0, each time scalameta will get bumped, we'll have to manually assess if it should result in a minor or patch release for scalafix-core. But I guess it will be more meaningful for Scalafix users?

project/ScalafixBuild.scala
julienrf: You probably also want to add:

~~~ scala
versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r)
~~~

See https://github.com/scalacenter/sbt-version-policy#how-to-integrate-with-sbt-dynver

julienrf (approved):

julienrf:
> One argument to stay on PVP would be that scalameta (the main dependency) is effectively on PVP and gets updated quite often. After scalafix reaches 1.0, each time scalameta will get bumped, we'll have to manually assess if it should result in a minor or patch release for scalafix-core.

I guess the defensive strategy would be to always bump the minor version number of scalafix. I don’t know if this is much an issue.

Also, why do you say that scalameta is on PVP? According to https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md it seems to use semantic versioning.

bjaglin:
> I guess the defensive strategy would be to always bump the minor version number of scalafix. I don’t know if this is much an issue.

Indeed, my only concern was that there is no tooling to enforce that (since sbt-version-policy's `versionPolicyReportDependencyIssues` won't help if I understand correctly)

> Also, why do you say that scalameta is on PVP? According to https://github.com/scalameta/scalameta/blob/main/COMPATIBILITY.md it seems to use semantic versioning.

Thanks for the pointer, I hadn't seen this. `x.y.z1 must be compatible with x.y.z2` was not strictly honored as [a lot of stuff got added in 4.4.x
](https://github.com/scalameta/scalameta/compare/v4.4.0...v4.4.35#diff-40552a8e530f2d1e2f5fc18d305466a7eb7a0af727ad0c88f75aa67d30958253), but https://github.com/scalameta/scalameta/pull/2504/files explains the rationale. I'll follow-up with a PR to define `versionScheme` there.

Thanks for your insights!

julienrf:
> Indeed, my only concern was that there is no tooling to enforce that (since sbt-version-policy's `versionPolicyReportDependencyIssues` won't help if I understand correctly)

Now that you say it, `versionPolicyReportDependencyIssues` should fail when you bump scalameta, if `versionPolicyIntention` is set to `BinaryAndSourceCompatible` and `libraryDependencySchemes` contains `"org.scalameta" %% "scalameta" % "pvp"`. If this is not the case, please report a bug in sbt-version-policy.

PR #1568, opened 2022-03-19
Post 0.10.0

bjaglin:


bjaglin:
```
[error] Incompatibilities with dependencies of scalafix-core:0.10.0
[error]   com.lihaoyi:pprint_2.12: missing dependency
```
It looks like sbt-version-policy does not handle the [`Runtime`](https://github.com/scalacenter/scalafix/blob/660877e92000c20b266e1e754cd01089d01c598d/build.sbt#L80) configuration properly.

bjaglin:
It seems to be the same problem as https://github.com/coursier/coursier/issues/2370. However, forcing an ivy `DependencyResolution` does not seem to help.
```scala
    versionPolicyInternal.versionPolicyDependencyResolution := {
      // Try to fallback to ivy to avoid https://github.com/coursier/coursier/issues/2370
      IvyDependencyResolution(ivySbt.value.configuration)
    },
```

bjaglin:
Both ivy and coursier seem to get runtime dependencies in the compile report
```
test := {
  //val depRes = sbt.librarymanagement.ivy.IvyDependencyResolution(ivySbt.value.configuration)
  val depRes = dependencyResolution.value
  val update0 = depRes.update(depRes.wrapDependencyInModule("ch.epfl.scala" % "scalafix-core_2.12" % "0.10.0"), updateConfiguration.value, (update / unresolvedWarningConfiguration).value, sLog.value)
  println(update0.right.get.configuration(Compile).get.modules.filter(_.toString.contains("com.lihaoyi:pprint")))
}
```

bjaglin:
Worked around it by excluding pprint from the check

project/ScalafixBuild.scala
```diff
@@ -175,10 +175,12 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
   )
 
   override def buildSettings: Seq[Setting[_]] = List(
+    // https://github.com/sbt/sbt/issues/5568#issuecomment-1094380636
+    versionPolicyIgnored += "com.lihaoyi" %% "pprint",
     versionPolicyIgnoredInternalDependencyVersions :=
       Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r),
     versionScheme := Some("early-semver"),
-    versionPolicyIntention := Compatibility.None // TODO: harden after 0.10.0
+    versionPolicyIntention := Compatibility.BinaryCompatible
```
bjaglin, line 183: since we are still in 0.x.y, we'll need to bump a minor anyway for the next release (and not a "semantic" patch), so there is not interest in checking `BinaryAndSourceCompatible`

mlachkar (approved):

PR #1578, opened 2022-03-27
`ExplicitResultTypes`: backquote types when needed

### LINKED ISSUE #1219: ExplicitResultTypes should backquote identifiers if necessary

danicheg:
This fixes #1219. I was struggled with a working fix for all type-symbols but didn't succeed. I'd be glad to get help in moving this onward.

### DIFF
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
@@ -114,6 +114,9 @@ object Identifier {
     else s
   }
 
-  def backtickWrapWithoutCheck(typeName: String, typeArgs: Option[String]): String =
+  def backtickWrapWithoutCheck(
+      typeName: String,
+      typeArgs: Option[String]
+  ): String =
     typeArgs.fold('`' + typeName + '`')(ta => '`' + typeName + '`' + ta)
 }
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
@@ -114,9 +114,5 @@ object Identifier {
     else s
   }
 
-  def backtickWrapWithoutCheck(
-      typeName: String,
-      typeArgs: Option[String]
-  ): String =
-    typeArgs.fold('`' + typeName + '`')(ta => '`' + typeName + '`' + ta)
+  def backtickWrapWithoutCheck(typeName: String): String = '`' + typeName + '`'
 }
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
@@ -106,6 +106,7 @@ object Identifier {
   def backtickWrap(name: Global#Name): String = {
     backtickWrap(name.decoded.trim)
   }
+
   def backtickWrap(s: String): String = {
     if (s.isEmpty) "``"
     else if (s(0) == '`' && s.last == '`') s
@@ -113,4 +114,6 @@ object Identifier {
     else s
   }
 
+  def backtickWrapWithoutCheck(typeName: String, typeArgs: Option[String]): String =
+    typeArgs.fold('`' + typeName + '`')(ta => '`' + typeName + '`' + ta)
 }
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
@@ -169,6 +169,35 @@ class ScalafixGlobal(
       .filterKeys(_ != NoSymbol)
       .toMap
 
+  // currently wrapping in backticks supported only for traits and case classes
+  private def backtickify(
+      tpe: Type,
+      loop: (Type, Option[ShortName]) => Type,
+      withPrefix: Boolean
+  ): Type = {
+    val supportedTypes: Symbol => Boolean =
+      s => s.isCaseClass || s.isTrait
+
+    tpe match {
+      case TypeRef(pre, sym, args)
+          if supportedTypes(sym) && Identifier.needsBacktick(sym.decodedName) =>
+        val rawPrettyName = Identifier.backtickWrapWithoutCheck(sym.decodedName)
+        val prefix = if (withPrefix) pre.prefixString else ""
+        if (args.isEmpty) {
+          new PrettyType(prefix + rawPrettyName)
+        } else {
+          val typeArgsStr =
+            args.map(arg => loop(arg, None)).map(_.toString()).mkString(", ")
+          new PrettyType(prefix + s"$rawPrettyName[$typeArgsStr]")
+        }
+      case TypeRef(p, sym, args) =>
+        val pre = if (withPrefix) p else NoPrefix
+        TypeRef(pre, sym, args.map(arg => loop(arg, None)))
+      case _ =>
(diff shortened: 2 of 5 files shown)


bjaglin:
Thanks for the PR, that's already a step forward! I didn't have a look at it yet, but can you maybe elaborate what challenges you faced with other type symbols? You can open a draft PR with failing tests as a starting point, it's always valuable for others to pick up where you left off.

scalafix-rules/src/main/scala/scalafix/internal/rule/CompilerTypePrinter.scala
```diff
@@ -109,7 +109,26 @@ class CompilerTypePrinter(g: ScalafixGlobal, config: ExplicitResultTypesConfig)(
     }
     val preProcessed = preProcess(toLoop, inverseSemanticdbSymbol).widen
     val shortType = g.shortType(preProcessed, history)
-    val short = shortType.toString()
+    val shortTypeStr = shortType.toString()
+
+    // currently wrapping in backticks supported only for case classes
+    val isCaseClass = shortType.typeSymbol.isCaseClass
+    val typeSymbolName = shortType.typeSymbol.decodedName
```
danicheg, line 116: @bjaglin For example, I don't get why in case of traits `shortType.typeSymbol.decodedName` is producing `<none>`. Also, it was easy to break the current semantic when no backticks are needed  (function types, path-dependent types, etc).

mlachkar:
Hello! 
thanks for your contribution. 
This code should not be outside of the method `g.shortType(preProcessed, history)`, because this method is recursive. 
There is a test example that would break with your fix:

```
object ExplicitResultTypesBackQuote {
  case class `Foo-Bar`(i: Int)
}

object TestBackQuote {
  val foobar = ExplicitResultTypesBackQuote.`Foo-Bar`(42)
}
```
it would output:
```
object TestBackQuote {
  val foobar: `ExplicitResultTypesBackQuote.Foo-Bar` = ExplicitResultTypesBackQuote.`Foo-Bar`(42)

}
```

It's tempting to fix the issue, by taking the `shortType` and pattern match on it, but it can't work, because you need to do that recursively as it's done in the method `def shortType`.  You can check this PR, where we discussed how to fix issues on this rule https://github.com/scalacenter/scalafix/pull/1495

If you have any question, don't hesitate!

danicheg:
@mlachkar Thanks for the review. I moved impl to `ScalafixGlobal#shortType`, so now traits and nested case classes/traits support backticks.

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
```diff
@@ -169,6 +169,33 @@ class ScalafixGlobal(
...
       .toMap
 
+  // currently wrapping in backticks supported only for traits and case classes
+  private def backtickify(
+    tpe: Type,
+    loop: (Type, Option[ShortName]) => Type,
+    withPrefix: Boolean
+  ): Type = {
+    val supportedTypes: Symbol => Boolean =
+      s => s.isCaseClass || s.isTrait
```
bjaglin, line 179: I am still curious to know what's the limitation with other simple types like a regular class?

scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBackQuote.scala
```diff
@@ -0,0 +1,33 @@
...
+    case class `Quux !`(i: Int)
+  }
+
+  val foobar = `Foo-Bar`(42)
+  val fooTrait = null.asInstanceOf[`Foo-Trait`]
+  val barbaz = `Bar[,?! $42Baz`[List, Option](List(42), Some(""))
+  val qux = Nested.Inner.`Qux Qux`(42)
+  val quxqux = Nested.Inner.`Qux.Qux`[List](List(42))
+  val quux = `Quux !`(42)
+  val nestedTrait = null.asInstanceOf[Nested.Inner.`Nested-Trait`[List]]
```
bjaglin, line 32: maybe that's something we want to exercise also:
```suggestion
  val nestedTrait = null.asInstanceOf[Nested.Inner.`Nested-Trait`[`Foo-Bar`]]
```

bjaglin (commented):
Thanks for the iteration, it's shaping up! Don't forget to run `./bin/scalafmt`.

scalafix-rules/src/main/scala/scala/meta/internal/pc/Identifier.scala
```diff
@@ -106,11 +106,17 @@ object Identifier {
...
   def backtickWrap(s: String): String = {
     if (s.isEmpty) "``"
     else if (s(0) == '`' && s.last == '`') s
     else if (needsBacktick(s)) '`' + s + '`'
     else s
   }
 
+  def backtickWrapWithoutCheck(
+      typeName: String,
+      typeArgs: Option[String]
```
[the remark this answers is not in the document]
danicheg: Ouch. Yes. It's left from the previous iteration. Thanks! But the name `backtickWrap` is already taken, so I left it as is.

scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBackQuote.scala
danicheg: Good point, done.

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
danicheg: Changing `isCaseClass` -> `isClass` leads to failed test on `(=> Any) => Any` case. So, it brings redundant backticks. And this could mean there are other cases when it also would lead to redundant backticks (which are missing in unit tests).

danicheg: Though, maybe it's not so bad to add backticks for the `(=> Any) => Any`...  I'm just not sure wouldn't bring other uncontrolled changes in current behavior.

danicheg:
Thank you for your review, @bjaglin!

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
bjaglin: > So, it brings redundant backticks

I removed the `supportedTypes` filter to have a look at it, and the problem actually seems to be that we can't rely on `decodedName` for by-names.
```
###########> Diff       <###########

=======
=> Diff
=======
--- obtained
+++ expected
@@ -133,3 +133,3 @@
∙
-    val byNameType: `<byname>`[Any] => Any = ??? : ((=> Any) => Any)
+    val byNameType: (=> Any) => Any = ??? : ((=> Any) => Any)
     // FIXME: https://github.com/twitter/rsc/issues/144
         
[info] - test/explicitResultTypes/RscCompat.scala *** FAILED *** (273 milliseconds)
[info]   see above (AbstractSemanticRuleSuite.scala:91)
```

As a result, I took the liberty to add b2c88064a00cafe496cd84bf40187ff84caaf3c0 to your branch to unlock support for classes as well. Let me know what you think!

danicheg: Whoa! That's tremendous. Thank you so much.

mlachkar (approved):

tgodzik:
Thanks for the great work @danicheg ! I managed to backport the changes to Metals also https://github.com/scalameta/metals/pull/3829/files

PR #1582, opened 2022-03-29
Add Rule for Databricks(Hygiene rules)

i10416:


bjaglin (commented):
Thanks for the PR! However, I am hesitant to advertize these rules at the moment as they are very basic (2 of them could be implemented via configuration through `DisableSyntax.regex`). I suggest we revisit the addition if/when they grow in scope and/or complexity.

PR #1588, opened 2022-04-04
only warn when loading rules built with newer core

bjaglin:
Follow-up of https://github.com/scalacenter/scalafix/pull/1565 (https://github.com/scalacenter/scalafix/commit/c4c1f1e686367f1bcf7d627526d93416fa8de07b#r69611050)

### DIFF
--- a/docs/developers/api.md
+++ b/docs/developers/api.md
@@ -14,10 +14,12 @@ for binary compatiblity and aims at keeping source compatibility across
 versions.
 - Running a rule built against an older X or 0.Y version of Scalafix may cause
   either runtime errors, or false positive/negative tree selection. In that
-  case, a warning is issued when resolving the rules(s) artifact(s).
+  case, a warning calling the user to use a more recent version of the rule
+  or downgrading Scalafix is issued when resolving the rules(s) artifact(s).
 - Running a rule built against a more recent X.Y or 0.Y.Z version of Scalafix
-  is not possible. In that case, a `ScalafixException` is raised when resolving
-  the rules(s) artifact(s).
+  is also a potential source of incompatibility. In that case, a warning
+  calling the user to upgrade Scalafix is issued when resolving the rules(s)
+  artifact(s).
 
 ## Packages
 
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
@@ -101,12 +101,14 @@ final case class ScalafixArgumentsImpl(args: Args = Args.default)
           dependency.getVersion,
           Versions.nightly
         ) match {
-          case Incompatible =>
-            throw new ScalafixException(
-              s"Scalafix version ${Versions.nightly} cannot load the registered external rules, " +
-                s"please upgrade to ${dependency.getVersion} or later"
+          case TemptativeUp(compatibleRunWith) =>
+            args.out.println(
+              s"""Loading external rule(s) built against a recent version of Scalafix (${dependency.getVersion}).
+                |This might not be a problem, but in case you run into unexpected behavior, you
+                |should upgrade Scalafix to ${compatibleRunWith}.
+              """.stripMargin
             )
-          case Temptative(compatibleRunWith) =>
+          case TemptativeDown(compatibleRunWith) =>
             args.out.println(
               s"""Loading external rule(s) built against an old version of Scalafix (${dependency.getVersion}).
                 |This might not be a problem, but in case you run into unexpected behavior, you
--- a/scalafix-cli/src/main/scala/scalafix/internal/util/Compatibility.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/util/Compatibility.scala
@@ -7,8 +7,8 @@ object Compatibility {
 
   case object Compatible extends Compatibility
   case object Unknown extends Compatibility
-  case object Incompatible extends Compatibility
-  case class Temptative(compatibleRunWith: String) extends Compatibility
+  case class TemptativeUp(compatibleRunWith: String) extends Compatibility
+  case class TemptativeDown(compatibleRunWith: String) extends Compatibility
 
   private val Snapshot = """.*-SNAPSHOT""".r
   val XYZ: Regex = """^([0-9]+)\.([0-9]+)\.([0-9]+).*""".r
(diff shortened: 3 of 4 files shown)


scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
```diff
@@ -101,12 +101,14 @@ final case class ScalafixArgumentsImpl(args: Args = Args.default)
           dependency.getVersion,
           Versions.nightly
         ) match {
-          case Incompatible =>
-            throw new ScalafixException(
-              s"Scalafix version ${Versions.nightly} cannot load the registered external rules, " +
-                s"please upgrade to ${dependency.getVersion} or later"
+          case TemptativeUp(compatibleRunWith) =>
```
bjaglin, line 104: to clarify: this is code that will be triggered once external rules built against 0.10.1 and later are available

docs/developers/api.md
```diff
@@ -14,10 +14,12 @@ for binary compatiblity and aims at keeping source compatibility across
...
   either runtime errors, or false positive/negative tree selection. In that
-  case, a warning is issued when resolving the rules(s) artifact(s).
+  case, a warning calling the user to use a more recent version of the rule
+  or downgrading Scalafix is issued when resolving the rules(s) artifact(s).
 - Running a rule built against a more recent X.Y or 0.Y.Z version of Scalafix
-  is not possible. In that case, a `ScalafixException` is raised when resolving
-  the rules(s) artifact(s).
+  is also a potential source of incompatibility. In that case, a warning
+  calling the user to upgrade Scalafix is issued when resolving the rules(s)
+  artifact(s).
```
mlachkar, line 22: I find this case easier to deal with than using a more recent rule.
The second requires rule author to publish a new version of their rule for each major version. In the other hand, we don't publish a major version very often!

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
```diff
@@ -101,12 +101,14 @@ final case class ScalafixArgumentsImpl(args: Args = Args.default)
...
+            args.out.println(
+              s"""Loading external rule(s) built against a recent version of Scalafix (${dependency.getVersion}).
+                |This might not be a problem, but in case you run into unexpected behavior, you
+                |should upgrade Scalafix to ${compatibleRunWith}.
+              """.stripMargin
```
mlachkar, line 109: The warn messages are super clear! 
We can definitely turn that to an exception if many users complains that with scalafix 0.10.x, rules in 0.10.x x>0, it breaks. It would mean we don't pay too much attention to warn messages and in this case an exception is better.

mlachkar:
When we are going to publish `0.10.0`, most external rules will be in `0.9.x` , which will trigger the warn message of `Tentative.Down`.
Usually users that upgrade are the one that uses scala steward, so maybe the warn message won't be that visible if the CI is green.
Do we expect people not to upgrade to this version, but the rule authors to use this version to publish their rules ?

mlachkar (approved):

bjaglin:
> When we are going to publish `0.10.0`, most external rules will be in `0.9.x` , which will trigger the warn message of `Tentative.Down`.

Indeed, that's what I tried to explain in the [release notes](https://github.com/scalacenter/scalafix/releases/tag/v0.10.0-RC1). I updated the message in the draft https://github.com/scalacenter/scalafix/releases/tag/untagged-7be1f5cfc653f2fefbfc to reflect the changes in this PR.

>  Do we expect people not to upgrade to this version, but the rule authors to use this version to publish their rules ?

With or without Steward, it's likely that people first upgrade sbt-scalafix (and other clients, as I opened PR for them to target 0.10.0 as early as possible) since rules will lag a bit behind, so they will get a warning. The value of the warning is that it will encourage rule authors to build against 0.10.x, so that we can potentially drop support for 0.9.x-built rules in 0.11.x or 1.x. (https://github.com/scalacenter/scalafix/pull/1546 does the job but is hacky).

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
bjaglin: > We can definitely turn that to an exception if many users complains that with scalafix 0.10.x, rules in 0.10.x x>0, it breaks

What do you mean? 0.10.x runtimes will already be in the wild. we can't retroactively turn the warning into an exception.

docs/developers/api.md
bjaglin: >  The second requires rule author to publish a new version of their rule for each major version

With this PR and the exception turned warning, it _encourages_ them (since they will get report from users complaining about warnings, as discussed in https://github.com/scalacenter/scalafix/pull/1588#issuecomment-1088347809) but does not _require_ them.

bjaglin:
@tgodzik it might be hard to follow the various discussions, but your feedback would be welcome here (with both your Metals and scalameta hat).

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
mlachkar: What I mean, we can still turn it into an exception in 0.10.5 for example, if we think a warning is not enough. I don't know if it's ok to that. 


This message is only displayed when we have a tentativeUp starting `0.10.0`. 
The only case will be: A rule is published with `0.10.1` and the user has upgraded the rule, but not the scalafix version which is still `0.10.0`.
Maybe throwing an exception is ok in that case.

mlachkar:
I think @tgodzik is on vacation, but maybe we can have @ckipp01 opinion. Thanks in advance!

bjaglin:
> I think @tgodzik is on vacation, but maybe we can have @ckipp01 opinion. Thanks in advance!

On the metals front, my (unchecked) assumption is that a given metals release pins both scalafix & the organize import rule, in which case versioning/compatibility is not a concern. If that's not the case (user-defined rules?), then it might be worth looking at how/if metals would handle warnings.

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
bjaglin: > What I mean, we can still turn it into an exception in 0.10.5 for example, if we think a warning is not enough. I don't know if it's ok to that.

We can, but we would want to be more verbose in case 0.10.5 (or 0.11.0) brings silent changes important enough to prevent users to use an earlier version to run rules built/tested against it. But it would be too late as 0.10.4 and earlier are out!

mlachkar:
So on metals, your assumption is correct. 
Users can't define yet their rules that can be run through metals.

ckipp01:
> > I think @tgodzik is on vacation, but maybe we can have @ckipp01 opinion. Thanks in advance!
> 
> On the metals front, my (unchecked) assumption is that a given metals release pins both scalafix & the organize import rule, in which case versioning/compatibility is not a concern. If that's not the case (user-defined rules?), then it might be worth looking at how/if metals would handle warnings.

You're correct, we have both of those pinned, so if I'm following this all correctly this shouldn't impact stuff at all since the only external rule we're implemented with is the organize imports one. In the future _if_ we ever add the functionality to run user-defined rules, then we'll cross that bridge when we get there 😆

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixArgumentsImpl.scala
bjaglin: As discussed, let's for go for warn now. We can revisit for 1.x, maybe with an exception if build is newer by major or minor only, keeping a warn for patch diff.

PR #1596, opened 2022-04-11
pprint is only required on -cli for 0.9.x compat

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1546

Tested against regressions via https://github.com/scalacenter/scalafix/blob/c5633c67a449b0ed8658c26910805f6893485007/scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala#L84-L95 (`com.nequissimus::sort-imports:0.5.2` is built against [0.9.x](https://github.com/NeQuissimus/sort-imports/blob/v0.5.2/project/plugins.sbt#L2))

### DIFF
--- a/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
+++ b/scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
@@ -49,23 +49,10 @@ public class ScalafixCoursier {
             String scalafixVersion,
             String scalaVersion
     ) throws ScalafixException {
-        List<Dependency> dependencies = new ArrayList<Dependency>();
-        dependencies.add(
-            Dependency.parse(
-                "ch.epfl.scala:::scalafix-cli:" + scalafixVersion,
-                ScalaVersion.of(scalaVersion)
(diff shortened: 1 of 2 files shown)


scalafix-interfaces/src/main/java/scalafix/internal/interfaces/ScalafixCoursier.java
```diff
@@ -49,23 +49,10 @@ public static List<URL> scalafixCliJars(
...
-            Dependency.parse(
-                "ch.epfl.scala::scalafix-core:" + scalafixVersion,
-                ScalaVersion.of(scalaVersion)
-            ).withConfiguration("runtime")
-        );
-        return toURLs(fetch(repositories, dependencies, ResolutionParams.create()));
+        Dependency scalafixCli = Dependency
+				.parse("ch.epfl.scala:::scalafix-cli:" + scalafixVersion, ScalaVersion.of(scalaVersion))
+				.withConfiguration("runtime");
+        return toURLs(fetch(repositories, Collections.singletonList(scalafixCli), ResolutionParams.create()));
```
mlachkar, line 55: Nice that this code is simplified

PR #1600, opened 2022-04-15
Fix issues with backquotes in several case as seen in Metals

tgodzik:
I backported the changes to backquote types to Metals and found some issues. Some of that might not be needed for Scalafix, but it's better to be on the safe side.

Related https://github.com/scalameta/metals/pull/3829

### DIFF
--- a/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
@@ -177,8 +177,11 @@ class ScalafixGlobal(
     tpe match {
       case TypeRef(pre, sym, args)
           if Identifier.needsBacktick(sym.decodedName) &&
-            sym != definitions.ByNameParamClass // `=> T` is OK
-          =>
+            sym != definitions.ByNameParamClass && // `=> T` is OK
+            !sym.isPackageObject &&
+            !definitions.isRepeated(sym) &&
+            !sym.isAbstractType &&
+            sym.owner != definitions.ScalaPackageClass =>
         val rawPrettyName = Identifier.backtickWrapWithoutCheck(sym.decodedName)
         val prefix = if (withPrefix) pre.prefixString else ""
         if (args.isEmpty) {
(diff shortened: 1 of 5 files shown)


scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
```diff
@@ -177,8 +177,11 @@ class ScalafixGlobal(
     tpe match {
       case TypeRef(pre, sym, args)
           if Identifier.needsBacktick(sym.decodedName) &&
-            sym != definitions.ByNameParamClass // `=> T` is OK
-          =>
+            sym != definitions.ByNameParamClass && // `=> T` is OK
+            !sym.isPackageObject &&
+            !definitions.isRepeated(sym) &&
+            !sym.isAbstractType &&
```
bjaglin, line 183: it looks like only this new condition is tested - is that what you mean by "Some of that might not be needed for Scalafix, but it's better to be on the safe side" ?

bjaglin (approved):
I see there is still a pending discussion in https://github.com/scalameta/metals/pull/3829, but since this is addressing a regression from https://github.com/scalacenter/scalafix/pull/1578, I suggest we take that in as-is.

scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala
tgodzik: Repeated seems only to happen in in signatures of methods, which is not reproducible in here, so maybe I can remove it?

tgodzik: `sym.isPackageObject` was needed for Iterable, but it doesn't seem to break here, just in metals

tgodzik:
> I see there is still a pending discussion in [scalameta/metals#3829](https://github.com/scalameta/metals/pull/3829), but since this is addressing a regression from #1578, I suggest we take that in as-is.

I will take a further look into the MEtals PR also

mlachkar (approved):

PR #1623, opened 2022-06-24
Add an option to only add explicit result types to implicit definitions

OlegYch:
This is useful for scala3 migration since it requires explicit return types for implicits


bjaglin:
Related to https://github.com/scalacenter/scala3-migrate/issues/334

bjaglin:
Thanks for the PR, new configuration flag LGTM :+1: 

Your test seems to be specific to 2.13, I guess you can have a simpler one that compiles also on 2.11 & 2.12?

bjaglin (approved):
Thanks!

mlachkar (approved):

PR #1624, opened 2022-06-27
return custom error when no rule was requested

bjaglin:


scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -382,6 +382,7 @@ object MainOps {
           case _ =>
             ex.printStackTrace(args.args.out)
         }
+        println(e)
```
tanishiking, line 385: debug print?

mlachkar-da (approved):

Issue #1583, opened 2022-03-30
scala 3: ExplicitResultTypes


mlachkar:
# Adapt Explicit result types for Scala 3 

## The current state
`ExplicitResultTypes` rule calls the scala 2 presentation compiler to infer types for each `val`, `def` or `var`. To make this rule available for scala 3, we need to re-write it to interface this time with the Scala 3 presentation compiler, in other words, `ExplicitResultTypes` rule for Scala 3, need to be written in Scala 3. 

## How to implement ?

```mermaid
flowchart LR
A(Scalafix-cli) -->|depends on| B(Scalafix-rules)
```
The actual design of Scalafix requires that the module `scalafix-rules` uses the same Scala Version than the `Scalafix-cli` module. To be able to use the Scala 3 presentation compiler, the rule needs to be rewritten in Scala 3 and therefore `scalafix-cli` needs to be able to load a Scala 3 rule.

### Most likely implementation

 1. Create a Java interface for `ExplicitResultTypes` that provides the minimum needed methods for it to work. The interface would look like below:
```
public interface ExplictResultTypesInterface {
  List<PositionWithString> fix(Optional<Object>: global, Path: relativePathToSemanticDbFile)
  Object createGlobal(scalacClasspath, scalacOptions)
}
```
- Currently `ExplicitResultTypes` rule needs a `SemanticDocument`, that cannot be passed as a parameter to the java method, so each implementation of this rule will need to recreate it from its relative path.
- The `fix` method would return a list of `Position` and `String`, which should be enough to create a `scalafix.Patch`. 
- The `global` instance needs also to be created differently depending on the Scala Version. We can maybe stop using `ScalafixGlobal` and use directly `scala.tools.nsc.interactive.Global` (in Scala 2) and `dotty.tools.dotc.interactive.InteractiveDriver` (in Scala 3). `ScalafixGlobal` is almost only wrapping the `Global` from the presentation compiler. In this Java interface, it would be an Object, and then each implementation would cast it as a `Global` instance or `InteractiveDriver` instance depending on the Scala version.

[...]

Issue #1316, opened 2021-01-21
Scala 3 support


bjaglin:
Opening this for visibility and to track the upcoming work for supporting Scala 3
- [x] running syntactic & semantic rules against Scala 3 sources
  - scalafix
     - ~new parameter (optional for backward compatibility) to feed the input/output dialect~ reuse `scalaVersion`, see https://github.com/scalacenter/scalafix/pull/1392#pullrequestreview-655259487
     - support scala3 as input&output in testkit
       - sbt 1.5 must be used to compile input/output
       - bootstrap `scala-3[.0.0-RC2]` folder with all revelevant tests (maybe check if sbt can remove the RC suffix?) & move scala 2.12 syntax currently in `scala` to `scala-2.12.-x` folders
       - update https://github.com/scalacenter/scalafix.g8
     - document [what is not yet supported in Scala3](https://github.com/scalacenter/GoogleSummerOfCode2021/blob/main/README.md#add-synthetics-and-symbol-information-for-semanticdb-in-scala-3)
       - try out the most common community rules to prioritize adding support for what's needed 
   - sbt-scalafix
     - ~provide the  input/output dialect based on the build `scalaVersion`~ always provide `scalaVersion`, see https://github.com/scalacenter/scalafix/pull/1392#pullrequestreview-655259487
     -  support `scalaVersion := "3"` in `scalafixEnable`
     - document that to use scala3, the user must rely on `SemanticdbPlugin` to [abstract usage of semanticdb-scalac](https://scalacenter.github.io/scala-3-migration-guide/docs/tutorials/sbt-migration.html#semanticdbhttpsscalametaorgdocssemanticdbguidehtml)
- [ ] cross-building scalafix-cli , scalafix-core & scalafix-rules for Scala 3?
   - not a priority as the 2 other usages of the compiler in scalafix are not critical
     - on-demand source rules compilation can remain on 2.12 since they are not going to use the presentation compiler anyway
     - ~[on-demand compilation when semanticdb info is missing should probably be removed ?](https://github.com/scalacenter/scalafix/pull/1348#pullrequestreview-604155677)~
       - done in https://github.com/scalacenter/scalafix/pull/1376 
   - blocked by
     - [x] https://github.com/scalameta/metaconfig/pull/148 
     - [ ] https://github.com/scalameta/scalameta/issues/2485  
       - we are exploring using scalameta_2.13 in scalafix-core_3, see https://github.com/scalacenter/scalafix/issues/1583
[...]

PR #1562, opened 2022-03-12
better error for external rules ran with old scalafix-cli version

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1530

`metaconfig-pprint` brought by `metaconfig-core` 0.10.0 will not be available to external rules when classloaded by an old `scalafix-cli` version as the runtime `scalafix-core` and its dependencies are driven by `scalafix-cli` dependencies, not the rule dependencies.

We obviously have no control over old `scalafix-cli` versions, but since the reference to the potentially-missing type is from a macro-generated block, this wraps it to provide a more actionable error message.

PR #1629, opened 2022-07-06
Scalafix-core compiled in Scala 3

rvacaru:
This PR is the first step towards adapting ExplicitResultTypes for scala3 #1583 and part of the plan for scala3 support #1316. The idea is to cross compile all scalafix modules to scala3 and then introduce a new `ExplicitResultTypes` implementation for scala3.

Screenshot 2022-07-22 at 12 03 53

In this first step we're cross compiling `scalafix-core` to scala3


scalafix-core/src/main/scala-3/scalafix/util/Compat.scala
```diff
@@ -0,0 +1,9 @@
+package scalafix.util
+
+import scala.collection.immutable.IndexedSeq
+
+object Compat {
+  type View[T] = collection.SeqView[T] //, IndexedSeq[T]]
+  type SeqView[T] = collection.SeqView[T] //, IndexedSeq[T]]
```
rvacaru, line 7: if I provide `[T], IndexedSeq[T]]` it fails compiling saying there are too many type args and it only expects one [A] type arg.. would you know how to fix this?

rvacaru: Thanks @armanbilge, now it's more clear

build.sbt
```diff
@@ -64,18 +64,34 @@ lazy val core = projectMatrix
...
     moduleName := "scalafix-core",
     buildInfoSettingsForCore,
-    libraryDependencies ++= List(
-      scalameta,
-      googleDiff,
-      collectionCompat
+    scalacOptions --= (
+      if (isScala3.value)
+        Seq("-P:semanticdb:synthetics:on")
+      else Nil
```
bjaglin, line 70: this is repeated in many places, can we move it to a settings block covering all current usages?

build.sbt
```diff
@@ -64,18 +64,34 @@ lazy val core = projectMatrix
...
     ),
+    libraryDependencies ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) => List(
+          scalameta
+            .exclude("com.lihaoyi", "sourcecode_2.13")
+            .exclude("org.scala-lang.modules", "scala-collection-compat_2.13"),
+          googleDiff
```
bjaglin, line 78: extract this common one to another block

build.sbt
```diff
@@ -309,6 +325,11 @@ lazy val unit = projectMatrix
     }
   )
   .defaultAxes(VirtualAxis.jvm)
+  .jvmPlatform(
+    scalaVersions = Seq(scala3),
+    axisValues = Seq(TargetAxis(scala3)),
+    settings = Seq()
+  )
```
bjaglin, line 332: let's remove this for now, focusing on getting a green build

project/Dependencies.scala
```diff
@@ -53,7 +53,7 @@ object Dependencies {
   val nailgunServer = "com.martiansoftware" % "nailgun-server" % nailgunV
   val scalaXml = "org.scala-lang.modules" %% "scala-xml" % scalaXmlV
   val scalaXml211 = "org.scala-lang.modules" %% "scala-xml" % scalaXml211V
-  val scalameta = "org.scalameta" %% "scalameta" % scalametaV
+  val scalameta = ("org.scalameta" %% "scalameta" % scalametaV).cross(CrossVersion.for3Use2_13)
```
bjaglin, line 56:
```suggestion
  // https://github.com/scalameta/scalameta/issues/2485
  val scalameta = ("org.scalameta" %% "scalameta" % scalametaV).cross(CrossVersion.for3Use2_13)
```

scalafix-core/src/main/scala-2/scalafix/internal/util/MetaconfigCompatMacros.scala
```diff
@@ -1,3 +1,4 @@
+
 package scalafix.internal.util
```
bjaglin, line 2: It looks like https://github.com/scalacenter/scalafix/pull/1562 was only partially reverted - why? Thi seems to be used in the scala-2 version of `ConfigRulePatches` but should not longer be the case (so you won't need a separate scala-3 version).

For visibility, let's spin off the revert of https://github.com/scalacenter/scalafix/pull/1562 in a separate PR with the rationale (major clients have been released and scala-steward is back so the risk of using an old scalafix-cli with a rule compiled against a recent scalafix-core is limited).

project/ScalafixBuild.scala
```diff
@@ -184,7 +184,10 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     Compile / unmanagedSourceDirectories ++= {
       val sourceDir = (Compile / sourceDirectory).value
       CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, n)) if n >= 12 => Seq(sourceDir / "scala-2.12+")
+        case Some((2, n)) if n == 13 => Seq(sourceDir / "scala-2.13")
+        case Some((2, n)) if n == 12 => Seq(sourceDir / "scala-2.12")
+        case Some((2, n)) if n == 11 => Seq(sourceDir / "scala-2.11")
+        case Some((3, _))  => Seq(sourceDir / "scala-3")
```
bjaglin, line 190: AFAIK, all these are built-in in sbt, why did you add them here?

project/ScalafixBuild.scala
```diff
@@ -184,7 +184,10 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     Compile / unmanagedSourceDirectories ++= {
       val sourceDir = (Compile / sourceDirectory).value
       CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, n)) if n >= 12 => Seq(sourceDir / "scala-2.12+")
```
bjaglin, line 187: I still see files under these directories, so the removal of this makes them dead code (and probably compilation errors). This avoid duplication between 2.12/2.13 source directories, so why not keep it?

scalafix-core/src/main/scala-3/scalafix/internal/util/PrettyType.scala
```diff
@@ -0,0 +1,732 @@
+package scalafix.internal.util
```
bjaglin, line 1: These new files in `scala-3` are HUGE - so much duplicated code will be very costly to keep in sync. Instead try to keep the common code in `src/mains/scala` and extract the differences in a smaller file (`Compat.scala`).

build.sbt
```diff
@@ -64,18 +64,34 @@ lazy val core = projectMatrix
...
-      googleDiff,
-      collectionCompat
+    scalacOptions --= (
+      if (isScala3.value)
+        Seq("-P:semanticdb:synthetics:on")
+      else Nil
     ),
+    libraryDependencies ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) => List(
```
bjaglin, line 74: `if (isScala3.value)` is more readable if you just to branch based on the major version

scalafix-core/src/main/scala/scalafix/internal/util/Pretty.scala
```diff
@@ -5,6 +5,7 @@ import scala.meta.Lit
 import org.typelevel.paiges.Doc
 import scalafix.internal.v1.Types
 import scalafix.v1._
+import meta.XtensionSyntax
```
bjaglin, line 8: is meta the root package or is that a relative import?

scalafix-core/src/main/scala/scalafix/util/TokenClasses.scala
```diff
@@ -3,6 +3,7 @@ package scalafix.util
 import scala.meta.classifiers.Classifier
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
+import meta.classifiers.XtensionClassifiable
```
bjaglin, line 6: is meta the root package or is that a relative import?

bjaglin:
Another note: since `core3` is an orphan modules not used by any other, I don't think it's covered by CI, so you should add an explicit `unit3/compile` in https://github.com/scalacenter/scalafix/blob/59ed8c053a79cadfe3ef2f83a10561df7170a192/project/ScalafixBuild.scala#L131-L134
This will apply for all xxx3 modules until we move unit2_12Target3 to unit3.

scalafix-core/src/main/scala/scalafix/util/TokenClasses.scala
rvacaru: it's a relative import, will fix all of them

project/Dependencies.scala
rvacaru: added

scalafix-core/src/main/scala-2/scalafix/internal/util/MetaconfigCompatMacros.scala
rvacaru: Here's the separate PR for the revert, feel free to edit the description if needed: https://github.com/scalacenter/scalafix/pull/1632

bjaglin: merged, you can rebase

build.sbt
rvacaru: factored out

scalafix-core/src/main/scala-3/scalafix/internal/util/PrettyType.scala
rvacaru: refactored similarly as Compat, defining different types for different scala versions

scalafix-core/src/main/scala-3/scalafix/util/TokenList.scala
```diff
@@ -0,0 +1,77 @@
+package scalafix.util
+
+import scala.meta.tokens.Token
+import scala.meta.tokens.Tokens
+import scala.meta.classifiers.XtensionClassifiable
+
+import scalafix.util.Compat._
```
bjaglin, line 7: the idea of Compat classes is that
- they have different implementations depending on the version
- they are referenced in code common to all versions

In that case, can't you try to keep `TokenList` generic? I see the original one was moved to `scala-2` without changes - instead it should remain in `scala` with minor modifications.

scalafix-core/src/main/scala/scalafix/internal/util/PrettyType.scala
```diff
@@ -172,7 +173,7 @@ class PrettyType private (
         Nil,
         Pat.Var(Term.Name(info.displayName)) :: Nil,
         None,
-        q"???"
+        "???".parse[Term].get
```
bjaglin, line 176: why did you have to change that? The `.get` is suspicious, what about
```suggestion
        Term.Name("???")
```

project/ScalafixBuild.scala
```diff
@@ -130,6 +130,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     },
     commands += Command.command("ci-3") { s =>
       "unit2_12Target3/test" ::
+      "unit3/compile" ::
```
bjaglin, line 133:
```suggestion
      "core3/compile" ::
```

scalafix-core/src/main/scala-2/scalafix/internal/config/ConfigRulePatches.scala
```diff
@@ -0,0 +1,25 @@
...
+    addGlobalImports: List[AddGlobalImport] = Nil,
+    removeGlobalImports: List[RemoveGlobalImport] = Nil
+) {
+  def all: List[Patch] =
+    replaceSymbols ++ addGlobalImports ++ removeGlobalImports
+}
+
+object ConfigRulePatches {
+  implicit val surface: Surface[ConfigRulePatches] =
+    MetaconfigCompatMacros.deriveSurfaceOrig[ConfigRulePatches]
```
bjaglin, line 21: something went wrong with your rebase - this file and `MetaconfigCompatMacros` shouldn't be there, and `ConfigRulePatches` should remain in `scala`

build.sbt
```diff
@@ -26,6 +26,12 @@ def inferJavaHome() = {
   Some(actualHome)
 }
 
+lazy val scala3ScalacOption = Def.setting (
```
bjaglin, line 29: there is a much cleaner way to do this:
- since the decision must be taken at the project level and not at the build level, remove https://github.com/scalacenter/scalafix/blob/59ed8c053a79cadfe3ef2f83a10561df7170a192/build.sbt#L8
- add the opposite of this (i.e. adding `-P:semanticdb:synthetics:on` if not scala3) in https://github.com/scalacenter/scalafix/blob/59ed8c053a79cadfe3ef2f83a10561df7170a192/project/ScalafixBuild.scala#L180
- you can now get rid of all the customizations!

build.sbt
```diff
@@ -64,18 +70,31 @@ lazy val core = projectMatrix
...
     buildInfoSettingsForCore,
-    libraryDependencies ++= List(
-      scalameta,
-      googleDiff,
-      collectionCompat
-    ),
+    scalacOptions --= (scala3ScalacOption.value),
+    libraryDependencies += {
+      googleDiff
+    },
```
bjaglin, line 76:
```suggestion
    libraryDependencies += googleDiff
```

bjaglin (commented):
Last round of comments, and we'll be good to go (provided `core3/compile` is green!)

build.sbt
rvacaru: having the reverse logic of adding the option only when needed is more readable, i like it!

scalafix-core/src/main/scala-3/scalafix/util/TokenList.scala
rvacaru: my bad, looking at it again the problem was that explicitly defining types `View[T]` in the below methods prevented the `Compat` classes to do their work:

```
  def leadingSpaces(token: Token): View[T] =
    leading(token).takeWhile(_.is[Token.Space])

  def trailingSpaces(token: Token): View[T] =
    trailing(token).takeWhile(_.is[Token.Space])

```

by removing the types it now compiles for the different scala versions

scalafix-core/src/main/scala-3/scalafix/util/Compat.scala
```diff
@@ -0,0 +1,9 @@
+package scalafix.util
+
+import scala.collection.immutable.IndexedSeq
+
+object Compat {
```
bjaglin, line 5: unreferenced?

rvacaru: I don't think so, also for the other `Compat` classes (at least in VS code) I don't get any references on the class level. Moreover if I remove the class I get a compilation error:
```
sbt:scalafix> core3/compile
[info] compiling 125 Scala sources to /Users/razvan.vacaru@TMNL.nl/srepo/scalafix/scalafix-core/target/jvm-3/classes ...
[error] -- [E008] Not Found Error: /Users/razvan.vacaru@TMNL.nl/srepo/scalafix/scalafix-core/src/main/scala/scalafix/util/TokenList.scala:7:21 
[error] 7 |import scalafix.util.Compat._
[error]   |       ^^^^^^^^^^^^^^^^^^^^
[error]   |value Compat is not a member of scalafix.util, but could be made available as an extension method.
[error]   |
[error]   |The following import might make progress towards fixing the problem:
[error]   |
[error]   |  import sourcecode.Text.generate
[error]   |
[error] one error found
[error] (core3 / Compile / compileIncremental) Compilation failed
```

bjaglin: oh, sorry, I thought you had introduced a new Compat class, but this is just the scala-3 variant of an existing one :+1:

bjaglin (approved):

PR #1643, opened 2022-07-22
Scalafix-rules cross compiled in scala3

rvacaru:
Ready for review:
- moved rules `ExplicitResultTypes` (with all related dependencies) and `ProcedureSyntax` in `scala-2`
  - did not check if also other rules should be moved to `scala-2` 
- provided another `TPrintImplicits` in `scala-3`, exact copy as the one in `scala-2.12+`
- added `rules3/compile` in `ci-3` check


bjaglin (approved):

bjaglin:
rebased against https://github.com/scalacenter/scalafix/pull/1629 to ease the review (ignore my previous approval!)

bjaglin (approved):
Nice!

We'll have to reflect on how to have different versions of https://github.com/scalacenter/scalafix/blob/main/scalafix-rules/src/main/resources/META-INF/services/scalafix.v1.Rule based on the scala version, as the presence of the reference to a non-existing `ProcedureSyntax` will cause a runtime failure for https://github.com/scalacenter/scalafix/blob/d6645dab43845f3ab1e7a577719a812b082f6c14/scalafix-core/src/main/scala/scalafix/internal/v1/Rules.scala#L101-L108

But this can be handled later, when working on `unit3`.

Issue #1593, opened 2022-04-10
Break up `unit` the test monolith


bjaglin:
Having all tests in a single project brings challenges
- It's unclear which ones rely on testsInput/shared (see https://github.com/scalacenter/scalafix/pull/1528/files#r846795199)
- The suite is quite long, especially because of the prerequisite of `ScalafixSuite` (cross-publishing other modules)

Suggestion
- Many of them can distributed in existing projects, colocated with the SUT
- Integration tests between cli/reflect/core, etc could be in a given module/configuration
- Integration tests depending on shared & testInputs could be in another one

Issue #1172, opened 2020-06-19
scalafix-testkit users cannot use scalatest >= 3.2.0


bjaglin:
See https://github.com/scalacenter/scalafix/pull/1083

Experienced in https://github.com/scalacenter/sbt-scalafix/pull/131: scalatest 3.1.x users could still use scalafix-testkit as long as they didn't treat deprecations as warnings, but scalatest 3.2.x cannot as deprecated classes have been removed.

bjaglin:
Maybe we could publish several scalafix-testkit versions, one for each breaking change of scalatest?

bjaglin:
> Is there a stable ScalaTest API that we could use that's consistent between 3.0 and 3.1?

Good point, I'll have a look.

>  I don't know when we will invest in doing the 3.1.x upgrade

Out of curiosity, why isnt' https://github.com/scalatest/autofix/tree/master/3.1.x enough for a smooth migration? Other libraries built against scalatest like scalafix maybe?

> We could publish the 3.2.x artifacts with an additional _scalatest3.2 suffix and keep the current 3.0.x artifact unchanged (no suffix).

That's what I had in mind, but I agree that should be the last resort.

PR #1650, opened 2022-07-28
WIP - Scalafix-Reflect cross compiled to scala3

rvacaru:
WIP pr for having `reflect3` module


scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,88 @@
+package scalafix.internal.reflect
```
rvacaru, line 1: this is a first stub, disregard as it needs to be reimplemented with the new scala3 compiler APIs

scalafix-reflect/src/main/scala/scalafix/internal/reflect/RuleInstrumentation.scala
```diff
@@ -15,26 +15,26 @@ object RuleInstrumentation {
...
     object LambdaRule {
       def unapply(arg: Term): Boolean = arg match {
-        case q"Rule.syntactic($_)" => true
-        case q"Rule.semantic($_)" => true
-        case q"Rule.syntactic($_)" => true
-        case q"Rule.semantic($_)" => true
+        case Term.Select(Type.Name("Rule"), Term.Name("syntactic")) => true // q"Rule.syntactic($_)" => true
+        case Term.Select(Type.Name("Rule"), Term.Name("semantic")) => true
+        case Term.Select(Type.Name("Rule"), Term.Name("syntactic")) => true
+        case Term.Select(Type.Name("Rule"), Term.Name("semantic")) => true
```
mlachkar-da, line 37: you can delete this code I think

scalafix-reflect/src/main/scala/scalafix/internal/reflect/RuleInstrumentation.scala
```diff
@@ -15,26 +15,26 @@ object RuleInstrumentation {
...
 
         // v0
-        case Template(_, init"Rewrite" :: _, _, _) => true
-        case Template(_, init"Rule($_)" :: _, _, _) => true
-        case Template(_, init"SemanticRewrite($_)" :: _, _, _) => true
-        case Template(_, init"SemanticRule($_, $_)" :: _, _, _) => true
+        case Template(_, Init(Type.Name("Rewrite"), _, _) :: _, _, _) => true
+        case Template(_, Init(Type.Name("Rule"), _, List(List(_))) :: _, _, _) => true
+        case Template(_, Init(Type.Name("SemanticRewrite"), _, List(List(_))) :: _, _, _) => true
+        case Template(_, Init(Type.Name("SemanticRule"), _, List(List(_, _))) :: _, _, _) => true
```
mlachkar-da, line 21: @tgodzik can you please confirm it's correct. We are rewriting quasiqotes to be able to compile in scala 3. thank you

scalafix-reflect/src/main/scala/scalafix/internal/reflect/RuleInstrumentation.scala
```diff
@@ -15,26 +15,26 @@ object RuleInstrumentation {
...
     object LambdaRule {
       def unapply(arg: Term): Boolean = arg match {
-        case q"Rule.syntactic($_)" => true
-        case q"Rule.semantic($_)" => true
-        case q"Rule.syntactic($_)" => true
-        case q"Rule.semantic($_)" => true
+        case Term.Apply(Term.Select(Type.Name("Rule"), Term.Name("syntactic")), _)  => true // q"Rule.syntactic($_)" => true
+        case Term.Apply(Term.Select(Type.Name("Rule"), Term.Name("semantic")), _) => true
+        case Term.Apply(Term.Select(Type.Name("Rule"), Term.Name("syntactic")), _) => true
+        case Term.Apply(Term.Select(Type.Name("Rule"), Term.Name("semantic")), _) => true
```
rvacaru, line 37: to be confirmed if this block can be actually removed

scalafix-reflect/src/main/scala/scalafix/internal/reflect/RuleInstrumentation.scala
tgodzik, line 21: I run it in a worksheet and got:
```

import scala.meta._
init"Rewrite"
// res0: Init = Init(tpe = Type.Name(value = "Rewrite"), name = _, argss = List())
init"Rule(a)"
// res1: Init = Init(tpe = Type.Name(value = "Rule"), name = _, argss = List(List(Term.Name(value = "a"))))
init"SemanticRewrite(a)"
// res2: Init = Init(tpe = Type.Name(value = "SemanticRewrite"), name = _, argss = List(List(Term.Name(value = "a"))))
init"SemanticRule(a, b)"
// res3: Init = Init(
//   tpe = Type.Name(value = "SemanticRule"),
//   name = _,
//   argss = List(List(Term.Name(value = "a"), Term.Name(value = "b")))
// )
```

So I would say that we should tweak it a little bit:
```suggestion
        case Template(_, Init(Type.Name("Rewrite"), _, Nil) :: _, _, _) => true
        case Template(_, Init(Type.Name("Rule"), _, List(List(_))) :: _, _, _) => true
        case Template(_, Init(Type.Name("SemanticRewrite"), _, List(List(_))) :: _, _, _) => true
        case Template(_, Init(Type.Name("SemanticRule"), _, List(List(_, _))) :: _, _, _) => true
```

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,93 @@
...
+  // settings.unchecked.value = true // enable detailed unchecked warnings
+  // settings.outputDirs.setSingleOutput(target)
+  // settings.classpath.value = classpath
+
+  private lazy val reporter: StoreReporter = new StoreReporter()
+  private val compiler: Compiler = new Compiler()
+  private val classLoader: AbstractFileClassLoader =
+    new AbstractFileClassLoader(target, this.getClass.getClassLoader)
+
+  @throws(classOf[CompilationException])
```
rvacaru, line 38: here I'm changing the signature of `compile` by declaring  `throw CompilationException`, is it ok?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,93 @@
...
+  def compile(input: Input): Configured[ClassLoader] = {
+    val run: Run = compiler.newRun
+
+    run.compileSources(
+      List(
+        new SourceFile(
+          dotty.tools.io.AbstractFile.getFile(input.path),
```
rvacaru, line 45: we are using both `dotty.tools.io.AbstractFile` and `scala.reflect.io.AbstractFile`

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,93 @@
...
+        } finally {
+          stream.close()
+        }
+        AbsolutePath(tmp)
+      } else {
+        AbsolutePath(Paths.get(u.toURI))
+      }
+    }
+    paths.toList
+  }
```
rvacaru, line 91: this didn't change from the original scala2 source, it should be refactored out and reused across scala versions probably

rvacaru: done, in `RuleCompilerClasspath`

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
bjaglin: personal taste, but I would import this as top level with an alias (`DottyAbstractFile`) to make the dependency clear.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
bjaglin: There are not checked exceptions in scala so I am not sure this will change the signature - did you check?

But anyway, why throwing an exception when you have a `metaconfig.Configured` allowing you to capture errors?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,71 @@
...
+
+      if (reporter.errorCount > 1) {
+        val errMsg = reporter.allErrors(0).getMessage()
+        throw new CompilationException(
+          defaultMsg + "\nOriginal message: " + errMsg,
+          cause
+        )
+      }
+      throw new CompilationException(
+        defaultMsg,
```
bjaglin, line 64: "Multiple" ?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,71 @@
...
+        new SourceFile(
+          dotty.tools.io.AbstractFile.getFile(input.path),
+          input.chars
+        )
+      )
+    )
+
+    if (reporter.hasErrors) {
+      val defaultMsg = "Multiple compilation errors occurred for Scala 3, " +
+        "please try to use a CLI version which compiles into Scala 2"
```
bjaglin, line 53: This would be quite abstract/hard to understand for an end-user.

What about something like:
> Error compiling rule(s) from source using Scala 3 compiler; to use the Scala 2.x compiler instead, use the corresponding scalafix-cli artifact or force scalafixScalaBinaryVersion to 2.x in your build tool.

It's not perfect as scalafixScalaBinaryVersion is a sbt-scalafix concept, but
1. judging by the number of tickets/requests, sbt-scalafix is by far the most used client so far
1. build tools will have something similar (although maybe not named exactly the same) if they allow picking up a specific scalafix scala versiom.

build.sbt
```diff
@@ -114,13 +114,20 @@ lazy val reflect = projectMatrix
...
-    )
+    libraryDependencies ++= {
+      if (!isScala3.value)
+        List(
+          "org.scala-lang" % "scala-compiler" % scalaVersion.value,
+          "org.scala-lang" % "scala-reflect" % scalaVersion.value
+        )
+      else
+        List(
+          "org.scala-lang" %% "scala3-compiler" % "3.1.3"
```
bjaglin, line 125:
```suggestion
          "org.scala-lang" %% "scala3-compiler" % scala3
```
or maybe even
```suggestion
          "org.scala-lang" %% "scala3-compiler" % scalaVersion.value
```

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,71 @@
...
+class RuleCompiler(
+    classpath: String,
+    target: AbstractFile = new VirtualDirectory("(memory)", None)
+) {
+  // we are not configuring any settings!
+  // private val settings = new Settings()
+  // settings.deprecation.value = true // enable detailed deprecation warnings
+  // settings.unchecked.value = true // enable detailed unchecked warnings
+  // settings.outputDirs.setSingleOutput(target)
+  // settings.classpath.value = classpath
```
bjaglin, line 31: irrelevant with dotty's API?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,71 @@
...
+          cause
+        )
+      }
+      throw new CompilationException(
+        defaultMsg,
+        cause
+      )
+    }
+
+    Configured.Ok(classLoader)
```
bjaglin, line 69: I am not familiar with dotty, but I wonder how the compiled sources are made available to that classloader since I don't see it passed anywhere during the compilation phase?

bjaglin (commented):
Great to see the first little bits of dotty in here!

However, compilation alone won't give many guarantees on this kind of code/PR, so I am wondering if we should
1. tackle part of https://github.com/scalacenter/scalafix/issues/1593 first to extract some tests relevant to -reflect to -reflect!
1.  just accept that we'll merge something that might be completely broken (and we'll figure it out only in one or 2 weeks when cross-compiling/running `unit3`)
1. accept that this will be a larger PR than the others, including `scalafix-reflect_3`, `scalafix-cli_3` and finally `unit3` where we will know if this code works.

I would tend to favor (3), as long as this PR maintains a good git hygiene. Any other opinion?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: the main reason why I'm throwing directly an exception is because we discussed that in case of compilation errors we want to inform the user with a message. I probably misunderstood how we should send the message. 
I'll change this and use `metaconfig.Configured` with this message. Then the exception will be gone as well

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: good point, I'll check this

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: I've found a way to provide such settings, will remove the commented lines once we think RuleCompiler is good enough

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,88 @@
...
+  // settings.classpath.value = classpath
+  private val settings = "-unchecked" :: "-deprecation" :: Nil
+  private val driver = new InteractiveDriver(settings.distinct)
+  private val reporter: StoreReporter = new StoreReporter()
+  private given ctx: FreshContext = driver.currentCtx.fresh
+
+  ctx
+    .setReporter(reporter)
+    .setSetting(ctx.settings.outputDir, DottyAbstractFile.getFile(target.path))
+    .setSetting(ctx.settings.classpath, classpath)
```
rvacaru, line 45: I've found examples of dotty usage from which I've implemented the setup of the context and later the usage of the compiler / run for scala3:
- [scalameta](https://github.com/scalameta/mdoc/blob/c52156730d3bdd485dd213b9d195402bbe255174/mdoc/src/main/scala-3/mdoc/internal/markdown/MarkdownCompiler.scala#L56)
- [scala compiler-benchmark](https://github.com/scala/compiler-benchmark/blob/90f1b17f570bd66562a473f9d3de3f698e0fde28/compilation/src/main/dotc/scala/tools/benchmark/BenchmarkDriver.scala)
- [dotty SnippetCompiler](https://github.com/lampepfl/dotty/blob/8bd527ed54bed91294373f7e20ba73565b9324ff/scaladoc/src/dotty/tools/scaladoc/snippets/SnippetCompiler.scala)

Now it looks more reasonable

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,88 @@
...
+        )
+        .notOk
+    )
+    if (!errors.isEmpty) 
+      errors :+ ConfError.message(
+            "Error compiling rule(s) from source using Scala 3 compiler; " +
+            "to use the Scala 2.x compiler instead, use the corresponding " +
+            "scalafix-cli artifact or force scalafixScalaBinaryVersion " +
+            "to 2.x in your build tool"
+            ).notOk
```
rvacaru, line 81: here I'm not sure about the order of the errors, and if this error will actually be the first (which is what we want)

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: I don't have a strong opinion on this tbh, I don't see any problem with requiring users to set a `scalafixScalaBinaryVersion` as well. Could be useful to document this config somewhere tho.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: classpath it's now included in the initial context settings for dotty

rvacaru:
> Great to see the first little bits of dotty in here!
> 
> However, compilation alone won't give many guarantees on this kind of code/PR, so I am wondering if we should
> 
> 1. tackle part of #1593 first to extract some tests relevant to -reflect to -reflect!
> 2. just accept that we'll merge something that might be completely broken (and we'll figure it out only in one or 2 weeks when cross-compiling/running `unit3`)
> 3. accept that this will be a larger PR than the others, including `scalafix-reflect_3`, `scalafix-cli_3` and finally `unit3` where we will know if this code works.
> 
> I would tend to favor (3), as long as this PR maintains a good git hygiene. Any other opinion?

I don't like option 2 either, but then wouldn't it be the same for `core3`, `rules3` that we've already merged? I just want to understand if there's a difference and what.
Between 1 and 3 my thought is what's gonna be more fun or where am I gonna learn more? 

1 seems like creating a separate `unit-reflect` module and moving reflect tests from `unit` into this one. On Another note I'd like at some point to write some unit tests if possible (not the ones using input/output classes)

About 3 I don't mind the pr being bigger, but there's also ways to mitigate it via either branching off from this branch (`reflect3`) or by squashing commits and reviewing them one by one (eg. one commit for `reflect3`, one for `cli3`, etc)

build.sbt
```diff
@@ -137,13 +137,18 @@ lazy val cli = projectMatrix
     moduleName := "scalafix-cli",
     isFullCrossVersion,
     libraryDependencies ++= Seq(
-      java8Compat,
```
rvacaru, line 132: `java8Compat` dependency couldn't be found for `cli3/compile`, hence removing it

scalafix-cli/src/main/scala-3/scalafix/internal/v1/TPrintImplicits.scala
```diff
@@ -0,0 +1,39 @@
+package scalafix.internal.v1
```
rvacaru, line 1: copied from the `scala-2.12+` folder

build.sbt
```diff
@@ -188,21 +201,24 @@ lazy val testsOutput = projectMatrix
...
-    libraryDependencies ++= Seq(
-      googleDiff,
-      scalatest
-    )
+    libraryDependencies += googleDiff,
+    libraryDependencies ++= {
+      if (!isScala3.value)
+        Seq(scalatest)
+      else
+        Seq("org.scalatest" %% "scalatest" % "3.2.13")
```
rvacaru, line 217: a different version of scalatest has to be provided, otherwise the dependency can't be found:
- `sbt.librarymanagement.ResolveException: Error downloading org.scalatest:scalatest_3:3.0.8`

I'm not sure if we can use the same version for every scala version

bjaglin: For now, let's pick the latest scalatest for scala3 (can you please track it as `scalatest3V` in `Dependencies.scala` ?) and keep the old version for scala 2.x (we can align them in the next breaking release if needed).

scalafix-cli/src/main/scala-3/scalafix/internal/v1/TPrintImplicits.scala
bjaglin: What about including [scala-2.12+ folders](https://github.com/scalacenter/scalafix/blob/c31ffa416c3ab4241e0849dcfd204b2e4b32336e/project/ScalafixBuild.scala#L207) for scala 3? there might be cases where it's not what we want, but if it works for now, I'd favor that over duplicating a file.

build.sbt
```diff
@@ -129,13 +137,18 @@ lazy val cli = projectMatrix
...
     ),
     libraryDependencies ++= {
-      if (isScala211.value) Seq()
+      if (!isScala3.value)
+        Seq(java8Compat)
+      else
+        Seq()
+    },
+    libraryDependencies ++= {
+      if (isScala211.value || isScala3.value) Seq()
```
bjaglin, line 151: :+1:

build.sbt
```diff
@@ -114,13 +114,21 @@ lazy val reflect = projectMatrix
...
+      else
+        List(
+          "org.scala-lang" %% "scala3-compiler" % scalaVersion.value,
+          "org.scala-lang" %% "scala3-library" % scalaVersion.value
+        )
+    }
   )
   .defaultAxes(VirtualAxis.jvm)
-  .jvmPlatform(buildScalaVersions)
+  .jvmPlatform(buildScalaVersions :+ scala3, Seq(), p => p)
```
bjaglin, line 131: nit: not new to this PR, but why is different overload used when adding scala3 to the list?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
bjaglin: are you sure this has an effect? by having a quick look at https://github.com/scalameta/mdoc/blob/c52156730d3bdd485dd213b9d195402bbe255174/mdoc/src/main/scala-3/mdoc/internal/markdown/MarkdownCompiler.scala#L56-L72, it seems that `ctx` is immutable and `setSetting` is returning a copy (discarded above)

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
bjaglin: `:+` appends so it does not look right. You might want to check https://github.com/scalacenter/scalafix/blob/main/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala to see how `Configured.notOk` is handled since that's the entry point.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    ConfError
+      .fromResults(errors.toSeq)
```
bjaglin, line 77: Something is weird here as errors looks like a `Seq[Configured[_]]`. Shouldn't you remove the `notOk` above and use https://github.com/scalameta/metaconfig/blob/466f7cf86f6ad658312fa132f742178827d5c19c/metaconfig-core/shared/src/main/scala/metaconfig/ConfError.scala#L216 ?

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,81 @@
...
+    val errors = reporter.allErrors.map(error => 
+      ConfError
+        .message(error.getMessage)
+        .atPos(
+          if (error.position.isPresent()) {
+            val errPos = error.position.get()
+            Position.Range(input, errPos.start(), errPos.end())
+          } else
+              Position.None
+        )
```
bjaglin, line 65: the error coming from dotty is a source file position while `ConfError` expects a position in the config file
```suggestion
```

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,81 @@
...
+        new SourceFile(
+          DottyAbstractFile.getFile(input.path),
+          input.chars
+        )
+      )
+    )
+
+    val errors = reporter.allErrors.map(error => 
+      ConfError
+        .message(error.getMessage)
```
bjaglin, line 58: is that a multiline error? I guess we'll need unit tests to make sure this is not spamming the error logs for nothing

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,81 @@
...
+    classpath: String,
+    target: AbstractFile = new VirtualDirectory("(memory)", None)
+) {
+  // we are not configuring any settings!
+  // private val settings = new Settings()
+  // settings.deprecation.value = true // enable detailed deprecation warnings
+  // settings.unchecked.value = true // enable detailed unchecked warnings
+  // settings.outputDirs.setSingleOutput(target)
+  // settings.classpath.value = classpath
+  private val settings = "-unchecked" :: "-deprecation" :: Nil
```
bjaglin, line 30: I am not familiar with compiler options but aren't these making compilation stricter? Not sure what was the original intent in the Scala 2 counterpart, but in the context of Scala 3 I think we should be as permissive as possible.

rvacaru: In this case the two options don't make the compilation stricter, they just print out additional warnings, from the man page:
```
-deprecation  Emit warning and location for usages of deprecated APIs.

-unchecked  Enable additional warnings where generated code depends on assumptions.
```
It could also be that we don't really want the additional warnings in the beginning.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
rvacaru: indeed it returns a new context, I changed the implementation and reassign the context

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,68 @@
...
+) {
+  private val settings = "-unchecked" :: "-deprecation" :: "-classpath" :: classpath :: Nil
+  private val driver = new InteractiveDriver(settings)
+  private val reporter: StoreReporter = new StoreReporter()
+  private var ctx: FreshContext = driver.currentCtx.fresh
+  private val dottyVirtualDirectory = new DottyVirtualDirectory(target.name, None)
+  
+  ctx = ctx
+    .setReporter(reporter)
+    .setSetting(ctx.settings.outputDir, dottyVirtualDirectory)
```
rvacaru, line 33: Rule compiler init is now fixed, the console manual test works:
```scala
scala> val r = new scalafix.internal.reflect.RuleCompiler(scalafix.internal.reflect.RuleCompilerClasspath.defaultClasspath)
val r: scalafix.internal.reflect.RuleCompiler = scalafix.internal.reflect.RuleCompiler@6c20679c
``` 
This was inspired by the following article with links to code samples:
- https://www.chris-kipp.io/blog/an-intro-to-the-scala-presentation-compiler
- https://github.com/com-lihaoyi/Ammonite/blob/master/amm/compiler/src/main/scala-3/ammonite/compiler/Compiler.scala
- https://github.com/scalameta/metals/blob/main/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
bjaglin: if it's just warnings let's keep them for now :+1:

build.sbt
```diff
@@ -188,21 +201,24 @@ lazy val testsOutput = projectMatrix
...
-    libraryDependencies ++= Seq(
-      googleDiff,
-      scalatest
-    )
+    libraryDependencies += googleDiff,
+    libraryDependencies ++= {
+      if (!isScala3.value)
+        Seq(scalatest)
+      else
+        Seq("org.scalatest" %% "scalatest" % scalatest3V)
```
bjaglin, line 217: you can avoid repeating the artifact coordinates
```suggestion
        Seq("scalatest.withRevision(scalatest3V))
```
actually, since scalatest is used also in unit, we could probably have a `scalatestDep` in `ScalafixBuild`
```scala
    lazy val scalatestDeps = Def.setting {
      if (isScala3.value) Seq(scalatest.withRevision(scalatest3V))
      else Seq(scalatest)
    }
```

project/ScalafixBuild.scala
```diff
@@ -203,8 +206,10 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     Compile / doc / scalacOptions ++= scaladocOptions,
     Compile / unmanagedSourceDirectories ++= {
       val sourceDir = (Compile / sourceDirectory).value
+      val scala212PlusSourceDir = "scala-2.12+"
       CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, n)) if n >= 12 => Seq(sourceDir / "scala-2.12+")
+        case Some((2, n)) if n >= 12 => Seq(sourceDir / scala212PlusSourceDir)
+        case Some((3, _)) => Seq(sourceDir / scala212PlusSourceDir)
```
bjaglin, line 212: it would be nice to have this as a separate preliminary commit since it impacts stuff in -rules

rvacaru: history is now rewritten

build.sbt
rvacaru: I have a doubt, why in unit we have scalatest with version 3.2.0:
```
    libraryDependencies ++= List(
      jgit,
      coursier,
      scalatest.withRevision(
        "3.2.0"
      ), // make sure testkit clients can use recent 3.x versions
      scalametaTeskit
    ),
```

while in `Dependencies.scala` we have a different version:
- `  val scalatestV = "3.0.8"`

In your `scalatestDeps` proposal we'd be using either version 3.0.8 (scalatestV) or 3.2.13 (scalatest3V)

bjaglin: That hardcoded "3.2.0" has no reason to be pinned, it can/should follow the latest version. So instead of `scalatestV` & `scalatest3V`, we could name `scalatestMinV` & `scalatestLatestV`.

It's OK that the scala3 artifacts are not compatible with Scalatest 3.0.x since there are new so there are no concerns of breaking changes. See https://github.com/scalacenter/scalafix/issues/1172#issuecomment-646621781 for the context. For the next breaking change version of scalafix, I believe we should remove support for Scalafix 3.0.x (it's been nearly 3 years since the old Scalatest 3.0.x API is deprecated) and just build against the latest, but let's not change that in that PR.

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
```diff
@@ -28,7 +28,7 @@ import scalafix.internal.reflect.RuleCompilerClasspath
 import scalafix.test.StringFS
 import scalafix.testkit.DiffAssertions
 import scalafix.tests.util.ScalaVersions
-import scalafix.tests.util.SemanticdbPlugin
+import scalafix.tests.util.compat.SemanticdbPlugin
```
rvacaru, line 31: here we're also calling nsc.Main.process @ line 195:
`    val compileSucceeded = scala.tools.nsc.Main.process(scalacOptions)
`

scalafix-tests/unit/src/test/scala-2/scalafix/tests/util/compat/CompatSemanticdb.scala
```diff
@@ -0,0 +1,18 @@
+package scalafix.tests.util.compat
+
+object CompatSemanticdb {
+
+    def scalacOptions(opts: Array[String], src: java.nio.file.Path, target: java.nio.file.Path) : Array[String] = {
```
bjaglin, line 5: nitpicking, but I would not take the previous opts as parameter, this can be done on the callsite

scalafix-tests/unit/src/test/scala-2/scalafix/tests/util/compat/CompatSemanticdb.scala
```diff
@@ -0,0 +1,18 @@
...
+        val scalacOptions: Array[String] = Array[String] (
+            s"-Xplugin:${SemanticdbPlugin.semanticdbPluginPath()}",
+            "-Xplugin-require:semanticdb",
+            s"-P:semanticdb:sourceroot:$src",
+            s"-P:semanticdb:targetroot:$target"
+        )
+        opts ++ scalacOptions
+    }
+
+    def nscMainProcess(scalacOptions: Array[String]): Unit = {
```
bjaglin, line 15: nit: favor Seq over Array to be more generic?

scalafix-tests/unit/src/test/scala/scalafix/tests/cli/ScalafixImplSuite.scala
bjaglin: then you'll probably need to update the `scalacOptions` as well

scalafix-tests/unit/src/test/scala-2/scalafix/tests/util/compat/CompatSemanticdb.scala
bjaglin:
```suggestion
    def runScalac(options: Seq[String]): Unit = {
```

scalafix-tests/unit/src/test/scala-2/scalafix/tests/util/compat/CompatSemanticdb.scala
rvacaru: true! I'm such a dummy haha

project/ScalafixBuild.scala
```diff
@@ -220,6 +220,14 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
       }
     },
+    Compile / unmanagedResourceDirectories ++= {
+      val resourceDir = (Compile / resourceDirectory).value
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, _)) => Seq(resourceDir / "resource-2")
+        case Some((3, _)) => Seq(resourceDir / "resource-3")
+        case _ => Seq()
+      }
+    },
```
adpi2, line 230:
```
sbt:example> show Compile / unmanagedResourceDirectories
[info] * /home/piquerez/github/scalacenter/sbt-resource-dir/src/main/resources
[info] * /home/piquerez/github/scalacenter/sbt-resource-dir/src/main/resources/resource-3
```
I don't think it is a good idea to have `resource-3` inside `resources` because they are both `unmanagedResourceDirectories`. Maybe you should put `resource-3` on the same level as `resources`.

```diff
Compile / unmanagedResourceDirectories ++= {
      val resourceDir = (Compile / resourceDirectory).value
      CrossVersion.partialVersion(scalaVersion.value) match {
        case Some((2, _)) => Seq(resourceDir.getParentFile / "resource-2")
        case Some((3, _)) => Seq(resourceDir.getParentFile / "resource-3")
        case _ => Seq()
      }
    },
```

rvacaru: Thanks! I'm new to sbt and didn't realize I could actually check the dirs with `show Compile / unmanagedResourceDirectories`. 
On the multiple resource directories or not I'm not sure, we might have resources that are common and only some specific for each scala version. Right now it's not the case so this might be overcomplicating for [speculative generality](https://refactoring.guru/smells/speculative-generality)

What do you think @bjaglin ?

bjaglin: If I see a `resource-x` directory in a project, I would probably expect `resource` to be included as well, just like the built-in `scala-x`  pattern for `unmanagedSourceDirectories`. So based on my own experience with sbt projects, I would favor adding new directories rather than replacing the default one. Also, it means we can apply the setting across all modules (even those that don't have `resource-x` needs) instead of on a per-module basis.

build.sbt
```diff
@@ -231,14 +228,29 @@ lazy val unit = projectMatrix
...
+      } else {
+        List(
+          jgit,
+          coursier
+            .exclude("org.scala-lang.modules", "scala-xml_2.13"),
+          scalametaTeskit
+            .exclude("com.lihaoyi", "sourcecode_2.13")
+            .exclude("org.scala-lang.modules", "scala-collection-compat_2.13")
+            .exclude("org.scalameta", "munit_2.13"),
+          ("org.scalameta" %% "munit" % "0.7.29")
```
bjaglin, line 250: declare in `Dependencies.scala`?

build.sbt
```diff
@@ -231,14 +228,29 @@ lazy val unit = projectMatrix
...
-      coursier,
-      scalatest.withRevision(
-        "3.2.13"
-      ), // make sure testkit clients can use recent 3.x versions
-      scalametaTeskit
-    ),
+    libraryDependencies ++= {
+      if (!isScala3.value) {
+        List(
+          jgit,
```
bjaglin, line 234: extract as it's common

project/ScalafixBuild.scala
```diff
@@ -124,6 +124,11 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     lazy val buildInfoSettingsForRules: Seq[Def.Setting[_]] = Seq(
       buildInfoObject := "RulesBuildInfo"
     )
+
+    lazy val scalatestDeps = Def.setting {
```
bjaglin, line 128: looks like you return a single element in both cases, so remove the `Seq` and rename to `scalatestDep` ?

project/ScalafixBuild.scala
```diff
@@ -146,6 +151,10 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       "unit2_12Target3/test" ::
         "core3/compile" ::
         "rules3/compile" ::
+        "reflect3/compile" ::
+        "cli3/compile" ::
+        "testkit3/compile" ::
+        "unit3Target3/compile" :: // TODO replace core rules and reflect with unit3 in the end
```
bjaglin, line 157:
```suggestion
      "unit2_12Target3/test" ::
         "unit3Target3/test"
```

scalafix-tests/unit/src/test/scala/scalafix/tests/core/PatchSuite.scala
```diff
@@ -94,7 +95,13 @@ class PatchSuite extends AbstractSyntacticRuleSuite with AnyFunSuiteLike {
...
   case object AddGlobalImporter extends SyntacticRule("addGlobalImporter") {
     override def fix(implicit doc: SyntacticDocument): Patch = {
-      Patch.addGlobalImport(importer"scala.collection.{mutable => _}")
+      Patch.addGlobalImport(
+        Importer(
+          Term.Select(Term.Name("scala"), Term.Name("collection")),
+          List(Importee.Unimport(Name.Indeterminate("mutable")))
+        )
+      )
+      // (importer"scala.collection.{mutable => _}")
```
bjaglin, line 104:
```suggestion
```

build.sbt
```diff
@@ -188,21 +201,21 @@ lazy val testsOutput = projectMatrix
...
     moduleName := "scalafix-testkit",
     isFullCrossVersion,
-    libraryDependencies ++= Seq(
-      googleDiff,
-      scalatest
-    )
+    libraryDependencies += googleDiff,
+    libraryDependencies ++= {
+      scalatestDeps.value
+    }
```
bjaglin, line 215:
```suggestion
    libraryDependencies ++= scalatestDeps.value
```

build.sbt
```diff
@@ -188,21 +201,21 @@ lazy val testsOutput = projectMatrix
...
   )
   .defaultAxes(VirtualAxis.jvm)
-  .jvmPlatform(buildScalaVersions)
+  .jvmPlatform(buildScalaVersions :+ scala3)
```
bjaglin, line 218: is it time to add `scala3` to `buildScalaVersions` if we migrated everything?

scalafix-tests/unit/src/test/scala/scalafix/tests/util/ScalametaStructureSuite.scala
```diff
@@ -44,7 +56,18 @@ class ScalametaStructureSuite extends AnyFunSuite with DiffAssertions {
...
+          List(),
+          Term.Name("foo"),
+          List(),
+          List(List()),
+          Some(Type.Name("A")),
+          Term.Name("???")
+        )
+        .decltpe
+        .structureWidth(1),
+      // q"def foo: A = ???".decltpe.structureWidth(1),
```
bjaglin, line 70:
```suggestion
```

scalafix-rules/src/main/resources-3/META-INF/services/scalafix.v1.Rule
```diff
@@ -0,0 +1,7 @@
+scalafix.internal.rule.DisableSyntax
+# scalafix.internal.rule.ExplicitResultTypes # comment out once scala-3 ExplicitResultTypes is implemented
```
bjaglin, line 2: I would just leave it out, it will be obvious when we reach that point
```suggestion
```

bjaglin (commented):
Looking very good - only cosmetic/nitpicking comments.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -41,7 +40,7 @@ class RuleCompiler(
 
   private val compiler: Compiler = new Compiler()
   private val classLoader: AbstractFileClassLoader =
-    new AbstractFileClassLoader(target, this.getClass.getClassLoader)
```
bjaglin: When the target is virtual (which is the case if no argument is provided like in `ScalafixToolbox`)
https://github.com/scalacenter/scalafix/blob/ac41300fe922ac27931064fd881272b927c3c72e/scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala#L32
is creating a `dottyTargetDirectory` (used as compiler output) completely disconnected from `target` (used to try to classload later): they only share a name, but any memory space.

This commits fixes the problem before further refactoring to get rid of the line above.

bjaglin:
rebased against master to verify that https://github.com/scalacenter/scalafix/pull/1665 has a positive impact

scalafix-tests/unit/src/test/scala/scalafix/tests/reflect/RuleDecoderSuite.scala
```diff
@@ -26,13 +27,42 @@ class RuleDecoderSuite extends AnyFunSuite {
...
     val rules = decoder.read(Conf.Str(s"file:$abspath")).get
     assert(expectedName == rules.name.value)
   }
+
   test("relative resolves from custom working directory") {
     val rules = decoder.read(Conf.Str(s"file:$relpath")).get
     assert(expectedName == rules.name.value)
   }
+
+  test("resolved classes can be reloaded", SkipWindows) {
```
bjaglin, line 41: I would keep a separate commit for that particular change instead of merging it to the big unit one, as the reasoning for the change (both for scala 2 & scala3) is quite tricky.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -461,20 +471,15 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
-      "-Xplugin-require:semanticdb",
       "-classpath",
       s"${scalaLibrary.mkString(":")}",
-      s"-P:semanticdb:sourceroot:$src",
-      s"-P:semanticdb:targetroot:$target",
       "-d",
       d.toString,
       main.toString
-    )
+    ) ++ CompatSemanticdb.scalacOptions(src, target)
```
rvacaru, line 480: should `specificScalacOption2` be added here too?

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -66,22 +64,25 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
   val main: Path = src.resolve("Main.scala")
   val relativePath: Path = cwd.relativize(main)
 
+  val specificScalacOption2: Seq[String] =
+    if (!ScalaVersions.isScala3)
+      Seq(removeUnused)
+    else Nil
+
```
rvacaru, line 71: @mlachkar I thought we were handling scalacOptions in the [CompatSemanticdb](https://github.com/scalacenter/scalafix/pull/1650/files#diff-4cc852444bb821de2b756471dc5f8040c8a634fabf8c91e2ef4629229eb21cbc) classes in `scala2` and `scala3`. That's where I already moved `"-Yrangepos"`

project/Dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Dependencies {
   val scala211 = "2.11.12"
   val scala212 = "2.12.17"
   val scala213 = "2.13.9"
-  val scala3 = "3.1.3"
+  val scala3 = "3.2.0"
```
bjaglin, line 13: why do we need to bump this? it should be fine as the [compatibility plans are still being ironed](https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html) (so it's only starting 3.0 that we should pin), but I would not bump in this PR unless we fully understand the cost/benefits.

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -510,4 +524,12 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
   def scalaLibrary: Seq[AbsolutePath] = Classpaths.scalaLibrary.entries
 
+  def scalacOptions(
+      scalaVersion: String = BuildInfo.scalaVersion
+  ): Array[String] =
+    Array[String](
+      "-Yrangepos",
+      "-classpath",
+      s"${scalaLibrary.mkString(":")}"
+    )
```
bjaglin, line 534: unused?

scalafix-tests/unit/src/test/scala/scalafix/tests/util/ScalaVersions.scala
```diff
@@ -1,10 +1,14 @@
 package scalafix.tests.util
 
+import scalafix.tests.BuildInfo
```
bjaglin, line 3: :+1:

bjaglin:
Rebased against 0.10.4

bjaglin:
nearing merging state: down to 4 open, actionable comments

bjaglin:
> Rebased against 0.10.4

Regression in ci-3 since that. It looks like https://github.com/scalacenter/scalafix/pull/1683/commits/ad2f54471f93d8362625ddfae1dcc8d3ed70f085 triggers a failure in ScalafixArgumentsSuite that breaks testing infrastructure for further tests.
```
[info] ScalafixArgumentsSuite:
[info] - ScalafixArguments.evaluate with a semantic rule !!! CANCELED !!! (3 milliseconds)
[info]   org.scalatest.exceptions.TestCanceledException was thrown. (ScalafixArgumentsSuite.scala:83)
[info] - ScalafixArguments.evaluate getting StaleSemanticdb !!! CANCELED !!! (0 milliseconds)
[info]   org.scalatest.exceptions.TestCanceledException was thrown. (ScalafixArgumentsSuite.scala:165)
[info] - ScalafixArguments.evaluate doesn't take into account withMode and withMainCallback !!! CANCELED !!! (0 milliseconds)
[info]   org.scalatest.exceptions.TestCanceledException was thrown. (ScalafixArgumentsSuite.scala:200)
Warning: Unable to read from client, please check on client for futher details of the problem.
Reporter completed abruptly with an exception after receiving event: TestSucceeded(Ordinal(0, 71),ScalafixArgumentsSuite,scalafix.tests.interfaces.ScalafixArgumentsSuite,Some(scalafix.tests.interfaces.ScalafixArgumentsSuite),CommentFileNonAtomic retrieves 2 patches,CommentFileNonAtomic retrieves 2 patches,Vector(),Some(97),Some(IndentedText(- CommentFileNonAtomic retrieves 2 patches,CommentFileNonAtomic retrieves 2 patches,0)),Some(LineInFile(257,ScalafixArgumentsSuite.scala,Some(Please set the environment variable SCALACTIC_FILL_FILE_PATHNAMES to yes at compile time to enable this feature.))),Some(scalafix.tests.interfaces.ScalafixArgumentsSuite),None,pool-1-thread-1-ScalaTest-running-ScalafixArgumentsSuite,1665911264396).
java.net.SocketException: Broken pipe (Write failed)
	at java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:111)
	at java.net.SocketOutputStream.write(SocketOutputStream.java:155)
	at java.io.ObjectOutputStream$BlockDataOutputStream.drain(ObjectOutputStream.java:1877)
	at java.io.ObjectOutputStream$BlockDataOutputStream.setBlockDataMode(ObjectOutputStream.java:1786)
	at java.io.ObjectOutputStream.writeNonProxyDesc(ObjectOutputStream.java:1286)
	at java.io.ObjectOutputStream.writeClassDesc(ObjectOutputStream.java:1231)
	at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1427)
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1178)
	at java.io.ObjectOutputStream.writeFatalException(ObjectOutputStream.java:1577)
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:351)
	at org.scalatest.tools.SocketReporter.liftedTree1$1(SocketReporter.scala:46)
	at org.scalatest.tools.SocketReporter.apply(SocketReporter.scala:55)
	at org.scalatest.DispatchReporter.org$scalatest$DispatchReporter$Propagator$$_$run$$anonfun$1(DispatchReporter.scala:249)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
	at scala.collection.immutable.List.foreach(List.scala:333)
	at org.scalatest.DispatchReporter$Propagator.run(DispatchReporter.scala:249)
	at java.lang.Thread.run(Thread.java:748)
```
Will investigate.

bjaglin:
https://github.com/scalameta/scalameta/pull/2825 seems to be culprit (incompatible with scalatest_3 3.2.{13,14} for some unknown reason), by the following bisect
```
4.5.13+71-aca5dd6e-SNAPSHOT/ 	Tue Sep 20 07:02:07 UTC 2022 	OK
4.5.13+73-4db513a9-SNAPSHOT/ 	Tue Sep 20 11:37:57 UTC 2022 	KO  
```

bjaglin:
I tried to force runtime scala libs to 2.13.8 in unit, without any success
```
    dependencyOverrides += "org.scala-lang" % "scala-compiler" % "2.13.8",
    dependencyOverrides += "org.scala-lang" % "scala-reflect" % "2.13.8",
    dependencyOverrides += "org.scala-lang" % "scalap" % "2.13.8",
```

bjaglin:
> Next step would be to try a scalameta built against 2.13.10 to confirm/deny that hypothesis.

Using a local build of https://github.com/scalameta/scalameta/commits/6d9b59da427df20f79fa0786fc680e6596da1586 does fix the problem, so it does look related to https://github.com/scala/bug/issues/12650. That means that scalameta 4.6.1 should help, so we could just wait for that.

I am still puzzled to know which part of the scalameta code is involved when using scalatest's `cancel()`...

tgodzik:
> I am still puzzled to know which part of the scalameta code is involved when using scalatest's `cancel()`... Any clue @tgodzik ?

No idea, it's super weird :thinking:

bjaglin:
Adding `-Xlog:exceptions=info`, we can see the deserialization issue (most likely on `TestCanceledException` since that's the exception thrown on the client-side)

```
$ sbt -J-Xlog:exceptions=info "unit3Target3/testOnly *ScalafixArgumentsSuite"
...
70.108s][info][exceptions] Exception <a 'java/io/InvalidClassException'{0x00000000bf525de0}: local class incompatible: stream classdesc serialVersionUID = 1590627031578033034, local class serialVersionUID = 8944294921032069635>
 thrown in interpreter method <{method} {0x00007f3cc84a3ce0} 'readObject' '()Ljava/lang/Object;' in 'java/io/ObjectInputStream'>
 at bci 3 for thread 0x00007f3c2e92b000 (Thread-24)
[70.108s][info][exceptions] Exception <a 'java/io/InvalidClassException'{0x00000000bf525de0}: local class incompatible: stream classdesc serialVersionUID = 1590627031578033034, local class serialVersionUID = 8944294921032069635>
 thrown in interpreter method <{method} {0x00007f3b56599c20} 'react' '()V' in 'org/scalatest/tools/Framework$Skeleton$1$React'>
 at bci 15 for thread 0x00007f3c2e92b000 (Thread-24)
```

My assumption is that the client and the server sides of scalatest (per `fork := true`) run with a scala library that [has significantly different bytecode](https://github.com/scala/bug/issues/12650) for some core classes that `TestCanceledException` rely on, impacting the dynamic java `serialVersionUID` generation. I still don't know why, but I bet the server is driven by the build (2.13.10), while the client is driven by scalameta (2.13.9).

bjaglin:
I found the cause of the problem: sbt uses a [layered classloader by default](https://github.com/sbt/sbt/blob/1f29c9053cdc52128fa45cfe0b22e9a4d085cc1b/main/src/main/scala/sbt/internal/ClassLoaders.scala#L166-L167), which [looks at the version of scala-reflect](https://github.com/sbt/sbt/blob/1f29c9053cdc52128fa45cfe0b22e9a4d085cc1b/main/src/main/scala/sbt/internal/ClassLoaders.scala#L166-L167) in the classpath (2.13.8 in our case) to load the scala library. This library is different than the one explicitly declared in the classpath, which gets loaded in the forked process (2.13.9). Because of the breaking change in 2.13.9, serialization is impacted and the 2 processes can't communicate through classes serialized through dynamicly-generated `serialVersionUID` :facepalm: 

This only happens with unit3 as I guess sbt makes sure library, reflect & compiler are aligned when building with Scala 2, while in Scala 3 scala-library is just a regular dependency brought by `scala3-library_3` and other libs in the classpath with no specific checks.

Options
1. force scala-library:2.13.10 in the testkit_3 classpath
1. wait for scalameta to be published with scala-library:2.13.10
1. use `ClassLoaderLayeringStrategy.Flat` to have scala-library:2.13.9 on both scalatest client and server
1. disable forking

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
bjaglin: Actually this is not about semanticdb so I would keep it separated, we'll have to reconsider that when working on https://github.com/scalacenter/scalafix/issues/1680 anyway

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
bjaglin: This does not need `removeUnused` so no. This test will be removed with https://github.com/scalacenter/scalafix/pull/1666.

scalafix-reflect/src/main/scala-3/scalafix/internal/reflect/RuleCompiler.scala
```diff
@@ -0,0 +1,78 @@
...
+    val errors = reporter.allErrors.map(error =>
+      ConfError
+        .message(error.getMessage)
+    )
+    if (!errors.isEmpty)
+      errors :+ ConfError.message(
+        "Error compiling rule(s) from source using Scala 3 compiler; " +
+          "to use the Scala 2.x compiler instead, use the corresponding " +
+          "scalafix-cli artifact or force scalafixScalaBinaryVersion " +
+          "to 2.x in your build tool"
```
bjaglin, line 66: I don't manage to get that error when trying to compile source rules with quasiquotes with scalafix-cli_3

```
cs launch -r ivy2local ch.epfl.scala:scalafix-cli_3.1.3:0.10.4+22-b8cd292e+20221018-1241-SNAPSHOT -M scalafix.cli.Cli -- --rules=https://raw.githubusercontent.com/scalacenter/scalafix-named-literal-arguments/fb9f65ee9f09f167651b3d9bfe94a9be76818759/scalafix/rules/src/main/scala/fix/Namedliteralarguments_v1.scala
error: 2 errors
[E0]  Any is not a valid result type of an unapply method of an extractor.
[E1] Scala 2 macro cannot be used in Dotty. See https://dotty.epfl.ch/docs/reference/dropped-features/macros.html
To turn this error into a warning, pass -Xignore-scala2-macros to the compiler
```

bjaglin: Added https://github.com/scalacenter/scalafix/pull/1650/commits/d15af9a0a1eb545819d0fb53ae3306b2e90313a4
```
cs launch -r ivy2local ch.epfl.scala:scalafix-cli_3.1.3:0.10.4+23-d15af9a0+20221018-1320-SNAPSHOT -M scalafix.cli.Cli -- --rules=https://raw.githubusercontent.com/scalacenter/scalafix-named-literal-arguments/fb9f65ee9f09f167651b3d9bfe94a9be76818759/scalafix/rules/src/main/scala/fix/Namedliteralarguments_v1.scala
error: 3 errors
[E0]  Any is not a valid result type of an unapply method of an extractor.
[E1] Scala 2 macro cannot be used in Dotty. See https://dotty.epfl.ch/docs/reference/dropped-features/macros.html
To turn this error into a warning, pass -Xignore-scala2-macros to the compiler
[E2] Error compiling rule(s) from source using Scala 3 compiler; to use the Scala 2.x compiler instead, use the corresponding scalafix-cli artifact or force scalafixScalaBinaryVersion to 2.x in your build tool
```

mlachkar-da (approved):

PR #1653, opened 2022-08-12
Use scalameta's Discord rather than gitter for support

bjaglin:
I recently learnt about the #scalafix channel on the Scalameta servers (advertized in https://www.scala-lang.org/blog/2021/12/21/discord.html#what-other-chat-servers-exist). I am not fully convinced Discord is the way to go (sharing the same opacity concerns as https://alexn.org/blog/2022/04/09/scala-gitter-discord-mistake/), but I think it's better to stay close to the Scalameta community. We should delete the gitter room once this is merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Contributions to Scalafix are welcome! This document is a guide to help you get
 familiar with the Scalafix build, if you are unsure about anything, don't
-hesitate to ask in the [gitter channel](https://gitter.im/scalacenter/scalafix).
+hesitate to ask in the [Discord channel](https://discord.gg/8AHaqGx3Qj).
 
 ## Modules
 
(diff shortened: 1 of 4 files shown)


tgodzik:
Thanks for updating this! And sorry we didn't explicitely update the information on scalafix website :fearful: 

There hasn't been that much traffic on scalafix channel and we've been able to respond to most of it.

Btw. overall I am much more happier with discord, there is much more moderation options and feels a bit more of safe environment.

ckipp01:
>  I am not fully convinced Discord is the way to go (sharing the same opacity concerns as https://alexn.org/blog/2022/04/09/scala-gitter-discord-mistake/), but I think it's better to stay close to the Scalameta community. We should delete the gitter room once this is merged.

Just a heads up, especially for those that don't love Discord, I have bridged all these rooms. So if Matrix is more your jam, you can find the Matrix room for `scalacenter/scalafix` at https://matrix.to/#/#scalacenter:scalafix. You may want to mention that as well.

rvacaru (approved):

mlachkar (approved):

mvasyliv:
we have in source code line
val res = df1.unionAll(df2)
How to get types for df1 and df2 in scalafix?

Issue #1635, opened 2022-07-15
Drop support for scala 2.11


bjaglin:
Scalafix itself is being cross-published to Scala 3 as part of https://github.com/scalacenter/scalafix/issues/1583, which means more version-specific code.

The last maintenance release of scala 2.11 being almost 5 years old, it seems fair to drop support for it in upcoming releases, so that we can:
- get rid of the `scala-2.11` sources
- move `scala-2.12+` to `scala`
- simplify the build by getting rid of `metaconfig*For211`

PR #1661, opened 2022-08-25
Update unit module scalatest version to 3.2.13

rvacaru:
The scalatest version update, needed for the work done in https://github.com/scalacenter/scalafix/pull/1650 , specifically because the old scalatest version in not found anymore.

When updating the versions there's a lib version conflict on `scala-xml` which is resolved by ignoring with a `libraryDependencySchemes` as suggested in the conflict output below:
```
sbt:scalafix> unit2_12Target3/update
[error] stack trace is suppressed; run last unit2_12Target3 / update for the full output
[error] (unit2_12Target3 / update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error] 
[error]         * org.scala-lang.modules:scala-xml_2.12:2.1.0 (early-semver) is selected over {1.3.0, 1.0.6}
[error]             +- org.scalatest:scalatest-core_2.12:3.2.13           (depends on 2.1.0)
[error]             +- org.scala-lang:scala-compiler:2.12.16              (depends on 1.0.6)
[error]             +- io.get-coursier:coursier-core_2.12:2.0.0-RC5-6 (depends on xxx)
[error] this can be overridden using libraryDependencySchemes or evictionErrorLevel
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,7 +21,7 @@ object Dependencies {
 
   val bijectionCoreV = "0.9.7"
   val collectionCompatV = "2.8.0"
-  val coursierV = "2.0.0-RC5-6"
+  val coursierV = "2.1.0-M6-53-gb4f448130"
   val coursierInterfaceV = "1.0.8"
   val commontTextV = "1.9"
   val googleDiffV = "1.3.0"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,7 +21,7 @@ object Dependencies {
 
   val bijectionCoreV = "0.9.7"
   val collectionCompatV = "2.8.0"
-  val coursierV = "2.1.0-M6-53-gb4f448130"
+  val coursierV = "2.0.0-RC5-6"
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -21,7 +21,7 @@ object Dependencies {
 
   val bijectionCoreV = "0.9.7"
   val collectionCompatV = "2.8.0"
-  val coursierV = "2.0.0-RC5-6"
```
rvacaru, line 24: this is actually the latest version available also for scala 2.11 (see https://github.com/scalacenter/scalafix/runs/8031777807?check_suite_focus=true) Hence I think we can use 2 different versions of coursier instead, one for 2.11 and the other for the rest

bjaglin: let's not bump coursier and introduce more complexity, 2.11 is on its way out anyway https://github.com/scalacenter/scalafix/issues/1635

rvacaru: reverted

bjaglin (approved):

Issue #1663, opened 2022-08-31
RemoveUnused does not clean Scala 3 style wildcard import with `-Xsource:3`


xuwei-k:
## step

`build.sbt`

```scala
scalaVersion := "2.13.8"

semanticdbEnabled := true

scalacOptions ++= Seq("-Wunused", "-Xsource:3")
```

`project/build.properties`

```properties
sbt.version=1.7.1
```

`project/plugins.sbt`

```scala
addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.1")
```

`Main.scala`

```scala
import scala.collection.mutable
import scala.util._
import scala.io.*

object Main
```

run: `sbt "scalafix RemoveUnused"`

## actual

log

```
[warn] /Users/kenji/remove-unused-scala-3-syntax/Main.scala:1:25: Unused import
[warn] import scala.collection.mutable
[warn]                         ^
[warn] /Users/kenji/remove-unused-scala-3-syntax/Main.scala:2:19: Unused import
[warn] import scala.util._
[warn]                   ^
[warn] /Users/kenji/remove-unused-scala-3-syntax/Main.scala:3:17: Unused import
[warn] import scala.io.*
[warn]                 ^
[warn] three warnings found
[info] Running scalafix on 1 Scala sources
[success] Total time: 3 s, completed Aug 31, 2022, 4:36:11 AM
```

diff

```diff
@@ -1,5 +1,4 @@
-import scala.collection.mutable
-import scala.util._
+
 import scala.io.*
```

## expect

remove `import scala.io.*`

xuwei-k:
note: different position if Scala 3 style wildcard 🤔 

```
import scala.io.*
^^^^^^^^^^^^^^^^^
```

```
import scala.util._
                  ^
```

```diff
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/RemoveUnused.scala
@@ -65,6 +65,7 @@ class RemoveUnused(config: RemoveUnusedConfig)
 
     doc.diagnostics.foreach { diagnostic =>
       if (config.imports && diagnostic.message == "Unused import") {
+        println(diagnostic)
         isUnusedImport += diagnostic.position
``

xuwei-k:
[...]

PR #1664, opened 2022-09-05
support Scala 3 style wildcard import in RemoveUnused

xuwei-k:
- fix https://github.com/scalacenter/scalafix/issues/1663


bjaglin (approved):
Thanks for the fix!

I wonder if this [inconsistency](https://github.com/scalacenter/scalafix/issues/1663#issuecomment-1232508407) is something we might want to fix upstream, in scalameta or in the compiter @tgodzik?

If not, I think we could leverage the testkit for such tests, by adding a `inputSource3` sbt module (a bit like in [organize-imports](https://github.com/liancheng/scalafix-organize-imports/search?q=inputUnusedImports)), and put input/output sources in [`scala-2.12+`](https://github.com/scalacenter/scalafix/blob/f7944df30526326478cb1809fb4afd5db2f85812/project/ScalafixBuild.scala#L208). I can do that as a follow-up PR @xuwei-k if you don't have time/interest to try that out.

bjaglin:
> I can do that as a follow-up PR @xuwei-k if you don't have time/interest to try that out.

https://github.com/scalacenter/scalafix/pull/1666

PR #1681, opened 2022-10-01
Cross compile scalafix

mlachkar:


bjaglin:
re-integrated into https://github.com/scalacenter/scalafix/pull/1650, following-up there

Issue #1168, opened 2020-06-18
make undocumented replace rules safe to use


bjaglin:
https://github.com/scalacenter/scalafix/blob/d55b09bc18fe97686bd995fa44186e9ea18e9772/scalafix-tests/input/src/main/scala/test/ReplaceSymbol.scala#L1-L29 demonstrates usage of simple symbol replacements via `--rules=replace:from/to` and via the `patches.replaceSymbols` configuration key, backed by https://github.com/scalacenter/scalafix/blob/6811967f6e6059a59d8b04e7f2b76eb3c3ddcfbc/scalafix-core/src/main/scala/scalafix/patch/Patch.scala#L125-L152.

As of 0.9.17, this is not documented publicly as it doesn’t handle some important cases leading to generating broken code.

From https://gitter.im/scalacenter/scalafix?at=5eeb039a405be935cda9fe27
>  At the top of my head, renaming a method won’t rename overrides/supermethods. The rename functionality in metals and IntelliJ are more reliable since they handle such cases. There are some cases the replace rule can do stuff I haven’t seen in IntelliJ or metals (for example replace usage of com.foo.Bar with org.foo.Bar, not definition site) but it’s not obvious for users to evaluate what cases will generate valid rewrites and what cases generate broken code.

PR #1695, opened 2022-10-17
Scalafix rename that leads to a name collision breaks compilation

### LINKED ISSUE #1305: Scalafix rename that leads to a name collision breaks compilation
Problem
======

If a rename causes a collision in the renamed file, you can no longer compile the file.  Scalafix does not deal with the collision gracefully.

Proposed Solution
=============

We can infer if the Scalafix rename would cause a collision by inspecting the identifiers that are at the scope of the import when it's renamed.  This will be a little tricky (I believe imports are evaluated in order) as we want to know about all of the identifiers that come from imports.  If the rename would cause a collision, we inline the fully qualified name and delete the old import, instead of performing the rename.

Next Steps
========

If someone is inspired to tackle this problem, I would be very grateful!  If no one is feeling so inspired, I can take a stab at working on this, but I'll need guidance, since I'm a scalafix newb.

jmcrawford45:
Fixes #1305

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
@@ -10,6 +10,8 @@ import scalafix.patch.Patch
 import scalafix.patch.Patch.internal.ReplaceSymbol
 import scalafix.syntax._
 import scalafix.v0._
+import scala.annotation.tailrec
+
 
 object ReplaceSymbolOps {
   private object Select {
@@ -19,6 +21,20 @@ object ReplaceSymbolOps {
       case _ => None
     }
   }
+  private def extractImports(stats: Seq[Stat]): Seq[Import] = {
+    stats
+      .takeWhile(_.is[Import])
+      .collect { case i: Import => i }
+  }
+
+  @tailrec private final def getGlobalImports(ast: Tree): Seq[Import] =
+    ast match {
+      case Pkg(_, Seq(pkg: Pkg)) => getGlobalImports(pkg)
+      case Source(Seq(pkg: Pkg)) => getGlobalImports(pkg)
+      case Pkg(_, stats) => extractImports(stats)
+      case Source(stats) => extractImports(stats)
+      case _ => Nil
+    }
 
   def naiveMoveSymbolPatch(
       moveSymbols: Seq[ReplaceSymbol]
@@ -92,6 +108,7 @@ object ReplaceSymbolOps {
         case _ => None
       }
     }
+    
     val patches = ctx.tree.collect { case n @ Move(to) =>
       // was this written as `to = "blah"` instead of `to = _root_.blah`
       val isSelected = to match {
@@ -126,10 +143,20 @@ object ReplaceSymbolOps {
             if sig.name != parent.value =>
           Patch.empty // do nothing because it was a renamed symbol
         case Some(parent) =>
+          val causesCollision = getGlobalImports(ctx.tree).exists { importStmt =>
+            importStmt.importers.flatMap(_.importees).exists {
+              case Importee.Name(name) => name.value == to.signature.name
+              case Importee.Rename(_, rename) => rename.value == to.signature.name
+              case _ => false
+            }
+          }
           val addImport =
-            if (n.isDefinition) Patch.empty
+            if (n.isDefinition || causesCollision) Patch.empty
             else ctx.addGlobalImport(to)
-          addImport + ctx.replaceTree(n, to.signature.name)
+          if(causesCollision)
+            addImport + ctx.replaceTree(n, to.owner.syntax + to.signature.name)
+          else
+            addImport + ctx.replaceTree(n, to.signature.name)
         case _ =>
           Patch.empty
       }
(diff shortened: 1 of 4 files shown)


scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
```diff
@@ -126,10 +143,20 @@ object ReplaceSymbolOps {
             if sig.name != parent.value =>
           Patch.empty // do nothing because it was a renamed symbol
         case Some(parent) =>
+          val causesCollision = getGlobalImports(ctx.tree).exists { importStmt =>
```
jmcrawford45, line 146: It would be more performant to get the global imports from the start and manipulate them as patches are processed. We want to handle for `foo.Bar => foo.Baz` and `bar.Foo => bar.Baz` without breaking compilation, so we have to consider previous renames as well. Now that I think more on it, I guess for completeness we'd also want to handle non-import collisions. Open to suggestions on these.

scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
```diff
@@ -19,6 +21,20 @@ object ReplaceSymbolOps {
       case _ => None
     }
   }
+  private def extractImports(stats: Seq[Stat]): Seq[Import] = {
+    stats
+      .takeWhile(_.is[Import])
```
bjaglin, line 26: is this useful?

scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
bjaglin: > Now that I think more on it, I guess for completeness we'd also want to handle non-import collisions. Open to suggestions on these.

Indeed, although this would be more complex as  there is not enough information in the tree nor in semanticdb to list all locals for example. You'd have to start a presentation compiler instance (outside the scalafix API), just like [ExplicitResultTypes](https://github.com/scalacenter/scalafix/blob/e11610fb2abcf78355c57794b04626e2d2c5c24e/scalafix-rules/src/main/scala-2/scalafix/internal/rule/CompilerTypePrinter.scala), which would greatly increase code complexity and ease-of-use (as scalafix's Scala binary version must match the target's). Without going that way, there are some heuristics that could prevent non-import collisions (getting visibile members from super traits, getting local val names via the tree, etc), but it's a lot of work to cover them all.

scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
```diff
@@ -126,10 +143,20 @@ object ReplaceSymbolOps {
             if sig.name != parent.value =>
           Patch.empty // do nothing because it was a renamed symbol
         case Some(parent) =>
+          val causesCollision = getGlobalImports(ctx.tree).exists { importStmt =>
+            importStmt.importers.flatMap(_.importees).exists {
+              case Importee.Name(name) => name.value == to.signature.name
+              case Importee.Rename(_, rename) => rename.value == to.signature.name
+              case _ => false
```
bjaglin, line 150: Wildcards can also cause collisions (although we can't tell using Scalafix APIs). This is probably outside the scope of this PR, but we could have an opt-in "safe" flag that disable global imports whenever it finds wildcard imports.

scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
```diff
@@ -19,6 +21,20 @@ object ReplaceSymbolOps {
       case _ => None
     }
   }
+  private def extractImports(stats: Seq[Stat]): Seq[Import] = {
+    stats
+      .takeWhile(_.is[Import])
+      .collect { case i: Import => i }
+  }
+
+  @tailrec private final def getGlobalImports(ast: Tree): Seq[Import] =
```
bjaglin, line 30: this could be more granular and collect relevant scoped imports, all the way to the global ones at the top

scalafix-core/src/main/scala/scalafix/internal/patch/ReplaceSymbolOps.scala
bjaglin: > It would be more performant to get the global imports from the start

Indeed - is it possible to call `getGlobalImports` once outside the tree traversal?

bjaglin (commented):
Welcome to the project and thanks a lot for putting this together! This feature [has a lot of shortcomings](https://github.com/scalacenter/scalafix/issues/1168), and this does remove one :+1: 

Don't forget to run `scalafixAll` before committing (see CI failure https://github.com/scalacenter/scalafix/actions/runs/3266135538/jobs/5369523474).

bjaglin:
@jmcrawford45 will you have time to pursue this?

bjaglin:
@jmcrawford45 I would love to get that in, any chance you could address the outstanding comments? Thanks!

subhramit:
Hi @bjaglin, I don't know if this is still of interest, but I came across this stale PR and attempted to take it forward in https://github.com/scalacenter/scalafix/pull/2260.

PR #1697, opened 2022-10-19
Update RedundantSyntax.md

kimnilsson:
Typo in RedundantSyntax.md

### DIFF
--- a/docs/rules/RedundantSyntax.md
+++ b/docs/rules/RedundantSyntax.md
@@ -32,4 +32,7 @@ Example:
 
 - println(raw"Baz")
 + println("Baz")
+
+// No change as `raw` is not redundant.
+println(raw"Foo\nBar")
 ```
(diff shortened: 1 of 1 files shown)


docs/rules/RedundantSyntax.md
```diff
@@ -31,7 +31,5 @@ Example:
 + println("Bar")
 
 - println(raw"Baz")
-+ println("Foo")
-
-println(raw"Foo\nBar")
```
bjaglin, line 36: The intent of that line was to show that this wouldn't get rewritten. Did you remove it on purpose?

bjaglin (commented):
Thanks for taking the time to update the docs!

kimnilsson:
> Thanks for taking the time to update the docs!

You've made it so easy with the `Edit` button so I thought why not :+1:

docs/rules/RedundantSyntax.md
kimnilsson: Yes, that was my intention as it broke the pattern of the preceding three examples (or well, that's how I read it at least). But perhaps it's a good idea to keep it but with a comment saying that it's not rewritten as `raw` is indeed used.

bjaglin:
Thanks!

PR #1698, opened 2022-10-28
ExplicitResultTypes: honor skipSimpleDefinitions on implicits

### LINKED ISSUE #1628: ExplicitResultTypes.isRuleCandidate evaluation improvements/fixes
In [ExplicitResultTypes](https://github.com/scalacenter/scalafix/blob/main/scalafix-rules/src/main/scala/scalafix/internal/rule/ExplicitResultTypes.scala#L189) there's a block of code potentially broken and hard to read:

```
    def qualifyingImplicit: Boolean =
      isImplicit && !isFinalLiteralVal

    def matchesMemberKindAndVisibility: Boolean =
      matchesMemberKind && matchesMemberVisibility

    def qualifyingNonImplicit: Boolean = {
      !onlyImplicits &&
      hasParentWihTemplate &&
      !defn.hasMod(mod"implicit") &&
      !matchesSimpleDefinition()
    }

    matchesMemberKindAndVisibility && {
      qualifyingImplicit || qualifyingNonImplicit
    }
```

The `!matchesSimpleDefinition` predicate is related to the rule configuration `skipSimpleDefinitions` ([link](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html)). This predicate is only evaluated in the `qualifyingNonImplicit` block, while most likely it always needs to be evaluated, together with the `matchesMemberKindAndVisibility` block. 

First of all this assumption needs to be confirmed as part of this issue.
The same thoughts can be valid for `!onlyImplicits` block which is related to a new configuration of ExplicitResultTypes.

Given that the assumption holds true, moving `!matchesSimpleDefinition` out of `qualifyingNonImplicit` will make the `qualifyingNonImplicit` method name reflect what it actually does and will make this method more readable and maintainable.

This change was proposed in PR https://github.com/scalacenter/scalafix/pull/1627

Current scalafix version is 0.10.1

rvacaru:
Fixes #1628 

The `skipSimpleDefinitions` configuration wasn't always applied correctly (specifically when members are implicit the configuration wasn't considered)

- A unit test for this specific case was added
- Old unit tests are fixed based on the default value of `skipSimpleDefinitions` or on the intent of the test

### DIFF
--- a/scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
+++ b/scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
@@ -176,17 +176,16 @@ final class ExplicitResultTypes(
     def qualifyingImplicit: Boolean =
       isImplicit && !isFinalLiteralVal
 
-    def matchesMemberKindAndVisibility: Boolean =
-      matchesMemberKind && matchesMemberVisibility
+    def matchesConfig: Boolean =
+      matchesMemberKind && matchesMemberVisibility && !matchesSimpleDefinition
 
     def qualifyingNonImplicit: Boolean = {
       !onlyImplicits &&
       hasParentWihTemplate &&
-      !defn.hasMod(mod"implicit") &&
-      !matchesSimpleDefinition()
+      !defn.hasMod(mod"implicit")
     }
 
-    matchesMemberKindAndVisibility && {
+    matchesConfig && {
       qualifyingImplicit || qualifyingNonImplicit
     }
   }
--- a/scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesOnlyImplicits.scala
+++ b/scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesOnlyImplicits.scala
@@ -1,6 +1,7 @@
 /*
 rules = ExplicitResultTypes
 ExplicitResultTypes.onlyImplicits = true
+ExplicitResultTypes.skipSimpleDefinitions = []
  */
 package test.explicitResultTypes
 
--- a/scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesSingleton.scala
+++ b/scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesSingleton.scala
@@ -1,5 +1,6 @@
 /*
 rules = ExplicitResultTypes
+ExplicitResultTypes.skipSimpleDefinitions = []
  */
 package test.explicitResultTypes
 
--- a/scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBase.scala
+++ b/scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBase.scala
@@ -19,8 +19,8 @@ object ExplicitResultTypesBase {
   val k: (Int, String) = (1, "msg")
   implicit val L: List[Int] = List(1)
   implicit val M: Map[Int,String] = Map(1 -> "STRING")
-  implicit def D: Int = 2
-  implicit def tparam[T](e: T): T = e
+  implicit def D = 2
+  implicit def tparam[T](e: T) = e
   implicit def tparam2[T](e: T): List[T] = List(e)
   implicit def tparam3[T](e: T): Map[T,T] = Map(e -> e)
   class implicitlytrick {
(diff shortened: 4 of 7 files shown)


scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -176,17 +176,16 @@ final class ExplicitResultTypes(
     def qualifyingImplicit: Boolean =
       isImplicit && !isFinalLiteralVal
 
-    def matchesMemberKindAndVisibility: Boolean =
-      matchesMemberKind && matchesMemberVisibility
+    def matchesConfig: Boolean =
+      matchesMemberKind && matchesMemberVisibility && !matchesSimpleDefinition
```
rvacaru, line 180: looking at this line, wouldn't it make sense to have a `&& matchesSkipSimpleDefinition` predicate instead, with the not (!) inside of the predicate itself? 
At least to me it would look more readable, without having to worry on the content of the predicate

scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesSingleton.scala
```diff
@@ -1,5 +1,6 @@
 /*
 rules = ExplicitResultTypes
+ExplicitResultTypes.skipSimpleDefinitions = []
```
rvacaru, line 3: the singletons in this tests are a `Term.Ref`, hence for this unit test to work the config needs to be overwritten

scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/ImplicitMembersWithSkipSimpleDefinition.scala
```diff
@@ -0,0 +1,13 @@
+package test.explicitResultTypes
+
+object ImplicitMembersWithSkipSimpleDefinition {
+
+    val x = 42
+
+    implicit var y = 43
```
rvacaru, line 7: Literal definitions are skipped, regardless if it's implicit or not, as expected

scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesShort.scala
```diff
@@ -14,7 +14,7 @@ object ExplicitResultTypesShort {
   implicit def z(x: Int): List[String] = Nil
   implicit var zz: ListSet[String] = scala.collection.immutable.ListSet.empty[String]
   implicit val FALSE: Any => Boolean = (x: Any) => false
-  implicit def tparam[T](e: T): T = e
+  implicit def tparam[T](e: T) = e
```
rvacaru, line 17: by default `Term.Ref` is skipped, hence this is the correct output

scalafix-tests/output/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBase.scala
```diff
@@ -19,8 +19,8 @@ object ExplicitResultTypesBase {
   val k: (Int, String) = (1, "msg")
   implicit val L: List[Int] = List(1)
   implicit val M: Map[Int,String] = Map(1 -> "STRING")
-  implicit def D: Int = 2
-  implicit def tparam[T](e: T): T = e
+  implicit def D = 2
+  implicit def tparam[T](e: T) = e
```
rvacaru, line 23: by default `Lit` and `Term.Ref` are skipped, so this is the correct output

scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesOnlyImplicits.scala
```diff
@@ -1,6 +1,7 @@
 /*
 rules = ExplicitResultTypes
 ExplicitResultTypes.onlyImplicits = true
+ExplicitResultTypes.skipSimpleDefinitions = []
```
rvacaru, line 4: we're not testing `skipSimpleDefinitions` in this test, hence overwriting the default

scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: I personally like `!matchesSimpleDefinition` better - the "matches skip" is giving me headaches :sweat_smile:

bjaglin (approved):
Thanks for the fix!

PR #1737, opened 2023-02-23
Add textEdits method to ScalafixPatch

LaurenceWarne:
Hi!  I'd like to be able to obtain (LSP style)  text edit like information (essentially representing a diff as a start position, end position, and insert text) from the scalafix api (specifically a given `ScalafixFileEvaluation`) - AFAICS something like this is currently not exposed (?).

I've created an implementation (the method is similar to `PatchInternals.tokenPatchApply`).  Though I've found some rules don't give the greatest results, for example patches from the `OptionWhenUnless` rule (from https://github.com/xuwei-k/scalafix-rules):
```
Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Concat(Remove(if∙[142..144)),Remove(∙∙[144..145))),Remove((∙[145..146))),Remove(4∙[146..147))),Remove(∙∙[147..148))),Remove(>∙[148..149))),Remove(∙∙[149..150))),Remove(3∙[150..151))),Remove()∙[151..152))),Remove(∙∙[152..153))),Remove(Some∙[153..157))),Remove((∙[157..158))),Remove(43∙[158..160))),Remove()∙[160..161))),Remove(∙∙[161..162))),Remove(else∙[162..166))),Remove(∙∙[166..167))),Remove(None∙[167..171))),Add(if, if [142..144), ifOption.when(4 > 3)(43)))
```

Result in multiple overlapping text edits. Though I've had good results for many of the built in rules, for example a typical `LeakingImplicitVal` patch:

```
Add(val, val [202..205), private val)
```
results in a singular text edit.

Anyway, let me know if you are open to something like this, or if something like this is already implemented, thanks :slightly_smiling_face:

### DIFF
--- a/scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
+++ b/scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
@@ -517,6 +517,61 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
     assert(run.failed.toOption.map(_.getMessage) == Some(expectedErrorMessage))
   }
 
+  test("textEdits returns patch as edits") {
+    val cwd: Path = StringFS
+      .string2dir(
+        """|/src/Main2.scala
+          |import scala.concurrent.duration
+          |import scala.concurrent.Future
+          |
+          |object Main extends App {
+          |  import scala.concurrent.Await
+          |  println("test");
+          |  println("ok")
+          |}""".stripMargin,
+        charset
+      )
+      .toNIO
+    val src = cwd.resolve("src")
+
+    val run = api
+      .withRules(List("CommentFileNonAtomic", "CommentFileAtomic").asJava)
+      .withSourceroot(src)
+
+    val fileEvaluation = run.evaluate().getFileEvaluations.head
+    val patches = fileEvaluation.getPatches
+
+    // CommentFileNonAtomic produces two patches which in turn produce one
+    // token patch each, whilst CommentFileAtomic produces one patch which
+    // in turn produces two token patches
+    val Array(nonAtomicEditArray1, nonAtomicEditArray2, atomicEditArray) =
+      patches.map(_.textEdits()).sortBy(_.length)
+
+    // Check the above holds
+    assert(nonAtomicEditArray1.length == 1 && nonAtomicEditArray2.length == 1)
+    assert(atomicEditArray.length == 2)
+
+    // Check the offsets for the atomic edits look ok (e.g. they're zero-based)
+    val Array(atomicEdit1, atomicEdit2) =
+      atomicEditArray.sortBy(_.position.startLine)
+    assert(
+      atomicEdit1.position.startLine == 0 && atomicEdit1.position.startColumn == 0
+    )
+    assert(
+      atomicEdit2.position.startLine == 7 && atomicEdit2.position.startColumn == 1
+    )
+
+    // Check the same holds for the non-atomic edit pair
+    val Array(nonAtomicEdit1, nonAtomicEdit2) =
+      (nonAtomicEditArray1 ++ nonAtomicEditArray2).sortBy(_.position.startLine)
+    assert(
+      nonAtomicEdit1.position.startLine == 0 && nonAtomicEdit1.position.startColumn == 0
+    )
+    assert(
+      nonAtomicEdit2.position.startLine == 7 && nonAtomicEdit2.position.startColumn == 1
+    )
+  }
+
   def removeUnsuedRule(): SemanticRule = {
     val config = RemoveUnusedConfig.default
     new RemoveUnused(config)
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixFileEvaluationImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixFileEvaluationImpl.scala
@@ -143,7 +143,10 @@ object ScalafixFileEvaluationImpl {
       ctx: RuleCtx,
       index: Option[v0.SemanticdbIndex]
   ): ScalafixFileEvaluationImpl = {
-    val scalafixPatches = patches.map(ScalafixPatchImpl.apply)
+    val indexOrEmpty = index.getOrElse(v0.SemanticdbIndex.empty)
+    val scalafixPatches = patches.map { p =>
+      ScalafixPatchImpl(p)(args, ctx, indexOrEmpty)
+    }
     ScalafixFileEvaluationImpl(
       originalPath = originalPath,
       fixedOpt = fixed,
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixPatchImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixPatchImpl.scala
@@ -2,5 +2,26 @@ package scalafix.internal.interfaces
 
 import scalafix.Patch
 import scalafix.interfaces.ScalafixPatch
+import scalafix.interfaces.ScalafixTextEdit
+import scalafix.internal.patch.PatchInternals
+import scalafix.internal.v1.ValidatedArgs
+import scalafix.v0
+import scalafix.v0.RuleCtx
 
-case class ScalafixPatchImpl(patch: Patch) extends ScalafixPatch
+case class ScalafixPatchImpl(patch: Patch)(
+    args: ValidatedArgs,
+    ctx: RuleCtx,
+    index: v0.SemanticdbIndex
+) extends ScalafixPatch {
+
+  override def textEdits(): Array[ScalafixTextEdit] =
+    PatchInternals
+      .treePatchApply(patch)(ctx, index)
+      .map { t =>
+        ScalafixTextEditImpl(
+          PositionImpl.fromScala(t.tok.pos),
+          t.newTok
+        ): ScalafixTextEdit
+      }
+      .toArray
+}
--- a/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixTextEditImpl.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixTextEditImpl.scala
@@ -1,0 +1,9 @@
+package scalafix.internal.interfaces
+
+import scalafix.interfaces.ScalafixPosition
+import scalafix.interfaces.ScalafixTextEdit
+
+final case class ScalafixTextEditImpl(
+    position: ScalafixPosition,
+    newText: String
+) extends ScalafixTextEdit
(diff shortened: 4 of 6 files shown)


bjaglin:
Hi @LaurenceWarne! Thanks for putting this together, apologies for the (long) delay.

First of all, I am not very knowledgeable about patch internals as I never got the chance to change anything in there, so take my words with a pinch of salt.

I understand your idea is to let a user interactively apply patches, and that sounds like a cool feature! I see 4 options:

1. Iterate over on what you started, adding logic to merge contiguous `TokenPatch`s when possible (most rules use `Patch.replaceTree()`, which creates one patch per token - this is what you observed with `OptionWhenUnless`)
   * The signature makes it impossible to identify/group several `TextEdit`s belonging to the same patch (FTR, `ScalafixFileEvaluationImpl.patches` are all [atomic](https://scalacenter.github.io/scalafix/docs/developers/patch.html#atomic) since https://github.com/scalacenter/scalafix/pull/1261) - say many functions/variables are renamed by a rule, how could the user apply a consistent set of `TextEdit`s if you don't want all changes?
1. Keep the same signature, but rely on the overall file diff (like `previewPatchesAsUnifiedDiff()` does), and extract a set of `TextEdit`s based on that only (I assume 3rd party libraries exist for that?)
   * Same limitation as the last one
   * Merging contiguous changes would come for free (in terms of development cost and also run time as the target is already generated so we wouldn't need to traverse the tree again - this is the most performant option)
   * Risk of stitching unrelated changes, making it potentially impossible for the user to programmatically accept changes at the beginning of a line but rejecting the changes at the end of it (in the case where they came from several atomic patches)
1. Exposing `TextEdit`s on each `scalafix.interfaces.ScalafixPatch`
   * Removes the limitation from (1) and (2)
   * This would provide the biggest flexibility for the client to flatten them (effectively bringing back the first behavior) or let the user review patches independently
1. Exposing a unified diff on each `scalafix.interfaces.ScalafixPatch`
   * A different developer experience than the previous one, that might be more generic, but requires some more logic on the client side
   * No risk of stitching unrelated changes like in (2) since each patch is supposed to be atomic

My current thoughts
- I would rather explore (3) or (4) as the extra effort is likely limited
- I have a small preference for (4) as unidiff seems to be a much more generic/adopted format than a set of `TextEdit`s

Does this help? Let me know what you think. In any case, I'll be happy to help you with reviews (in a much more timely manner than this first one :sweat_smile:), but not much more considering the limited bandwidth I have.

LaurenceWarne:
> Hi @LaurenceWarne! Thanks for putting this together, apologies for the (long) delay.

Hi @bjaglin, no problem at all, thanks for the detailed response!

> Iterate over on what you started, adding logic to merge contiguous TokenPatchs when possible (most rules use Patch.replaceTree(), which creates one patch per token - this is what you observed with OptionWhenUnless)
>    The signature makes it impossible to identify/group several TextEdits belonging to the same patch (FTR, ScalafixFileEvaluationImpl.patches are all [atomic](https://scalacenter.github.io/scalafix/docs/developers/patch.html#atomic) since https://github.com/scalacenter/scalafix/pull/1261) - say many functions/variables are renamed by a rule, how could the user apply a consistent set of TextEdits if you don't want all changes?

I think taking advantage of atomic patches might be enough for my use case actually.  (you might have already guessed :slightly_smiling_face: ) I'm playing around with extending metals' scalafix integration to make it more fine-grained - and I realised I can use a [workspace edit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit) so multiple text edits can be applied at once (for a given code action) and they needn't overlap in any sense - we would only need to map a given atomic patch to a set of text edits if that makes sense.

So we could return a nested collection of text edits, with each individual text edit set corresponding to one atomic patch without any need to resort to heuristics around combining overlapping token patches, etc.  How does that sound?

One problem with this method I found is that it relies on rule authors using `.atomic`, for example, `OptionWhenUnless` doesn't use this (is there a reason that you might not want to use `.atomic` for a rule?).

I think the implementation is a lot easier as well, will push a commit :slightly_smiling_face:

bjaglin:
> So we could return a nested collection of text edits, with each individual text edit set corresponding to one atomic patch without any need to resort to heuristics around combining overlapping token patches, etc. How does that sound?

Good - that's effectively what I was suggesting in (3).
1. Could you just change the signature to expose an `Array[TextEdit]` on `scalafix.interfaces.ScalafixPatch` instead of having a nested array at the root?
1. Unfortunately, we don't have a great unit test coverage of the patch APIs, but could you add an integration test in `ScalafixArgumentsSuite`?

> One problem with this method I found is that it relies on rule authors using .atomic, for example, OptionWhenUnless doesn't use this (is there a reason that you might not want to use .atomic for a rule?).

As [documented](https://scalacenter.github.io/scalafix/docs/developers/patch.html#atomic), rules must use `.atomic` for [suppressions](https://scalacenter.github.io/scalafix/docs/users/suppression.html) to work.

One could argue that `Patch.replaceTree()` should be atomic by default, but I assume there were good reasons not to do it from the beginning. Challenging that would require an investigation of the potential impact on existing rules defining atomic boundaries at a higher level.

So I recommend that you open a PR to add `.atomic` around each `Patch.replaceTree()` for that particular rule, adding suppressions in the test suite to showcase the benefit... And repeat for any other rule that wouldn't play well with the Metals feature you are envisioning.

bjaglin:
> So I recommend that you open a PR to add `.atomic` around each `Patch.replaceTree()` for that particular rule, adding suppressions in the test suite to showcase the benefit...

Interestingly, it seems that `.atomic` [was removed on rules of the same repo](https://github.com/xuwei-k/scalafix-rules/commit/bab0ebc0bce1449dc9ad75647fa854bdecf906c0). If I had to guess why, it would be because the atomic was not granular enough, leading to false positives when suppressing patches.

LaurenceWarne:
> Could you just change the signature to expose an Array[TextEdit] on scalafix.interfaces.ScalafixPatch instead of having a nested array at the root?
    Unfortunately, we don't have a great unit test coverage of the patch APIs, but could you add an integration test in ScalafixArgumentsSuite?

Done!  The test currently checks against how many edits are returned per patch, I can change this to the actual content of the edits f you think this would be better.

> So I recommend that you open a PR to add .atomic around each Patch.replaceTree() for that particular rule, adding suppressions in the test suite to showcase the benefit... And repeat for any other rule that wouldn't play well with the Metals feature you are envisioning.

> Interestingly, it seems that .atomic [was removed on rules of the same repo](https://github.com/xuwei-k/scalafix-rules/commit/bab0ebc0bce1449dc9ad75647fa854bdecf906c0). If I had to guess why, it would be because the atomic was not granular enough, leading to false positives when suppressing patches.

Yeah I saw that too, I can ask the question at least :slightly_smiling_face: - Xuwei's rules are one of the most popular repos I think, but I'll try and test with others.  Do you maintain a curated list of community rules by any chance (I couldn't see mention of it on the microsite)?

Another nice to have would be if `ScalafixPatch`es, had some kind of reference to their rule (I think `RuleDiagnostics` have this, but not patches?), but that's another issue :slightly_smiling_face: .

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEdit.java
```diff
@@ -0,0 +1,13 @@
+package scalafix.interfaces;
+
+public interface ScalafixEdit {
```
bjaglin, line 3: Even if a bit pedantic, 
```suggestion
public interface ScalafixTextEdit {
```
would be more consistent

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEdit.java
```diff
@@ -0,0 +1,13 @@
...
+    int startLine();
+
+    int startColumn();
+
+    int endLine();
+
+    int endColumn();
```
bjaglin, line 10: I initially thought that this was https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit, however not only naming, but also the way the range is expressed differs.

Could we reuse `ScalafixPosition` which is quite well documented?

scalafix-tests/unit/src/test/scala/scalafix/tests/interfaces/ScalafixArgumentsSuite.scala
```diff
@@ -517,6 +517,20 @@ class ScalafixArgumentsSuite extends AnyFunSuite with DiffAssertions {
...
+    val run = api
+      .withRules(List("CommentFileNonAtomic", "CommentFileAtomic").asJava)
+      .withSourceroot(src)
+
+    val fileEvaluation = run.evaluate().getFileEvaluations.head
+    val patches = fileEvaluation.getPatches
+    val edits = patches.map(_.getTextEdits())
+    // CommentFileNonAtomic produces two patches which in turn produce one
+    // token patch each, whilst CommentFileAtomic produces one patch which
+    // in turn produces two token patches
```
bjaglin, line 530: :+1:

bjaglin (commented):
>  The test currently checks against how many edits are returned per patch, I can change this to the actual content of the edits f you think this would be better.

I think it would be good do check content, in order to guard the 0-based behavior of ranges.

>  Do you maintain a curated list of community rules by any chance (I couldn't see mention of it on the microsite)?

There is a [list](https://scalacenter.github.io/scalafix/docs/rules/community-rules.html#hygiene-rules), but it hasn't been updated for a while.

> Another nice to have would be if ScalafixPatches, had some kind of reference to their rule (I think RuleDiagnostics have this, but not patches?), but that's another issue slightly_smiling_face .

Agreed, let's keep that for another PR? For the record, I will cut a new minor release of Scalafix as soon as the new 2.12 and 2.13 patch releases are out (within a couple of weeks if I understand well, given that they depend on the release date of 3.3.0, for which RC5 seems final), so there is an opportunity to get this PR out very soon!

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixEdit.java
LaurenceWarne: sure, sounds good to me.

LaurenceWarne:
> There is a [list](https://scalacenter.github.io/scalafix/docs/rules/community-rules.html#hygiene-rules), but it hasn't been updated for a while.

Cool, I though I remembered one.

> Agreed, let's keep that for another PR? For the record, I will cut a new minor release of Scalafix as soon as the new 2.12 and 2.13 patch releases are out (within a couple of weeks if I understand well, given that they depend on the release date of 3.3.0, for which RC5 seems final), so there is an opportunity to get this PR out very soon!

Yes, sorry that's what I meant, and nice!

bjaglin (approved):
Thanks @LaurenceWarne !

bjaglin:
Sorry that it took longer than expected to cut the release with that!

Somewhat related effort: https://contributors.scala-lang.org/t/roadmap-for-actionable-diagnostics/6172

PR #1744, opened 2023-03-20
expose & fix scalameta deprecations WIP

bjaglin:


docs/developers/tutorial.md
```diff
@@ -181,7 +181,7 @@ To fix this bug, we first match function call nodes `Term.Apply` and pattern
 match only `Lit.Boolean` that appear in argument position
 
 ```scala
-case Term.Apply(_, args) =>
+case Term.Apply.After_4_6_0(_, args) =>
```
bjaglin, line 184: git history for https://github.com/scalacenter/scalafix-named-literal-arguments should be rewritten

scalafix-core/src/main/scala/scalafix/internal/util/PrettyType.scala
```diff
@@ -301,35 +311,44 @@ class PrettyType private (
...
+                            Term.ParamClause(Nil, _) :: Nil
+                          ) if !info.is(p.CASE) =>
                         // Remove redudant () for non-case classes: class Foo
                         ctor.copy(paramss = Nil)
                       case e: Ctor.Primary => e
                     }
                 }
                 .getOrElse {
-                  Ctor.Primary(Nil, Name(""), Seq.empty[Term.ParamClause])
+                  Ctor.Primary.After_4_6_0(
```
kitbellew, line 334: i thought specific versions were only needed for unapply

bjaglin: Good catch, I didn't even remember an apply was created for the specific versions. I guess I just changed that because of a bad search, but there is no reason to. In any case, this PR needs so much rebase that I would (will?) probably start from scratch.

Issue #1662, opened 2022-08-26
Remove support for scalatest 3.0.x in testkit


bjaglin:
From https://github.com/scalacenter/scalafix/pull/1650#discussion_r948473083
> See https://github.com/scalacenter/scalafix/issues/1172#issuecomment-646621781 for the context of why we build/publish against an old version. For the next breaking change version of scalafix, I believe we should remove support for Scalatest 3.0.x (it's been nearly 3 years since the old Scalatest 3.0.x API is deprecated) by just building against the latest, but let's not change that in that PR.

PR #1750, opened 2023-04-16
testkit: build against ScalaTest 3.2

bjaglin:
This is a source-breaking change for testkit users that were relying on the ScalaTest 3.0 package hierarchy, deprecated 3 years ago in 3.1 and removed in 3.2 (despite no major bump).

Scalafix rules were made available to ease that transition, see https://github.com/scalatest/autofix/tree/master.

Closes https://github.com/scalacenter/scalafix/issues/1662.

### DIFF
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -118,11 +118,6 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       buildInfoObject := "RulesBuildInfo"
     )
 
-    lazy val scalatestDep = Def.setting {
-      if (isScala3.value) scalatest.withRevision(scalatestLatestV)
-      else scalatest
-    }
-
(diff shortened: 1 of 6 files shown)


project/ScalafixBuild.scala
```diff
@@ -211,7 +207,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
     // don't publish scala 3 artifacts for now
     publish / skip := (if ((publish / skip).value) true
                        else scalaBinaryVersion.value == "3"),
-    versionPolicyIntention := Compatibility.BinaryCompatible,
+    versionPolicyIntention := Compatibility.None,
```
bjaglin, line 210: I want to merge this just before tagging 0.11.0, otherwise other breaking changes that we still want to minimize will go undetected

PR #1729, opened 2023-01-28
drop support for Scala 2.11

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/1635

PR #1751, opened 2023-04-19
Add Sonatype statistics

julienrf:
Fetch statistics from the Sonatype API and write the results in the repository as CSV files. Also show the data as bar charts in an HTML page available at `sonatype-stats/index.html` (it can directly be opened in a web browser, no need to set up an HTTP server).

Set up a monthly GitHub workflow that updates the stats.

The charts look like the following:

newplot

This one is for `scalafix-core_2.13` only. The script could possible be adjusted to merge both `scalafix-core_2.12` and `scalafix-core_2.13` into a single chart.

Ref: #1729


.github/scripts/sonatype-stats.scala
```diff
@@ -0,0 +1,181 @@
...
+  val start = YearMonth.now(ZoneOffset.UTC)
+
+  val cutOff = start.minusMonths(4L)
+
+  // Note: this assumes the current working directory is the repository root directory!
+  val base = Paths.get("sonatype-stats")
+
+  val artifacts = Set(
+    "scalafix-core_2.12",
+    "scalafix-core_2.13"
```
bjaglin, line 49: `scalafix-interfaces` would also be interesting to better understand how it is used. I guess I could add it later and run an adhoc job via a PR to backfill data for it before next month, but if you have the secret locally it might be easier to do it in this PR?

bjaglin (approved):
Thank you!

bjaglin:
Latest stats are now accessible via https://raw.githack.com/scalacenter/scalafix/main/sonatype-stats/index.html.

PR #1776, opened 2023-06-07
Remove redundant build setting

bjaglin:
By default, the behavior of `publishLocal` is following `publish / skip`.

(found out about this while browing 1.9.0 changelog, where `publishLocal / skip` is now honored but [defaults to `publish / skip`](https://github.com/sbt/sbt/blob/cab5c48e244543c92ae1e0a59d1a23778c626aed/main/src/main/scala/sbt/Defaults.scala#L2625))


bjaglin:
Confirmed in the [CI logs](https://github.com/scalacenter/scalafix/actions/runs/5204718112/jobs/9389308039?pr=1776) that the only occurrences of `published scalafix-cli_` are for 2.12 & 2.13 :ok_hand:

PR #1780, opened 2023-06-11
Fix compilation issues with sun.misc.Unsafe

bjaglin:
Work around https://bugs.openjdk.org/browse/JDK-8206937


bjaglin:
While https://bugs.openjdk.org/browse/JDK-8206937 is real, it seems that it is already mitigated by 
https://github.com/sbt/sbt/issues/6817.

I had the problem locally on a new laptop, but it turned out that the root cause was an empty `/home/brice/.sbt/1.0/java9-rt-ext-eclipse_adoptium_17_0_7/rt.jar` (probably some race condition when booting sbt for the first time?).

PR #1784, opened 2023-06-11
dogfooding: run scalafixAll with local scalafix-interfaces

bjaglin:
This adds a command to replace the `scalafix-interfaces` brought in by `sbt-scalafix` with the local one. Rationale: observe impact of changes & detect regressions.

The implementation is not great as it's slow and mutates state, but I ran into dead-ends for the 2 alternatives I explored:
- injecting a custom `scalafix-interfaces.properties` to force the default scalafix-interfaces to load a custom scalafix-cli would require changes in the classloading logic to favor the custom property file over the one embedded in the JAR
- sbt-scalafix's `scalafixInterfacesProvider` could be rewired to `cli2_12/fullClasspath` with [some minor addition](https://github.com/scalacenter/sbt-scalafix/pull/351), but it would require turning `scalafixInterfacesProvider` into a task key, which is not possible because the parser's completion [can only depend on setting keys](https://github.com/scalacenter/sbt-scalafix/blob/main/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L196-L197)

### DIFF
--- a/project/ScalafixBuild.scala
+++ b/project/ScalafixBuild.scala
@@ -171,6 +171,25 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
         "interfaces/doc" ::
         s
     },
+    commands += Command.command("dogfoodScalafixInterfaces") { state =>
+      val extracted = Project.extract(state)
+      val v =
+        (ThisBuild / version)
+          .get(extracted.structure.data)
+          .get
+      val suffix =
+        (ThisBuild / scalafixScalaBinaryVersion)
+          .get(extracted.structure.data)
+          .get
+          .replace('.', '_')
+
+      s"all cli$suffix/publishLocalTransitive interfaces/publishLocal" ::
+        "reload plugins" ::
+        s"""set dependencyOverrides += "ch.epfl.scala" % "scalafix-interfaces" % "$v"""" :: // as documented in installation.md
+        "session save" ::
+        "reload return" ::
+        state
+    },
     Test / publishArtifact := false,
     licenses := Seq(
       "Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")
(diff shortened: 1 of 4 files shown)


project/ScalafixBuild.scala
```diff
@@ -171,6 +171,23 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
...
+      val suffix =
+        (ThisBuild / scalafixScalaBinaryVersion)
+          .get(extracted.structure.data)
+          .get
+          .replace('.', '_')
+
+      s"all cli$suffix/publishLocalTransitive interfaces/publishLocal" ::
+        "reload plugins" ::
+        s"""set dependencyOverrides += "ch.epfl.scala" % "scalafix-interfaces" % "$v"""" :: // as documented in installation.md
+        "session save" ::
```
bjaglin, line 187: this pollutes a bit the history locally, but it shouldn't hurt
```
sbt:project> show dependencyOverrides
[info] * org.scala-sbt:scripted-plugin:1.8.3
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+10-1a516a0c+20230611-2036-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+10-1a516a0c+20230611-2037-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+10-1a516a0c+20230611-2038-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+11-61c0c9e4+20230611-2100-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+15-54c3f8dd+20230611-2111-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+15-ee65ae33+20230611-2113-SNAPSHOT
[info] * ch.epfl.scala:scalafix-interfaces:0.11.0+15-e1c41e9f+20230611-2116-SNAPSHOT
```

PR #1813, opened 2023-07-06
fix: ensure scala 3 dirs use scala3 dialect for scalafmt

ckipp01:
I think this is the cause of the error we're seeing in the [Scala
Steward](https://github.com/scalacenter/steward/actions/runs/5470172748/jobs/9959938109#step:5:870)
where the scala213 dialect has issues with the Scala 3 code. This small
changes just adds in some fileOverrides to make sure the scala-3 dirs
set the runner.dialect as scala3 instead of scala213.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -12,3 +12,9 @@ project.excludeFilters = [
   scalafix-tests/shared/src/main/scala/test/PrettyTest.scala
   scalafix-sbt/src/main/scala/scalafix/internal/sbt/ScalafixRuleNames.scala
 ]
+
+fileOverride {
+  "glob:**/scala-3/**" {
+     runner.dialect = scala3
+  }
+}


ckipp01:
Odd, `./bin/scalafmt --check` fails in CI, but locally when running `.bin/scalafmt` it doesn't format anything... Also does that get dynamically updated in CI or something? I'm not really sure why my change would impact this though 🤔 .

bjaglin:
> Also does that get dynamically updated in CI or something?

:exploding_head:  https://github.com/scalacenter/scalafix/pull/1784#discussion_r1225908483 ([ran just before scalafmt](https://github.com/scalacenter/scalafix/blob/d1f8cfc902dc7e2f746821f488f021a5e05b8d8c/.github/workflows/ci.yml#L41)) does mutate the faulty `project/plugins.sbt`. I'll open a PR to run it last (it makes sense anyway as scalafmt is faster to run and more likely to fail).

2 things remain mysterious to me:
- locally, `dogfoodScalafixInterfaces`' `session save` does not touch `project/plugins.sbt`, why does it do it on CI?
- why did the `--test` picks up the missing LF in this PR (and this PR only)?

bjaglin:
Regarding the original trigger for this PR: we are actually [back to green](https://github.com/scalacenter/steward/actions).

Failing runs (and failing runs only) have "ERROR Steward scalacenter/bloop-config failed", so I assume the problem is related to that.
```
  org.scalasteward.core.io.process$ProcessFailedException: '"SBT_OPTS=-Xmx2048m -Xss8m -XX:MaxMetaspaceSize=512m" mill -i -p /home/runner/scala-steward/workspace/repos/scalacenter/bloop-config/scala-steward.sc show org.scalasteward.mill.plugin.StewardPlugin/extractDeps' exited with code 1.
   Cannot resolve -p. Try `mill resolve _` to see what's available.
```

I am not sure what changed in bloop-config/steward on July 6th though...

.scalafmt.conf
```diff
@@ -12,3 +12,18 @@ project.excludeFilters = [
...
+  }
+  "glob:**/scalafix-reflect/src/main/scala-3/**" {
+     runner.dialect = scala3
+  }
+  "glob:**/scalafix-core/src/main/scala-3/**" {
+     runner.dialect = scala3
+  }
+  "glob:**/scalafix-tests/shared/src/main/scala-3/**" {
+     runner.dialect = scala3
+  }
```
bjaglin, line 28: can the glob match arbitrary depth?
```suggestion
  "glob:**/scala-3/**" {
     runner.dialect = scala3
  }
```

ckipp01: I believe so yes, that's how we have it in Metals, which is how I copied these over.

bjaglin: could we take in my suggestion then to avoid hardcoding module paths?

ckipp01: Hey, yup, sorry for the delay.

bjaglin (approved):
:bow:

PR #1817, opened 2023-07-11
Allow package when referencing rules in github

RustedBones:
Referencing scalafix rules does not currently allow rules to be located in a package other than `fix`.
Eg. in [scio](https://github.com/spotify/scio/tree/main/scalafix/rules/src/main/scala/fix) we've grouped rules in packages corresponding to the breaking version.

This change allows to reference a rule based on its package name (stripping the required `fix` package prefix)

### DIFF
--- a/docs/developers/tutorial.md
+++ b/docs/developers/tutorial.md
@@ -654,6 +654,7 @@ The expansion rules for `github:org/repo` are the following:
 | `github:org/repo`                      | `scalafix/rules/src/main/scala/fix/Repo.scala`                           |
 | `github:org/some-repo`                 | `scalafix/rules/src/main/scala/fix/SomeRepo.scala`                       |
 | `github:org/repo/RuleName`             | `scalafix/rules/src/main/scala/fix/RuleName.scala`                       |
+| `github:org/repo/package.RuleName`     | `scalafix/rules/src/main/scala/fix/package/RuleName.scala`               |
 | `github:org/repo/RuleName?sha=HASH125` | (at commit `HASH125`) `scalafix/rules/src/main/scala/fix/RuleName.scala` |
 
 ## Publish the rule to Maven Central
--- a/docs/developers/tutorial.md
+++ b/docs/developers/tutorial.md
@@ -650,11 +650,12 @@ scalafix --rules=github:scalacenter/named-literal-arguments
 The expansion rules for `github:org/repo` are the following:
 
 | Before                                 | After                                                                    |
-| -------------------------------------- | ------------------------------------------------------------------------ |
-| `github:org/repo`                      | `scalafix/rules/src/main/scala/fix/Repo.scala`                           |
-| `github:org/some-repo`                 | `scalafix/rules/src/main/scala/fix/SomeRepo.scala`                       |
-| `github:org/repo/RuleName`             | `scalafix/rules/src/main/scala/fix/RuleName.scala`                       |
-| `github:org/repo/package.RuleName`     | `scalafix/rules/src/main/scala/fix/package/RuleName.scala`               |
+|----------------------------------------|--------------------------------------------------------------------------|
+| `github:org/repo`                      | (on `master`) `scalafix/rules/src/main/scala/fix/Repo.scala`             |
+| `github:org/some-repo`                 | (on `master`) `scalafix/rules/src/main/scala/fix/SomeRepo.scala`         |
+| `github:org/repo/RuleName`             | (on `master`) `scalafix/rules/src/main/scala/fix/RuleName.scala`         |
+| `github:org/repo/com.example.RuleName` | (on `master`) `scalafix/rules/src/main/scala/com/example/RuleName.scala` |
+| `github:org/repo/RuleName?sha=main`    | (on `main`) `scalafix/rules/src/main/scala/fix/RuleName.scala`           |
 | `github:org/repo/RuleName?sha=HASH125` | (at commit `HASH125`) `scalafix/rules/src/main/scala/fix/RuleName.scala` |
 
 ## Publish the rule to Maven Central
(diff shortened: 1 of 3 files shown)


scalafix-reflect/src/main/scala/scalafix/internal/reflect/GitHubUrlRule.scala
```diff
@@ -3,20 +3,24 @@ package scalafix.internal.reflect
...
+import scala.util.Try
+
 import metaconfig.Conf
 import metaconfig.ConfError
 import metaconfig.Configured
 import metaconfig.Configured.Ok
 
 object GitHubUrlRule {
 
+  private val DefaultBranch = "master"
```
RustedBones, line 15: Should this actually be changed to `main` ?

scalafix-tests/integration/src/test/scala/scalafix/tests/config/GitHubUrlRuleSuite.scala
```diff
@@ -37,6 +37,11 @@ class GitHubUrlRuleSuite extends AnyFunSuite with DiffAssertions {
     "https://raw.githubusercontent.com/someorg/some-repo/master/scalafix/rules/" +
       "src/main/scala/fix/RuleName.scala"
   )
+  check(
+    "github:someorg/some-repo/version.RuleName",
```
bjaglin, line 41: could we add a test with several segments? like `my.package.RuleName`

scalafix-reflect/src/main/scala/scalafix/internal/reflect/GitHubUrlRule.scala
```diff
@@ -32,17 +36,21 @@ object GitHubUrlRule {
   private def guessGitHubURL(
       org: String,
       repo: String,
-      filename: String,
+      rule: String,
```
bjaglin, line 39: maybe `className` (here and above) to highlight that `.` is used as separator and that extension is not provided? 
```suggestion 
      className: String,
```

scalafix-reflect/src/main/scala/scalafix/internal/reflect/GitHubUrlRule.scala
```diff
@@ -54,35 +62,27 @@ object GitHubUrlRule {
   private val GitHubFallback =
     """github:(.*)""".r
 
-  private val alphanumerical = "[^a-zA-Z0-9]"
+  private val NonAlphaNumeric = "[^a-zA-Z0-9]"
```
bjaglin, line 65: :+1:

docs/developers/tutorial.md
```diff
@@ -654,6 +654,7 @@ The expansion rules for `github:org/repo` are the following:
 | `github:org/repo`                      | `scalafix/rules/src/main/scala/fix/Repo.scala`                           |
 | `github:org/some-repo`                 | `scalafix/rules/src/main/scala/fix/SomeRepo.scala`                       |
 | `github:org/repo/RuleName`             | `scalafix/rules/src/main/scala/fix/RuleName.scala`                       |
+| `github:org/repo/package.RuleName`     | `scalafix/rules/src/main/scala/fix/package/RuleName.scala`               |
```
bjaglin, line 657: if package is explicit, shouldn't we remove the `fix` prefix?
```suggestion
| `github:org/repo/package.RuleName`     | `scalafix/rules/src/main/scala/package/RuleName.scala`               |
```

scalafix-tests/integration/src/test/scala/scalafix/tests/config/GitHubUrlRuleSuite.scala
bjaglin: also, a test with the sha maybe? (it's not documented but I would expect the most verbose/complicated syntax to support all features)

bjaglin (commented):
Thanks for the PR @RustedBones, and sorry for the delay. I am definitely in favor of merging this after a few minor adjustments :+1:

docs/developers/tutorial.md
RustedBones: Looks that scalafix requires the root package to be `fix`.
When only the rule name is provided, it searches in the `fix` package, this changes keep the same behavior

bjaglin: > Looks that scalafix requires the root package to be fix.

I am not aware of this constraint in scalafix-core, and none of the built-in rules is rooted in `fix`. Can you elaborate where you observed this limitation?

>  When only the rule name is provided, it searches in the fix package, this changes keep the same behavior

I agree that the existing behavior should remain, but it feels strange that the user provides a package that is only a suffix. If the example would look like this, would it make more sense?
```suggestion
| `github:org/repo/fix.package.RuleName`     | `scalafix/rules/src/main/scala/fix/package/RuleName.scala`               |
```

RustedBones: My bad, I double checked `scalafix` uses the rule name only, regardless of the class name/package.
I agree with the suggestion. Will update the PR

docs/developers/tutorial.md
```diff
@@ -650,10 +650,12 @@ scalafix --rules=github:scalacenter/named-literal-arguments
...
-| -------------------------------------- | ------------------------------------------------------------------------ |
-| `github:org/repo`                      | `scalafix/rules/src/main/scala/fix/Repo.scala`                           |
-| `github:org/some-repo`                 | `scalafix/rules/src/main/scala/fix/SomeRepo.scala`                       |
-| `github:org/repo/RuleName`             | `scalafix/rules/src/main/scala/fix/RuleName.scala`                       |
+|----------------------------------------|--------------------------------------------------------------------------|
+| `github:org/repo`                      | (on `master`) `scalafix/rules/src/main/scala/fix/Repo.scala`             |
+| `github:org/some-repo`                 | (on `master`) `scalafix/rules/src/main/scala/fix/SomeRepo.scala`         |
+| `github:org/repo/RuleName`             | (on `master`) `scalafix/rules/src/main/scala/fix/RuleName.scala`         |
+| `github:org/repo/com.example.RuleName` | (on `master`) `scalafix/rules/src/main/scala/com/example/RuleName.scala` |
+| `github:org/repo/RuleName?sha=main`    | (on `main`) `scalafix/rules/src/main/scala/fix/RuleName.scala`           |
```
bjaglin, line 658: :+1:

bjaglin (approved):
Thanks!

PR #1830, opened 2023-07-20
Update OrganizeImports.md

liancheng:
Fix a link to the `INTELLIJ_2020_3` option.


bjaglin (approved):
Thanks! There were a lot of manual steps in the ascii -> markdown conversion so I wouldn't be surprised if there were other broken links :/

PR #1832, opened 2023-07-24
Try to parse all valid Testkit HOCON comments

mdedetrich:
Resolves: https://github.com/scalacenter/scalafix/issues/1828

Pinging @bjaglin since you commented on the original issue. The only thing that is missing is a test, its my first time working on scalafix so not entirely sure where to put such a test and what type of test.

### DIFF
--- a/scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
+++ b/scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
@@ -31,36 +31,47 @@ object RuleTest {
       val input = test.toInput
       val dialect = args.scalaVersion.dialect(args.sourceScalaVersion)
       val tree = dialect(input).parse[Source].get
-      val comment = SemanticRuleSuite.findTestkitComment(tree.tokens)
-      val syntax = comment.syntax.stripPrefix("/*").stripSuffix("*/")
-      val conf = Conf.parseString(test.testName, syntax).get
-      val scalafixConfig = conf.as[ScalafixConfig].get
-      val doc = v1.SyntacticDocument(
-        tree.pos.input,
-        LazyValue.now(tree),
-        DiffDisable.empty,
-        scalafixConfig
-      )
-      val sdoc =
-        v1.SemanticDocument.fromPath(
-          doc,
-          test.semanticdbPath,
-          classLoader,
-          symtab
-        )
-      val decoderSettings =
-        RuleDecoder.Settings().withConfig(scalafixConfig)
-      val decoder = RuleDecoder.decoder(decoderSettings)
-      val rulesConf = ConfGet
-        .getKey(conf, "rules" :: "rule" :: Nil)
-        .getOrElse(Conf.Lst(Nil))
-      val config = Configuration()
-        .withConf(conf)
-        .withScalaVersion(args.scalaVersion.value)
-        .withScalacOptions(args.scalacOptions)
-        .withScalacClasspath(args.classpath.entries)
-      val rules = decoder.read(rulesConf).get.withConfiguration(config).get
-      (rules, sdoc)
+      // Its possible to have comments which isn't valid HOCON (i.e.
+      // license headers), so lets parse until we find valid HOCON.
+      val allComments = SemanticRuleSuite.filterPossibleTestkitComments(tree.tokens)
+      val lazilyParsedComments = allComments.view.map{ comment =>
+        val syntax = comment.syntax.stripPrefix("/*").stripSuffix("*/")
+        for {
+          conf <- Conf.parseString(test.testName, syntax).toEither
+          scalafixConfig <- conf.as[ScalafixConfig].toEither
+        } yield {
+          val doc = v1.SyntacticDocument(
+            tree.pos.input,
+            LazyValue.now(tree),
+            DiffDisable.empty,
+            scalafixConfig
+          )
+          val sdoc =
+            v1.SemanticDocument.fromPath(
+              doc,
+              test.semanticdbPath,
+              classLoader,
+              symtab
+            )
+          val decoderSettings =
+            RuleDecoder.Settings().withConfig(scalafixConfig)
+          (RuleDecoder.decoder(decoderSettings), conf, sdoc)
+        }
+      }
+
+      lazilyParsedComments.collectFirst {
+        case Right((decoder, conf, sdoc)) =>
+          val rulesConf = ConfGet
+            .getKey(conf, "rules" :: "rule" :: Nil)
+            .getOrElse(Conf.Lst(Nil))
+          val config = Configuration()
+            .withConf(conf)
+            .withScalaVersion(args.scalaVersion.value)
+            .withScalacOptions(args.scalacOptions)
+            .withScalacClasspath(args.classpath.entries)
+          val rules = decoder.read(rulesConf).get.withConfiguration(config).get
+          (rules, sdoc)
+      }.getOrElse(throw new IllegalArgumentException(s"Expected a single comment at ${test.testPath} with a valid rule or rules key defined"))
     }
 
     new RuleTest(test, run)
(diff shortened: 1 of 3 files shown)


scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -31,36 +30,48 @@ object RuleTest {
...
+            .getOrElse(Conf.Lst(Nil))
+          val config = Configuration()
+            .withConf(conf)
+            .withScalaVersion(args.scalaVersion.value)
+            .withScalacOptions(args.scalacOptions)
+            .withScalacClasspath(args.classpath.entries)
+          val rules = decoder.read(rulesConf).get.withConfiguration(config).get
+          (rules, sdoc)
+      } match {
+        case None =>
```
bjaglin, line 72: it looks like you forgot to unpack the happy path

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -31,36 +30,48 @@ object RuleTest {
...
+            Some((RuleDecoder.decoder(decoderSettings), conf, sdoc))
+          case Configured.NotOk(_) => None
+        }
+      }
+
+      lazilyParsedComments.collectFirst {
+        case Some((decoder, conf, sdoc)) =>
+          val rulesConf = ConfGet
+            .getKey(conf, "rules" :: "rule" :: Nil)
+            .getOrElse(Conf.Lst(Nil))
```
bjaglin, line 63: per the message in the exception, this should be lifted up to the `lazilyParsedComments` closure to effectively skip comments that are HOCON but do not contain `rule` or `rules`

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -31,36 +30,48 @@ object RuleTest {
...
+          val config = Configuration()
+            .withConf(conf)
+            .withScalaVersion(args.scalaVersion.value)
+            .withScalacOptions(args.scalacOptions)
+            .withScalacClasspath(args.classpath.entries)
+          val rules = decoder.read(rulesConf).get.withConfiguration(config).get
+          (rules, sdoc)
+      } match {
+        case None =>
+          throw new IllegalArgumentException("Expected a single comment with a valid rule or rules key defined")
```
bjaglin, line 73: adding the context (path in `test`) would help here

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -2,8 +2,7 @@ package scalafix.testkit
 
 import scala.meta._
 import scala.meta.internal.symtab.SymbolTable
-
-import metaconfig.Conf
+import metaconfig.{Conf, Configured}
```
bjaglin, line 5: make sure to run the [formatters](https://github.com/scalacenter/scalafix/blob/main/CONTRIBUTING.md#formatting) to please the CI

bjaglin (commented):
Thanks for the PR!

For testing, I would suggest an integration test, simply copy/pasting https://github.com/scalacenter/scalafix/blob/main/scalafix-tests/input/src/main/scala-2/test/LintAsserts.scala (the `Disable` rule is useless though) in its current directory, and adding a license header.

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
mdedetrich: So I thought the failure in parsing is done in `Conf.parseString(test.testName, syntax)` (which is why I pattern match against it). The idea behind my solution is that we create a lazy collection called `lazilyParsedComments` of successfully/unsuccessfully parsed comments and then we do `collectFirst` to get the first successful parsing.

In other words, the logic of figuring out what is a correct HOCON comment has already happened by the time we get to `rulesConf`

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
mdedetrich: The idea is that we ignore the unhappy path, i.e. we only care about the first success case which occurs on `collectFirst`.

Or am I missing something?

mdedetrich:
@bjaglin Thanks for reviewing the PR, I addressed some of the comments but I may have an incorrect understanding of what the problem is since it appears that `RuleSuite` is not passing, investigating it more

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
mdedetrich: Done

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
mdedetrich: Done

mdedetrich:
@bjaglin Okay so I have done some improvements to the PR, on my local machine `sbt testOnly scalafix.tests.rule.RuleSuite` is passing, some other tests are failing but that appears to be unrelated (jgit/git/signing related stuff).

Would it be possible to re-run the CI just to confirm I didn't create any regressions?

bjaglin:
> Would it be possible to re-run the CI just to confirm I didn't create any regressions?

Done. I also changed [the default](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories), so I shouldn't need to approve execution at next push :crossed_fingers:

bjaglin (commented):
You can commit the test file and run the formatters and we should be good to go :+1:

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
bjaglin: > This is definitely an improvement and should cover your original use-case of a long comment

Actually, it does not - the test I added in fd35cf2b6e63fb251e73a152a0badede1b93ab75 shows it. I will try to do what I was suggesting in another commit and see if it helps.

mdedetrich:
@bjaglin Thanks for helping out, I will try and have a look at why the test is failing in the next few days

scalafix-testkit/src/main/scala/scalafix/testkit/RuleTest.scala
```diff
@@ -31,36 +33,54 @@ object RuleTest {
...
-          symtab
+      // Its possible to have comments which isn't valid HOCON (i.e.
+      // license headers), so lets parse until we find valid HOCON.
+      val allComments =
+        SemanticRuleSuite.filterPossibleTestkitComments(tree.tokens)
+      val lazilyParsedComments = allComments.view.map { comment =>
+        val syntax = comment.syntax.stripPrefix("/*").stripSuffix("*/")
+        for {
+          conf <- Try(Conf.parseString(test.testName, syntax)).toEither
+            .flatMap(_.toEither)
```
mdedetrich, line 44: @bjaglin I didn't realize you were working on this right now, thanks for fixing the bug!

I think there is a slight improvement that can be done here, this expression ends up getting the type `Either[Serializable, _]` because the first `toEither` has `Throwable` on the left where as the second has `ConfError`, should I fix this?

bjaglin: I am planning to cut a release tonight ahead of the official release of scala 2.13.12, that's why I wanted to get this out ASAP.

As for the type, I am not sure it really matters since we never unpack the left? I'll merge this if it's green in an hour or so, feel free to push if you see further improvements.

mdedetrich: I already have the fix written, its quite simple. Just running scalafmt/scalafix locally and ill push immediately after

mdedetrich:
@bjaglin Just pushed the fix I was talking about, tests are passing locally so all should be good.

bjaglin:
Thanks @mdedetrich !

PR #1851, opened 2023-08-27
RedundantSyntax: distinguish single and triple quotes

bjaglin:
:exploding_head: 

```scala
Welcome to Scala 2.13.11 (OpenJDK 64-Bit Server VM, Java 17.0.7).
Type in expressions for evaluation. Or try :help.

scala> s""" \n """ == """ \n """ 
val res0: Boolean = false

scala> s" \n " == " \n "
val res1: Boolean = true

scala> f""" \n """ == """ \n """
val res2: Boolean = false

scala> f" \n " == " \n "
val res3: Boolean = true

scala> raw""" \n """ == """ \n """
val res4: Boolean = true

scala> raw" \n " == " \n "
val res5: Boolean = false
```

Test input/output had to be duplicated because 2.12 does not support `s"my \"quoted\" string"`

### DIFF
--- a/scalafix-tests/input/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
+++ b/scalafix-tests/input/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
@@ -1,0 +1,44 @@
+/*
+rules = RedundantSyntax
+RedundantSyntax.stringInterpolator = true
+*/
+
+package test.redundantSyntax
+
+class StringInterpolator {
+
+  val a = 42d
+
+  var b = ""
+  b = "foo"
+  b = s"foo"
+  b = s"foo $a bar"
+  b = s"my \"quoted\" string"
+
+  b = """foo"""
+  b =
+    s"""foo
+       |bar"""
+  b = s"""my \"quoted\" string"""
+  b = s"""foo $a bar"""
+  b = s"""$a"""
+
+  b = f"foo"
+  b = f"foo $a%2.2f"
+  b = f"foo \n bar"
+
+  b = raw"foo $a \nbar"
+  b = raw"""foo\nbar\\"""
+  b = raw"foo\nbar\\"
+  b = raw"foo bar"
+  b = raw"a\*b\+"
+
+  b = my"foo"
+  b = my"foo $a bar"
+
+  b = s"foo" // scalafix:ok
+
+  implicit class MyInterpolator(sc: StringContext) {
+    def my(subs: Any*): String = sc.toString + subs.mkString("")
+  }
+}
--- a/scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
+++ b/scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
@@ -1,8 +1,3 @@
-/*
-rules = RedundantSyntax
-RedundantSyntax.stringInterpolator = true
-*/
-
 package test.redundantSyntax
 
 class StringInterpolator {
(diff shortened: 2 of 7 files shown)


bjaglin:
Thanks for the report @bmarker, this was a "fun" fix to make!

scalafix-tests/output/src/main/scala/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,21 @@ class StringInterpolator {
   b = "foo"
   b = "foo"
   b = s"foo $a bar"
+  b = "my \"quoted\" string"
 
   b = """foo"""
   b =
     """foo
        |bar"""
+  b = s"""my \"quoted\" string"""
```
bjaglin, line 17: this is the actual fix

scalafix-tests/output/src/main/scala/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,21 @@ class StringInterpolator {
   b = "foo"
   b = "foo"
   b = s"foo $a bar"
+  b = "my \"quoted\" string"
```
bjaglin, line 11: this makes sure we still remove the prefix when we can

bjaglin:
Oh, it's even funnier than I thought, it works in 2.13, but the compiler has a different behavior in 2.12 & 3 so we need a better heuristic to capture usage of `"""` with scalameta...

scalafix-tests/input/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -13,18 +13,22 @@ class StringInterpolator {
   b = "foo"
   b = s"foo"
   b = s"foo $a bar"
+  b = s"my \"quoted\" string"
```
bjaglin, line 16: this tests the actual fix for the reported problem

scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,22 @@ class StringInterpolator {
...
+  b = s"""my \"quoted\" string"""
   b = s"""foo $a bar"""
   b = s"""$a"""
 
   b = "foo"
   b = f"foo $a%2.2f"
+  b = "foo \n bar"
 
   b = raw"foo $a \nbar"
+  b = """foo\nbar\\"""
```
bjaglin, line 26: this makes sure we still remove the interpolator when we can

scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,22 @@ class StringInterpolator {
...
   b =
     """foo
        |bar"""
+  b = s"""my \"quoted\" string"""
   b = s"""foo $a bar"""
   b = s"""$a"""
 
   b = "foo"
   b = f"foo $a%2.2f"
+  b = "foo \n bar"
```
bjaglin, line 23: this makes sure we still remove the interpolator when we can

scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,22 @@ class StringInterpolator {
   b = "foo"
   b = "foo"
   b = s"foo $a bar"
+  b = "my \"quoted\" string"
```
bjaglin, line 11: this makes sure we still remove the interpolator when we can

scalafix-tests/output/src/main/scala-2.13/test/redundantSyntax/StringInterpolator.scala
```diff
@@ -8,18 +8,22 @@ class StringInterpolator {
...
 
   b = """foo"""
   b =
     """foo
        |bar"""
+  b = s"""my \"quoted\" string"""
```
bjaglin, line 17: this tests the actual fix for the reported problem

bjaglin:
cc @gontard as the original rule author for visiblity

gontard:
👋 Thanks for the fix!

PR #1864, opened 2023-09-10
improve handling of testkit semantic comments

bjaglin:
- strip THE semantic comment and not others for input/output diff (follow-up of https://github.com/scalacenter/scalafix/issues/1828)
- ignore HOCON-valid comments that do not contain a rules attribute


bjaglin:
@mdedetrich FYI, I am improving a bit more the resilience and the visibility of testkit semantic comments

scalafix-tests/input/src/main/scala-2/NoSemanticComment.scala
```diff
@@ -0,0 +1,4 @@
+package test
+
+// This has no testkit semantic comment and should be gracefully ignored by the testkit
+object NoSemanticComment
```
mdedetrich, line 4: Nit: newline

bjaglin: FTR, such files should actually raise an error per the current implementation, so I removed it

scalafix-tests/shared/src/main/scala/org/scalatest_autofix/Matchers.scala
```diff
@@ -1,6 +1,3 @@
-/*
-ignore = true
```
bjaglin, line 2: A HOCON-parseable comment was enough to please the teskit comment parser. This file was relying that behavior to define a helper that should not be evaluated as an input. This is error-prone as the user might have just forgotten to add the rule to be evaluated, so this PR skips HOCON-parseable comments that do not contain a `rules` or `rule` attribute.

PR #1865, opened 2023-09-10
allow Scala nightlies via system properties

bjaglin:
https://github.com/scala/community-build/issues/1673

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,9 +7,9 @@ import scala.util.Try
 /* scalafmt: { maxColumn = 120 }*/
 
 object Dependencies {
-  val scala212 = "2.12.18"
-  val scala213 = "2.13.12"
-  val scala3 = "3.3.1"
+  val scala212 = sys.props.getOrElse("scala212.nightly", "2.12.18")
+  val scala213 = sys.props.getOrElse("scala213.nightly", "2.13.12")
+  val scala3 = sys.props.getOrElse("scala3.nightly", "3.3.1")
 
   val buildScalaVersions = Seq(scala212, scala213, scala3)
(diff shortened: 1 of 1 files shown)


bjaglin:
cc @SethTisue

project/Dependencies.scala
```diff
@@ -8,7 +8,7 @@ import scala.util.Try
 
 object Dependencies {
   val scala212 = "2.12.18"
-  val scala213 = "2.13.12"
+  val scala213 = sys.props.getOrElse("scala.nightly", "2.13.12")
```
bjaglin, line 11: not sure how scala-steward will handle that, but since the bump is manual during the embargo, it does not matter much

project/Dependencies.scala
```diff
@@ -7,9 +7,9 @@ import scala.util.Try
 /* scalafmt: { maxColumn = 120 }*/
 
 object Dependencies {
-  val scala212 = "2.12.18"
-  val scala213 = "2.13.12"
-  val scala3 = "3.3.1"
+  val scala212 = sys.props.getOrElse("scala212.nightly", "2.12.18")
+  val scala213 = sys.props.getOrElse("scala213.nightly", "2.13.12")
```
bjaglin, line 11: @SethTisue I have suffixed the system property with the scala version to make it possible to tweak any scala version, I assume this is not a problem as it's explicitly forwarded [there](https://github.com/SethTisue/community-build/blob/dcf2a870c2803a39006ae7a03ab7ddab2b65c29f/core/scalafix.conf#L10) so it can easily be changed?

SethTisue: yup that's fine

SethTisue (approved):

Issue #1889, opened 2023-11-08
Cannot ignore DisableSyntax.regex which starts with .*


tubo28:
This maybe bug report, but we cannot ignore `DisableSyntax.regex` rules with the pattern starts with `.*` by `scalafix:ok`.

It reproduces with `Main.scala` like this:

```scala
package example

object Hello extends Greeting with App {
  println(greeting)
}

trait Greeting {
  lazy val greeting: String = "hello"
  val foo = 0 // scalafix:ok
  val bar = 0 // scalafix:ok
}
```

and `.scalafmt.conf` like this:

```hocon
rules = [
  DisableSyntax
]

DisableSyntax.regex = [
  {
    id = FooChecker
    pattern = ".*foo" // somothing goes wrong; we cannot ignore this
    message = foo is bad
  }
  {
    id = BarChecker
    pattern = "bar" // ok
    message = bar is bad
  }
]
```

Then result:

```sbt
sbt:scalafixtmp> scalafix
[info] Running scalafix on 1 Scala sources
[error] /Users/user/ghq/github.com/tubo28/scalafixtmp/src/main/scala/example/Hello.scala:9:1: error: [DisableSyntax.FooChecker] foo is bad
[error]   val foo = 0 // scalafix:ok
[error] ^^^^^^^^^
[warn] /Users/user/ghq/github.com/tubo28/scalafixtmp/src/main/scala/example/Hello.scala:9:15: warning: [UnusedScalafixSuppression] Unused Scalafix suppression, this can be removed
[warn]   val foo = 0 // scalafix:ok
[warn]               ^^^^^^^^^^^^^^
[error] (Compile / scalafix) scalafix.sbt.ScalafixFailed: LinterError
[error] Total time: 0 s, completed 2023/11/09 3:01:04
```

Minimum project to reproduce this is here https://github.com/tubo28/scalafixtmp

Scala 2.13.12
Sbt: 1.9.7
Scalafix: `"sbt-scalafix" % "0.11.1"`

bjaglin:
Thanks for the clear report!

What happens here is that the wildcard capture whitespace characters before the literal, so the suppression mechanism is not working. You can see that by looking at the carets highlighting all heading chars until `foo` in the error message.

I'll see if we could make it a bit more robust, but for the time being, if you want to catch literals ending with `foo` AND be able to suppress them, `pattern = "[^\\s]*foo"` should do the trick.

skitanoooo:
[...]

PR #1890, opened 2023-11-09
rules capturing leading tokens can be suppressed with scalafix:ok

bjaglin:
Fix https://github.com/scalacenter/scalafix/issues/1889


scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
```diff
@@ -422,9 +423,11 @@ object EscapeHatch {
             case comment @ Token.Comment(ScalafixOkRgx(rules)) =>
               if (!visitedFilterExpression.contains(comment.pos)) {
                 val rulesPos = rulesWithPosition(rules, comment)
-                val start = EscapeOffset(t.pos.start)
+                // include line start in case the rule captures leading tokens
+                val start = EscapeOffset(t.pos.start - t.pos.startColumn)
```
bjaglin, line 427: this is the actual fix - the rest of the changes is to satisfy `scalafix-tests/input/src/main/scala-2/test/escapeHatch/Issue694.scala`

PR #1892, opened 2023-11-16
Add isAtomic to ScalafixPatch interface

LaurenceWarne:
Hi, I thought it might be helpful to surface whether or not a patch is atomic in the `ScalafixPatch` API.  The idea is this indicates whether or not this patch makes sense to apply in isolation (e.g. through the `textEdits` method).

Thanks!

### DIFF
--- a/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
+++ b/scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
@@ -8,4 +8,12 @@ public interface ScalafixPatch {
     default ScalafixTextEdit[] textEdits() {
         throw new UnsupportedOperationException("textEdits() is not implemented");
     }
+
+    /**
+     *
+     * @return True if this patch is atomic, else false.
+     */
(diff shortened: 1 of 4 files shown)


scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixPatch.java
```diff
@@ -8,4 +8,12 @@ public interface ScalafixPatch {
     default ScalafixTextEdit[] textEdits() {
         throw new UnsupportedOperationException("textEdits() is not implemented");
     }
+
+    /**
+     *
+     * @return True if this patch is atomic, else false.
+     */
+    default boolean isAtomic() {
+        throw new UnsupportedOperationException("isAtmoic() is not implemented");
```
bjaglin, line 17:
```suggestion
        throw new UnsupportedOperationException("isAtomic() is not implemented");
```

LaurenceWarne: Ah good spot :+1:

Issue #1894, opened 2023-11-21
OrganizeImports uses wildcard pattern "_" instead of asteriks in Scala 3


longliveenduro:

bjaglin:
Syntax of wildcard is currently preserved (see [this](https://github.com/liancheng/scalafix-organize-imports/issues/127) and [this](https://github.com/liancheng/scalafix-organize-imports/pull/280)), so I assume you are referring to the wildcards generated by `coalesceToWildcardImportThreshold`, which you expect to be `*` in Scala 3 (and Scala 2 with `-Xsource:3`?), but `_` otherwise?

longliveenduro:
@bjaglin Thanks for coming back to this. Yes you are right.

I use `coalesceToWildcardImportThreshold` but the generated groups have the `_` wildcard instead of the `*` wildcard of Scala 3

bjaglin:
@longliveenduro I am about to introduce a new configuration flag in https://github.com/scalacenter/scalafix/pull/1896 to address that, let me know what you think

longliveenduro:
@bjaglin That is a big PR. Where is the relevant part for this issue?

bjaglin:
@longliveenduro The PR is the fix for this issue, nothing more. I understand the implementation is a bit technical, so I was only asking for your feedback on the UI - namely what's documented in the [first markdown file appearing in the diff](https://github.com/scalacenter/scalafix/pull/1896/files#diff-5792f094ca4afed0d8ab56e05ce17edba168c5147e59863a851908013489298f).

longliveenduro:
@bjaglin Thanks for your reply. That would do the job for me. LGTM. Thanks!

PR #1666, opened 2022-09-10
add support for -Xsource:3 testkit input & output

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1664#pullrequestreview-1096671581

This adds adds some complexity in the inter-project lookup mechanism so that some projects introduced in https://github.com/scalacenter/scalafix/pull/1756 can be removed, in order to reduce (almost cancel) the extra build cardinality introduced by this PR .

```diff
--- main
+++ this PR
-[info] resolving key references (53896 settings) ...
+[info] resolving key references (56220 settings) ...
 sbt:scalafix> projects
 [info] In file:/home/brice/git/scala/scalafix/
 [info] 	   cli2_12
 [info] 	   cli2_13
 [info] 	   cli3
 [info] 	   core2_12
 [info] 	   core2_13
 [info] 	   core3
 [info] 	   docs2_13
 [info] 	   expect2_12Target2_12_16
 [info] 	   expect2_12Target2_12_17
 [info] 	   expect2_12Target2_12_18
+[info] 	   expect2_12Target2_12_18xsource3
 [info] 	   expect2_12Target3_3_1
 [info] 	   expect2_13Target2_13_10
 [info] 	   expect2_13Target2_13_11
 [info] 	   expect2_13Target2_13_12
+[info] 	   expect2_13Target2_13_12xsource3
 [info] 	   expect2_13Target3_3_1
 [info] 	   expect3Target3_3_1
 [info] 	   input2_12_16
 [info] 	   input2_12_17
 [info] 	   input2_12_18
+[info] 	   input2_12_18xsource3
 [info] 	   input2_13_10
 [info] 	   input2_13_11
 [info] 	   input2_13_12
+[info] 	   input2_13_12xsource3
 [info] 	   input3_3_1
 [info] 	   integration2_12
 [info] 	   integration2_13
 [info] 	   integration3
 [info] 	   interfaces
-[info] 	   output2_12
+[info] 	   output2_12_18
+[info] 	   output2_12_18xsource3
-[info] 	   output2_13
+[info] 	   output2_13_12
+[info] 	   output2_13_12xsource3
-[info] 	   output3
+[info] 	   output3_3_1
 [info] 	   reflect2_12
 [info] 	   reflect2_13
 [info] 	   reflect3
 [info] 	   rules2_13
 [info] 	   rules3
 [info] 	 * scalafix
-[info] 	   shared2_12_16
-[info] 	   shared2_12_17
-[info] 	   shared2_12_18
[...]

PR #1896, opened 2023-11-23
OrganizeImports.targetDialect: wildcard/rename syntax & curly braces stripping

bjaglin:
Fixes https://github.com/scalacenter/scalafix/issues/1894

### Context

Before this PR, Scala2 syntax (`_` & `Foo => F`, surrounded by curly braces even for single-importee importers) was used, except for importers not semantically updated by an `OrganizeImports` run (allowing Scala3 syntax `*` & `Foo as F`, without surrounding braces for single-importee importers).

### Goal of this PR

This adds a new configuration flag which controls which dialect is used consistently, for all importers. Optional braces are stripped whenever possible.

### Impact 

That new flag defaults to `StandardLayout` to maximize compatibility. Note however that is NOT similar to the previous behavior, as it's more strict: Scala 3 syntax for wildcard and rename importees (`*` & `Foo as F`) is forced for `scala-3` sources.

Scala3-only builds (i.e not cross-compiling to Scala2 withtout `-Xsource;3`) should switch to ` Scala3`. For example, [this](https://github.com/bjaglin/dotty/pull/1/files) is the effect of `targetDialect = Scala3` vs scalafix 0.11.1 on the dotty codebase (see also [a note on `Ascii` ordering](https://github.com/bjaglin/dotty/pull/1/files#r1461240117)).

### Testing strategy

Leaving the default (`StandardLayout`) for most of the tests to reduce the amount of files, unless the test is exercising renames and/or multiple importees, in which case we use `Auto` in the version-agnostic input, and with version-specific outputs.


scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -776,48 +847,6 @@ object OrganizeImports {
...
-    importer.pos match {
-      case pos: Position.Range =>
-        // Position found, implies that `importer` was directly parsed from the source code. Returns
-        // the original parsed text to preserve the original source level formatting.
-        pos.text
-
-      case Position.None =>
-        // Position not found, implies that `importer` is derived from certain existing import
-        // statement(s). Pretty-prints it.
-        val syntax = importer.syntax
```
bjaglin, line 806: This was capturing the implicit `Dialect` always in scope (`Dialect.current`) which matches the runtime Scalafix Scala version. This one is not necessarily the Scala version of the sources (as Scalafix never runs in Scala 3 yet, and as we can have `-Xsource:3` in Scala 2.13).

Moving it outside the companion object allows to capture the freshly-added instance `dialect`, inferred from the `Configuration`.

scalafix-tests/output/src/main/scala-3/test/organizeImports/DeduplicateImportees.scala
```diff
@@ -0,0 +1,7 @@
+package test.organizeImports
+
+import scala.collection.immutable.Vector
+import scala.collection.immutable.{Map => Dict}
+import scala.collection.immutable.{Set as _, *}
```
bjaglin, line 5: The first line is [preserved](https://github.com/liancheng/scalafix-organize-imports/issues/127) while the second is generated by Scalafix, thus the inconsistency. We need to fix that - just not sure if it has to be in this PR?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -776,48 +848,6 @@ object OrganizeImports {
...
```
bjaglin, line 806: This was capturing the implicit `Dialect` always in scope (`Dialect.current`) which matches the runtime Scalafix Scala version. This one is not necessarily the Scala version of the sources (as Scalafix never runs in Scala 3 yet, and as we can have `-Xsource:3` in Scala 2.13).

Moving it outside the companion object allows to capture the freshly-added instance `dialect`, inferred from the `Configuration`.

scalafix-tests/output/src/main/scala-3/test/organizeImports/DeduplicateImportees.scala
bjaglin, line 5: The first line is [preserved](https://github.com/liancheng/scalafix-organize-imports/issues/127) while the second is generated by Scalafix, thus the inconsistency. We need to fix that - just not sure if it has to be in this PR as it does not look trivial.

bjaglin: Actually, preserving original syntax is also what is causing a few tests to fail as:
- the sorting is done on pretty-printed importers even though the original syntax remains
- `foo.{bar => qux}` is rewritten `foo.bar as qux` with the Scala 3 dialect

As a result, the order is changed for no apparent reason: a preserved `foo.{bar => qux}` no longer follows `import foo.wiz` if the source is Scala 3.

scalafix-tests/input/src/main/scala/test/organizeImports/ExplodeImportsFormatPreserving.scala
```diff
@@ -7,6 +7,6 @@ OrganizeImports.groupedImports = Explode
 package test.organizeImports
 
 import test.organizeImports.ExplodeImports.FormatPreserving.g1.{ a, b }
-import test.organizeImports.ExplodeImports.FormatPreserving.g2.{ c => C, _ }
+import test.organizeImports.ExplodeImports.FormatPreserving.g2.{ c  =>  C, _ }
```
bjaglin, line 10: add whitespaces on purpose to see how rewriting with `as` behaves when the patch does not have the same length as what's being replaced

scalafix-tests/input/src/main/scala-3/test/organizeImports/GroupedGivenImportsMergeUnimports.scala
```diff
@@ -5,14 +5,14 @@ OrganizeImports.groupedImports = Merge
...
+import test.organizeImports.GivenImports.{alpha as _, given}
 import test.organizeImports.GivenImports.{given Beta}
-import test.organizeImports.GivenImports.{gamma => _, given}
+import test.organizeImports.GivenImports.{gamma as _, given}
 import test.organizeImports.GivenImports.{given Zeta}
 
-import test.organizeImports.GivenImports2.{alpha => _}
-import test.organizeImports.GivenImports2.{beta => _}
+import test.organizeImports.GivenImports2.{alpha as _}
+import test.organizeImports.GivenImports2.beta as _
```
bjaglin, line 15: feeding both notations (braces are optional for a single rename or unimport)

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -717,8 +842,29 @@ object OrganizeImports {
       }
     }
 
+    // TODO: add an option to force a dialect with many tests
```
bjaglin, line 845: `importSyntax = Scala2 | Scala3 | Detect`

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -685,6 +694,122 @@ class OrganizeImports(config: OrganizeImportsConfig)
...
+          if (stripEnclosingBraces)
+            (
+              syntax.take(start).lastIndexOf('{'),
+              syntax.indexOf('}', end)
+            ) match {
+              case (from, to) if from != -1 && to != -1 =>
+                syntax.delete(end, to + 1)
+                syntax.delete(from, start)
+              case _ =>
+            }
```
bjaglin, line 741:
```diff
-import foo.{qux as bar}
+import foo.qux as bar
```

Still unsure if we want to force that as it goes against the format-preserving features.

If we don't strip, we end up having unexpected sorting in Scala 3 as we use pretty-printed brace-less syntax as sort key with the current implementation:
```
import foo.{qux as bar} // sorted as foo.qux
import foo.{bar, qux}
```

tgodzik: I think stripping {} here is fine. It's actually much easier to just do as, so I would recommend that

scalafix-tests/input/src/main/scala/test/organizeImports/CurlyBracedSingleImportee.scala
```diff
@@ -1,10 +1,12 @@
 /*
 rules = [OrganizeImports]
 OrganizeImports.removeUnused = false
+OrganizeImports.targetDialect = Auto
  */
 package test.organizeImports
 
 import scala.collection.{Map}
+import scala.collection.{Seq => _}
```
bjaglin, line 9: exercise unimport alone

docs/rules/OrganizeImports.md
```diff
@@ -1328,3 +1328,86 @@ object RemoveUnused {
...
+
+Use `*` as wildcard and `as` for renames.
+
+### Default value
+
+`Scala2`
+
+Rationale: `Auto` is not a safe default for projects cross-compiling with
+Scala 3.x and Scala 2.x without `-Xsource:3`, as the Scala 3.x Scalafix run
+would introduce Scala 2.x compilation errors.
```
bjaglin, line 1362: It's really a shame we can't default to `Auto` :disappointed: 

@tgodzik maybe metals has more information and could force `Auto` whenever we know for sure that sources won't be cross-compiled with Scala2? I am not convinced as we would get a lot of false negatives with the checks I have in mind
- source files rooted in a `scala-3` directory
- source files using other Scala 3 features (I am not sure we can pull this information for the tree - as far as I know the only way would be to try and fail to parse with a Scala 2 dialect...)

We could do something similar in sbt-scalafix (although this would be the first time we mingle with rule configuration) with fairly low risk when detecting that we both have
- no usage of sbt-projectmatrix
- no `crossScalaVersions` set

scalafix-tests/input/src/main/scala-2-xsource3/test/organizeImports/TargetDialectAuto.scala
```diff
@@ -0,0 +1,13 @@
+/*
+rules = [OrganizeImports]
+OrganizeImports {
+  removeUnused = false
+  targetDialect = Auto
```
bjaglin, line 5: testing `-Xsource:3` detection thanks to https://github.com/scalacenter/scalafix/pull/1666

scalafix-tests/input/src/main/scala-3/test/organizeImports/ExpandGiven.scala
```diff
@@ -1,12 +1,13 @@
...
 rules = [OrganizeImports]
 OrganizeImports.removeUnused = false
+OrganizeImports.targetDialect = Scala3
  */
 package test.organizeImports
 
 import test.organizeImports.GivenImports.Beta
 import test.organizeImports.GivenImports.Alpha
-import test.organizeImports.GivenImports.{given Beta, given Alpha}
+import test.organizeImports.GivenImports.{given Beta, given test.organizeImports.GivenImports.Gamma, given Alpha}
```
bjaglin, line 10: exercising a given with FQDN

bjaglin:
@tgodzik I am gonna merge this soon, ahead of the upcoming release for the new Scala2/3 releases. I updated the PR description and rule docs to try to spell out the changes. Let me know if you see any negative impact for Metals.

docs/rules/OrganizeImports.md
tgodzik: I don't think we have that information even in Metals especially that with crossScalaVersion only one version would be imported. 

The alternative to Auto would be to have StandardLayout which would do Scala2 target under `scala` and `scala-2`, Scala3 for `scala-3` directories.

Anyway, I think having Scala2 as default makes sense

bjaglin: It turned out `StandardLayout` wasn't so hard to implement so I added it and made it the default. Thanks :+1:

tgodzik: Awesome! Thanks!

PR #1897, opened 2023-11-24
Add description to the OrganizeImports rule

marcelocenerine:
The [Built-in Rules](https://scalacenter.github.io/scalafix/docs/rules/overview.html) section of the docs shows an empty description for the `OrganizeImports` rule.


bjaglin (approved):
Thanks!

PR #1924, opened 2024-02-05
pull stats for sbt-scalafix

bjaglin:


.github/scripts/sonatype-stats.scala
```diff
@@ -47,7 +47,8 @@ object Params {
   val artifacts = Set(
     "scalafix-core_2.12",
     "scalafix-core_2.13",
-    "scalafix-interfaces"
+    "scalafix-interfaces",
+    "sbt-scalafix_2.12_1.0"
```
bjaglin, line 51:
```suggestion
    "sbt-scalafix_2.12_1.0",
    "sbt-scalafix"
```

bjaglin:
it seems that we can only get stats on POM-consistent downloads, so it's not really representative at the moment


Screenshot from 2024-02-05 12-35-15

PR #1966, opened 2024-03-21
test JDK22

bjaglin:


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: coursier/setup-action@v1
         with:
-          jvm: temurin:${{ matrix.jvm }}
+          jvm: zulu:${{ matrix.jvm }}
```
bjaglin, line 23: JDK22 is only available for Zulu for now with coursier

PR #1968, opened 2024-03-31
docs: remove dead link for `cross-publish-custom-rules`

xuwei-k:
- https://github.com/scalacenter/scalafix/pull/1939
- https://github.com/scalacenter/scalafix/blob/112fd902227a9a90351af66761777997a9c793ba/docs/developers/local-rules.html#L229
- https://scalacenter.github.io/scalafix/docs/developers/local-rules.html

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -22,8 +22,7 @@
       "developers/setup",
       "developers/before-you-begin",
       "developers/tutorial",
-      "developers/local-rules",
-      "developers/cross-publish-custom-rules"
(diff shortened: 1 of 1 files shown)


bjaglin (approved):
This is a static reference unchecked at build time :( Thanks!

bjaglin:
FTR, there is what [seems to be a transient failure when publishing](https://github.com/scalacenter/scalafix/actions/runs/8497875822), so this update is not effective yet

PR #1952, opened 2024-03-03
run more expect tests against Scala 3

bjaglin:
Rationale: ensure good coverage & detect potential Scalameta regressions

PR #1976, opened 2024-04-17
move last escapeHatch test to be cross-tested

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/1952, now that https://github.com/scalameta/scalameta/issues/3689 is fixed


bjaglin:
following up in https://github.com/scalacenter/scalafix/pull/2011

PR #1979, opened 2024-04-21
fix UnusedScalafixSuppression warnings on Scala 3 scalafix invocations

bjaglin:


scalafix-core/src/main/scala/scalafix/v0/Signature.scala
```diff
@@ -50,10 +50,8 @@ object Signature {
   }
 
   final case class Self(name: String) extends Signature {
-    override def syntax =
-      throw new UnsupportedOperationException(
-        "No longer supported."
-      ) // scalafix:ok
+    override def syntax: String =
```
bjaglin, line 53:
```
[error] Module scalafix-core:0.12.0+48-ad1d9456-SNAPSHOT is not binary compatible with scalafix-core_2.12:0.12.0.
[error] You have to relax our compatibility intention by changing the value of versionPolicyIntention, or to fix the incompatibilities.
[error] We found the following incompatibilities:
[error]    * method syntax()scala.runtime.Nothing# in class scalafix.v0.Signature#Self does not have a correspondent in current version
[error]      filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scalafix.v0.Signature#Self.syntax")
```

bjaglin: for some reason, the annotation was removed in https://github.com/scalacenter/scalafix/commit/96e3c607c499f9179e2a350e2bf12676548f81a2#diff-8bcb1fea5512e923315adba5f40c197a178dda1697225f8c955bce1db2f0a2f4

bjaglin:
```
[error] Module scalafix-core:0.12.0+48-ad1d9456-SNAPSHOT is not binary compatible with scalafix-core_2.12:0.12.0.
[error] You have to relax our compatibility intention by changing the value of versionPolicyIntention, or to fix the incompatibilities.
[error] We found the following incompatibilities:
[error]    * method syntax()scala.runtime.Nothing# in class scalafix.v0.Signature#Self does not have a correspondent in current version
[error]      filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scalafix.v0.Signature#Self.syntax")
```

it's an insignificant breaking change, moreover on an old API version, so it's fine to ignore it

Issue #2000, opened 2024-05-16
Support for non-zero exit codes from CLI when there are warnings


mrdziuban:
Prompted by [my question in Discord](https://discord.com/channels/632642981228314653/632683673417678860/1238510377541697568), I wanted to open this to discuss potential approaches.

I see that there's a [`fatalWarnings` field on the `ScalafixConfig` type](https://github.com/scalacenter/scalafix/blob/main/scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala#L14) that defaults to `true`. Am I misunderstanding, or should that mean that warnings should be treated as errors?

bjaglin:
Hi @mrdziuban, thanks for reporting back here and sorry for the delay.

That `fatalWarnings` flag seems unused and is not exposed as a CLI arg. Judging by the git history, it controlled the reporting behavior when rules had issues introspecting the source files, but not the linter diagnostics.

I'll draft a PR to propose a new CLI or configuration flag.

bjaglin:
@mrdziuban any feedback on that?

mrdziuban:
Sorry for the delay @bjaglin! I was able to get this working for my use case, which is running all syntactic rules during a pre-commit hook.

The one catch I still had to account for was making sure that any `UnusedScalafixSuppression` errors don't cause a non-zero exit code -- there might be unused suppressions for semantic rules that weren't run.

I was able to use `--settings.lint.error` like this to get it working:

```
--settings.lint.error '^((?!UnusedScalafixSuppression).)*$'
```

Ideally I could pass a flag to disable the `UnusedScalafixSuppression` checks completely (#1271 possibly related) but this works for now. Thanks for the help!

bjaglin:
Thanks for the feedback. I documented our respective findings (as well as the `.scalafix.conf` / CLI args duality) through https://github.com/scalacenter/scalafix/pull/2062 . It's live at https://scalacenter.github.io/scalafix/docs/users/configuration.html#overriding-the-linting-behavior and below. Feel free to open a PR if that's not clear or if it does not work for you!

> Ideally I could pass a flag to disable the UnusedScalafixSuppression checks completely (https://github.com/scalacenter/scalafix/issues/1271 possibly related) but this works for now

Yes let's follow up there if ever needed. The `includes/excludes` syntax it's a bit verbose, but at least I find it clear about the intent, so I am not sure we need more, especially now that it's documented.

PR #2012, opened 2024-06-26
remove unused attributes

bjaglin:
https://github.com/scalacenter/scalafix/issues/2000#issuecomment-2191815501

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
@@ -4,14 +4,10 @@ import scala.meta._
 
 import metaconfig._
 import metaconfig.generic.Surface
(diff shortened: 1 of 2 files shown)


scalafix-core/src/main/scala/scalafix/internal/config/ScalafixConfig.scala
```diff
@@ -4,14 +4,10 @@ import scala.meta._
 
 import metaconfig._
 import metaconfig.generic.Surface
-import scalafix.Versions
 import scalafix.internal.config.ScalafixConfig._
 
 case class ScalafixConfig(
-    version: String = Versions.version,
```
bjaglin, line 11: this was added in https://github.com/olafurpg/scalafix/commit/3a73ba04378f2de7bc850aa4d6be9e43169c5e14 but I don't understand why

PR #2022, opened 2024-07-24
refactor: Remove Unsafe at it's not accessible on all JDKs

tgodzik:


bjaglin (approved):
Did you run into troubles compiling with a specific JDK? Is this about https://openjdk.org/jeps/471? I guess the real problem is the runtime, so I am curious about it... I did look into that [a while ago](https://github.com/scalacenter/scalafix/pull/1780), but I had concluded it was not trivial to find a replacement for fetching system class loader URLs...

tgodzik:
> Did you run into troubles compiling with a specific JDK? Is this about https://openjdk.org/jeps/471? I guess the real problem is the runtime, so I am curious about it... I did look into that [a while ago](https://github.com/scalacenter/scalafix/pull/1780), but I had concluded it was not trivial to find a replacement for fetching system class loader URLs...

The problem was that it did not compile if sbt or Bloop was run on a newer version of the JDK. The release flag didn't help at all here. This way it should compile on all possible JDKs

tgodzik:
> So I guess the classpath-related tests don't pass on these versions either, and users will have a similar issue? Or is that because the build tool is putting stricter constraints on usage of deprecated classes?

Wait, actually, this doesn't compile in Bloop only. Seems sbt is doing some clever things and adding rt.jar for the current JDK under ~/.sbt/0.13/java9-rt-ext-eclipse_adoptium_17_0_9/rt.jar This might be missing in Bloop only then.

PR #1926, opened 2024-02-06
Scala3: test against both LTS & Next, build with LTS

bjaglin:
Rationale
- test against LTS & Next -> test the 2 Scala3 versions we can expect in the wild
- build with LTS -> prepare for https://github.com/scalacenter/scalafix/issues/1680

TODO
- [ ] port something similar to https://github.com/scalacenter/scalafix.g8

```diff
 sbt:scalafix> projects
 [info] In file:/home/brice/git/scala/scalafix/
 [info] 	   cli2_12
 [info] 	   cli2_13
 [info] 	   cli3
 [info] 	   core2_12
 [info] 	   core2_13
 [info] 	   core3
 [info] 	   docs2_13
 [info] 	   expect2_12Target2_12_18
 [info] 	   expect2_12Target2_12_19
 [info] 	   expect2_12Target2_12_19xsource3
 [info] 	   expect2_13Target2_13_12
 [info] 	   expect2_13Target2_13_13
 [info] 	   expect2_13Target2_13_14
 [info] 	   expect2_13Target2_13_14xsource3
+[info] 	   expect2_13Target3_3_4-RC1
 [info] 	   expect2_13Target3_4_2
-[info] 	   expect3Target3_4_2
+[info] 	   expect3Target3_3_4-RC1
 [info] 	   input2_12_18
 [info] 	   input2_12_19
 [info] 	   input2_12_19xsource3
 [info] 	   input2_13_12
 [info] 	   input2_13_13
 [info] 	   input2_13_14
 [info] 	   input2_13_14xsource3
+[info] 	   input3_3_4_RC1
 [info] 	   input3_4_2
 [info] 	   integration2_12
 [info] 	   integration2_13
 [info] 	   integration3
 [info] 	   interfaces
 [info] 	   output2_12_19
 [info] 	   output2_12_19xsource3
 [info] 	   output2_13_14
 [info] 	   output2_13_14xsource3
-[info] 	   output3_4_2
+[info] 	   output3_3_4_RC1
 [info] 	   reflect2_12
 [info] 	   reflect2_13
 [info] 	   reflect3
 [info] 	   rules2_12
 [info] 	   rules2_13
 [info] 	   rules3
 [info] 	 * scalafix
 [info] 	   shared2_12
 [info] 	   shared2_13
 [info] 	   shared3
 [info] 	   testkit2_12
 [info] 	   testkit2_13
 [info] 	   testkit3
 [info] 	   unit2_12
 [info] 	   unit2_13
 [info] 	   unit3
```

Issue #1680, opened 2022-10-01
Publish cli_3 & rules_3, built against core_2.13


mlachkar:
Follows https://github.com/rvacaru/GsoC-2022-Scalafix

- Testing
  - [x] [reenable testing scala latest with Scala 3 build](https://github.com/scalacenter/scalafix/pull/1926/files#r1479349441) 
  - [x] many of tests in ScalafixArgumentsSuite rely on semanticRule, and we don't have any yet for scala 3. We need to create a simple semanticrule and test the ignored tests for scala 3 (search for "i1680" in code)
    - https://github.com/scalacenter/scalafix/pull/1756
  - [x] update [integration tests](https://github.com/scalacenter/scalafix/blob/dc2b126a35d24f427742880d60d2a6ae8be74f78/scalafix-tests/integration/src/test/scala/scalafix/tests/interfaces/ScalafixSuite.scala#L108)
- ~core_3~
  - [x] ~wait for https://github.com/scalameta/scalameta/pull/3347 acceptable to start without support of quasiquotes and getting 2.13 artifacts in the classpath?~ we'll proceed with cli_3/reflect_3/rules_3 builtcore_2.13 to unblock ExplicitResultTypes, and follow-up with core_3 in https://github.com/scalacenter/scalafix/issues/2041
- ~reflect_3~
  - [x] ~add warning that quasiquotes are not supported and that compiler behavior might be different, proposing a fallback to 2.13 artefacts (or throw an exception that cli understands, which would reboot with 2.13 automatically?)~ let's stick to reflect_2.13 instead, and follow-up in https://github.com/scalacenter/scalafix/issues/2041
- rules_3
  - [x] make sure all rules are available (even the one that is irrelevant for Scala 3) to maximise compatibility of cli_3
  - [x] start testing against 2_13 sources to maximise compatibility of cli_3
- cli_3
  - [x]  ~do we _really_ need full-cross-version in Scala 3? Consider https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html~ we do for ExplicitResultTypes to be efficient for LTS & Next, see https://github.com/scalacenter/scalafix/pull/2023#discussion_r1711317793
  - [x] build with Scala Next on top of Scala LTS, and add corresponding expect (with minor suffix) - recap with a https://github.com/dwijnand/sbt-project-graph augmented with non-standard dependencies and persist it in CONTRIBUTING.md
[...]

PR #2023, opened 2024-07-25
feature: ExplicitResultTypes for Scala 3

tgodzik:
Approach: use the [stable Scala 3 presentation compiler](https://github.com/scala/scala3/pull/17528) available since 3.3.2

There is still plenty to do, and I am not 100% sure about the approach, but it's a start. Tests are passing but some outputs are different the Scala 2 ones. I also found some issues in some of the tests, I removed them until I can fix them in the compiler.

TODO:
- [x] extract common parts for both Scala 2 and 3 to a trait/abstract class
- [x] ~fix issues in the compiler~ (non-blocking?)
- [x] download specific compiler for a specific version instead of explicitly setting dependency
- [x] ~separate tests for LTS and Next~ with both on the latest it seem ok, but we might need some compat at some point
- [x] make sure that each patch is sequentially one after another (currently all are done one the same code, which might cause issues)


bjaglin:
Exciting! I will have a look at it some time this week.

scalafix-rules/src/main/scala-3/scalafix/internal/rule/Embedded.scala
```diff
@@ -0,0 +1,94 @@
...
+      classloader
+    )
+  }
+
+  private def supportPresentationCompilerInDotty(scalaVersion: String) = {
+    scalaVersion.split("\\.").take(3).map(_.toInt) match {
+      case Array(3, minor, patch) => minor > 3 || minor == 3 && patch >= 4
+      case _ => false
+    }
+  }
```
bjaglin, line 55: Since we control the runtime of scalafix, and we'll first [publish](https://github.com/scalacenter/scalafix/issues/1680) scala3 artifacts with [3.3.4 or later](https://github.com/scalacenter/scalafix/pull/1926), this should always be true. If that's correct, I assume we don't need coursier and we can declare the dependency statically in scalafix-rules_3?

bjaglin (commented):
That looks promising, thanks for putting this together!

In terms of workflow, we could merge this quite early and mutualize code / enable more tests as we go, since nothing is published for Scala 3 anyway.

I will try to dust off https://github.com/scalacenter/scalafix/issues/1680 in mid/late August since that remains a prerequisite. I had put it on hold as I was watching https://github.com/scalameta/scalameta/pull/3347 (do you know if you or anyone else will allocate time to it by the way?), but I guess we could release experimental Scala 3 versions that would only allow rules written without quasiquotes, with the 2.13/3 sandwitch.

tgodzik:
Ok, I am getting closer to a finish. I left some TODOs I want to still address, but there is nothing major. I've seen some test failure on one of JDK versions earlier so I will also check it out (I think we need to use JDK 11 at a minimum for the Scala 3 tests, so might need to disable the tests there)

The biggest change I did is extracting most of the methods to a common base trait, but also enabled to use presentation compiler as a fallback for Scala 2. I noticed why we might be getting some failures in the tests (those removed for now), so I will also take a look.

> I had put it on hold as I was watching https://github.com/scalameta/scalameta/pull/3347 (do you know if you or anyone else will allocate time to it by the way?), but I guess we could release experimental Scala 3 versions that would only allow rules written without quasiquotes, with the 2.13/3 sandwitch.

I need to get back to it, I want to get something ready here and will be back to scalameta.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
```diff
@@ -0,0 +1,106 @@
...
+    val deps =
+      scala3PresentationCompilerDependencies(scalaVersion)
+    val jars = Fetch
+      .create()
+      .addDependencies(deps: _*)
+      .addRepositories(
+        MavenRepository.of(
+          "https://oss.sonatype.org/content/repositories/snapshots"
+        )
+      )
```
bjaglin, line 29: This might not be enough in corporate/CI environments. Custom repositories provided at Scalafix invocation are not available to rules, so I am not sure how to honor them without changing the API.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
```diff
@@ -0,0 +1,106 @@
...
+      .map(_.toInt) match {
+      case Array(3, minor, patch) => minor > 3 || minor == 3 && patch >= 4
+      case _ => false
+    }
+  }
+
+  private def scala3PresentationCompilerDependencies(version: String) =
+    if (supportPresentationCompilerInDotty(version)) {
+      val dep = Dependency
+        .of("org.scala-lang", "scala3-presentation-compiler_3", version)
```
bjaglin, line 64: As a follow up to my comments about repositories, I am trying to understand what forces us to have that dynamic coursier resolution which brings a lot of complexity. I guess we need a PC matching as much as possible the compiler used by the client? Would it be an option to cross-build scalafix_rules to each minor version of Scala 3 to have build-time dependencies instead?

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import scala.meta.*
+
+import metaconfig.Configured
+import scalafix.internal.pc.ExplicitResultTypesFallback
+import scalafix.patch.Patch
+import scalafix.v1.*
+
+
+final class ExplicitResultTypes(
```
bjaglin, line 11: As I understand it from a very quick review as I am time-constrained but I see you progress very well, the 2.13 version is able to support 3 sources via the dynamic resolution. So we might not need to publish scalafix-rules_3 at all if we stick to that strategy (pending my other comment)?

bjaglin: I guess that's what you meant in your previous comment

> Also I commented out using 2.13 compiled rules for Scala 3, but I can make it work with coursier.

bjaglin: That would allow to push this faster as https://github.com/scalacenter/scalafix/issues/1680 wouldn't be blocking.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: Setting COURSIER_REPOSITORIES globally should help, but it's always a bit of a problem. The other options would be to somehow reuse sbt resolution? And just bring in the exact presentation compiler version via that route?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: > I guess we need a PC matching as much as possible the compiler used by the client?

Ideally yes, the compiler API itself is not binary compatible, so we need to have the same version of PC and the compiler and if possible also the same Scala version used by the user. For Scala 2 we are kind of cheating, since we should be doing the same thing. For example some bug could be fixed in a later version and if someone has an earlier one the rule might break things for them (granted I haven't seen too many issues like this)

> Would it be an option to cross-build scalafix_rules to each minor version of Scala 3 to have build-time dependencies instead

It would be possible for sure, but might be a lot of effort and we would need to release every time a new release is out, which this would avoid (we can still keep dynamic resolution as fallback in that case). We could also release rules for each minor, which would be a bit less work for sure.
Possibly we could

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
tgodzik: >  So we might not need to publish scalafix-rules_3 at all if we stick to that strategy (pending my other comment)?

Yep, it wouldn't be needed, but dynamic resolution in this case is a must.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: That might be possible, and it would solve it for sbt-scalafix, but not for other clients. Anyway, it's not a blocker for this PR.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: > We could also release rules for each minor

My assumption here is that the 3.x.0 PC is able to handle what the 3.x.y compiler can. Is that the case?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: I guess we could just depend on the presentation compiler and override compiler version in sbt (metals should be able to do that also I think). For all other clients we could just publish LTS/Next which in most cases should be enough.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: I think you're right. Maybe I am being too cautious? We can start with releasing for LTS and Next.

bjaglin: Using LTS for 3.4 can be problematic since new source-level features won't be handled. But maybe we can build scalafix-rules_3 with latest (and scalafix-core_3 with LTS for maximum compatibility for external rule authors), and always use that? That would be very similar to what we do for 2.12 and 2.13 actually.

tgodzik: ok, let's do that. So scalafix-rules_3 to be build with 3.3.x, 3.4.x and 3.5.x and core just for 3.3.x, right?

bjaglin: I would say: build scalafix-rules_3 with NEXT (to statically bring the most powerful/up-to-date PC, that should be able to handle bytecode/TASTY compiled by a prior version), and scalafix-core with LTS (for rule authors to be flexible).

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: I would favor static resolution over ability to run ExplicitTypes with scalafix-rules_2.13. I can dedicate my full Tuesday next week to https://github.com/scalacenter/scalafix/issues/1680.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: We can think of a more dynamic solution later (and how to honour custom repositories), which would allow scalafix to handle future/upcoming Scala 3 minor versions, but it has been this way for Scala 2.x so users won't find it weird to update as Scala versions come out.

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
tgodzik: Should I wait with the changes here until your PR is finished then?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: The only thing I am wondering if the type could change in between major version (I think it can) and in the case using latest PC under Next could break your code which might be under LTS

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
bjaglin: I think both PR are orthogonal, it's just that your implementation will only be available to users after my upcoming PR. So you can iterate now on this one if you have time, and we can merge it as soon as it's ready independently of mine.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: Oh I see, the latest Next PC would infer a type that LTS would not support 😞  I am not sure how common that use case is, and whether it justifies the complexity at build time and also at resolution time, since we'd have to suffix artefacts ourselves as I don't think coursier (the one already existing in scalafix-interfaces) supports "3.x" lookups?

tgodzik: At a minimum we need to publish for 3.3, 3.4, 3.5 etc. since we can get changes that even remove certain syntax, so PC would no longer be able to work with that. Sorry for the back and forth but I am doing a lot of pondering here.

My opinion about final implementation:
- core published with LTS
- rules published for each minor to be on the safe side -> not sure how to solve this build wise, maybe full compat needed
- leave fallback for anything that we don't know about -> that should be safe enough as the alternative for people is that it will not work.

tgodzik: >  since we'd have to suffix artefacts ourselves as I don't think coursier (the one already existing in scalafix-interfaces) supports "3.x" lookups?

We would need to do a full binary suffix and release one for each minor. So now we would have `rules_3.3.4` for 3.3 line, `rules_3.4.2` for 3.4 line and `rules_3.5.0` for 3.5 line.

It might be easier at this point to deal with coursier resolution :sweat_smile:

tgodzik:
Another fun observation is that running tests with static dependency on the compiler seems much faster than the dynamic one. I am not sure what it's caused by, I was pretty sure that separate classloaders are almost zero cost.

tgodzik:
> Another fun observation is that running tests with static dependency on the compiler seems much faster than the dynamic one. I am not sure what it's caused by, I was pretty sure that separate classloaders are almost zero cost.

Ach, I know what is happening, we are loading a separate classloader each time for each new file :fearful:

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: scalafix-cli (whic brings in scalafix-rules) is fully cross-built already, so that wouldn't be a problem at least.

What I am concerned about is that [ScalafixInterface  only asks for a binary version](https://github.com/scalacenter/scalafix/blob/5a02480f5ee126a3804ac3bbd6e3d13cb864b02a/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L105) to know which scalafix-cli to load, so it looks like we would have to update all clients (which rely on scalafix-interfaces only, example [sbt-scalafix](https://github.com/scalacenter/sbt-scalafix/blob/9bf8b8b1b21ad098ef75995622b20f263dd050fb/src/main/scala/scalafix/sbt/ScalafixPlugin.scala#L372-L374)) to pass more information than just "3".

I will try to reflect on the tradeoffs...

bjaglin:
> Ach, I know what is happening, we are loading a separate classloader each time for each new file 😨

Right, we'd have to amortise the classloading cost across files, and ideally across invocations, to keep CPU/IO/RAM at a reasonable level when `scalafixOnCompile := true`. And shared/global state outside a build tool is scary 🙈.

tgodzik:
Ok, switched to using presentation compiler statically if the minor version of Scala 3 is the same and for dynamic loading I added caching which should fix things.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: Couldn't we basically require 3.3 etc ? That wouldn't be a super huge change for the clients (just throw an exception let them know) and we can translate it to the known artifacts at the release of the interface. If the interface doesn't know that version, it just loads LTS, which resolves things dynamically as a last resort?

bjaglin: Right, it wouldn't be a breaking change anyway since the full version is not needed for 2.x, and that's the only thing supported by existing clients. Adding this to https://github.com/scalacenter/scalafix/issues/1680.

build.sbt
```diff
@@ -85,6 +85,14 @@ lazy val rules = projectMatrix
     moduleName := "scalafix-rules",
```
bjaglin, line 85: As discussed, we should add scala Next to the cross-build versions, and probably add a expect3_5Target3_5_x. This will require some tweaks in the build system to include the minor version version in `expect`, so I can handle that in another PR, within https://github.com/scalacenter/scalafix/issues/1680.

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -0,0 +1,84 @@
...
+          Option {
+            if (
+              toBinaryVersion(config.scalaVersion) == toBinaryVersion(
+                compilerScalaVersion
+              )
+            )
+              PcExplicitResultTypes
+                .static(config, new ScalaPresentationCompiler())
+            else
+              PcExplicitResultTypes.dynamic(config)
```
bjaglin, line 54: I think we could condition that to a configuration flag that makes the user aware that it's sub-optimal. If that flag is not set and we come here, we can provide an actionnable error message.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
```diff
@@ -0,0 +1,112 @@
...
+    val classloader = presentationCompilers.getOrElseUpdate(
+      scalaVersion,
+      newPresentationCompilerClassLoader(scalaVersion)
+    )
+    val presentationCompilerClassname =
+      if (supportPresentationCompilerInDotty(scalaVersion)) {
+        "dotty.tools.pc.ScalaPresentationCompiler"
+      } else {
+        "scala.meta.pc.ScalaPresentationCompiler"
+      }
```
bjaglin, line 32: This is needed only if running against Scala <3.4.0 since that's when mtags when upstreamed with a different name if I understand well. I would rather fail and ask for users to upgrade to LTS - I believe it's a fair thing to ask - or use cli_2x if sources are Scala 2. That, together with the inability to target Scala 3 sources from rules_2 discussed above would remove the need for mtags all together right?

scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -0,0 +1,84 @@
...
+      implicit val printer = new Scala3Printer(fallback)
+      unsafeFix()
+    } catch {
+      case _: Throwable if !config.fatalWarnings =>
+        Patch.empty
+    }
+
+}
+
+class Scala3Printer(
```
bjaglin, line 71: nit: as it looks now, this is also handling Scala2 right?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
```diff
@@ -0,0 +1,112 @@
...
+      val dep = Dependency
+        .of("org.scala-lang", "scala3-presentation-compiler_3", version)
+
+      // some versions of the presentation compiler depend on versions only build for JDK 11
+      dep.addExclusion("org.eclipse.lsp4j", "org.eclipse.lsp4j")
+      dep.addExclusion("org.eclipse.lsp4j", "org.eclipse.lsp4j.jsonrpc")
+      // last built with JDK 8
+      val lsp4jDep =
+        Dependency.of("org.eclipse.lsp4j", "org.eclipse.lsp4j", "0.20.1")
+      List(dep, lsp4jDep)
```
bjaglin, line 63: In build.sbt, these exclusions are applied to mtags and mtags only. Maybe we don't test properly? 🤔

bjaglin (commented):
Thanks for all the work!

I had a look at infrastructure rather than actual code of the rule, since we can iterate on that easily after.

I'd like to sort out 2 things (mentioned in comments) before we merge:
- whether we want mtags or not (I think we can do without it)
- move input/output files to `scala` instead of copying to `scala-3` when possible

scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -69,184 +74,64 @@ final class ExplicitResultTypes(
     if (
       config.scalacClasspath.nonEmpty && inputBinaryScalaVersion != runtimeBinaryScalaVersion
     ) {
-      Configured.error(
-        s"The ExplicitResultTypes rule needs to run with the same Scala binary version as the one used to compile target sources ($inputBinaryScalaVersion). " +
-          s"To fix this problem, either remove ExplicitResultTypes from .scalafix.conf or make sure Scalafix is loaded with $inputBinaryScalaVersion."
-      )
```
bjaglin, line 75: If not for scala 3. we still need that to prevent trying to load Scala 2.12 against 2.13 or the other way around, right?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: Not sure if I understand, I added those exclusions there as well. It was failing previously with those

bjaglin: Here we exclude the lsp4j that the compiler brings transitively, while in build.sbt, we exclude the lsp4j that mtags-interfaces brings transitively. I did not check if lsp4j was actually a dependency from one or another (or both?), but I just find this suspicious to have different logic for static vs dynamic resolution.

Maybe coursier does the exclusion for all artifacts involved for a single resolution so it's equivalent? The API just seems to show that exclusion is selective.

scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
```diff
@@ -3,23 +3,28 @@ package scalafix.internal.rule
...
 
 final class ExplicitResultTypes(
-    config: ExplicitResultTypesConfig,
-    global: LazyValue[Option[ScalafixGlobal]]
-) extends SemanticRule("ExplicitResultTypes") {
-
-  def this() = this(ExplicitResultTypesConfig.default, LazyValue.now(None))
+    val config: ExplicitResultTypesConfig,
+    global: LazyValue[Option[ScalafixGlobal]],
+    fallback: LazyValue[Option[PcExplicitResultTypes]]
```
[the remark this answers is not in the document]
tgodzik: Honestly, I would keep the dynamic resolution, but for sure we can limit the places where it's used to maybe just a new minor we didn't yet support?

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
tgodzik: I can remove it, but I don't think removing the dynamic fallback is something we should do. I will add a config option

tgodzik:
I will probably be back to this next week, sorry for the delay!

bjaglin:
> I will probably be back to this next week, sorry for the delay!

No worry, it gave me some time to finalize my PR about publishing cli_3.x.y https://github.com/scalacenter/scalafix/pull/2034, in light of all discussions here. I am gonna merge it very soon since it's already way too big.

Commits should be self-describing, but here are the highlights to help you rebase (I can do it if you prefer?)
- No core_3, let's stick to core_2.13 until scalameta_3 is out
  - this allows to classload source & published 2.13 community rules from cli_3.x.y, providing a completely transparent upgrade from cli_2.x.y to cli_3.x.y when Scala 3 sources are targeted, in the next patch release of scalafix/sbt-scalafix
  - I hacked something to fix metaconfig macros in cli_3.x.y (and removed cosmetic usage of sourcecode internally as it was causing even more headaches) - it's not pretty, but I think we can live with that as artifacts that use it are not meant to be used directly and binary compatibility will be broken anyway as soon as scalameta_3 is out
- rules, cli & tests are built with both LTS & Next
   - their sbt project name is now suffixed with the full version (just like their full cross-version publishing strategy, so it's a good move anyway)
   - scalafix-interfaces will load cli_LTS if 3.3.y is requested (clients need to be updated to (1) stop falling back to 2.z and (2) provide the full version information) or cli_Next otherwise
- since cli_Next will become the default for `cs install scalafix`, I made sure all rules were available/tested against Scala 2 sources via `expect3_5_0Target2_13_14` (at the moment, there is not reason that it shouldn't work, but it's more of a precaution for any upcoming changes)
  - I added a dirty hack to ignore ExplicitResultTypes expect tests when running in cli_3.x.y - we should be able to remove it in this PR (using the flag controlling usage of dynamic PC loading via coursier)
  - there is no reason to actively test cli_2.x.y against Scala 3 sources any longer since upcoming versions of sbt-scalafix & other clients should load cli_3.x.y whenever Scala 3 sources are targeted
  - with this PR, the service loader catalog `scalafix.v1.Rule` should now be the same for all Scala versions so no need to have separate instances

bjaglin:
@tgodzik let me know if you don't have bandwidth to handle it this week, I can try to find time for this during the weekend, ahead of the upcoming Scalafix release next week (for the new Scala 2.x and 3 LTS releases).

tgodzik:
I was thinking of finishing https://github.com/scalameta/scalameta/pull/3939 first, so probably I would get back to it later. If you have time and the will you are welcome to try and finish this PR :+1:

bjaglin:
Ok, ni problem, I'll invest (some) time by Monday (most likely tonight and Saturday), we'll see how far I can get!

> I was thinking of finishing https://github.com/scalameta/scalameta/pull/3939 first

This would mean scalameta would be published for Scala 3 but quasiquotes wouldn't be available right? I think for scalafix it wouldn't change anything (switching to these new artifacts would make backward compatibility with community rules in 2.13 harder actually). I'd like to provide ExplicitResultTypes first, and then look at https://github.com/scalacenter/scalafix/issues/2041 in the coming week/months since it's much more tricky to have a smooth transition for rule authors.

tgodzik:
> This would mean scalameta would be published for Scala 3 but quasiquotes wouldn't be available right? 

We have quasiquotes partly ready, so I would want to unblock that

bjaglin:
I'll try to address outstanding comments and merge this PR this week-end since I have some bandwidth

bjaglin:
> I'll try to address outstanding comments and merge this PR this week-end since I have some bandwidth

Well, it clearly didn't happen (life happened), but I have a 15h train ride this Friday 27th, so I will take care of this before cutting a release.

tgodzik:
> > I'll try to address outstanding comments and merge this PR this week-end since I have some bandwidth
> 
> Well, it clearly didn't happen (life happened), but I have a 15h train ride this Friday 27th, so I will take care of this before cutting a release.

No worries. I also planned to get back to it, but dealing with other bugs proved to be more urgent :/

bjaglin:
2c9e4f9 -> 1153521
- spinned off https://github.com/scalacenter/scalafix/pull/2084
- squashed
- consolidated as much as I could into `scala/test/explicitResultTypes` to make it easier to see what's missing/different
- rebased against 7f471d0
 
I'll address comments as separate commits this afternoon. I am targeting to merge this tonight and cut a release now that 2.12.20, 2.13.15, and 3.3.4 are out.

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: I am removing mtags and direct references to mtags-interfaces for now

scalafix-rules/src/main/scala/scalafix/internal/pc/Embedded.scala
bjaglin: I am removing mtags and direct references to mtags-interfaces for now, keeping dynamic loading of the Scala 3 PC behind a flag

scalafix-rules/src/main/scala-2/scalafix/internal/rule/ExplicitResultTypes.scala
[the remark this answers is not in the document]
bjaglin: I am removing support for targeting Scala 3 sources from cli_2 for now, since it's not a scenario clients will encounter

project/ScalafixBuild.scala
```diff
@@ -58,7 +58,10 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
           val xsource3 = TargetAxis(sv, xsource3 = true)
 
           (prevVersions :+ xsource3).map((sv, _))
-        } :+ (scala3Next, TargetAxis(scala213))
+        } ++ Seq(
+          (scala3Next, TargetAxis(scala213)),
+          (scala3Next, TargetAxis(scala3LTS))
```
bjaglin, line 63: Next vs LTS should not happen in practice because [interface should load the right version](https://github.com/scalacenter/scalafix/blob/590db62f1984848a2765a834865cc96b3f8c83ed/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java#L128-L134), but this is an opportunity to test dynamic loading

bjaglin: The problem is that this means enabling the flag for all cases, including the ones where it's not needed. So effectively not testing the regular path. I'll test dynamic loading with an adhoc, non-expect test I guess.

tgodzik:
Thanks @bjaglin for your work on this! I can do a follow ups with any issues that show up. Or especially getting all the tests working the same or similar enough like in Scala 2

PR #2048, opened 2024-08-24
core: reduce amount of scalac dependencies

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/2047 / https://github.com/scalameta/scalameta/pull/3747

See initial discussion in https://github.com/scalacenter/scalafix/pull/2024#discussion_r1703407620

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,7 @@ lazy val core = projectMatrix
       googleDiff,
       metaconfig,
       scalametaFor3Use2_13,
-      semanticdbScalacCore,
+      semanticdbSharedFor3Use2_13,
       collectionCompat
     )
   )
@@ -95,8 +95,7 @@ lazy val core3 = project
       metaconfig
     ) ++ Seq(
       scalametaFor3Use2_13,
-      // CrossVersion.for3Use2_13 would only lookup a binary version artifact, but this is published with full version
-      semanticdbScalacCore.cross(CrossVersion.constant(scala213))
+      semanticdbSharedFor3Use2_13
     ).map { mod =>
       mod
         .exclude("com.lihaoyi", "sourcecode_2.13")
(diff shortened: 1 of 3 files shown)


scalafix-core/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala
```diff
@@ -0,0 +1,10 @@
+package scala.meta.internal.symtab
```
bjaglin, line 1: copy of https://github.com/scalameta/scalameta/blob/v4.9.9/semanticdb/scalac/library/src/main/scala/scala/meta/internal/symtab/SymbolTable.scala

It's not great as drift between this version and the upstream one may cause runtime errors in reflect and cli, but I think it's worth it to keep core clean until we contractualize with scalameta what could be exposed as a public API.

project/Dependencies.scala
```diff
@@ -41,6 +41,7 @@ object Dependencies {
   val scalatest = "org.scalatest" %% "scalatest" % scalatestV
   val munit = "org.scalameta" %% "munit" % munitV
   val semanticdbScalacCore = "org.scalameta" % "semanticdb-scalac-core" % scalametaV cross CrossVersion.full
+  val semanticdbSharedFor3Use2_13 = "org.scalameta" % "semanticdb-shared" % scalametaV cross CrossVersion.for3Use2_13
```
bjaglin, line 44: https://github.com/scalameta/scalameta/pull/3887

build.sbt
```diff
@@ -149,6 +148,7 @@ lazy val reflect = projectMatrix
     moduleName := "scalafix-reflect",
     isFullCrossVersion,
     libraryDependencies ++= Seq(
+      semanticdbScalacCore,
```
bjaglin, line 151: I wonder how much of the current implementation of `GlobalSymbolTable` works against class files produced by dotty.... We haven't got any issue but I guess it's just because the Symbol lookup API is not used much on  symbols that are not in the document.

https://github.com/scalacenter/scalafix/blob/c6d741174abf76c01862a2aac1b52379b106c5d9/scalafix-core/src/main/scala/scalafix/internal/v1/InternalSemanticDoc.scala#L80

bjaglin: https://github.com/scalacenter/scalafix/issues/2049

PR #2058, opened 2024-08-26
Clarify class name vs rule name in ServiceLoader section

mjhoy:
([discord context](https://discord.com/channels/632642981228314653/632683673417678860/1275968994830188644)) I ran into this while developing my own scalafix rule, foolishly calling my class something different than the string I passed to `SemanticRule`, and was confused when the reference-by-name didn't seem to work (when I was actually passing it the class name).

Took a stab at writing something that might have helped me, not sure if it's actually that clear though.


bjaglin:
Thanks for taking the time to open PR!

Since setup.md doesn't talk much about _running_ rules, I am wondering if your addition wouldn't be better if it was integrated to the current docs about the service loader (which are not really clear) https://github.com/scalacenter/scalafix/blob/90d0edb8600ce3076839ecaea2069e1a2b941340/docs/developers/tutorial.md?plain=1#L353-L357

Based on your experience/discovery of the docs, would that have been enough?

Also, maybe [adding a one-liner comment to the g8 template](https://github.com/scalacenter/scalafix.g8/blob/8080d4a1648be44b59400b032495f2b67a95c488/src/main/g8/scalafix/rules/src/main/scala/fix/%24repo__Camel%24.scala#L6) would be helpful? Did you use that template?

mjhoy:
@bjaglin That makes sense. What do you think of d5d90b23ea51bf5b31c3e9669de8e9ce0cf18b96?

I don't _think_ I used the template.

bjaglin (approved):
Perfect. Thanks for taking the time to improve the docs!

PR #2062, opened 2024-08-31
docs: explain how users can customize linting behavior of rules

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/2000

### DIFF
--- a/docs/users/configuration.md
+++ b/docs/users/configuration.md
@@ -57,3 +57,71 @@ triggered.rules = [
   DisableSyntax
 ]
 ```
+
+## Overriding the linting behavior
+
+By default, only lints raised with `LintSeverity.Error` will result in a
+non-zero exit code. To force Scalafix to return a non-zero exit code for other
+diagnostics, it is possible to promote non-error diagnostics to the error level
+without touching the rules(s) source code, thanks to `lint.error`.
+
+This key accepts as value either
+* a string, interpreted as a regular expression to match diagnostics using the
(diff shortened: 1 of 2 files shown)


docs/users/configuration.md
```diff
@@ -57,3 +57,71 @@ triggered.rules = [
...
+lint.error.includes = ".*"
+lint.error.excludes = "UnusedScalafixSuppression"
+```
+
+is equivalent to the following CLI arguments
+
+```
+scalafix ... \
+  --settings.DisableSyntax.noFinalize=true \
+  --settings.lint.error.includes=.* \
```
bjaglin, line 125: In a shell context, this needs to be wrapped in single quotes so that `*` is not processed. However, this syntax is not supported on sbt-scalafix, so I'd rather document the sbt-scalafix one, and let CLI users figure out by themselves.

bjaglin, line 125: In a shell context, this needs to be wrapped in single quotes so that `*` is not processed.

```suggestion
  '--settings.lint.error.includes=.*' \
```

However, this syntax is not supported on sbt-scalafix, so I'd rather document the sbt-scalafix one, and let CLI users figure out by themselves (instead of being very verbose about it here and confusing sbt-scalafix users).

bjaglin: It turns out that multiline syntax does not work on sbt, so I am taking back what I wrote above and will target CLI/shell syntax instead in another PR

Issue #2107, opened 2024-10-18
`ScalaVersion` rejects Scala 2 nightlies


SethTisue:
I'm trying to build and test more of Scalafix in the Scala 2 community build, but I keep hitting this:

```
[scalafix-rules] [info]   java.lang.Exception: 2.13.16-bin-ce78754 not a valid Scala Version.
[scalafix-rules] [info]   at scalafix.internal.config.ScalaVersion$.from(ScalaVersion.scala:138)
```

could the check be relaxed to allow nightlies?

(doc on nightlies: https://docs.scala-lang.org/overviews/core/nightlies.html)

SethTisue:
context is https://github.com/scala/community-build/pull/1723 fyi @xuwei-k

PR #2109, opened 2024-10-22
don't reject Scala 2 nightlies for community build

bjaglin:
Closes https://github.com/scalacenter/scalafix/issues/2107


bjaglin:
@SethTisue is that enough to get you further?

SethTisue (approved):
yes, I tried this in a local run and that gets me unstuck, thank you!

PR #2136, opened 2024-12-01
remove scala-xml libraryDependencySchemes setting in plugins.sbt

xuwei-k:


bjaglin (approved):
Thanks!

This could apparently have been removed as early as https://github.com/scalacenter/scalafix/pull/1720, thanks to https://github.com/scala/scala/pull/10108 & https://github.com/sbt/sbt/pull/7021.

Issue #2138, opened 2024-12-11
scalafix can't parse new scala 3.6.2 new given synstax


lenguyenthanh:
The changes is here: https://github.com/lichess-org/lila-fishnet/pull/376/commits/1163f62523bf027ebb0b9a28c6e0e4a78535ac63

And the error is below:

```
info] Running scalafix on 18 Scala sources
[error] xxx/lichess/lila-fishnet/app/src/main/scala/http/MkHttpServer.scala:19:9: error: abstract givens cannot be anonymous
[error]   given Logger[IO] => MkHttpServer = new:
[error]         ^
[error] (app / scalafixAll) scalafix.sbt.ScalafixFailed: ParseError
[error] Total time: 0 s, completed 11 Dec 2024, 7:15:23 pm
```

this pr in scalameta may fix this issue: https://github.com/scalameta/scalameta/pull/4068

lenguyenthanh:
thanks! I tried it fixed all but one issue:

```
[error] {redacted}/lila-search/modules/app/src/main/scala/http.server.scala:20:9: error: abstract givens cannot be anonymous
[error]   given Logger[IO] => MkHttpServer = new:
[error]         ^
```

I think this is still scalameta issue, you can check the code here: https://github.com/lichess-org/lila-search/pull/392/files#diff-665a50336c22a54ed11c39141cc1649c40ec1cd141de9adc3754b95e000f7891R20

ivan-klass:
similarly, new context bound syntax gives an error:
```
[error] MyFile.scala:34:21: error: illegal start of declaration
[error]   def apply[F[_]: { Async, Tracer, Parallel }](
[error]                     ^^^^^
```

tgodzik:
I will take a look at those, thanks!

tgodzik:
> def apply[F[_]: { Async, Tracer, Parallel }](

This seems to parse ok, so maybe it's just not working in the released version.

bjaglin:
> > def apply[F[_]: { Async, Tracer, Parallel }](
> 
> This seems to parse ok, so maybe it's just not working in the released version.

@ivan-klass could you confirm it works with the[ latest snapshot](https://scalacenter.github.io/scalafix/docs/users/installation.html#snapshot)?

bjaglin:
> thanks! I tried it fixed some but still at least some issue:
> 
> ```
[...]

PR #2151, opened 2024-12-25
use latest scalameta for better SIP-64 support

bjaglin:
Brings in https://github.com/scalameta/scalameta/pull/4092
Fixes https://github.com/scalacenter/scalafix/issues/2138

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -23,7 +23,7 @@ object Dependencies {
   val metaconfigV = "0.14.0"
   val nailgunV = "0.9.1"
   val scalaXmlV = "2.2.0"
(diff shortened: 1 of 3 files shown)


lenguyenthanh:
scalameta published new version, and I can confirm that it works for my cases at least.

Dessix:
@lenguyenthanh This may not cover all scenarios- `scalafmt` just merged this version of `scalameta` and it doesn't seem to cover [at least one non-parametric named-given scenario](https://github.com/scalameta/scalafmt/issues/4621#issuecomment-2564289218). Also, SIP-64 Context Bounded Type Members are unsupported.

Issue #2166, opened 2025-01-14
`ExplicitResultTypes` breaks compilation when using symbolic methods


dhalilov:
Hi folks,

`ExplicitResultTypes` doesn't seem to work correctly when the def/val uses symbols, as it inserts `:` right next to the symbol, which fools the Scala parser. Adding a space before the `:` should resolve the issue.

I'm not sure if this is part of the known limitations mentionned in the [documentation page](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html#known-limitations), as I haven't found an issue describing the problem.

# How to reproduce
Here is a minimal example, using Scala 3.5.2 and sbt 1.10.7.

`src/main/scala/Main.scala`:
```scala
object A {
  // Using the non-trivial example from the documentation to make sure that ExplicitResultsType kicks in
  def ! = 1.to(10).map(i => i -> i.toString()).toMap
}

@main def main(): Unit =
  println(A.!)
```

`build.sbt`:
```scala
lazy val root = project
  .in(file("."))
  .settings(
    scalaVersion := "3.5.2",
    semanticdbEnabled := true
  )
```

`project/plugins.sbt`:
```scala
addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.13.0")
```

`.scalafix.conf`:
```
rules = [
  ExplicitResultTypes
]
```

Running `sbt scalafix` produces the following line:
```scala
def !: Map[Int, String] = 1.to(10).map(i => i -> i.toString()).toMap
```
which fails to compile. The following version:
```scala
def ! : Map[Int, String] = 1.to(10).map(i => i -> i.toString()).toMap
```
would compile correctly.

tgodzik:
Thanks for reporting, looks like an issue in the presentation compiler. We need to fix it upstream, but should be a quick fix.

bjaglin:
Supposedly fixed in the latest dotty snapshot thanks to https://github.com/scala/scala3/pull/22485. It won't be backported to 3.5.x since it's EOL (and we follow the minor version of the sources to pick the presentation compiler), so I am afraid you will have to bump to 3.6.4 (hoping it does get in there).

dhalilov:
[...]

PR #2190, opened 2025-02-06
ExplicitResultTypes: annotate members with symbolic names on Scala 2

rochala:
TypeName / TermName constructor does not run NameTransformer. The issue was that the type had encoded names while we used plain: Ours `TermName(-)` vs `TermName($minus)`

### DIFF
--- a/scalafix-rules/src/main/scala-2/scalafix/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala-2/scalafix/internal/pc/ScalafixGlobal.scala
@@ -5,6 +5,7 @@ import java.{util => ju}
 
 import scala.collection.compat._
 import scala.collection.mutable
+import scala.reflect.NameTransformer
 import scala.reflect.internal.{Flags => gf}
 import scala.reflect.io.VirtualDirectory
 import scala.tools.nsc.Settings
@@ -17,7 +18,6 @@ import scala.{meta => m}
 
 import scala.meta.internal.semanticdb.scalac.SemanticdbOps
 import scala.meta.io.AbsolutePath
-import scala.reflect.NameTransformer
 
 object ScalafixGlobal {
   def newCompiler(
--- a/scalafix-rules/src/main/scala-2/scalafix/internal/pc/ScalafixGlobal.scala
+++ b/scalafix-rules/src/main/scala-2/scalafix/internal/pc/ScalafixGlobal.scala
@@ -17,6 +17,7 @@ import scala.{meta => m}
 
 import scala.meta.internal.semanticdb.scalac.SemanticdbOps
 import scala.meta.io.AbsolutePath
+import scala.reflect.NameTransformer
 
 object ScalafixGlobal {
   def newCompiler(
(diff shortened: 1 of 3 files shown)


scalafix-tests/input/src/main/scala-2/test/explicitResultTypes/ExplicitResultTypesBase.scala
```diff
@@ -36,6 +36,7 @@ object ExplicitResultTypesBase {
     x + 2
   object ExtraSpace {
     def * = "abc".length
+    def ! = "abc".length
     def foo_ = "abc".length
     def `x` = "abc".length
     def `x ` = "abc".length
```
rochala, line 42: This case is not working because semanticdb symbol for this is:

`x` => test/explicitResultTypes/ExplicitResultTypesBase.ExtraSpace.x().
`x ` => test/explicitResultTypes/ExplicitResultTypesBase.ExtraSpace.x(+1).

The inverse method is too naive with this and the methhod fails because the type has following members:

`List(<init>, x, x$u0020)`

We can't easily map `x(+1)` into `x$u0020`. @tgodzik pinging you to let you know

rochala:
We need to investigate whether we need the same fix for Scala 2 inverseSemanticdb method in metals and for Scala 3. I'm almost sure that yes.

scalafix-rules/src/main/scala-2/scalafix/internal/pc/ScalafixGlobal.scala
```diff
@@ -72,6 +73,13 @@ class ScalafixGlobal(
     import scala.meta.internal.semanticdb.Scala._
     if (!symbol.isGlobal) return Nil
 
+    def EncodedTermName(str: String): TermName = TermName(
+      NameTransformer.encode(str)
+    )
+    def EncodedTypeName(str: String): TypeName = TypeName(
+      NameTransformer.encode(str)
```
bjaglin, line 80: I wonder if inverse lookup of types is actually needed for ExplicitResultTypes (it might be a lot of code that was copied from metals)

bjaglin (approved):
Thanks for the PR!

It looks like this was addressed in metals via https://github.com/scalameta/metals/pull/5481, in a slightly different and more exhaustive way, but let's get that in anyway.

rochala:
> Thanks for the PR!
> 
> It looks like this was addressed in metals via [scalameta/metals#5481](https://github.com/scalameta/metals/pull/5481), in a slightly different and more exhaustive way, but let's get that in anyway.

Oh, thanks for pointing the PR. It also answered my question about backporting it to Scala 3 :D https://github.com/scalameta/metals/pull/5481#issuecomment-1677283987

bjaglin:
> It also answered my question about backporting it to Scala 3 :D [scalameta/metals#5481 (comment)](https://github.com/scalameta/metals/pull/5481#issuecomment-1677283987)

I am not sure the answer was correct, see https://github.com/scalacenter/scalafix/issues/2166 - in any case, it's been taken care of.

Issue #1937, opened 2024-02-21
RemoveUnused: detect unused function parameters in Scala 3


bjaglin:
Despite the presence of `-Wunused:params` and the explicit mention of it in https://github.com/lampepfl/dotty/pull/16157, I did not manage to get diagnostics (nor warnings) for unused parameters with scala 3.4.0-RC2, so the feature is currently documented/tested as Scala2-only.

We probably need to report this upstream.

```
$ metap ./scalafix-tests/input/target/jvm-3.4.0-RC2/meta/META-INF/semanticdb/scalafix-tests/input/src/main/scala/test/removeUnused/RemoveUnusedParams.scala.semanticdb 
scalafix-tests/input/src/main/scala/test/removeUnused/RemoveUnusedParams.scala
------------------------------------------------------------------------------

Summary:
Schema => SemanticDB v4
Uri => scalafix-tests/input/src/main/scala/test/removeUnused/RemoveUnusedParams.scala
Text => empty
Language => Scala
Symbols => 11 entries
Occurrences => 28 entries

Symbols:
local0 => param unused: String
local1 => param unused: String
local2 => param used: String
local3 => param unused: Long
local4 => implicit param string: String
test/removeUnused/UnusedParams. => final object UnusedParams extends Object { self: UnusedParams.type => +5 decls }
test/removeUnused/UnusedParams.f. => val method fFunction1[String, Unit]
test/removeUnused/UnusedParams.ff. => val method ffFunction1[String, Unit]
test/removeUnused/UnusedParams.fs. => val method fsFunction2[String, Long, Unit]
test/removeUnused/UnusedParams.g(). => method g(x: Function1[String, Unit]): Unit
test/removeUnused/UnusedParams.g().(x) => param x: Function1[String, Unit]

Occurrences:
[3:8..3:12) => test/
[3:13..3:25) <= test/removeUnused/
[5:7..5:19) <= test/removeUnused/UnusedParams.
[6:6..6:7) <= test/removeUnused/UnusedParams.f.
[6:9..6:15) => scala/Predef.String#
[6:19..6:23) => scala/Unit#
[6:26..6:32) <= local0
[6:36..6:43) => scala/Predef.println(+1).
[7:6..7:8) <= test/removeUnused/UnusedParams.ff.
[7:12..7:18) <= local1
[7:20..7:26) => scala/Predef.String#
[7:31..7:38) => scala/Predef.println(+1).
[...]

PR #2211, opened 2025-03-13
build and test against scala 3.7.0

bjaglin:
* Fixes https://github.com/scalacenter/scalafix/issues/1937 via https://github.com/scala/scala3/pull/20894
* Fixes https://github.com/scalacenter/scalafix/issues/2166 via https://github.com/scala/scala3/pull/22485


scalafix-tests/output/src/main/scala-3next/test/removeUnused/RemoveUnusedParams.scala
```diff
@@ -0,0 +1,9 @@
+package test.removeUnused
+
+object UnusedParams {
+  val f: String => Unit = _ => println("f")
+  val ff = (_: String) => println("f")
+  val fs = (used: String, _: Long) => println(used)
+  def g(x: String => Unit): Unit = ???
+  g{implicit string => println("g")}
```
bjaglin, line 8: there is no diagnostic for this on the semanticdb emitted by scala 3.7.0

docs/rules/RemoveUnused.md
```diff
@@ -81,14 +84,14 @@ object Main {
...
+    case AB(a, b) => println("Not used, but canonical names")
+  }
 }
 ```
 
-On Scala 3, `-Wunused:unsafe-warn-patvars` is required.
-
-On Scala 2.13.15+, canonical patterns (vars with the same names as the
-attributes) do not trigger unused warnings, so the input above will not
-be rewritten. See https://github.com/scala/bug/issues/13035.
```
bjaglin, line 91: since this is now the default behavior on latest 2.13.x and 3.x, I am removing the "exception" note and documenting it via an example, as it can be surprising

scalafix-tests/input/src/main/scala-3next/test/explicitResultTypes/ExplicitResultTypesBase.scala
```diff
@@ -0,0 +1,46 @@
...
+  private def h(a: Int) = ""
+  private var i = 22
+  val k = (1, "msg")
+  def comment(x: Int) =
+    // comment
+    x + 2
+  object ExtraSpace {
+    def * = "abc".length
+    def ! = "abc".length
+    //def foo_ = "abc".length
```
bjaglin, line 29: this works fine on Scala 2 but generates invalid code on Scala 3

```scala
// scala 2
def foo_ : Int = "abc".length

// scala 3.7.0
def foo_: Int = "abc".length
```

bjaglin: it looks like https://github.com/scalacenter/scalafix/issues/309 hasn't been addressed in the PC

bjaglin: Scalameta handled it via https://github.com/scalameta/scalameta/commit/d2acb56632f205171f10f8fec60909a9526de2ac

tgodzik: Might be worth an issue I think?

bjaglin: I wanted to open an issue against the PC indeed, but I can't reproduce it on scala3 main, [the heading space does get added there](https://github.com/bjaglin/dotty/commit/843b65c201539562b2766ce89006809a6b4ea276). I need to continue the investigation.

tgodzik: It looks like it was fixed in the meantime. Still broken in 3.7.0-RC2. Pretty sure there was a PR about it, but I can't find it currently.

scalafix-tests/output/src/main/scala-3/test/removeUnused/RemoveUnusedParams.scala
```diff
@@ -0,0 +1,9 @@
+package test.removeUnused
+
+object UnusedParams {
+  val f: String => Unit = _ => println("f")
+  val ff = (_: String) => println("f")
+  val fs = (used: String, _: Long) => println(used)
+  def g(x: String => Unit): Unit = ???
+  g{implicit string => println("g")}
```
bjaglin, line 8: there is no diagnostic for this on the semanticdb emitted by scala 3.7.0

scalafix-tests/input/src/main/scala-3next/test/explicitResultTypes/ExplicitResultTypesBase.scala
bjaglin: Oh, you're right! Thanks, I was going crazy, it didn't occur to me that it could have been fixed after 3.7.0 was branched off, because the file where the fix was made didn't change. But it's actually `isOperatorName` that was improved in https://github.com/scala/scala3/pull/22854/commits/094eaee2af7ab9b1bacca51098c9ee9242a0409e.

scalafix-tests/output/src/main/scala-3/test/removeUnused/RemoveUnusedParams.scala
tgodzik: Looks like a regression from the huge PR about unused settings. I think this might not only be semanticdb.

PR #2227, opened 2025-04-14
build and test against scala 3.7.1

bjaglin:


scalafix-tests/input/src/main/scala-3lts/test/explicitResultTypes/EnumerationValue.scala
```diff
@@ -0,0 +1,24 @@
...
+  }
+  import Bool._
+  def day(d: Day.Value): Unit = ???
+  val d =
+    if (true) Day.Weekday
+    else Day.Weekend
+  day(d)
+  val b =
+    if (true) True
+    else False
```
bjaglin, line 23: This was copied to skip scala3-lts, as the 3.7.1 PC is trying to be smarter when shortening types but gets fooled by the `import Bool.` somehow , and we end up with code that does not compile

```suggestion
  val d: Value =
    if (true) Day.Weekday
    else Day.Weekend
  day(d)
  val b: Value =
    if (true) True
    else False
```

```
[error] -- [E007] Type Mismatch Error: /Users/brice.jaglin/git/scala/scalafix/scalafix-tests/output/src/main/scala-3next/test/explicitResultTypes/EnumerationValue.scala:15:18
[error] 15 |    if (true) Day.Weekday
[error]    |              ^^^^^^^^^^^
[error]    |      Found:    (test.explicitResultTypes.EnumerationValue.Day.Weekday :
[error]    |        test.explicitResultTypes.EnumerationValue.Day.Value)
[error]    |      Required: test.explicitResultTypes.EnumerationValue.Bool.Value
[error]    |
[error]    | longer explanation available when compiling with `-explain`
[error] -- [E007] Type Mismatch Error: /Users/brice.jaglin/git/scala/scalafix/scalafix-tests/output/src/main/scala-3next/test/explicitResultTypes/EnumerationValue.scala:16:13
[error] 16 |    else Day.Weekend
[error]    |         ^^^^^^^^^^^
[error]    |      Found:    (test.explicitResultTypes.EnumerationValue.Day.Weekend :
[error]    |        test.explicitResultTypes.EnumerationValue.Day.Value)
[error]    |      Required: test.explicitResultTypes.EnumerationValue.Bool.Value
[error]    |
[error]    | longer explanation available when compiling with `-explain`
[error] -- [E007] Type Mismatch Error: /Users/brice.jaglin/git/scala/scalafix/scalafix-tests/output/src/main/scala-3next/test/explicitResultTypes/EnumerationValue.scala:17:6
[error] 17 |  day(d)
[error]    |      ^
[error]    |      Found:    (test.explicitResultTypes.EnumerationValue.d :
[error]    |        test.explicitResultTypes.EnumerationValue.Bool.Value)
[error]    |      Required: test.explicitResultTypes.EnumerationValue.Day.Value
[error]    |
[error]    | longer explanation available when compiling with `-explain`
```

bjaglin: I just went through [the recent changes of the PC](https://github.com/scala/scala3/commits/release-3.7.1/presentation-compiler) and https://github.com/scala/scala3/pull/22898 **might** be a candidate for the regression.

@tgodzik since you are the author, maybe you can confirm/infirm? I will end up filing an issue upstream eventually, but my knowledge of the dotty repo is limited so it will take me a while to write a unit test :)

tgodzik: It might be some side effect of that PR yeah. It was a tricky part of the compiler. If you could just put the example in a issue I will take a look.

tgodzik: Ach, I think I know what is going on. Those are path dependent types and it's not handled at all properly :/

tgodzik: I've got something working, but I broke another test:
https://github.com/scala/scala3/compare/main...tgodzik:scala3:fix-path-dep?expand=1

I am not sure how to differentiate types that are actually in Scope from those that are not if they actually have the same symbol.

tgodzik: Ok, let's try get https://github.com/scala/scala3/pull/23124 merged, took me longer than expected

bjaglin: Thanks! Taking note to simplify the test once 3.7.2 is tagged

PR #2250, opened 2025-05-15
Update JVM versions in contributing docs

subhramit:
Tests such as
```bash
> unit3_3_4 / test
```
as mentioned in `CONTRIBUTING.md` don't readily run.
```bash
sbt:scalafix> unit3_3_4 / test
[error] Expected ID character
[error] Not a valid command: unit3_3_4
[error] Expected project ID
[error] Expected configuration
[error] Expected ':'
[error] Expected key
[error] Not a valid key: unit3_3_4
[error] unit3_3_4 / test
[error]     
```
Version `3.3.4` is no longer available in the jvm versions.
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,23 +29,23 @@ sbt shell.
 
 ```sh
 # Fast unit tests for core, rules, reflect, cli & testkit.
-> unit3_3_4 / test
+> unit3_3_6 / test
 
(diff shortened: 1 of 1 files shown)


bjaglin (approved):
Thanks! Unfortunately, this is meant to drif as there is no CI enforcing that it's updated each time we bump, but we might as well do a one-shot now!

PR #2291, opened 2025-10-05
use Sonatype Central Snapshots instead of OSSRH public/snapshots

bjaglin:
Follows https://github.com/scalacenter/scalafix/pull/2257

### DIFF
--- a/scalafix-rules/src/main/scala-3/pc/Embedded.scala
+++ b/scalafix-rules/src/main/scala-3/pc/Embedded.scala
@@ -9,7 +9,6 @@ import scala.meta.pc.PresentationCompiler
 
 import coursierapi.Dependency
 import coursierapi.Fetch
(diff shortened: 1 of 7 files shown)


scalafix-rules/src/main/scala-3/pc/Embedded.scala
```diff
@@ -61,12 +60,6 @@ object Embedded {
     val jars = Fetch
       .create()
       .addDependencies(deps: _*)
-      .addRepositories(
-        // for some versions, the presentation compiler depends on mtags-interfaces SNAPSHOTs
-        MavenRepository.of(
-          "https://oss.sonatype.org/content/repositories/snapshots"
-        )
-      )
```
bjaglin, line 69: looks like it's no longer the case... we can reintroduce it later if needed

PR #2306, opened 2025-10-30
Support case-insensitive sorting

### LINKED ISSUE #2305: OrganizeImports: Support case-insensitive sorting
For example, I want to be able to get this order when running scalafix:
```
import org.apache.spark.sql.streaming.TimeMode
import org.apache.spark.sql.streaming.TTLConfig
```

Note that by ascii ordering TTLConfig comes first, but in case-insensitve mode TimeConfig comes first.
I checked https://scalacenter.github.io/scalafix/docs/rules/OrganizeImports.html and it seems like this is not currently supported.

I see 3 ways to represent this in config:
1. Add new value for option importsOrder, call it AsciiCaseInsensitive (I'd prefer this).
2. Add a new boolean option "caseInsensitive" (default false).
3. Add a new boolean option "caseSensitive" (default true).

I'll be happy to contribute this, but please let me know if scuh contribution will be accepted.

fedimser:
This PR adds an option `importsOrder = AsciiCaseInsensitive` that allows to sort imports alpabetically, but ignoring case.

Resolves #2305.

Tested locally: `sbt "expect3_7_3Target3_7_3 / testOnly -- -z organizeImports"`

### DIFF
--- a/docs/rules/OrganizeImports.md
+++ b/docs/rules/OrganizeImports.md
@@ -1127,10 +1127,10 @@ that the IntelliJ IDEA Scala import optimizer picks ("lexicographically"
 option).
 
 #### `AsciiCaseInsensitive`  
-Sort import statements by ASCII codes, but not distinguishing uppercase 
-and lowercase letters. Characters between uppercase and lowercase letters
-in ASCII table are ordered before any letter. If two strings are identical
-up to case, case-sensitive order used to tie breaks.
+Sort import statements by ASCII codes, without distinguishing between uppercase
+and lowercase letters. Characters that appear between uppercase and lowercase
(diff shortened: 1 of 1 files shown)


scalafix-tests/shared/src/main/scala/test/organizeImports/Case.scala
```diff
@@ -0,0 +1,19 @@
+package test.organizeImports
+
+object Case {
```
bjaglin: Could you test/document where literals starting with characters between uppercase and lowercase in the ASCII table (like `_foo`) end up?

bjaglin (commented):
Thanks for the contribution. Could you look into testing the corner case I mention in the comment?

PR #2308, opened 2025-11-02
Add the ForbiddenSymbol rule to community rules

tanin47:
See more here: https://github.com/tanin47/scalafix-forbidden-symbol

Here's a relevant discussion: https://users.scala-lang.org/t/scalafix-forbidding-imports-and-methods-from-being-used/10307/12


bjaglin:
> I haven't tried scaluzzi because I didn't know it would support that ... based on its description.

OK, I guess that makes an argument for advertizing yours then, even though it's more limited.

> I didn't know we could do the expectation of the linters. I thought the failure on test was expected because the scalafix rule is supposed to fail haha.

Perfect. You can also omit the output files, since they are useful only for rewriting rules.

PR #2319, opened 2025-11-21
build and test against scala 3.8.1

bjaglin:
https://contributors.scala-lang.org/t/scala-3-8-0-release-thread/7291


tgodzik:
Ok, looks like we will need to use a separate classloader. I can dig into it later next week.

bjaglin:
> I can dig into it later next week.

Thanks for the offer, but I finally have some spare this next week so I want to give it a go, now that it's clear it's needed. Will circle back if I need help. Thanks!

Issue #1747, opened 2023-04-04
Better support for cross compilation


steinybot:
It is kind of a pain to use scalafix with an sbt project that is cross compiled for Scala 2 and 3 because scalafix has a hard fail if given a rule that isn't supported for that Scala version. For example:
```
[error] (web / Compile / scalafix) scalafix.sbt.InvalidArgument: 2 errors
[error] [E0] This rule is specific to Scala 2, because the compiler option `-Ywarn-unused` is not available yet in scala 3 To fix this error, remove RemoveUnused from .scalafix.conf
[error] [E1] This rule is specific to Scala 2, since procedure syntax is not supported in Scala 3. To fix this error, remove ProcedureSyntax from .scalafix.conf
```

Removing the rule from Scala 2 is unsatisfactory so the only other solution seems to be to have two separate configuration files. I didn't see any include functionality so this would have to be a copy/paste and both files have to be kept in sync.

It would be nice to have better support for having a single configuration that is used for multiple Scala versions.

A few thoughts come to mind:
- Change it to a warning instead of an error
- Add an `include` feature to the config so that common configuration could be shared
- Make parts of the configuration conditional based on the Scala version

bjaglin:
Thanks for reaching out!

First of all, it's worth noting that [initial support for RemoveUnused is around the corner, in time for Scala 3.3.0](https://github.com/scalacenter/scalafix/pull/1728), so hopefully this will be a little less needed soon.

> Add an `include` feature to the config so that common configuration could be shared

Actually, scalafix relies on metaconfig, which [has precisely support for that through HOCON](https://github.com/scalameta/metaconfig/blob/c7a1aa5da2d43bd7c2fe8ebbfa6697b8abd5ff8e/metaconfig-tests/jvm/src/test/scala/metaconfig/typesafeconfig/TypesafeConfig2ClassSuite.scala#L54-L108). So you should be able to create a common configuration file included by scala-specific configuration files, and use them with something like
```
scalafixConfig := Some(file(s".scalafix-scala${scalaBinaryVersion.value}.conf"))
```

Would that work for you? If so, we could add a dedicated section to document that in https://github.com/scalacenter/scalafix/blob/main/docs/users/installation.md. PR with a reference to a real example usage would be very much welcome!

PR #2323, opened 2025-11-26
docs: cross-building guide

arnavsharma990:
Fixes scalacenter#1747

This PR adds a new documentation page explaining best practices for using
Scalafix in cross-compiled Scala 2 and Scala 3 projects.

### It covers:
- Why Scalafix config fails across different Scala versions
- How to structure a shared `base.conf` plus per-version overrides
- Using the “include” pattern for shared configuration
- SBT configuration to automatically pick the correct config file
- CLI/CI examples for selecting per-version configs
- Ideas for future improvements (conditional config, include support, etc.)

This is a documentation-only update.

### DIFF
--- a/docs/users/cross-building.md
+++ b/docs/users/cross-building.md
@@ -13,7 +13,7 @@ given Scala version, a different set of rules or rule features might have to be
 selected for each run, in order to avoid failing the entire run and preventing
 relevant rules to run.
 
-Since nothing in the current API allows rules to advertize their limitations
+Since nothing in the current API allows rules to advertise their limitations
 programmatically to Scalafix, this page explains how to do it manually.
 
 ## Why separate configs are required


docs/users/cross-compilation.md
```diff
@@ -0,0 +1,144 @@
+---
+id: cross-compilation
+title: Cross compilation setups
+---
+
+Cross-building the same codebase with Scala 2.13 and Scala 3.3+ is common, but
+it collides with how Scalafix resolves rules and compiler options. A rule or
+`--scalac-options` flag that only works on Scala 2 can crash Scalafix when the
+Scala 3 compilation unit is processed, and vice versa. Until Scalafix grows
+conditional configuration, the safest way to avoid those failures is to split
```
bjaglin: There is no plan to do conditional configuration so the "Until" can be misleading

docs/users/cross-compilation.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Your CI job can loop over each target Scala version, selecting the matching
+config before running `scalafix --check`.
+
+## Recommendations and current limitations
+
+* Keep rules that truly work on both versions inside `common.conf`.
+* Isolate risky scalac options (`-Wunused`, `-Ywarn-unused-import`, `-P:semanticdb`)
+  inside each version-specific file or sbt setting.
```
bjaglin: Same comment as above - scalac-options are not part of the config so I don't understand that part

docs/users/cross-compilation.md
```diff
@@ -0,0 +1,144 @@
...
+ergonomics for cross compilation. Potential improvements include:
+
+* Conditional configuration blocks directly inside `.scalafix.conf` (for example
+  `if scalaVersion.startsWith("3.")`).
+* First-class support for including multiple files via CLI flags.
+* Allowing rule selection based on the detected input Scala dialect.
+
+Until those land, the include-based layout above is the recommended, battle-tested
+approach.
+
```
bjaglin: I don't think we should advertize potential improvements in the docs, as it's unclear if there is a need for it. As a matter of fact, I was hoping to close https://github.com/scalacenter/scalafix/issues/1747 with this doc update.

docs/users/installation.md
```diff
@@ -443,6 +443,23 @@ cs launch scalafix:0.13.0 -- --version # Should say 0.13.0
 
 ```
 
+### Multiple values for CLI flags
```
bjaglin: this is merged, you can rebase against main

docs/users/cross-compilation.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+`scala2.conf`
+```scala
+include "common.conf"
+
+rules += RemoveUnused
+
+// Scala 2 only compilers flags or rule settings
```
bjaglin: As mentioned elsewhere, we should focus on compatibility (it's not necessarily about compiler flags but it can be new language features - although that's mostly for community rules, that you don't mention)
```suggestion
// Scala 2 only rules or rule settings
```

docs/users/cross-compilation.md
```diff
@@ -0,0 +1,144 @@
+---
+id: cross-compilation
+title: Cross compilation setups
+---
+
+Cross-building the same codebase with Scala 2.13 and Scala 3.3+ is common, but
```
bjaglin: I have a personal preference for cross-building rather than cross-compilation as it puts less emphasis on the compile phase - let's just make sure we use the same term in the header and the body

bjaglin (commented):
Thanks for taking the time to write that down, docs are very important! Have a look at my suggestions, I think the doc could be a bit more concise.

Also, I think the amount of people using the CLI is very low, so I wonder if we couldn't just make a paragraph explaining the sbt setup in the main page (which is sbt-centric)?

docs/users/cross-building.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### Multiple include files
+
+You may split out even more granular snippets (for example `linting.conf`,
+`rewrites.conf`) and include them from both `scala2.conf` and `scala3.conf`. The
+HOCON syntax supports nested includes, so feel free to create the hierarchy that
+matches your team conventions.
+
+## Selecting the right config in sbt
```
bjaglin:
```suggestion
## Selecting the right configuration file

 ### sbt
```

bjaglin:
> any update on merge?

Sorry, I have been busy. I'll merge this now, thanks for your contribution!

bjaglin:
Live at https://scalacenter.github.io/scalafix/docs/users/cross-building.html

PR #2324, opened 2025-11-28
test(organizeimports): add failing IO tests for inline comment behavior (issue #2267)

### LINKED ISSUE #2267: OrganizeImports does not retain inline comments
Hi!

I noticed today that if, for some desperate reason, you need to have a comment against an import, organize imports does not respect which import the comment was against.

Before:

```scala
import cats.Applicative
import cats.Functor
import cats.effect.kernel.Sync
import cats.effect.kernel.Temporal
import cats.kernel.Eq
import cats.syntax.eq.*
import tyrian.runtime.CmdHelper
import cats.kernel.Monoid // A comment
```

After

```scala
import cats.Applicative
import cats.Functor
import cats.effect.kernel.Sync
import cats.effect.kernel.Temporal
import cats.kernel.Eq
import cats.kernel.Monoid
import cats.syntax.eq.*
import tyrian.runtime.CmdHelper
```

Probably version specific, but I have also observed this 'after' state, too, where the comment is retained, but attached to the wrong import:

```scala
import cats.Applicative
import cats.Functor
import cats.effect.kernel.Sync
import cats.effect.kernel.Temporal
import cats.kernel.Eq
import cats.kernel.Monoid
import cats.syntax.eq.*
import tyrian.runtime.CmdHelper // A comment
```

Apologies is this has already been resolved, I had a look for similar issues but found none.

arnavsharma990:
Fixes #2267

This PR adds new IO test cases for OrganizeImports to capture the incorrect handling of inline comments (e.g., comments to the right of imports being lost, reordered incorrectly, or not associated with the correct import). These test
s encode the expected behavior so the underlying rule can later be updated to preserve inline comments consistently.

Locally, the new tests pass, but some unrelated expect*/testOnly suites occasionally behave inconsistently on my machine (sometimes showing passed, canceled, or failed). I’ve attached a screenshot in the discussion, but this inconsistency does not involve the new OrganizeImports tests themselves.

Please review the added test cases and confirm whether this structure matches what the maintainers expect before I proceed to implement the actual fix to OrganizeImports.
Screenshot 2025-11-28 at 4 17 07 PM
Screenshot 2025-11-28 at 4 04 01 PM

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -173,14 +173,17 @@ class OrganizeImports(
     val removalEnd = {
       // Look for trailing comment tokens after the last import
       val tokensAfter = allTokens.drop(lastImportTokenEnd)
-      val trailingCommentEnd = tokensAfter.takeWhile { t =>
-        t.is[Token.Comment] || t.is[Token.Space] || t.is[Token.Tab]
-      }.lastOption.collect {
-        case c: Token.Comment => allTokens.indexOf(c) + 1
-      }
+      val trailingCommentEnd = tokensAfter
+        .takeWhile { t =>
+          t.is[Token.Comment] || t.is[Token.Space] || t.is[Token.Tab]
+        }
+        .lastOption
+        .collect { case c: Token.Comment =>
+          allTokens.indexOf(c) + 1
+        }
       trailingCommentEnd.getOrElse(lastImportTokenEnd)
     }
-    
+
     val removalPatch = Patch.removeTokens(
       allTokens.slice(imports.head.tokens.start, removalEnd)
     )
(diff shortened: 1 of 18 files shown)


arnavsharma990:
Thanks for the clarification and for reviewing the tests!
I’ll trim the test cases as suggested.

Also, should I start working on the actual fix for the issue now?

tgodzik:
Sure, just make sure that if you use AI that you properly vet the output and try to minimize the changes.

bjaglin (commented):
That's a great start 👍 I think we could add heading/trailing examples, as well as multi-line `/* */` comments to have a better coverage.

Feel free to start working on implementation.

bjaglin (commented):
Thanks for the taking a stab at the implementation @arnavsharma990 !
- Could you have a look at the remaining comments? I find them all relevant. As noted [here](https://github.com/scalacenter/scalafix/issues/2267#issuecomment-3901564070), there is a new[ scalameta feature available on main](https://scalameta.org/docs/trees/examples.html#coments-metatreecomments) that should make the implementation much simpler
- Could you update the title and the description now that this PR is actually an implementation?

bjaglin:
- implementation superseeded by https://github.com/scalacenter/scalafix/pull/2428
- tests included in https://github.com/scalacenter/scalafix/pull/2428/changes/a07088cedfb6fcbe65008a34bee2050d3232e4fb

PR #2355, opened 2026-01-22
Include scala-library on Scalafix CLI classpath by default

### LINKED ISSUE #1712: scalafix CLI should include scala-library on the classpath by default
In order to test out the `ExplicitResultTypes` semantic rule with scalafix CLI like the [example in the docs](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html), one currently has to include `scala-library` in the `--classpath` argument.

```
$ scalafix --scala-version 2.13.10 --verbose --rules ExplicitResultTypes --classpath=/tmp/scala-library-2.13.10.jar --semanticdb-targetroots . --stdout
info: Processing (0/1) /tmp/scalafix-sandbox/Main.scala
package com.example
object Main {
  def myComplexMethod: Map[Int,String] = 1.to(10).map(i => i -> i.toString).toMap
}
```

I think it'd be a smoother developer experience to include `scala-library` (and possibly other `org.scala-lang.*` libraries) on the classpath by default so that small examples with semantic rules work when people are using scalafix CLI as a means to try out the tooling for the first time.

originally surfaced in https://github.com/scalameta/scalameta/issues/2979

arnavsharma990:
Fixes #1712

This PR improves Scalafix CLI UX by automatically including the Scala standard library on the classpath when \`--scala-version\` is provided and the classpath does not already contain it.

The implementation is intentionally conservative:
- Resolution is only attempted when a fully specified Scala version (major.minor.patch) is provided, to avoid ambiguous resolution for partial versions (e.g. \`2.13\`).
- User-provided classpaths are never overridden, only extended when necessary.
- Resolution failures are handled gracefully.
- The change only affects the CLI and does not impact sbt, bloop, or programmatic integrations.

### DIFF
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -358,13 +358,14 @@ case class Args(
     else Configured.error(s"--sourceroot $path is not a directory")
   }
 
-  private def containsScalaLibrary(classpath: Classpath): Boolean = {
-    classpath.entries.exists { path =>
-      Option(path.toNIO.getFileName).exists { fileName =>
-        val fileNameStr = fileName.toString
-        fileNameStr.contains("scala-library") || fileNameStr.contains("scala3-library")
+  private def containsScalaLibrary(classpath: Classpath): Boolean =
+  classpath.entries.exists { path =>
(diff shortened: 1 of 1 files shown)


bjaglin:
Hey @arnavsharma990, thanks for the contribution - can you address both copilot comments? They are relevant.

PR #2366, opened 2026-02-13
scalameta: upgrade to v4.15.2

kitbellew:
This version contains a bugfix for comments attached to a tree. We should have attempted to use that new functionality and fixing any bugs, before removing AssociatedComments.

Requires the following changes:
- EscapeHatch: don't use AssociatedComments
- AssociatedComments: add a skeleton, remove use
- scalameta protobuf: use new message structure

Closes #2372.


kitbellew:
@tgodzik @bjaglin ptal

Issue #2041, opened 2024-08-21
Publish core_3


bjaglin:
Spin-off from https://github.com/scalacenter/scalafix/issues/1680

- [ ] Consider our options for https://github.com/scalacenter/scalafix/issues/1949
- [ ] Wait for https://github.com/scalameta/scalameta/issues/2485
- [ ] Find a way to test rules built against core3 in test kit
- [ ] Publish it as a patch release
- [ ] Go through the listed community rules and make sure Scala Steward rolls out the change (even if the _3 artifact won't be fetched, it will prepare the field for the upcoming minor release)
- [ ] Advocate for community rules to cross-compile to scala 3 by updating https://github.com/scalacenter/scalafix.g8 & examples in the docs (use scala center twitter account?)
- [ ] Instead of testing Scala 3 targets with Scala 2.12, test Scala 2.13 targets with Scala 3 in https://github.com/scalacenter/scalafix.g8 since that's what cli3 will take by default 
- [ ] Improve messaging when _3 artifacts are not found for 
- [ ] Start using it from cli3 in a minor release

bjaglin:
https://github.com/scalameta/scalameta/pull/3347 is now merged :tada: 
```
$ cs resolve org.scalameta:scalameta_3:4.12.5
com.lihaoyi:sourcecode_2.13:0.4.2:default
org.scala-lang:scala-library:2.13.14:default
org.scala-lang:scala3-library_3:3.3.4:default
org.scalameta:common_2.13:4.12.5:default
org.scalameta:parsers_3:4.12.5:default
org.scalameta:scalameta_3:4.12.5:default
org.scalameta:trees_2.13:4.12.5:default
```

org.scalameta:common_2.13 -> ?
[transversers missing](https://github.com/scalameta/scalameta/blob/f029adb3e0e3422fdf17ce84c5072a8888801f9d/build.sbt#L248) -> https://github.com/scalameta/scalameta/issues/4146
org.scalameta:trees_2.13 -> https://github.com/scalameta/scalameta/issues/4145

tgodzik:
> so https://github.com/scalameta/scalameta/issues/4145 is not coming any time soon

Not likely at this point, the code being generated is quite complicated. If we ever switch, we would try to just generate the same bytecode or if that proves impossible release 5.x

We could also rethink the trees and just write them anew. Do 5.x afterwards. Though to be honest I am a bit scared to touch that 😅

PR #2401, opened 2026-04-03
Bump jgit version

matthughes:
In 7.0.0, jgit started supporting git work trees.  I've been using this workflow a lot lately and am used to relying on `scalafix --diff` but that doesn't work when you're in a worktree.  It doesn't think you're in a git repo.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -22,7 +22,7 @@ object Dependencies {
   val coursierInterfaceV = "1.0.28"
   val commontTextV = "1.15.0"
   val googleDiffV = "1.3.0"
-  val jgitV = "5.13.5.202508271544-r"
+  val jgitV = "7.6.0.202603022253-r"
   val metaconfigV = "0.18.4"
   val nailgunV = "0.9.1"
   val scalaXmlV = "2.4.0"


project/Dependencies.scala
```diff
@@ -22,7 +22,7 @@ object Dependencies {
   val coursierInterfaceV = "1.0.28"
   val commontTextV = "1.15.0"
   val googleDiffV = "1.3.0"
-  val jgitV = "5.13.5.202508271544-r"
+  val jgitV = "7.6.0.202603022253-r"
```
tgodzik, line 25: I think this was to support JDK 8, but we could run JDK 17 with the -release option.

tgodzik: Looks like we would have drop JDK 17 altogether for sbt plugin. Maybe one of the newer 5 versions also support worktrees?

matthughes:
Not sure I follow.  7.0.0 is what introduced supposed for worktrees.
Release notes show it requires jdk17.  Why do you say we’d have to drop 17
altogether?

tgodzik:
Ach sorry, drop 8 to 11. We would have to switch to 17

matthughes:
Obviously your call but I’d be surprised if that change affected many
people.  It’s easy enough  bump your jdk for your ci.

tgodzik:
@bjaglin what do you think, is it the time to drop JDKs pre 17?

bjaglin:
> @bjaglin what do you think, is it the time to drop JDKs pre 17?

I was hoping to coordinate the next minor / breaking scalafix/sbt-scalafix release with https://github.com/scalacenter/scalafix/issues/2041, so I'd rather wait a little bit.

However, what we can do without breaking compatibility is bump jgit to 7.x only for the scalafix-cli 3.8.x which already targets JDK17 (since it requires the presentation compiler that itself targets JDK17). @matthughes would that be a good first step / workaround for your use-case?

project/Dependencies.scala
bjaglin, line 25: independently of the current discussion, please remove https://github.com/scalacenter/scalafix/blob/dda28c73dabac74928127977e94862132e7ec1b7/.scala-steward.conf#L2-L5 to keep upgrading that in the future

bjaglin:
I'll tag a release in a couple of weeks when Scala 3.8.4 comes out, until then we can discuss if it's worth breaking bytecode compatibility to improve compatibility for a single command.

I know worktrees are common now with agents, so it might deserve a minor, breaking-signaling release. @matthughes is that your use-case? sbt-scalafix has built-in caching so such invocations are not very useful.

matthughes:
Yes I encountered this as I started playing around with git worktree.  Not
having the agents create them but I tend to juggle more PRs than usual and
have found them useful.

PR #2403, opened 2026-04-13
Don't use accidentally public Tokens fields

kitbellew:
- PresentationCompilerTypeInferrer: iterated over all tokens, not just those referred to by `ctx.tokens`

### DIFF
--- a/scalafix-rules/src/main/scala-3/pc/PresentationCompilerTypeInferrer.scala
+++ b/scalafix-rules/src/main/scala-3/pc/PresentationCompilerTypeInferrer.scala
@@ -10,8 +10,7 @@ import scala.meta.*
 import scala.meta.inputs.Input.File
 import scala.meta.inputs.Input.VirtualFile
 import scala.meta.pc.PresentationCompiler
-import scala.meta.trees.Origin.DialectOnly
-import scala.meta.trees.Origin.Parsed
+import scala.meta.trees.Origin
 
 import scalafix.internal.v1.LazyValue
 import scalafix.patch.Patch
@@ -44,9 +43,7 @@ final class PresentationCompilerTypeInferrer private (
       pc <- pc.value
     } yield {
       ctx.tree.origin match {
-        case _: DialectOnly => empty
-        case scala.meta.trees.Origin.None => empty
-        case parsed: Parsed =>
+        case parsed: Origin.Parsed =>
           val text = parsed.source.input.text
           val uri = parsed.source.input match {
             // case Ammonite(input) =>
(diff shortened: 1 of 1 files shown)


kitbellew:
@bjaglin PTAL. also, #2402 will need to rebased if this is merged.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -1174,4 +1171,20 @@ object OrganizeImports {
...
   }
+
+  private def tokensWithTrailingTrivia(tree: Tree): Iterator[Token] = {
+    val tokens = tree.tokens
+    val endidx = tokens.length
+    val wsidx = {
+      val eolidx = tokens.skipWideIf(_.is[Token.HTrivia], endidx)
+      val ewsidx = tokens.skipWideIf(_.is[Token.Trivia], eolidx)
+      tokens.getWideOpt(ewsidx).fold(ewsidx) {
+        case _: Token.KwImport | _: Token.KwExport => ewsidx
```
bjaglin: I am having a hard time following the flow - why do these 2 particular tokens deserve a specific handling?

bjaglin (approved):
Thanks! I must say the token traversal is not obvious, but the intent is right and we have a decent coverage of expect tests.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: the way it worked before is, they deleted all tokens between the first token of the first import and the last token of the last import, including any blank lines and comments in between.

to support this behaviour, we delete the import and all horizontal spaces and comments on the same line. then we skip all spaces and comments and check the next token; if it starts another import (hence the two magic tokens), then we delete these intervening tokens as well.

kitbellew: if this is harder to understand, i can add a method in scalameta to join consecutive token spans, but that will require combining the dependency update and the code adjustments into one commit (which i usually like to separate).

bjaglin: Following up on the `KwExport` part specifically: I believe it's not only unnecessary, but potentially harmful.

`collectImports` uses `stats.span(_.is[Import])` to collect global imports as a contiguous leading prefix of `Import` nodes. In Scalameta, Scala 3's `export` is an `Export` tree node (distinct from `Import`), so it breaks the span. This means:

1. **For intermediate imports** (not the last one): `export` can never appear between two imports in the collected list, so `KwExport` is unreachable here.

2. **For the last import**, if it's followed by an `export` statement:
   ```scala
   import a.B
   
   export c.D
   ```
   The `KwExport` match causes `wsidx = ewsidx`, meaning the `\n\n` between the import and the export is included in the removal. That collapses the blank line, giving:
   ```scala
   import a.B
   export c.D
   ```
   Whereas `case _ => eolidx` (what you'd get without the `KwExport` arm) preserves it correctly.

So `KwExport` is unreachable in the intermediate case and wrong in the trailing case. Shouldn't only `KwImport` be in that pattern match?

kitbellew: for now, i have descoped this commit. i will address the other part later, differently.

PR #2404, opened 2026-04-13
TokenListSuite: use parse before tokens

kitbellew:
In scalameta v4.16.0, one must provide an implicit Tokenize instance if not parsing.


kitbellew:
Blocks #2402. @bjaglin PTAL.

PR #2419, opened 2026-04-21
OrganizeImports: refactor sorting importers

kitbellew:
This allows to avoid relying on `importerSyntax` which can later be removed.


kitbellew:
@bjaglin PTAL

PR #2428, opened 2026-04-22
OrganizeImports: preserve comments, too

### LINKED ISSUE #2267: OrganizeImports does not retain inline comments

kitbellew:
For that, we need to access the `.originalPrototype()` tree that was parsed out from input, before we generated partial copies with various combinations of importees and importers (each such copy also creates a copy of the fields with the new parent set).

That original prototype is what can be used to access the parent import statement and its attached comments.

Fixes #2267. Borrowed tests from #2324.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -715,7 +715,59 @@ class OrganizeImports(
   private def prettyPrintImportGroups(
       groups: Seq[ImportGroup]
   ): Seq[(Int, Seq[String])] = {
-    groups.map(g => g.index -> g.imports.map("import " + treeSyntax(_)))
+    groups.map { group =>
+      val res = Seq.newBuilder[String]
+
+      group.imports.foreach { i =>
+        val sb = new StringBuilder
+
+        val single =
+          if (i.importees.lengthCompare(1) == 0) i.importees.head else null
+
+        sb.append("import ").append(treeSyntax(i.ref)).append('.')
+        if (single != null) {
+          val isCurly = single.isCurlyBraced
+          val useOuterSpace = isCurly && single
+            .originalPrototype()
+            .parent
+            .exists(_.hasSpaceInCurly)
+          if (isCurly) {
+            sb.append('{')
+            if (useOuterSpace) sb.append(' ')
+          }
+          sb.append(treeSyntax(single))
+          if (isCurly) {
+            if (useOuterSpace) sb.append(' ')
+            sb.append('}')
+          }
+        } else {
+          val lines = i.importees.iterator.map(_.pos.startLine).filter(_ >= 0)
+          val isMultiline = lines.hasNext && {
+            val line = lines.next()
+            lines.exists(_ != line)
+          }
+          val useOuterSpace = !isMultiline && i.importees
+            .flatMap(_.originalPrototype().parent)
+            .distinct
+            .exists(_.hasSpaceInCurly)
+          sb.append('{')
+          val sep = if (isMultiline) "\n  " else " "
+          if (isMultiline) sb.append(sep)
+          else if (useOuterSpace) sb.append(' ')
+          val sblen = sb.length
+          i.importees.foreach { i2 =>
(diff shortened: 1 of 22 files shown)


kitbellew:
requires a fix in scalameta.

kitbellew:
Blocked on #2443.

bjaglin:
> Blocked on #2443.

is that for https://github.com/scalameta/scalameta/pull/4595 ?

kitbellew:
> > Blocked on #2443.
> 
> is that for [scalameta/scalameta#4595](https://github.com/scalameta/scalameta/pull/4595) ?

yessir!

kitbellew:
@bjaglin this one is finally ready to go.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -715,7 +718,46 @@ class OrganizeImports(
...
+        val sb = new StringBuilder
+        def appendEndComment(pc: Tree.Comments): Unit =
+          if (commentsPrinted.add(pc))
+            pc.values.foreach(x => sb.append(' ').append(x.syntax))
+
+        ps.foreach(_.begComment.foreach(appendBegComment))
+        i.begComment.foreach(appendBegComment)
+        sb.append("import ").append(treeSyntax(i))
+        i.endComment.foreach(appendEndComment)
+        ps.foreach(_.endComment.foreach(appendEndComment))
```
bjaglin, line 801: nit: I am not even sure if this is valid syntax (I am on my phone), nor that the added implementation complexity/overhead is worth it, but I would have naively expected that when `groupedImports = Merge`, comments stick to the importee rather than the importer when there is a single importee
```scala
import scala.collection.mutable.Buffer // keep with Buffer
import scala.collection.mutable.ArrayBuffer // keep with ArrayBuffer
```

```scala
import scala.collection.mutable.{
  ArrayBuffer, // keep with ArrayBuffer
  Buffer // keep with Buffer
}
```
WDYT? it doesn't look tested

kitbellew: we can add more tests. 

the example you gave is indeed one where the comment is after the importee (nested level 3).

however, if the comment is before `import `, it's import (level 1); if after `}`, then importer (level 2), and import if this is the last importer.

this code does not really cover all this complexity but could. judging by your comment, you prefer less complexity so weird cases would not be handled. 

let me know what you prefer.

kitbellew: - added tests with embedded comments within importers/around importees
- implemented custom pretty-printing (scalameta-provided is not flexible enough)

bjaglin: Sorry for the silence, I think having test was necessary to guard the behavior.

As for the implementation, it was a bonus and it does look tricky indeed, but I see no reason not to bring it in now that you invested time on it.

bjaglin (approved):

PR #2429, opened 2026-04-22
Fix various warnings (unused, implicits, match)

kitbellew:


kitbellew:
@bjaglin couldn't fix all of them (and some are from `integrations` which could be intentional).

bjaglin (approved):
Thanks, was your goal to enable -XFatal-warnings to remove warts?

There is a potentially more impactful journey I barely started long time ago with https://github.com/scalacenter/scalafix/pull/1744, in order to identify feature gaps as the scalameta trees evolved. I will probably throw an agent at it some time this year.

bjaglin:
> Thanks, was your goal to enable -XFatal-warnings to remove warts?

sorry I missed the Wconf

kitbellew:
> Thanks, was your goal to enable -XFatal-warnings to remove warts?

that would be nice. i had a simpler, purely "egotistical" motivation: while working on the OrganizeImports problem, i would run tests locally and then would have to scroll through lots and lots of warnings, to find the needle (error) in that haystack. so i decided to clean it up a little :)

PR #2432, opened 2026-04-24
TestRegistration: deprecated, AnyFunSuite instead

kitbellew:


kitbellew:
@bjaglin PTAL

scalafix-testkit/src/main/scala/scalafix/testkit/AbstractSemanticRuleSuite.scala
bjaglin: to be updated

scalafix-testkit/src/main/scala/scalafix/testkit/AbstractSyntacticRuleSuite.scala
bjaglin: to be updated

bjaglin (commented):
As far as I can see, this might break source compatibility for clients using a different testing style, but I think that's acceptable to do that even in the next release which likely won't need to bump the minor version 👍 

- A [few usages of AnyFunSuiteLike](https://github.com/search?q=repo%3Ascalacenter%2Fscalafix%20AnyFunSuiteLike&type=code) can probably be simplified.
- https://github.com/scalacenter/scalafix-named-literal-arguments/blob/ba6604ad44a458f8840eabeea211a0642afa65f3/scalafix/tests/src/test/scala/fix/RuleSuite.scala#L6-L8 (indirectly referenced from https://github.com/scalacenter/scalafix/blob/main/docs/developers/tutorial.md) could also be simplified, I am taking notes to take care of that cumbersome task after cutting the release containing this PR (seizing the opportunity to rewrite the history to target the latest versions of everything from the very first commit).

kitbellew:
all mentions of `AnyFunSuiteLike` removed. can't reply to comments inline as they are no longer shown on the rebased patch.

bjaglin (approved):

PR #2433, opened 2026-04-24
CompilerTypePrinter: rewrite writing Defn.Def

kitbellew:
Supports multiple parameter clauses -- and none as well. Also avoids using deprecated methods when accessing params.


kitbellew:
@bjaglin PTAL

bjaglin (approved):

PR #2437, opened 2026-04-26
Mask a few additional warnings, including tests

kitbellew:

PR #2438, opened 2026-04-26
SBT: add -Werror and -deprecation, with exceptions

kitbellew:
Scalafix apparently depends on warnings to operate (although it would be ideal to separate generation of diagnostics from printing them out), so these stricter options cannot be applied to input, output, integration.

Depends on #2437, #2436, #2435, #2434, #2433, #2432, #2431, #2430.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,7 @@ lazy val interfaces = project
 // Scala 3 macros vendored separately (i.e. without runtime classes), to
 // shadow Scala 2.13 macros in the Scala 3 compiler classpath, while producing
 // code valid against Scala 2.13 bytecode
+// TODO: after replacing with metaconfig_3, remove -Wconf filter in ScalafixBuild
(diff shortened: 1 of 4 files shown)


kitbellew:
@bjaglin PTAL. added one last deprecation fix for URL ctor.

build.sbt
```diff
@@ -319,6 +322,7 @@ lazy val integration = projectMatrix
   .settings(
     noPublishAndNoMima,
     Test / parallelExecution := false,
+    scalacOptions --= ScalafixBuild.werrorOptions,
```
bjaglin: not sure if that would work, but could we try to keep linting the test code? 
```suggestion
    Compile / scalacOptions --= ScalafixBuild.werrorOptions,
```

bjaglin (approved):
Great work!

build.sbt
kitbellew: done

kitbellew: it did work.

PR #2440, opened 2026-04-29
SBT: build core and reflect for scala3

kitbellew:
Use scala3 versions of metaconfig, scalameta, semanticdbShared.


bjaglin:
I know this is still in draft, but I had a quick look as the title caught my attention.

I believe we shouldn't start building cli3 against scalameta3 as it will prevent loading all community rules that are built against rules2.13/scalameta2. My thinking so far about what we should migrate cli3 and expose a proper rules3 smoothly is laid out in https://github.com/scalacenter/scalafix/issues/2041. There is an early draft of the first step in https://github.com/scalacenter/scalafix/pull/2311 - I believe your PR could actually supersedes that one.

kitbellew:
> I know this is still in draft, but I had a quick look as the title caught my attention.
> 
> I believe we shouldn't start building cli3 against scalameta3 as it will prevent loading all community rules that are built against rules2.13/scalameta2. My thinking so far about what we should migrate cli3 and expose a proper rules3 smoothly is laid out in #2041. There is an early draft of the first step in #2311 - I believe your PR could actually supersedes that one.

I was just trying to figure out what's needed to build this using scala3, and one of the changes was to pubilsh semanticdbShared in scalameta for scala3.

but i got stuck with the build failures here anyway, couldn't really figure out how to fix them...

PR #2442, opened 2026-05-07
build and test against scala 3.3.8

bjaglin:


project/ScalafixBuild.scala
```diff
@@ -87,10 +87,7 @@ object ScalafixBuild extends AutoPlugin with GhpagesKeys {
       if (isScala3.value && minor >= 7)
         Seq("-Wunused:all")
       else if (isScala3.value)
-        Seq(
-          "-Wunused:all",
-          "-Wunused:unsafe-warn-patvars"
```
bjaglin, line 92: This was useless as it was removed back in [3.3.7](https://github.com/scala/scala3/commit/76e7558421c24ca35d0cb36c4c54e9a679d2f280) / [3.8.0](https://github.com/scala/scala3/commit/2f02456e5d5176a8ea3dd954bef99c3c6b7f4845), and redundant with `all`. It just looks like there is a stricter validation now in 3.3.8 (but interestingly not in 3.8.4 🤷 ), so we must remove it

tgodzik: Might be some mismatch when backporting changes. I can take a look if it's an issue.

tgodzik:
Btw. The fix in https://github.com/scala/scala3/pull/25469 is also ported to 3.3.8

project/ScalafixBuild.scala
bjaglin: That's definitely surprising to see that stricter validation exposing a now-deprecated/useless flag in a LTS patch release, but I don't think it's a big deal.

bjaglin:
### Failure analysis

https://github.com/scala/scala3/pull/25446 was backported in 3.3.8-RC2, making it impossible to load the presentation compiler from other versions <3.8.4 (where it is also present).

This was caught by `FetchScala3CompilerArtifactsOnVersionMismatch` in `expect3_8_3Target3_3_8-RC2` as the rule is a no-op. With additional logs, the problem is clear:

```scala
sbt:scalafix> expect3_8_3Target3_3_8-RC2/test
...
java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:74)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:483)
  | => sat scalafix.internal.pc.Embedded$.serviceLoader(Embedded.scala:44)
        at scalafix.internal.pc.Embedded$.presentationCompiler(Embedded.scala:29)
...
Caused by: java.lang.NoSuchMethodError: 'scala.meta.pc.SourcePathMode scala.meta.pc.PresentationCompilerConfig.sourcePathMode()'
        at dotty.tools.pc.ScalaPresentationCompiler.<init>(ScalaPresentationCompiler.scala:151)
        at dotty.tools.pc.ScalaPresentationCompiler.<init>(ScalaPresentationCompiler.scala:73)
        at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
```

### Solution

There is not much we can do as there is no forward-compatibility guarantee on the PC. Scalafix should still be verbose and explicit about the error, rather than silently ignoring it, so I'll add a commit in this PR so that future occurences are more clear.

### Impact
- 3.8.4 should come out before 3.3.8, so the failure in this branch will disappear as soon as it is rebased against https://github.com/scalacenter/scalafix/pull/2398, unblocking the merge of this PR
- ExplicitResultTypes users with current scalafix versions will experience the same problem if they run against code compiled with the upcoming 3.3.8, so they will need to upgrade to the upcoming scalafix-cli_3.8.4 (or ideally use scalafix-cli_3.3.8 when it's out to avoid dynamic loading all together).

tgodzik:
Could we not update the mtags-interface version to the latest? That should solve the problem. We can force a newer version when fetching

bjaglin:
> Could we not update the mtags-interface version to the latest? That should solve the problem. We can force a newer version when fetching

We might, but is it really worth it?

tgodzik:
> > Could we not update the mtags-interface version to the latest? That should solve the problem. We can force a newer version when fetching
> 
> We might, but is it really worth it?

Maybe not, yeah

scalafix-rules/src/main/scala-3/pc/PresentationCompilerTypeInferrer.scala
```diff
@@ -18,13 +18,11 @@ import scalafix.patch.Patch.empty
...
-    pc: LazyValue[Option[PresentationCompiler]]
+    pc: LazyValue[PresentationCompiler]
 ) {
 
   def shutdownCompiler(): Unit = {
-    pc.value.foreach {
-      _.shutdown()
-    }
+    pc.foreach(_.shutdown())
   }
```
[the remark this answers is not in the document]
bjaglin: not a new problem

scalafix-tests/integration/src/test/scala-3/scalafix/tests/rules/ExplicitResultTypesSuite.scala
```diff
@@ -58,7 +58,7 @@ class ExplicitResultTypesSuite extends AnyFunSuite {
     "fetchScala3CompilerArtifactsOnVersionMismatch unlocks dynamic loading"
   ) {
     val neitherLTSNorNext =
-      config.withScalaVersion("3.5.0")
+      config.withScalaVersion("3.6.4")
```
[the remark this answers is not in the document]
bjaglin: this remains tested on most minor so not worth changing

PR #2443, opened 2026-05-07
Dependencies: scalameta 4.17.0, metaconfig 0.18.6

kitbellew:
Closes #2444.


kitbellew:
@bjaglin PTAL. doesn't use scala2-specific traversers.

bjaglin:
Changes LGTM per-se, but I wonder what's your end goal here. I imagine this is a step to make https://github.com/scalameta/scalameta/issues/4146 irrelevant in the move towards scalameta3? See related discussion in https://github.com/scalacenter/scalafix/issues/2041#issuecomment-3488134626.

As a follow-up to this PR, we could invite all current community rules users to move towards these new scalameta3-ready traversal helpers (providing a Scalafix rule [here](https://github.com/scalacenter/scalafix.g8/tree/main/migration-rules) to have Scala Steward do it automatically). But it's a shame that we force that change on an API that hadn't moved for 6+ years, since the early stable releases of Scalafix. I also wonder if/how we could deprecate the Scalameta2 traversers if we don't want to maintain them and port them to Scala 3?

I am not sure if that's possible, but the best would be to achieve binary and source compatibility without relying on the scalameta traversers, through some shading maybe?

kitbellew:
hmm... i should probably move these methods into scalameta, that would be a more central place.

since implementing this code, i also moved out Traverser/Transformer into a scala2-only sub-project, while SimpleTraverser is now built with scala3, so perhaps it's even less critical.

ok, move back to draft for now.

kitbellew:
> hmm... i should probably move these methods into scalameta, that would be a more central place.

moved to scalameta. instead, this pr is upgrading to new version of scalameta, with a minor optional change to take advantage of that.

bjaglin (approved):

bjaglin:
FTR, particularly relevant for next PRs towards https://github.com/scalacenter/scalafix/issues/2041: https://github.com/scalameta/scalameta/pull/4599

Issue #2054, opened 2024-08-25
Partial semanticdb information for sources compiled with Scala 3


bjaglin:
This is just a placeholder for visibility for rule authors , to track the most important tickets upstream that will bring parity between Scala 2 and Scala 3 (or full visibility for features introduced in Scala 3).
- https://github.com/scala/scala3/issues/12963
- https://github.com/scala/scala3/issues/12766
- https://github.com/scala/scala3/issues/13291
- https://github.com/scala/scala3/issues/14828
- Synthetics
  - https://github.com/scala/scala3/issues/22934
  - https://github.com/scala/scala3/issues/22935

PR #2449, opened 2026-05-25
OrganizeImports: optionally move `given` imports to end

### LINKED ISSUE #2255: Organizing imports with ...{given Decoder[X]} fails
Given the following imports:

```
import io.circe.Json
import io.circe.parser.*
import io.circe.syntax.*
import io.circe.{Decoder, Encoder}
import org.scalatest.flatspec.AnyFlatSpec
import chimp.mcp.JSONRPCMessage.{given Decoder[JSONRPCMessage], given Encoder[JSONRPCMessage]}
import org.scalatest.matchers.should.Matchers
```

Organizing them yields:

```
import chimp.mcp.JSONRPCMessage.given Decoder[JSONRPCMessage]
import chimp.mcp.JSONRPCMessage.given Encoder[JSONRPCMessage]
import io.circe.Decoder
import io.circe.Encoder
import io.circe.Json
import io.circe.parser.*
import io.circe.syntax.*
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers
```

Which is incorrect, since Decoder is imported after the given.

This has been observed in Metals 1.5.3

kitbellew:
Fixes #2255.

### DIFF
--- a/docs/rules/OrganizeImports.md
+++ b/docs/rules/OrganizeImports.md
@@ -58,7 +58,7 @@ Known limitations:
     may result in incorrect imports.
 
 2.  The
-    [`groupExplicitlyImportedImplicitsSeparately`](OrganizeImports.md#groupexplicitlyimportedimplicitsseparately)
+    [`groupSeparately=ByNameImplicits`](#bynameimplicits)
     option has no effect.
 
 Configuration
@@ -374,130 +374,6 @@ import scala.util.control.NonFatal
 import sun.misc.BASE64Encoder
 ```
 
-`groupExplicitlyImportedImplicitsSeparately`
---------------------------------------------
-
-This option provides a workaround to a subtle and rarely seen
-correctness issue related to explicitly imported implicit names.
-
-The following snippet helps illustrate the problem:
-
-```scala
-package a
-
-import c._
-import b.i
-
-object b { implicit def i: Int = 1 }
-object c { implicit def i: Int = 2 }
-
-object Imports {
-  def f()(implicit i: Int) = println(1)
-  def main() = f()
-}
-```
-
-The above snippet compiles successfully and outputs `1`, because the
-explicitly imported implicit value `b.i` overrides `c.i`, which is made
-available via a wildcard import. However, if we reorder the two imports
-into:
-
-```scala
-import b.i
-import c._
-```
-
-The Scala compiler starts complaining:
-
-```
-error: could not find implicit value for parameter i: Int
-  def main() = f()
-                ^
-```
-
-This behavior could be due to a Scala compiler bug since [the Scala
-language
-specification](https://scala-lang.org/files/archive/spec/2.13/02-identifiers-names-and-scopes.html)
-requires that explicitly imported names should have higher precedence
-than names made available via a wildcard.
-
-Unfortunately, Scalafix is not able to surgically identify conflicting
-implicit values behind a wildcard import. In order to guarantee
-correctness in all cases, when the
-`groupExplicitlyImportedImplicitsSeparately` option is set to `true`,
-all explicitly imported implicit names are moved into the trailing
-order-preserving import group together with relative imports, if any
-(see the [trailing order-preserving import
-group](OrganizeImports.md#groups) section for more
-details).
-
-> In general, order-sensitive imports are fragile, and can easily be
-> broken by either human collaborators or tools (e.g., the IntelliJ IDEA
-> Scala import optimizer does not handle this case correctly). They should
-> be eliminated whenever possible. This option is mostly useful when you
-> are dealing with a large trunk of legacy codebase, and you want to
-> minimize manual intervention and guarantee correctness in all cases.
-
-> The `groupExplicitlyImportedImplicitsSeparately` option has currently no
-> effect on source files compiled with Scala 3, as the [compiler does not
-> expose full signature
-> information](https://github.com/scala/scala3/issues/12766), preventing
-> the rule to identify imported implicits.
-
-### Value type
-
-Boolean
-
-### Default value
-
-`false`
-
-Rationale:
-
-1.  Although setting it to `true` avoids the aforementioned correctness
-    issue, the result is unintuitive and confusing for many users since
-    it looks like the `groups` option is not respected.
-
-    E.g., why my `scala.concurrent.ExecutionContext.Implicits.global`
-    import is moved to a separate group even if I have a `scala.` group
-    defined in the `groups` option?
-
-2.  The concerned correctness issue is rarely seen in real life. When it
-    really happens, it is usually a sign of bad coding style, and you
-    may want to tweak your imports to eliminate the root cause.
-
-### Examples
-
-```conf
-OrganizeImports {
-  groups = ["scala.", "*"]
-  groupExplicitlyImportedImplicitsSeparately = true // not supported in Scala 3
-}
-```
-
-Before:
-
-```scala
-import org.apache.spark.SparkContext
-import org.apache.spark.RDD
-import scala.collection.mutable.ArrayBuffer
-import scala.collection.mutable.Buffer
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.sys.process.stringToProcess
-```
-
-After:
-```scala
-import scala.collection.mutable.ArrayBuffer
-import scala.collection.mutable.Buffer
-
-import org.apache.spark.RDD
-import org.apache.spark.SparkContext
-
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.sys.process.stringToProcess
-```
-
 `groupedImports`
 ----------------
 
@@ -755,8 +631,8 @@ same group and sorted by the order defined by the
 > 1.  The `expandRelative` option is set to `false` and there are relative
 >     imports.
 > 
-> 2.  The `groupExplicitlyImportedImplicitsSeparately` option is set to
->     `true` and there are implicit names explicitly imported.
+> 2.  The `groupSeparately` parameter is set, and there are
+>     [matching imports](#groupseparately).
 > 
 > This special import group is necessary because the above two kinds of
 > imports are order sensitive:
@@ -770,9 +646,9 @@ same group and sorted by the order defined by the
 > import util.control
 > import control.NonFatal
 > ```
-> #### Explicitly imported implicit names  
+> #### Explicitly imported implicit names
 > Please refer to the
-> [`groupExplicitlyImportedImplicitsSeparately`](OrganizeImports.md#groupexplicitlyimportedimplicitsseparately)
+> [`groupSeparately=ByNameImplicits`](#bynameimplicits)
 > option for more details.
 
 ### Value type
@@ -930,7 +806,7 @@ import control.NonFatal
 ```conf
 OrganizeImports {
   groups = ["re:javax?\\.", "scala.", "*"]
-  groupExplicitlyImportedImplicitsSeparately = true
+  groupSeparately = [ByNameImplicits]
 }
 ```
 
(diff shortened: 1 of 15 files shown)


kitbellew:
@bjaglin ptal

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
```diff
@@ -0,0 +1,15 @@
+/*
+rules = [OrganizeImports]
+OrganizeImports {
+  groupedImports = Keep
+  removeUnused = false
+  targetDialect = Scala3
```
bjaglin: nit: that is implicit since the harness (whether the testkit here or sbt-scalafix in real life) injects the scala version which is then used to infer the dialect
```suggestion
```

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
```diff
@@ -0,0 +1,15 @@
...
+}
+ */
+package test.organizeImports
+
+import GivensGeneric.AA
+import test.organizeImports.Givens.A
+import test.organizeImports.Givens.{given AA[A], given A}
+
+object RelativeGenericGivens
```
bjaglin, line 16: although tricky and thus good to test, the "relative" part seems to be a corner case - the reported issue was with fully qualified references, so I think we should have another test where the given would end up _before_ the generic it references (because of import order) without the fix

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
```diff
@@ -0,0 +1,15 @@
+/*
+rules = [OrganizeImports]
+OrganizeImports {
+  groupedImports = Keep
```
bjaglin: is this needed to reproduce the problem? it looks orthogonal
```suggestion
```

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -225,11 +225,14 @@ class OrganizeImports(
...
     val noImplicitImporters = Seq.newBuilder[Importer]
-    val separateImplicits = config.groupExplicitlyImportedImplicitsSeparately
+    val separateImplicits = config.groupExplicitlyImportedImplicitsSeparately ||
+      targetDialect.allowGivenUsing
     if (separateImplicits) importers.foreach { importer =>
       val (implicits, noImplicits) = importer.importees.partition {
         case i: Importee.Name =>
+          config.groupExplicitlyImportedImplicitsSeparately &&
           i.symbol.infoNoThrow.exists(_.isImplicit)
+        case i: Importee.Given => i.tpe.is[Type.Apply]
```
bjaglin: can you document this new scenario in https://github.com/scalacenter/scalafix/blob/main/docs/rules/OrganizeImports.md?plain=1#L757-L758 ?

bjaglin (commented):
Thanks! Implementation LGTM, a few comments on docs & testing.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: will do

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
kitbellew: the default for `groupedImports` is `Explode`. i used `Keep` (or could have used `Merge`) to request that the grouped imports below on line 13 should be preserved, but they will nevertheless be separated since there's an overriding reason.

i can remove it as the difference will be a blank line between the groups, but this looked a bit more emphatic (and in fact i might even use a more aggressive setting, just to be clear).

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
kitbellew: done

scalafix-tests/input/src/main/scala-3/test/organizeImports/RelativeGenericGivens.scala
kitbellew: by "relative" i mean that `AA` used in `given AA[A]` is relative (instead of fully-qualified).

the "trick" with using relative `GivensGeneric.AA` was so that it's "sorted" after the given. let me vary the cases a bit.

kitbellew:
i think this is the wrong approach :( `given` can import absolutely any type, even if it's not `A[B]`, so correct logic needs to figure out the right fully-qualified type of what it is importing. `.symbol` on these imported types returns empty; does that mean semanticdb is incomplete in scala3?

bjaglin:
> does that mean semanticdb is incomplete in scala3?

Yes - https://github.com/scalacenter/scalafix/issues/2054

> i think this is the wrong approach :( given can import absolutely any type, even if it's not A[B], so correct logic needs to figure out the right fully-qualified type of what it is importing.

From my understanding, the current logic in this PR might bring false positives (imports sorted last for "nothing"), but it's still better than generating invalid code, no?

albertikm:
> > does that mean semanticdb is incomplete in scala3?
> 
> Yes - #2054
> 
> > i think this is the wrong approach :( given can import absolutely any type, even if it's not A[B], so correct logic needs to figure out the right fully-qualified type of what it is importing.
> 
> From my understanding, the current logic in this PR might bring false positives (imports sorted last for "nothing"), but it's still better than generating invalid code, no?

the problem is that in reality, short of putting all givens in that separate section, we can't really fix anything that is likely to occur.

the given declaration that we're importing can be of any type (including simple Type.Name), and that type can be coming from any package(s); I thought I could access `.symbol` on all those types, to figure out where to sort the given, but symbols are empty for some reason.

perhaps you could briefly explain how symbols or occurrences can work, and what the current limitations are? I'm also not sure where scala3 semantic information is coming from, and whether it is compatible or vastly different from the scalameta version.

i can send a patch of what i experimented with, to see if it can be improved somehow.

kitbellew:
oops... again used other profile above.

kitbellew:
> > does that mean semanticdb is incomplete in scala3?
> 
> Yes - #2054

Oy vey. And imports are not mentioned there, perhaps they are covered by one of them.

> From my understanding, the current logic in this PR might bring false positives (imports sorted last for "nothing"), but it's still better than generating invalid code, no?

In that case, I would suggest the following: if `groupExplicitlyImportedImplicitsSeparately` is set, move all givens to the end, together with implicits. (Or we could *always* do it.)

But I don't like the logic I have proposed so far, it solves one particular rare case, but doesn't solve other problematic cases which are, in my opinion, even more likely.

kitbellew:
added another option to `groupSeparately` parameter (which was renamed), and moved all givens if the new option is set. i don't see a better solution until semanticdb can help.

docs/rules/OrganizeImports.md
```diff
@@ -437,12 +450,23 @@ details).
...
+
+If this value is specified, any given imports will be put in that
+final catch-all group.
+
+The reason is that a `given` import refers to a type `T`, and that type can
+come from any package `P` and must be imported earlier
+- however, our grouping and sorting logic could put `import P.T` later
+- at the moment, scala 3 semanticdb doesn't provide symbol information for `T`
+
 ### Value type
```
bjaglin: this and the lines below look obsolete - I think we should follow the structure of https://github.com/scalacenter/scalafix/blob/63be364161e828ac5b2bf0e32d9a52255811a2a1/docs/rules/OrganizeImports.md?plain=1#L777-L813 to document that new flag that can take an array (even though we obviously only have a limited set of possible values here)

bjaglin: Also, I think the new feature deserves a new example

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -227,11 +227,15 @@ class OrganizeImports(
...
+    val separateImplicits = config.groupSeparately
+      .contains(GroupSeparately.ExplicitlyImportedImplicits)
+    val separateGivens = targetDialect.allowGivenUsing &&
+      config.groupSeparately.contains(GroupSeparately.Givens)
+    if (separateImplicits || separateGivens) importers.foreach { importer =>
       val (implicits, noImplicits) = importer.importees.partition {
         case i: Importee.Name =>
-          i.symbol.infoNoThrow.exists(_.isImplicit)
+          separateImplicits && i.symbol.infoNoThrow.exists(_.isImplicit)
+        case _: Importee.Given | _: Importee.GivenAll => separateGivens
```
bjaglin: Do we have any problem to address with `GivenAll`? My understanding is that we don't have any information about Y in `import X.given Y`, but we do have information about X to sort `import X.given` properly.
```suggestion
        case _: Importee.Given => separateGivens
```
In any case, we are lacking a test for `GivenAll`.

docs/rules/OrganizeImports.md
```diff
@@ -437,12 +450,23 @@ details).
...
 > minimize manual intervention and guarantee correctness in all cases.
 
-> The `groupExplicitlyImportedImplicitsSeparately` option has currently no
+> This option has currently no
 > effect on source files compiled with Scala 3, as the [compiler does not
 > expose full signature
 > information](https://github.com/scala/scala3/issues/12766), preventing
 > the rule to identify imported implicits.
 
+`groupSeparately`: `Givens`
```
bjaglin: - I think we can clarify that this only impacts [by-type givens](https://docs.scala-lang.org/scala3/reference/contextual/given-imports.html#importing-by-type) (and document the preserved behavior of by-name givens with tests)
- Even though it's obvious, this doc is being very verbose about what's Scala 3 or not, so we might just add a note
```suggestion
`groupSeparately`: ` ByTypeGivens` (Scala 3)
```

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImportsConfig.scala
```diff
@@ -58,11 +58,20 @@ object TargetDialect {
...
+  implicit val codec: ConfCodecEx[GroupSeparately] = OrganizeImportsConfig
+    .getCodecFrom(ExplicitlyImportedImplicits, Givens)
+}
+
 final case class OrganizeImportsConfig(
     blankLines: BlankLines = BlankLines.Auto,
     coalesceToWildcardImportThreshold: Option[Int] = None,
     expandRelative: Boolean = false,
-    groupExplicitlyImportedImplicitsSeparately: Boolean = false,
+    groupSeparately: Seq[GroupSeparately] = Nil,
```
bjaglin: I think we should make this the default so we guarantee correctness for whoever uses by-type givens. This is a breaking change for the lucky ones that have been using scalafix and had the proper ordering by chance without a special group but I think it's a tiny price to pay.
```suggestion
    groupSeparately: Seq[GroupSeparately] = Seq(GroupSeparately.ByTypeGivens),
```

docs/rules/OrganizeImports.md
```diff
@@ -437,12 +450,23 @@ details).
...
+--------------------------------------------
+
+If this value is specified, any given imports will be put in that
+final catch-all group.
+
+The reason is that a `given` import refers to a type `T`, and that type can
+come from any package `P` and must be imported earlier
+- however, our grouping and sorting logic could put `import P.T` later
+- at the moment, scala 3 semanticdb doesn't provide symbol information for `T`
```
bjaglin: I went through https://github.com/scalacenter/scalafix/issues/2054 & the Scala 3 upstream issues, and I couldn't even find a ticket tracking that gap. I think it's worth creating one and linking it from the docs.

bjaglin (commented):
The metaconfig foo is pretty cool! And the implementation is now very clear about what's going on with that special group.

Yet, if we confirm the fix is only needed for the niche use-case of by-type givens (i.e. not by-names nor wildcard ones), and make it the default for correctness, I wonder if there is really a need for a flag to opt-out?

docs/rules/OrganizeImports.md
kitbellew: what are by-name givens?

docs/rules/OrganizeImports.md
kitbellew: submitted scala/scala3#26255, could you please add it to #2054?

docs/rules/OrganizeImports.md
kitbellew: not sure if i fully understood what you meant, made a small change (mentioned the default and used `####` prefix on values).

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: GivenAll is not needed, perhaps; must have confused with wildcard import which cannot be separated from Unimport/Rename. On the other hand, if we have GivenAll, why would we need explicit givens?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImportsConfig.scala
kitbellew: you da boss :)

kitbellew: this default exposed a possibly overlooked earlier problem, at least one which would otherwise result in massive changes to existing tests. see #2456, need to merge it first, before rebasing this one.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
bjaglin: > On the other hand, if we have GivenAll, why would we need explicit givens?

for a given package, we don't - I was just pointing out that GivenAll don't need special treatment since they don't have a RHS where information is missing.

docs/rules/OrganizeImports.md
bjaglin: looking at https://docs.scala-lang.org/scala3/reference/contextual/given-imports.html, it's a regular import, referencing a member with the given keyword

bjaglin: actually, it's precisely what ExplicitlyImportedImplicit is for, although in Scala 3 we can't distinguish them

kitbellew:
> ... I wonder if there is really a need for a flag to opt-out?

if we ever get the fix for semanticdb occurrences, we'll be able to sort those givens (and move to catch-all group only those whose dependency is not sorted before them). i don't know if it's ok just to add that new sorting logic and shuffle imports automatically, without a user being able to say "i am ok with them unsorted as before".

but let me know if you think it's ok to omit this. scalafmt used to be very opinionated, but now the opinionated ones are the users :)

docs/rules/OrganizeImports.md
bjaglin: sorry, I wasn't clear - I just added a7d6810f27fa81618fc2c04e8584b93cb2995489 to try to improve consistency with the rest of the doc

docs/rules/OrganizeImports.md
bjaglin: now that makes me wonder if `ExplicitlyImportedImplicits` shouldn't be called `ByNameImplicits` for symmetry with `ByTypeGivens`

kitbellew: actually, a very sane thought. very much aligned with my brand of OCD :)

bjaglin:
> if we ever get the fix for semanticdb occurrences, we'll be able to sort those givens (and move to catch-all group only those whose dependency is not sorted before them). i don't know if it's ok just to add that new sorting logic and shuffle imports automatically, without a user being able to say "i am ok with them unsorted as before".

Good point. Let's keep it 👍

docs/rules/OrganizeImports.md
albertikm: did you want me to rename this?

bjaglin: i think it would make understanding easier, I can do it before merging if you prefer

kitbellew: give me a minute to send in a new revision

kitbellew: all done

bjaglin (approved):
Thanks!

PR #2450, opened 2026-05-26
OrganizeImports: expand relative implicits as well

kitbellew:
Helps with #2255.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -127,9 +127,14 @@ class OrganizeImports(
     // See https://github.com/liancheng/scalafix-organize-imports/issues/30
     // for why implicits require special handling.
     val orderPreservingGroup = {
-      val relatives = if (config.expandRelative) Nil else relativeImporters
+      val importers =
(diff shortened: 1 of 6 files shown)


kitbellew:
@bjaglin PTAL

scalafix-tests/output/src/main/scala-3/test/organizeImports/ExpandRelativeImplicits.scala
```diff
@@ -0,0 +1,6 @@
+package test.organizeImports
+
+import test.organizeImports.Implicits.a
+import test.organizeImports.Implicits.a.intImplicit
+
+object ExpandRelativeImplicits
```
bjaglin: nit: the output is the same for scala 2 and scala 3 so we could have only one file (unlike when groupExplicitlyImportedImplicitsSeparately=true since it's not properly supported on Scala 3)

bjaglin (approved):

scalafix-tests/output/src/main/scala-3/test/organizeImports/ExpandRelativeImplicits.scala
kitbellew: done

bjaglin (approved):

PR #2451, opened 2026-05-26
OrganizeImports: refactor partitionImplicits

kitbellew:
- don't "explode" them automatically, ignoring `groupedImports`
- use one pass, don't use positions
Helps with #2255.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -102,13 +102,10 @@ class OrganizeImports(
   )(
       imports: Seq[Import]
   )(implicit doc: SemanticDocument): Patch = {
-    val noUnused = imports flatMap (_.importers) flatMap (
-      removeUnusedImporters(unusedImporteePositions)(_).toSeq
-    )
-
-    val (implicits, noImplicits) =
-      if (!config.groupExplicitlyImportedImplicitsSeparately) (Nil, noUnused)
-      else partitionImplicits(noUnused)
+    val noUnusedIterator = imports.iterator
+      .flatMap(_.importers)
+      .flatMap(removeUnusedImporters(unusedImporteePositions))
+    val (implicits, noImplicits) = partitionImplicits(noUnusedIterator)
 
     val (fullyQualifiedImporters, relativeImporters) =
       noImplicits partition isFullyQualified(diagnostics)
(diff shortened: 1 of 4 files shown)


kitbellew:
@bjaglin PTAL

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -219,23 +216,29 @@ class OrganizeImports(
...
-      case importer @ Importer(_, importees) =>
-        importees collect {
-          case i: Importee.Name if i.symbol.infoNoThrow exists (_.isImplicit) =>
-            importer.copy(importees = i :: Nil) -> i.pos
+    val excludeBuilder = Seq.newBuilder[Importer]
+    val includeBuilder = Seq.newBuilder[Importer]
+    importers.foreach { importer =>
+      val excludeImporteesBuilder = List.newBuilder[Importee]
+      importer
+        .filterImportees { i =>
```
bjaglin: Usage of that filter method to partition importers as-is (not sure that behavior is tested btw) or split them makes it a bit hard to reason IMHO, especially with the exclude/include prefix of the builders. WDYT of something like this 
```scala
    val separatedBuilder = Seq.newBuilder[Importer]
    val otherBuilder = Seq.newBuilder[Importer]
    importers.foreach { importer =>
      val (separatedImportees, otherImportees) = importer.importees.partition {
        case i: Importee.Name =>
          config.groupExplicitlyImportedImplicitsSeparately &&
          i.symbol.infoNoThrow.exists(_.isImplicit)
        case _ => false
      }

      (separatedImportees, otherImportees) match {
        case (Nil, Nil) =>
        case (Nil, _) => otherBuilder += importer
        case (_, Nil) => separatedBuilder += importer
        case (separatedImportees, otherImportees) =>
          separatedBuilder += importer.copy(importees = separatedImportees)
          otherBuilder += importer.copy(importees = otherImportees)
      }
    }
    (separatedBuilder.result(), otherBuilder.result())
```

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -102,13 +102,10 @@ class OrganizeImports(
   )(
       imports: Seq[Import]
   )(implicit doc: SemanticDocument): Patch = {
-    val noUnused = imports flatMap (_.importers) flatMap (
-      removeUnusedImporters(unusedImporteePositions)(_).toSeq
-    )
-
-    val (implicits, noImplicits) =
-      if (!config.groupExplicitlyImportedImplicitsSeparately) (Nil, noUnused)
```
bjaglin, line 110: since groupExplicitlyImportedImplicitsSeparately is false by default and probably rarely used, I wonder if it's not worth keeping an early return in partitionImplicits to avoid the loop on importees? maybe it's negligible because we need to turn the iterator into a concrete Seq anyway...

bjaglin (approved):
A perf question and readability suggestion in case you already gave it some thought (otherwise I am fine merging as-is).

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: there are two reasons:
- i changed the input to an iterator, so we still have to go through it to put into a sequence
- in the next PR (#2449), there will be an additional case to separate the implicits (givens, really) which does not depend on that flag.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: i can either rename some stuff and/or add a comment, or use your proposed version if it's more readable. include/exclude was part of my initial more generic logic which never materialized.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: modified a little, so that it sometimes skips importee partitioning.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: used something very similar to your proposed version.

PR #2453, opened 2026-06-03
OrganizeImports: expand relative, dedup first

kitbellew:
We apply expansion to both implicits and non-implicits, so let's do it before the split. Plus we might later need the fully-qualified importer to detect implicits.

Follow-up to #2450. Blocks #2449.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -102,20 +102,31 @@ class OrganizeImports(
   )(
       imports: Seq[Import]
   )(implicit doc: SemanticDocument): Patch = {
-    val noUnusedIterator = imports.iterator
-      .flatMap(_.importers)
-      .flatMap(removeUnusedImporters(unusedImporteePositions))
-    val (implicits, noImplicits) = partitionImplicits(noUnusedIterator)
+    val (fullyQualifiedImporters, otherImporters) = {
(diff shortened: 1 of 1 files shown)


kitbellew:
@bjaglin ptal

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -102,20 +102,31 @@ class OrganizeImports(
...
-    val (implicits, noImplicits) = partitionImplicits(noUnusedIterator)
+    val (fullyQualifiedImporters, otherImporters) = {
+      val noUnusedIterator = imports.iterator
+        .flatMap(_.importers)
+        .flatMap(removeUnusedImporters(unusedImporteePositions))
 
-    val (fullyQualifiedImporters, relativeImporters) =
-      noImplicits partition isFullyQualified(diagnostics)
+      val otherImporters = new ArrayBuffer[Importer]
+      val fullyQualifiedIterator =
```
bjaglin, line 111: I personnally find the mix of imperative & functional quite hard to grasp (compared to a single foreach like in partition), and quite a departure from the current style of this rule, but I guess it's because you want to keep an iterator for building fullyQualifiedImporters?

bjaglin (approved):

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
kitbellew: yes...

PR #2454, opened 2026-06-03
OrganizeImportsConfig: use Conf{Codec,Decoder}Ex

kitbellew:
Allows handling presets by passing the right initial state, rather than via encoding, merge and decoding.

Don't need to use:
- custom ReaderUtil as ConfCodecEx defines it already
- presets as a separate step as we can transform Conf.

The only nuance is that for scala3 the build uses sourcecode_2.13, so we can't use the macro syntax.

Blocks #2449. This will allow modifying `groupExplicitlyImportedImplicitsSeparately` parameter to allow additional options.


kitbellew:
@bjaglin ptal

bjaglin (approved):

PR #2456, opened 2026-06-08
OrganizeImports: merge or explode catch-all group

kitbellew:
We didn't merge/explode importers, nor coalesce/sort importees, within the final catch-all group. Let's fix that.
Blocks #2449.


kitbellew:
@bjaglin ptal. the failed checks seem transient, if you could retrigger them.

bjaglin (approved):
restores stable, pre-https://github.com/scalacenter/scalafix/pull/2451 behavior

Issue #2049, opened 2024-08-24
Improve semantic information for classpath types compiled by dotty


bjaglin:
The SymbolTable required by core for finding information that is not available in SemanticDB is injected by cli, and is currently backed by scalameta's internal semanticdb-scalac-core which was written for Scala 2 class files.

We need to migrate to https://github.com/scalacenter/tasty-query or equivalent to provide exhaustive information about types compiled by dotty.

tgodzik:
An alternative is to use info method that we added to the presentation compiler. It provides a similar interface as semanticdb.

bjaglin:
Interesting! Do you know what's the PC overhead? In terms of start up time (I guess in Metals it's warm no it does not matter much) and RAM footprint?

PR #2461, opened 2026-06-16
Add `OrganizeImports.expandWildcardImportThreshold` to expand wildcard imports

### LINKED ISSUE #182: turn wildcard imports into qualified imports
not sure how easy that is

jozanek:
## Summary

Closes #182: `OrganizeImports` can expand a wildcard import into the explicit members actually used - the inverse of `coalesceToWildcardImportThreshold`.

```diff
- import com.foo._
+ import com.foo.{A, B}
```

Opt-in, off by default:

```conf
OrganizeImports.expandWildcardImportThreshold = 5
```

Expands only when fewer than N names would result (keep it `<= coalesceToWildcardImportThreshold` so the two don't fight).

## Safety

Conservative by design — the wildcard is left untouched unless expansion is provably safe:

- only **unqualified** uses (and `implicit`/`given` via synthetics) drive it; fully-qualified and ordinary/inherited member calls don't, so it never clashes with a higher-precedence explicit import;
- inherited / package-object members are included; unresolvable scopes, extension-method usages, and `import p.{X => _, _}` are left untouched; `given` wildcards are preserved; Scala 3 `*` is handled like `_`.

Adds testkit fixtures (Scala 2 + 3), a config round-trip test, and docs.


scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,288 @@ class OrganizeImports(
...
+  ): Option[Symbol] =
+    symbol.infoNoThrow.flatMap { info =>
+      info.signature match {
+        case ValueSignature(tpe) => headSymbol(tpe)
+        case MethodSignature(_, _, returnType) => headSymbol(returnType)
+        case TypeSignature(_, _, upperBound) => headSymbol(upperBound)
+        case _: ClassSignature => symbol.asNonEmpty
+        case _ => None
+      }
+    }
```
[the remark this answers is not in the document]
bjaglin: 8ae5eccc

bjaglin:
FTR, I have pushed commits to address the Copilot feedback and make as many tests version-agnostic as possible, ahead of a later deeper review to identify/document gaps.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+          val expanded = owners.iterator
+            .flatMap(usedByOwner.getOrElse(_, Set.empty[Symbol]))
+            .filter(canExpandTo)
+            .map(_.displayName.stripSuffix("_="))
+            // Skip synthetic / compiler-internal names (`<init>`,
+            // `foo$default$1`, `$anonfun`, …) that synthetics may surface but
+            // that are not importable members.
+            .filterNot(name =>
+              name.isEmpty || name.startsWith("<") ||
+                name.contains("$") || alreadyImported(name)
```
bjaglin: **Correctness (low likelihood, high impact):** This filter drops every used symbol whose display name contains `$`, but the wildcard is still replaced whenever another candidate survives. A legal backticked source identifier such as `val `a$b`` can therefore disappear from `import p._` when it is used alongside a normal name, breaking compilation.

Could we treat any filtered synthetic candidate (except `alreadyImported`) as a reason to leave the importer unchanged, or otherwise distinguish compiler-generated names from source names? Please add the `$` case to `ExpandWildcardQuotedIdent`.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+            )
+            .toList
+            .distinct
+            .sorted
+          if (expanded.isEmpty || expanded.lengthCompare(threshold) >= 0)
```
bjaglin: **Threshold semantics:** The docs say the number of resulting names stays below the threshold, but `expanded.lengthCompare(threshold)` counts only newly added names. For `import p.{a, b, c, _}` with threshold 5 and four additional used names, the result has seven explicit importees but still expands. It can then be immediately coalesced because `coalesceImportees` compares total `importees.length`, so the documented guidance to keep the thresholds ordered does not prevent this.

Could the check include existing `names` and `renames` (and any other resulting importees), or should the docs define the threshold in terms of new names instead?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+   * — its owner is recorded so [[expandWildcardImporter]] leaves any wildcard
+   * exposing it untouched, as it cannot be rendered as a precise explicit
+   * import.
+   */
+  private def classifySelectedMember(
+      qualifier: Term,
+      member: Symbol,
+      unmodeled: mutable.Set[Symbol]
+  )(implicit doc: SemanticDocument): Unit =
+    if (member.isGlobal && !isMemberOfType(qualifier.symbol, member))
```
bjaglin: **Conservative behavior worth documenting:** For a chained call like `a.b().c`, the qualifier of `c` is a `Term.Apply`; its symbol is `None`, so `isMemberOfType` returns false and `c`'s owner is added to `unmodeledOwners`. That can block an unrelated wildcard whenever the wildcard exposes that owner or one of its ancestors. This is safe, but method chaining is common enough that it may materially reduce expansion.

Could the docs mention this, with a possible refinement of consulting synthetics for the receiver type before giving up?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -117,7 +134,23 @@ class OrganizeImports(
             ok
           }
       val dedupedIterator = deduplicateImportees(fullyQualifiedIterator)
-      val mergedIterator = mergeOrExplodeImporters(diagnostics)(dedupedIterator)
+      val expandedIterator = config.expandWildcardImportThreshold match {
```
bjaglin: **Undocumented limitation:** Relative wildcard imports never reach expansion. With `expandRelative = false`, relative importers are diverted to `relativeImporters` before this iterator; with `true`, `expandRelative` synthesizes a fully-qualified ref without symbol information, so `expandWildcardImporter` sees `Symbol.None` and `exposedOwners` cannot model the scope. Both outcomes are conservative, but the docs should state this behavior and a fixture should pin it.

bjaglin: **Design/maintainability:** The wildcard-expansion implementation is self-contained, but it adds roughly 290 lines and eight methods plus a constant to this already large rule. The only integration point is this call site.

Could we extract it into a dedicated private `WildcardImportExpander` class that owns `threshold`, `usedByOwner`, and `unmodeledOwners`? That would remove the current parameter/out-parameter threading and give `exposedOwners` a natural per-run cache, addressing the repeated traversal noted below. The call site could then stay close to `expandWildcardImportThreshold.fold(dedupedIterator)(t => deduplicateImportees(dedupedIterator.map(new WildcardImportExpander(t).apply)))`.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+   * excluded, since their members need no import). For a package this also
+   * walks its package object, whose members — including inherited ones — are
+   * visible through the package wildcard.
+   *
+   * Returns `None` when the scope cannot be modeled precisely — an ancestor
+   * whose definition is not on the classpath — so that
+   * [[expandWildcardImporter]] can leave the wildcard untouched rather than
+   * risk dropping a member declared in that unmodeled ancestor.
+   */
+  private def exposedOwners(prefix: Symbol)(implicit
```
bjaglin: **Simplification and performance:** `exposedOwners` walks the full ancestor hierarchy and performs `infoNoThrow` symbol-table lookups for each owner. `isMemberOfType` invokes it for every selected member, and `expandWildcardImporter` invokes it again for each wildcard.

A per-run `Map[Symbol, Option[Set[Symbol]]]` cache, naturally owned by the proposed `WildcardImportExpander`, would avoid these repeated traversals. The same extraction would also provide a good home for consolidating `canExpandTo`, `displayName.stripSuffix("_=")`, and the string-level `filterNot` into one documented `importableName(symbol: Symbol): Option[String]`; that is also where the earlier `$`-identifier correctness fix could live.

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,95 @@ import scala.util.control.NonFatal
...
+> -   Expansion includes members **inherited** from supertypes (and, for a
+>     package, from its package object and that object's supertypes), not just
+>     the directly declared ones. If any contributing supertype's definition is
+>     not available on the classpath, the scope cannot be modeled precisely and
+>     the wildcard is **left untouched** rather than risk dropping an inherited
+>     member.
+> -   A wildcard whose scope provides a member used through an **extension
+>     method** (Scala 3 `extension`, or a member call whose receiver type cannot
+>     be resolved) is also left untouched, since such a use cannot be turned into
+>     a precise explicit import. (A Scala 2 implicit-class conversion is detected
```
bjaglin, line 404: **Coverage gap:** The Scala 2 claim is specifically about implicit-class conversions being found through synthetics, but `ExpandWildcardImplicit` only exercises an implicit value (`needsInt`).

Could we add a fixture with an implicit class in the wildcard scope and a use such as `receiver.op`, so this synthetic conversion path is actually pinned?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+      threshold: Int,
+      usedByOwner: Map[Symbol, Set[Symbol]],
+      unmodeledOwners: Set[Symbol],
+      importer: Importer
+  )(implicit doc: SemanticDocument): Importer = {
+    if (!importer.hasWildcard) importer
+    else
+      exposedOwners(importer.ref.symbol) match {
+        case None => importer // scope can't be modeled precisely -> leave as-is
+        case Some(owners) if owners.exists(unmodeledOwners) =>
```
bjaglin: **Coverage gap:** This `unmodeledOwners` guard is not tested non-vacuously by the Scala 3 fixture: object-prefix wildcards do not expand on Scala 3 in the first place, so the output is unchanged even without this branch.

Could we add a Scala 2 fixture using the unresolved-receiver path instead, for example a wildcard on `M extends Base` plus an unqualified use and `makeBase().inherited`, and assert that expansion remains blocked?

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,95 @@ import scala.util.control.NonFatal
...
+> are set, keep `expandWildcardImportThreshold <=
+> coalesceToWildcardImportThreshold`. Otherwise an expansion large enough
+> to be re-coalesced is simply turned back into a wildcard within the same
+> run, so expansion only takes effect below the coalesce threshold.
+> 
+> **Scala 3:** a `*` wildcard is expanded just like `_`. A `given` wildcard
+> is always preserved and never expanded, because `given`s are not brought
+> into scope by `*` (they require a `given` import).
+> 
+> Symbols referenced only inside a macro or `inline` expansion are not counted:
```
bjaglin, line 436: **Coverage gap:** There is no fixture exercising the macro/inline exclusion described here.

Could we add a small test with an `inline` def or macro call where a referenced symbol is present only in the expansion, and assert it does not drive wildcard expansion? Without that, the claim that fully-resolved expansion symbols do not depend on caller imports is unverified.

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,95 @@ import scala.util.control.NonFatal
...
+> such expansions are typed with fully-resolved symbols and do not depend on the
+> caller's imports, so they neither require nor drive wildcard expansion.
+> 
+> Note: like `expandRelative`, the names introduced by this option are
```
bjaglin, line 440: **Coverage gap:** No fixture sets `groupSeparately` to include `ByNameImplicits`, so this documented limitation is currently unverified.

Please add a case where a wildcard-sourced implicit is expanded and confirm that the synthetic import is not moved into the separate implicit group.

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,95 @@ import scala.util.control.NonFatal
...
+>     member.
+> -   A wildcard whose scope provides a member used through an **extension
+>     method** (Scala 3 `extension`, or a member call whose receiver type cannot
+>     be resolved) is also left untouched, since such a use cannot be turned into
+>     a precise explicit import. (A Scala 2 implicit-class conversion is detected
+>     through synthetics and handled like any other implicit.)
+> -   A wildcard combined with an unimport (e.g. `import p.{X => _, _}`) is
+>     never expanded, so its hiding semantics are preserved.
+> 
+> If both `expandWildcardImportThreshold` and
```
bjaglin, line 418: **Coverage gap:** The docs explicitly describe same-run expansion/coalescing behavior, but no fixture configures both thresholds.

Could we add one with `expandWildcardImportThreshold` and `coalesceToWildcardImportThreshold` set together, asserting the documented re-coalescing result?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+
+  /**
+   * Replaces a standalone wildcard with the explicit members of its prefix that
+   * are actually used in the document. The importer is left untouched when the
+   * prefix's scope cannot be modeled precisely (see [[exposedOwners]]), when
+   * the scope exposes a member used through an unmodelable selection (an
+   * extension method — see `unmodeledOwners`), when nothing from it is used, or
+   * when the expansion would reach `threshold` names (beyond which a wildcard
+   * is preferable, and would anyway be re-introduced by
+   * `coalesceToWildcardImportThreshold`). Renames, `given` imports and a
```
bjaglin: **Coverage gap:** The scaladoc promises that renames are preserved and that rename sources participate in the `alreadyImported` exclusion, but the current fixtures do not cover either behavior for wildcard expansion. Could we add a case with a rename alongside a wildcard and verify both that the rename survives and that its source is not re-added as an explicit import?

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,95 @@ import scala.util.control.NonFatal
...
 ```
 
+`expandWildcardImportThreshold`
+-------------------------------
+
+Expand a wildcard import into explicit imports of the members actually
+used in the file, as long as the number of resulting names stays below
+the threshold. This is the inverse of
+[`coalesceToWildcardImportThreshold`](OrganizeImports.md#coalescetowildcardimportthreshold)
+and requires SemanticDB (`OrganizeImports` is a semantic rule).
```
bjaglin: Implementation detail, irrelevant to users docs
```suggestion
[`coalesceToWildcardImportThreshold`](OrganizeImports.md#coalescetowildcardimportthreshold).
```

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -165,6 +198,289 @@ class OrganizeImports(
...
+            }
+          case None =>
+            false // unresolvable ancestor -> cannot model precisely
+        }
+      }
+    val value = prefix.value
+    if (value.endsWith("/")) {
+      // Package: direct members are owned by the package itself; further members
+      // may be declared (or inherited) by a package object, which we must walk.
+      owners += prefix
```
bjaglin: **Scala 2/3 expect-test limitation and package-object unsoundness:** The split expect tests are currently showing a classpath symbol-resolution limitation rather than a SemanticDB limitation. On Scala 3, the document-side data is present: wildcard references, occurrences, and implicit-argument synthetics are emitted. However, scalameta `GlobalSymbolTable`/metacp reads Scala 2 pickles and Java classfiles, not Scala 3 TASTy. Consequently `Symbol.info` is `None` for classpath-defined prefixes, so `walk(prefix)` fails for `ExpandWildcard`, `WithExplicit`, `QuotedIdent`, `Inherited`, and `Implicit`; `InstanceCall` additionally cannot resolve `typeSymbolOf(other)`. The extension fixture is different: its extension type is defined in the same file, so `exposedOwners` succeeds and the `unmodeledOwners` guard is genuinely exercised.

The package-prefix branch here has a separate correctness issue. `packageObject.infoNoThrow.isEmpty` is interpreted as “no package object”, but on Scala 3 it can mean “a package object exists but is unreadable”. For `test.organizeImports.pkgobj._`, adding a plain `SomeClass` alongside a package-object `directMember` could make expansion emit `pkgobj.SomeClass` while dropping `directMember`, leaving broken code. The four Scala 3 fixtures that do expand are sound only because those packages genuinely have no package object; the documented fail-safe guarantee does not hold for an unreadable package object.

Could we make an unresolved `<pkg>/package.` symbol fail closed on Scala 3 by treating it as unmodelable, add that `directMember` plus `SomeClass` regression fixture, and document that [until TASTy-aware or classpath-SemanticDB lookup exists](https://github.com/scalacenter/scalafix/issues/2049), Scala 3 expansion is limited to same-file prefixes and plain packages?

bjaglin (commented):
Here is an AI-assisted review. The biggest finding (and blocker) is that there are several limitations on Scala 3 (explicitly documented by e8571f6e0b771dba009976517db1c4fdbbac9461). We must document these to lower expectations (as [fixing them is beyond the scope of the PR](https://github.com/scalacenter/scalafix/issues/2049)).

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Fixed in e7ed6839: a non-exempt candidate whose name contains `$` (or is empty) now fails closed — the wildcard is left untouched instead of silently dropping the name. Constructors, `given`s and compiler-internal `<...>` names remain principled exemptions in `importableName`, since a wildcard replacement never needs to cover them. `ExpandWildcardQuotedIdent` now uses `` `a$b` `` alongside a plain name: the `DollarIdent` importer stays a wildcard while the `QuotedIdent` importer still expands.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Fixed in d2e8c48a: the check now bounds the size of the resulting importee list (pre-existing names/renames/givens plus the expanded names), mirroring how `coalesceImportees` counts, so the documented `expandWildcardImportThreshold <= coalesceToWildcardImportThreshold` guidance now guarantees an expansion is never re-coalesced in the same run. Pinned by `ExpandWildcardThresholdTotal` (`{a, b, _}` with threshold 4 and two more used names → not expanded), plus the interplay fixtures on the other thread.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Documented in 9811b724 — with one correction to the premise discovered while writing the fixture: for `a.b().c` the qualifier's symbol resolves through the applied method, so `typeSymbolOf` models the receiver via `b`'s result type and ordinary chaining does **not** inhibit expansion (a plain `makeBase().inherited` expands fine). What does hit the conservative path is a receiver whose type cannot be named nominally — a literal, or a structural/unresolvable result type. The docs now describe exactly that, and `ExpandWildcardUnmodeledReceiver` uses a structural result type to exercise the guard non-vacuously.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Documented and pinned in 9811b724 — empirically only half the premise held: with `expandRelative = false` relative importers are indeed diverted and never expanded (`ExpandWildcardRelative`); but with `expandRelative = true` the synthesized fully-qualified ref does resolve its symbol, so the wildcard **is** expanded like any other (`ExpandWildcardRelativeExpanded` pins `import MergeImports.Wildcard1._` → `import test.organizeImports.MergeImports.Wildcard1.a` on Scala 2). Docs updated to state both behaviors.

jozanek: Done in 98251227: `WildcardImportExpander` owns `threshold`, the document-wide usage model (`usedByOwner`/`unmodeledOwners`) and the memoized scope lookups, removing the parameter threading. The call site is now `expandWildcardImportThreshold.fold(dedupedIterator)(threshold => deduplicateImportees(dedupedIterator.map(new WildcardImportExpander(threshold).apply)))`.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Done in 98251227: `exposedOwners` is memoized per run inside the expander (one cache shared by the model-building traversal and all importers), and `canExpandTo` + `stripSuffix("_=")` + the name filter are consolidated into a documented `importableName`. The `$` correctness fix landed there in a follow-up commit, as suggested.

docs/rules/OrganizeImports.md
jozanek: Added in c9e15773: `Implicits.d.IntOps` (implicit class) plus `ExpandWildcardImplicitClass` with a `1.incremented` use — the conversion is found through synthetics and the wildcard expands to `import test.organizeImports.Implicits.d.IntOps` on Scala 2 (left untouched on Scala 3).

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Added in 9befdc8a. One adjustment to the suggested shape: `makeBase().inherited` with a nominal result type resolves through the method's return type and does not trip the guard (see the chained-call thread), so the fixture gives `makeBase()` a structural result type to make the receiver genuinely unresolvable. On Scala 2 the guard then blocks an expansion that the unqualified `direct` use would otherwise have driven — non-vacuous as requested.

docs/rules/OrganizeImports.md
jozanek: Added in 04cb69e3: `ExpandWildcardInline` (Scala 3) uses same-file prefixes so expansion is live on Scala 3. `InlineUser.*` expands to `.reveal`, while `InlineMod2.*` — whose member is referenced only through the inline expansion — is left untouched, pinning that expansion internals neither require nor drive expansion.

docs/rules/OrganizeImports.md
jozanek: Added in 3a218706: `ExpandWildcardImplicitSeparate` sets `groupExplicitlyImportedImplicitsSeparately = true` and pins that the expanded `intImplicit` import stays in the main group (sorted ahead of the `Wildcard1.a` import) instead of moving to the trailing implicits group, since the synthesized importee carries no symbol.

docs/rules/OrganizeImports.md
jozanek: Added in b5a2d82c: `ExpandWildcardCoalesce` (recommended ordering, expand = coalesce = 4: expansion to 3 names sticks) and `ExpandWildcardCoalesceMisordered` (expand 5 > coalesce 2: the expansion is re-coalesced back to a wildcard within the same run, i.e. net unchanged) — both with the two options set together as requested.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Added in b566c36d: `ExpandWildcardRename` pins that `{a => a1, _}` expands to `{a => a1, b}` — the rename survives, and `a` is not re-added as an explicit import even though the `a1` use resolves to it (rename-source exclusion).

docs/rules/OrganizeImports.md
jozanek: Applied in 9811b724.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
jozanek: Addressed in 0770aaac: an unresolvable `<pkg>/package.` now fails closed when the document itself proves the package object exists — that is, when a used (or unmodeled) symbol is owned by it. That keeps plain packages expandable on Scala 3 while guaranteeing an unreadable package object can never cause a used member to be dropped. `ExpandWildcardPackageObjectMixed` pins the `directMember` + `PlainClass` scenario: full expansion on Scala 2, wildcard untouched on Scala 3. The docs now state the TASTy limitation (same-file prefixes and plain packages) and link scalacenter/scalafix#2049.

jozanek:
All review feedback addressed in 11 commits (one per thread, replies inline):

**Correctness** — the blocker: an unresolvable `<pkg>/package.` now fails closed when the document's own references prove the package object exists, so Scala 3 can never drop a used package-object member (docs state the TASTy limitation, linking #2049). Also: `$`-named identifiers block expansion instead of being silently dropped, and the threshold counts all resulting selectors, aligning with `coalesceImportees`.

**Refactor** — expansion is extracted into `WildcardImportExpander`, which owns the usage model, memoizes `exposedOwners`, and consolidates `importableName`.

**Coverage** — new fixtures for implicit-class conversions, a non-vacuous `unmodeledOwners` guard, inline-expansion exclusion, implicits grouping, renames, both-thresholds interplay, and relative imports.

Two premises turned out subtly different in practice (details in the threads): ordinary chained calls `a.b().c` resolve through the method's result type and do *not* inhibit expansion (only non-nominal receiver types do), and `expandRelative = true` *does* make relative wildcards expandable. Docs and fixtures reflect the actual behavior.

bjaglin (commented):
Thanks for addressing the comments!

I tried to dogfood the option by adding `expandWildcardImportThreshold = 5` on top of this PR and got a lot of compilation failures following the rewrites.

Could you have a look at it? The flag is new and opt-in so it doesn't need to be perfect, but I don't think we should ship it until it is proven to work on this repo and a couple of popular Scala libs using wildcards imports.

jozanek:
Thanks for dogfooding — that was a very productive failure. I reproduced it, found seven distinct causes (four on this repo, three on other projects), fixed each and pinned them with fixtures:

1. **Names bound by an enclosing scope** (own/inherited members, anonymous classes, self types, same-file package members) were imported explicitly and then reported unused — fatal here. Now skipped.
2. **Implicit-providing scopes**: `scala.jdk.CollectionConverters._` expanded to different member names on 2.12 and 2.13, each breaking the other; and implicits can be summoned inside macro expansions, which SemanticDB doesn't record. A scope that declares an implicit/given is never expanded now (this replaces the earlier synthetics-based handling of implicit classes).
3. **Scala 3 top-level definitions** (owned by `File$package`) were dropped; owners are normalized to the package.
4. **Scala 3 package objects** are unreadable TASTy, so members inherited through them were dropped. The classpath is now consulted: `<pkg>/package.class` present but unreadable ⇒ untouched.
5. **`new C(..)` with inferred type arguments** has no occurrence in Scala 3 SemanticDB (found on scala-xml). Unresolved references now fail the file closed; this shape is recovered from the classpath.
6. **A member named `*`** rendered as a wildcard (`ImportMatcher.*`); `*`, `_`, `given` are treated like `$` names.
7. **Crash** (`MissingSymbolException`) when a classpath entry can't be read (found on Metals, pre-existing): all symbol-table lookups are non-throwing now, failing closed.

Plus two guards: a name also resolved through a nested wildcard import blocks expansion (would become ambiguous), and operator calls count as member selections.

One caveat that isn't the rule's: `sbt scalafixAll` runs the matrix rows concurrently, so later rows read stale SemanticDB — that's where the mangled rewrites in your run came from. Running rows one at a time is deterministic; noted in the docs.

PR #2427, opened 2026-04-22
OrganizeImports: use scalameta to pretty-reprint

kitbellew:
Blocks #2428.

PR #2468, opened 2026-06-25
OrganizeImports: preserve trailing comma in multi-line imports

mlachkar:
Since the switch to scalameta pretty-reprinting ([0d232c9f](https://github.com/scalacenter/scalafix/pull/2427)), the rule stopped re-emitting the trailing comma when reprinting a multi-line import, even when nothing else changed. 
On an already-sorted import the rule would still produce a diff, just to strip the comma.

This makes it effectively impossible to roll OrganizeImports out on a large code base that uses trailing commas: every multi-line import loses its comma on the first run, producing a huge, noisy diff. Worse, it fights with scalafmt — scalafmt is configured to add the trailing commas back, so the two tools keep undoing each other, and the formatting never converges (CI flip-flops, pre-commit hooks loop).

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -794,10 +794,16 @@ class OrganizeImports(
             val line = lines.next()
             lines.exists(_ != line)
           }
-          val useOuterSpace = !isMultiline && i.importees
+          val origImporters = i.importees
             .flatMap(_.originalPrototype().parent)
             .distinct
-            .exists(_.hasSpaceInCurly)
+          val useOuterSpace =
+            !isMultiline && origImporters.exists(_.hasSpaceInCurly)
+          // Preserve a trailing comma if the (single) source importer had one.
+          val trailingComma = isMultiline && (origImporters match {
+            case Seq(i: Importer) => i.hasTrailingComma
+            case _ => false
+          })
           sb.append('{')
           val sep = if (isMultiline) "\n  " else " "
           if (isMultiline) sb.append(sep)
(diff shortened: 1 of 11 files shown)


kluen:
Hi @mlachkar, the pretty-printing you referenced seems to cause an issue in my company's codebase that seems to be related to what this PR fixes.

We run into an endless loop between Scalafix and Scalafmt on single-importee renamed imports that Scalafmt wraps once they pass `maxColumn`. Scalafmt writes:

```
import com.example.deeply.nested.generated.definitions.{
  OriginalName => RenamedName,
}
```
OrganizeImports then reprints it on one line:
```
import com.example.deeply.nested.generated.definitions.{OriginalName => RenamedName}
```
Scalafmt re-wraps it, and the two tools keep undoing each other.

I intended to contribute a PR to fix my issue, which would be similar to this one, but then I wondered if we are only patching the symptoms and if the code should probably be fixed on a different, more structural level. Happy to hear your thoughts.

francesconero:
We're also having the same problem in several repositories, hope it'll get merged soon!

mlachkar:
I applied the comments and added a test case for imports with comments.

bjaglin (approved):
Thanks!

mlachkar:
@kluen I m fixing the second case here: https://github.com/scalacenter/scalafix/pull/2507

You're right, we should not only "patch symptoms".
Since the switch to Scalameta pretty-printing (0.14.7), the rule deleted every import and re-inserted a fresh print of its tree, whether or not it had changed anything — and then recovered, one detail at a time, what the printer had dropped (spaces in braces, trailing commas, and now wrapped single importees). There was always going to be one more.

The fix goes at that level: an import that is already organized — same qualifier, same importees, same order after applying the config — and wrapped over several lines is now re-emitted from its source text verbatim.

PR #2475, opened 2026-07-10
Handle scala3 placeholder symbols

jackkoenig:
This is a workaround for https://github.com/scala/scala3/issues/26521 (assuming it is a bug as I think it is)

Basically, Scala 3 semanticdb will include placeholder symbols sometimes which causes `Patch.replaceSymbols` to crash.

I split this into 2 PRs because if you check out the commit with just the tests will pass with Scala 2.13.18, but fail with 3.3.8
```sh
# Passes
sbt "expect2_13_18Target2_13_18 / test"
# Fails
sbt "expect3_3_8Target3_3_8 / test"
```

If you include the workaround commit, then the Scala 3.3.8 tests also pass.

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/v0/LegacySemanticdbIndex.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/v0/LegacySemanticdbIndex.scala
@@ -67,7 +67,13 @@ final class LegacySemanticdbIndex(val doc: v1.SemanticDocument)
     }
 
   override def symbol(position: Position): Option[v0.Symbol] =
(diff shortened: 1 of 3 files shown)


jackkoenig:
Force pushed because I forgot to run scalafmt before, it had passed all other tests. I should've just pushed a new commit instead of force pushing 🙈.

tgodzik (approved):
Makes sense to have this workaround, I will take a look at the compiler issue, but we can't fix it for the already released versions. LGTM

bjaglin:
Thanks for the workaround @jackkoenig 👍

PR #2478, opened 2026-07-14
feat(OrganizeImports): add groupRelativeImports option

### LINKED ISSUE #2100: `OrganizeImports.groups` not correctly grouping imports if doing a nested import statement
Here is the [PR in context](https://github.com/mdedetrich/pekko-streams-circe/pull/12) and here is the specific `.scalafix` configuration 

https://github.com/mdedetrich/pekko-streams-circe/pull/12/files#diff-8fe42aaf3e5d1a3f8804efa64a6206f5af6a826dea8f6ce51ad0ce8105188858R1-R32

My current expectation of that configuration is that `java`/`scala` imports go to the bottom of the import list and all other imports gravitate to the top but as you can see from https://github.com/mdedetrich/pekko-streams-circe/pull/12/files#diff-ba5a4a197dba09e4d184afb14c8a2f50b68a870a27cfd9ceebc328fd2a96618b the various `pekko.*` imports are being placed after the scala/java imports.

I tried adding `pekko`/`org.apache.pekko` to the group list but that didn't have any effect, this is just a suspicion but maybe the underlying issue is that there is a `import org.apache.pekko` and then after that import statement we do `import pekko.NotUsed` etc etc

He-Pin:
## Summary
- Add `groupRelativeImports` configuration option to `OrganizeImports` rule
- When set to `Grouped` (with `expandRelative = false`), relative imports participate in regular group matching instead of being separated into an order-preserving group
- Eliminates the unwanted blank line between fully qualified and relative imports that belong to the same logical group (e.g., `org.apache.pekko` and `pekko.*` in the Apache Pekko project)

## Motivation
In projects like Apache Pekko, imports use both fully qualified (`org.apache.pekko.*`) and short (`pekko.*`) forms. The short form resolves to the same package but is classified as a "relative import" by the semantic analyzer because `pekko` resolves to `org.apache.pekko` (whose owner is `org.apache`, not root).

Currently, `OrganizeImports` always places relative imports in a separate order-preserving group with a mandatory blank line separator. This creates an unwanted blank line between `import org.apache.pekko` and `import pekko.annotation.*`, even when the user configures them to be in the same group.

## Modification
- Added `GroupRelativeImports` sealed trait with two values:
  - `KeepOrdered` (default): preserves existing behavior — relative imports go to order-preserving group
  - `Grouped`: relative imports participate in regular group matching using their textual form
- Modified `organizeGlobalImports` in `OrganizeImports.scala` to skip filtering relative imports when `groupRelativeImports = Grouped`
- Added documentation and test case

## Example
```conf
OrganizeImports {
  expandRelative = false
  groupRelativeImports = Grouped
  groups = ["re:javax?\\.", "scala.", "*"]
}
```

Before (blank line between pekko groups):
```scala
import org.apache.pekko

import pekko.annotation.ApiMayChange
import pekko.annotation.InternalApi
```

After (no blank line, same group):
```scala
import org.apache.pekko
import pekko.annotation.ApiMayChange
import pekko.annotation.InternalApi
```

## Test plan
- [x] Added `GroupRelativeImportsGrouped` input/output test case
- [x] All 159 existing expect tests pass
- [x] All 3 config unit tests pass
- [x] Code compiles for Scala 2.13

### DIFF
--- a/docs/rules/OrganizeImports.md
+++ b/docs/rules/OrganizeImports.md
@@ -374,6 +374,77 @@ import scala.util.control.NonFatal
 import sun.misc.BASE64Encoder
 ```
 
+
+`groupRelativeImports`
+----------------------
+
+Control how relative imports are handled when `expandRelative` is `false`.
+
+### Value type
+
+Enum: `KeepOrdered | Grouped`
+
+#### `KeepOrdered`
+Relative imports are moved to a separate order-preserving group that is
+appended after all the other configured import groups. This is the default
+behavior and preserves the original relative ordering.
+
+#### `Grouped`
+Relative imports participate in regular group matching. They are matched
+against the configured `groups` patterns using their textual form (as
+written in the source code) and sorted together with fully qualified
+imports. This avoids a blank line between fully qualified and relative
+imports that belong to the same logical group.
+
+> **Note**: When using `Grouped`, relative imports are sorted by their
+> textual form, not their fully qualified form. For example, `pekko.actor._`
+> sorts under `p`, not under its resolved path `org.apache.pekko.actor._`.
+
+### Default value
+
+`KeepOrdered`
+
+### Examples
+
+```conf
+OrganizeImports {
+  expandRelative = false
+  groupRelativeImports = Grouped
+  groups = ["re:javax?\\.", "scala.", "*"]
+}
+```
+
+Before:
+
+```scala
+import scala.util
+import util.control
+import control.NonFatal
+import java.time.Clock
+```
+
+After (`Grouped` — relative imports mixed into regular groups):
+
+```scala
+import java.time.Clock
+
+import scala.util
+
+import control.NonFatal
+import util.control
+```
+
+After (`KeepOrdered` — default, relative imports in separate group):
+
+```scala
+import java.time.Clock
+
+import scala.util
+
+import util.control
+import control.NonFatal
+```
+
 `groupedImports`
 ----------------
 
@@ -628,8 +699,8 @@ same group and sorted by the order defined by the
 > order-preserving import group may appear after all the configured import
 > groups when:
 > 
-> 1.  The `expandRelative` option is set to `false` and there are relative
->     imports.
+> 1.  The `expandRelative` option is set to `false`, `groupRelativeImports`
+>     is `KeepOrdered` (the default), and there are relative imports.
 > 
 > 2.  The `groupSeparately` parameter is set, and there are
 >     [matching imports](#groupseparately).
@@ -1303,6 +1374,7 @@ OrganizeImports {
   blankLines = Auto
   coalesceToWildcardImportThreshold = null
   expandRelative = false
+  groupRelativeImports = KeepOrdered
   groupSeparately = []
   groupedImports = Explode
   groups = [
@@ -1332,6 +1404,7 @@ OrganizeImports {
   blankLines = Auto
   coalesceToWildcardImportThreshold = 5
   expandRelative = false
+  groupRelativeImports = KeepOrdered
   groupSeparately = []
   groupedImports = Merge
   groups = [
--- a/docs/rules/OrganizeImports.md
+++ b/docs/rules/OrganizeImports.md
@@ -374,7 +374,6 @@ import scala.util.control.NonFatal
 import sun.misc.BASE64Encoder
 ```
 
-
 `groupRelativeImports`
 ----------------------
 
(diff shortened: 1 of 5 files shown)


He-Pin:
@bjaglin hi， sorry  for the ping, I think pekko need this to handle the style in pekko, thanks.

bjaglin:
Hi @He-Pin, I am planning to invest some time in the coming days but I can't promise anything as there is a lot of contribution to review. If I understand well, this is a new feature, why is it so urgent?

He-Pin:
No rush, just do whatever you like. Ideally, it should support the style of pekko: `import org.apache.pekko` and then `import pekko.*`, and be sorted by type without spaces.

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,78 @@ import scala.util.control.NonFatal
...
 
+`groupRelativeImports`
+----------------------
+
+Control how relative imports are handled when `expandRelative` is `false`.
+
+### Value type
+
+Enum: `KeepOrdered | Grouped`
+
```
bjaglin, line 385: I think the UX needs to be revisited
- the flag sounds like a boolean, yet it's an enum
- having one of the enum value repeat "group" which is the name of the flag is confusing
- I am unsure how `Ordered` can qualify a grouping strategy - should it be just `Keep`?

bjaglin: Actually, it looks like what you want to achieve is to let user opt-out from the separate group of relatives. So I would suggest adding a `Relative` option to the `groupSeparately` flag, which is on by default when expandRelative is false (current behavior, well documented), but that users can opt-out from.

This would be a breaking change for users explicitly setting groupSeparately (since they would need to add that flag to preserve behavior), but it's a very recent option so I think that's acceptable as long as we highlight it in the release notes.

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,78 @@ import scala.util.control.NonFatal
...
+> Relative imports can depend on imports that appear before them. `Grouped`
+> allows those imports to be regrouped and reordered, so moving a relative
+> import before its prerequisite can make the rewritten source fail to
+> compile. Use `KeepOrdered` unless the configured `groups` and
+> `importsOrder` preserve those dependencies.
+>
+> **Note**: When using `Grouped`, relative imports are sorted by their
+> textual form, not their fully qualified form. For example, `pekko.actor._`
+> sorts under `p`, not under its resolved path `org.apache.pekko.actor._`.
```
bjaglin, line 406: can you please add a test demonstrating that?

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,78 @@ import scala.util.control.NonFatal
...
+against the configured `groups` patterns using their textual form (as
+written in the source code) and sorted together with fully qualified
+imports. This avoids a blank line between fully qualified and relative
+imports that belong to the same logical group.
+
+> Relative imports can depend on imports that appear before them. `Grouped`
+> allows those imports to be regrouped and reordered, so moving a relative
+> import before its prerequisite can make the rewritten source fail to
+> compile. Use `KeepOrdered` unless the configured `groups` and
+> `importsOrder` preserve those dependencies.
```
bjaglin, line 402: can you add a test matching this scenario?

docs/rules/OrganizeImports.md
```diff
@@ -374,6 +374,78 @@ import scala.util.control.NonFatal
...
+Relative imports are moved to a separate order-preserving group that is
+appended after all the other configured import groups. This is the default
+behavior and preserves the original relative ordering.
+
+#### `Grouped`
+Relative imports participate in regular group matching. They are matched
+against the configured `groups` patterns using their textual form (as
+written in the source code) and sorted together with fully qualified
+imports. This avoids a blank line between fully qualified and relative
+imports that belong to the same logical group.
```
bjaglin, line 396: With the natural pekko config groups = ["org.apache.pekko", "*"], import org.apache.pekko matches group 1 and import pekko.annotation.X falls to *, so the blank line would come back with the current implementation. Could we sort based on the resolved reference but still print the relative one?

bjaglin (commented):
Thanks for your PR!

Sorry for the long delay getting back to you. I have 2 concerns with alternative implementations detailed in the inline comments: the overall UX & using the text representation for grouping,

By the way, https://github.com/scalacenter/scalafix/issues/2100 looks like exactly what this PR is trying to solve so I linked the PR to that issue.

bjaglin:
Hey @He-Pin! I am going to cut another release soon, do you have any bandwidth any time soon to follow up on my comments?

PR #2479, opened 2026-07-14
Support custom `onTestFailure` message on `--check` failure

### LINKED ISSUE #549: scalafixTest should support custom error message on failure
In scalafmt, the `onTestFailure: String` setting can be customized to print out a message on `--test` failure. See http://scalameta.org/scalafmt/#Other

It would be nice if scalafix did the same, so that contributors get actionable instructions on how to fix failures from `scalafix --test`.

jozanek:
Closes #549.

Like scalafmt's setting of the same name, prints project-specific instructions when `scalafix --check` fails:

```hocon
// .scalafix.conf
onTestFailure = "To fix this, run: sbt scalafixAll"
```

The message is printed after the diffs, and only when the failure is entirely auto-fixable (exit status exactly `TestError`) — not when the run also hits linter errors, parse errors, or invalid inputs that re-running Scalafix can't fix.

**Notes for reviewers:**

- Exit-code change: a nonexistent input path now merges `UnexpectedError` into the exit status, so `--check existing.scala missing.scala` exits `TestError+UnexpectedError` instead of plain `TestError`. Other exit codes unchanged.
- `ScalafixEvaluation` is untouched; surfacing the message via sbt-scalafix's `scalafixTest` needs a follow-up there.

### DIFF
--- a/docs/users/configuration.md
+++ b/docs/users/configuration.md
@@ -100,6 +100,24 @@ Similarly, diagnostics can be controlled through `lint.ignore`, `lint.info`,
 and `lint.warning`, which respectively suppress matching diagnostics, force
 them to be at info level, and force them to be at warning level.
 
+## Customizing the message on `--check` failure
+
+When running in check mode (`--check`/`--test`), Scalafix prints a unified
+diff and exits with a non-zero code if some files would be modified by the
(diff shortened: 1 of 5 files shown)


bjaglin (approved):
Thanks!

PR #2484, opened 2026-08-11
Upgrade to scalameta 4.17.3 and stop vendoring its SymbolTable FQCN

jozanek:
Upgrades scalameta from 4.17.0 to 4.17.3.

Not a one-line bump: scalafix-core vendored its own `scala.meta.internal.symtab.SymbolTable` — scalameta's exact FQCN — declaring only `info`, so core could compile without depending on `semanticdb-scalac-core` (which pulls in scala-compiler). scalafix-reflect *does* depend on the real artifact, so both end up on the classpath and the vendored copy shadows it.

That was harmless while the upstream trait was purely abstract. 4.17.1 gave it concrete members, so it now emits a `$init$` that `GlobalSymbolTable`'s constructor calls — and the shadowing copy has none:

```
java.lang.NoSuchMethodError: 'void scala.meta.internal.symtab.SymbolTable.$init$(scala.meta.internal.symtab.SymbolTable)'
  at scala.meta.internal.symtab.GlobalSymbolTable.<init>(GlobalSymbolTable.scala:18)
  at scalafix.internal.reflect.ClasspathOps$.newSymbolTable(ClasspathOps.scala:30)
```

Every semantic operation hits this, so scalafix cannot currently move past 4.17.0. It reproduces on `main` with only the version constant changed.

The fix keeps core compiler-free but drops the name collision:

- `scalafix.internal.symtab.SymbolTable` — scalafix-owned trait with the single `info` core needs.
- `ClasspathOps.fromScalameta` — adapts scalameta's table onto it, in reflect.

The rest is mechanical: imports move to the new name, plus one `scala.meta.internal.trees._` import in `ReplaceSymbolOps` that the upgrade leaves unused (`-Werror`).

No public API change; scalafix-core's dependencies are unchanged.

Verified on JDK 17 across the full matrix — `unit` and `integration` on 2.12.21, 2.13.18, 3.3.8, 3.5.2, 3.6.4, 3.7.4 and 3.8.4, plus the built-in rule `expect` suites and `versionPolicyCheck` on core (2.12/2.13), cli, rules and testkit. All green.


bjaglin (approved):
Thanks!

Long live https://github.com/scalacenter/scalafix/pull/2048#discussion_r1730163351, https://github.com/scalameta/scalameta/pull/4639 put a nail in its coffin.

PR #2485, opened 2026-08-12
Scope rule compilation caching to Scalafix instances (fix #782)

### LINKED ISSUE #782: Remove evil global cache for compiling rules
Scalafix uses a global mutable cache for compiling rules
https://github.com/scalacenter/scalafix/blob/6a0c120ce3e917acdb10e8ad4e69f52b26d468eb/scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala#L23-L26

We should replace this cache by requiring end-users to instantiate an instance of "Scalafix API". This gives end-users control over when to clear/start caching results.

jozanek:
Fixes #782

The old global caches never actually cached — `getOrDefault` evaluates its default eagerly, so every call recompiled — they only retained classloaders and serialized compilation behind a global lock.

`ScalafixToolbox` is now a class: each `Scalafix` instance owns one, shared by all its `ScalafixArguments`, so clients control the cache lifecycle by creating/dropping instances; everything else gets a private toolbox. Results are keyed by source path + content + tool classpath, so edited rules recompile instead of going stale.

No public API changes (`versionPolicyCheck` passes without exclusions). New `RuleDecoderSuite` tests pin cache hits, invalidation, and per-instance isolation.


scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixImpl.scala
```diff
@@ -4,15 +4,21 @@ import buildinfo.RulesBuildInfo
...
 import scalafix.interfaces.Scalafix
 import scalafix.interfaces.ScalafixArguments
+import scalafix.internal.reflect.ScalafixToolbox
+import scalafix.internal.v1.Args
 import scalafix.internal.v1.MainOps
 
 final class ScalafixImpl extends Scalafix {
 
+  // One toolbox per Scalafix instance: dropping the instance drops its
+  // compiled-rule caches (https://github.com/scalacenter/scalafix/issues/782).
```
bjaglin: I don't see any test for that actually - could you add a test in ScalafixSuite? An idea might be to rely on the fact that a companion object initializer runs once per classloader/compilation, so we can capture the number of compilations using the same mechanism as CaptureScalaVersion.

scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
```diff
@@ -1,55 +1,73 @@
...
+  }
+}
+
+object ScalafixToolbox {
+  case class CompiledRules(classloader: ClassLoader, fqns: Seq[String])
+
+  // input carries the source identity (path), which the compilers consume
+  private case class RuleKey(
+      input: Input,
+      content: String,
```
bjaglin: Having the content in the key looks like a source of leak when a user recompiles the same path repeatedly in the same sbt session. Could we instead keep a hash of it on the value side alongside CompilerRules, adding an additional check when that hash changes to trigger invalidation?

bjaglin (commented):
Thanks for the PR. Could you try drafting a test to document and guard the purpose of the original ticket and look at a performance improvement suggestion?

scalafix-reflect/src/main/scala/scalafix/internal/reflect/ScalafixToolbox.scala
jozanek: Done — content hash moved to the value side; on change the entry is replaced, so one entry per path.

scalafix-cli/src/main/scala/scalafix/internal/interfaces/ScalafixImpl.scala
jozanek: Added. It caught a bug too: `newArguments()` created a fresh default tool classpath each time (part of the cache key), so the instance now shares that as well.

bjaglin (approved):
Thanks!

PR #2496, opened 2026-08-21
Organize Imports: order import selectors by kind for IntelliJ preset

### LINKED ISSUE #2483: OrganizeImports with INTELLIJ_2020_3 uses different ordering for renamed imports than IntelliJ IDEA does
I'm using `OrganizeImports` with the `INTELLIJ_2020_3` preset, and it is using a different ordering than the latest version of IntelliJ IDEA (IntelliJ IDEA 2026.2.0.1) does.  I don't know if this is a change between the 2020 and 2026 IDEA behavior, or an inconsistency between the plugin and IDEA 2020.3.  But I would want them to be consistent so that optimizing imports in IDEA doesn't cause my build to fail when running `sbt "scalafix --check"`.

IDEA:

```scala
import com.codahale.metrics.{MetricRegistry, Gauge => DropwizardGauge}
```

Scalafix:

```scala
import com.codahale.metrics.{Gauge => DropwizardGauge, MetricRegistry}
```

# Error output

```
$ sbt "scalafix --check"
```

```
[error] --- /path/to/my-project/my-subproject/src/main/scala/invalid/test/MyClass.scala
[error] +++ <expected fix>
[error] @@ -1,5 +1,5 @@
[error]  package invalid.test
[error] -import com.codahale.metrics.{MetricRegistry, Gauge => DropwizardGauge}
[error] +import com.codahale.metrics.{Gauge => DropwizardGauge, MetricRegistry}
[error]  import org.apache.kafka.common.Metric
[error]  import org.apache.kafka.common.metrics.KafkaMetric
[error]  
[error] (mySubproject / Compile / scalafix) scalafix.sbt.ScalafixFailed: TestError
[error] Total time: 395 s (0:06:35.0), completed Aug 7, 2026, 4:23:10 PM
```

# Configuration

**`.scalafix.conf`**

```
rules = [OrganizeImports]

OrganizeImports.preset = INTELLIJ_2020_3
```

# Additional observations

Here's another case I've encountered:

IDEA:

```scala
import org.mockito.Mockito.{verify, when, timeout => mockitoTimeout}
```

Scalafix:

```scala
import org.mockito.Mockito.{timeout => mockitoTimeout, verify, when}
```

subhranshudash13-dotcom:
Problem : 

Previously, grouped import selectors under `INTELLIJ_2020_3` were ordered purely lexicographically (ASCII), leading to discrepancies with IntelliJ IDEA's import optimizer which applies selector-kind precedence before sorting lexicographically within each kind:
1. Plain imported names (`a`, `b`)
2. Renames / aliases (`c as C`, `timeout => mockitoTimeout`)
3. Unimports / exclusions (`d => _`)
4. Wildcards (`*`, `_`)
5. Givens (`given A`)
6. Given wildcards (`given`)
For example:
```scala
// IntelliJ IDEA:
import org.mockito.Mockito.{verify, when, timeout => mockitoTimeout}
// Scalafix previous output:
import org.mockito.Mockito.{timeout => mockitoTimeout, verify, when}

What i had worked on (solution) : 

- Added ImportSelectorsOrder.IntelliJ to categorize importees by selector kind before sorting.
- Configured INTELLIJ_2020_3 preset to default to importSelectorsOrder = IntelliJ.
- Added unit and regression test fixtures for Scala 2 and Scala 3.
- Updated docs/rules/OrganizeImports.md to document the new IntelliJ value for importSelectorsOrder.

Summary :

Fixes the import selector sorting behavior in `OrganizeImports` when using the `INTELLIJ_2020_3` preset or explicit IntelliJ selector ordering.

Please review the changes and let me know if anything needs adjustment or further testing. If everything looks good, feel free to merge. Thank you!

### DIFF
--- a/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
+++ b/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
@@ -5,7 +5,7 @@ import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.A
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.B
 
 // All six selector categories in a single import
-import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, *, given A, given B, given}
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, given A, given B, *, given}
 
 object SortImportSelectorsIntelliJScala3 {
   object fixtures {
--- a/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
+++ b/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
@@ -3,9 +3,8 @@ package test.organizeImports
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.*
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.A
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.B
-
 // All six selector categories in a single import
-import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, given A, given B, *, given}
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, *, given A, given B, given}
 
 object SortImportSelectorsIntelliJScala3 {
   object fixtures {
--- a/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
+++ b/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
@@ -4,7 +4,7 @@ import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.*
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.A
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.B
 
-// All six categories in a single import: givens, unimports, renames, wildcard, names, givenAll
+// All six selector categories in a single import
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, *, given A, given B, given}
 
 object SortImportSelectorsIntelliJScala3 {
--- a/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
+++ b/scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
@@ -3,12 +3,9 @@ package test.organizeImports
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.*
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.A
 import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.B
-// Category test 2: names (a, b), renames (c as C), unimports (d => _), givens (given A, given B)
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala [hunk 1]
```diff
@@ -0,0 +1,22 @@
+package test.organizeImports
+
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.*
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.A
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.B
+
+// All six selector categories in a single import
+import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, given A, given B, *, given}
```
bjaglin: based on https://github.com/JetBrains/intellij-scala/blob/14ac6dc689a6754603e30794e5028332dc97ee16/scala/scala-impl/src/org/jetbrains/plugins/scala/editor/importOptimizer/ScalaImportOptimizer.scala#L709-L723, shouldn't this be
```suggestion
import test.organizeImports.SortImportSelectorsIntelliJScala3.fixtures.{a, b, c as C, d as _, *, given A, given B, given}
```

bjaglin (commented):
Thanks for the PR Can you confirm the wildcards position? Also, please check the test failures.

scalafix-tests/output/src/main/scala-3/test/organizeImports/SortImportSelectorsIntelliJScala3.scala
subhranshudash13-dotcom: Thanks @bjaglin for the review and for pointing me to the IntelliJ implementation. I’ve updated the selector ordering to match IntelliJ’s behavior, with the wildcard placed before typed givens. 
{a, b, c as C, d as _, *, given A, given B, given}
I also identified that the Scala 3 test was being affected by groupSeparately, so I isolated the selector-ordering case with groupSeparately = [].
I’ve also run "git diff --check", which passes cleanly. Thanks again for the guidance.

bjaglin: >  I also identified that the Scala 3 test was being affected by groupSeparately, so I isolated the selector-ordering case with groupSeparately = [].

can you elaborate?

subhranshudash13-dotcom: Sure, I noticed that the default "groupSeparately" setting includes "ByTypeGivens" because of that, the given A selector in this fixture was being moved into a separate import instead of staying with the other selectors. Since this test is meant to check the IntelliJ ordering of all the selector types together, I set groupSeparately = [] for this fixture. That way the typed givens stay in the same import and the test actually exercises the ordering between *, typed givens, and given.
I also verified the updated test locally on Scala 3.3.8 and 3.8.4.

bjaglin: Does it mean groupSeparately should be set to [] in the preset? To better stick to IntelliJ behavior, but at the risk of generating invalid code?

subhranshudash13-dotcom: I'm actually not exactly sure whether groupSeparately should be changed in the preset. I added groupSeparately = [] because the imports were being split apart by the default ByTypeGivens behavior, and i believe the tests demanded all six selector categories together in a single import. So, as per my understanding this was only needed as a test-specific configuration and shouldn't require changing the preset itself. I may be missinh something, so I’d appreciate your guidance on whether this should remain test-specific or be reflected in the preset.

bjaglin: I don't have IntelliJ myself, but I am wondering what's the default behavior if it was presented the input file of the test - if it's keeping everything on the same line, then I think we should change the value of groupSeparately for that preset. But that's a somewhat different story than the original issue so let's park that for later (or never if there is no interest/bandwith to follow-up).

bjaglin (approved):
Thanks!

PR #2500, opened 2026-08-28
fix(organizeImports): preserve qualifier escaping for soft keywords #2480

### LINKED ISSUE #2480: `OrganizeImports` drops backticks around Scala 3 soft keywords in shared sources
`OrganizeImports` in 0.14.7 can rewrite valid imports into code that no longer compiles when an import path contains a Scala 3 soft keyword, for example, a Java package segment named `export`.

### Example:

  ```diff
  -import io.opentelemetry.sdk.trace.`export`.SimpleSpanProcessor
  -import io.opentelemetry.sdk.trace.`export`.SpanExporter
  +import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor
  +import io.opentelemetry.sdk.trace.export.SpanExporter
```

  This happened in a shared `src/main/scala` source compiled with Scala 3.

  Config:
```hocon
rules = [
  OrganizeImports
]

OrganizeImports {
  importsOrder = SymbolsFirst
  importSelectorsOrder = SymbolsFirst
  removeUnused = false
}
```

### Expected behavior: keep the escaped package segment:

```scala
import io.opentelemetry.sdk.trace.`export`.SimpleSpanProcessor
```

###  Actual behavior: backticks are removed:

```scala
import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor
```

I bisected this between `0.14.6` and `0.14.7`. First bad commit:

  0d232c9f649050b8e62a3c1aa8345acb18174e98 : OrganizeImports: use scalameta to pretty-reprint

The issue seems tied to shared sources: with `targetDialect = StandardLayout`, `src/main/scala` is printed with the Scala 2 dialect, so Scala 3 soft keywords like export can be emitted unescaped.

___

See:
- https://github.com/typelevel/otel4s/pull/1305
- https://github.com/typelevel/otel4s/actions/runs/29270323397/job/86886105703

subhranshudash13-dotcom:
Problem : 
Since 0.14.7, OrganizeImports pretty-prints the full importer using Scalameta. For shared src/main/scala sources, StandardLayout uses the Scala 2 dialect, which can cause necessary backticks in import qualifiers to be removed.

For example :
"import io.opentelemetry.sdk.trace.`export`.SimpleSpanProcessor"

can be rewritten to :
"import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor"
which is invalid when the source is compiled with Scala 3.

Fix :
Preserve the original source text of an existing import qualifier when it has a source position, while continuing to use Scalameta pretty-printing for synthesized qualifiers.

This restores the source-preserving behavior that existed before "0d232c9" without introducing a new configuration option or dialect.

Tests : 
Added a regression test covering a Scala 3 soft keyword (export) in a shared source using StandardLayout.

Verified with :
Scala 2 and Scala 3 expect tests
sbt testWindows
sbt ci-docs
sbt versionPolicyCheck
./bin/scalafmt --test
git diff --check

Fixes #2480

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -387,7 +387,7 @@ class OrganizeImports(
     type SortFunc = (StringBuilder, Boolean, List[Importee]) => Unit
     def sortSyntax(f: SortFunc)(imp: Importer): String = {
       implicit val sb = new StringBuilder
-      sb.append(refSyntax(imp.ref)).append('.')
+      sb.append(treeSyntax(imp.ref)).append('.')
       f(sb, imp.isCurlyBraced, imp.importees)
       sb.toString()
     }
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -780,7 +780,7 @@ class OrganizeImports(
         ps.foreach(_.begComment.foreach(appendBegComment))
         i.begComment.foreach(appendBegComment)
         if (single != null) single.begComment.foreach(appendBegComment)
-        sb.append("import ").append(treeSyntax(i.ref)).append('.')
+        sb.append("import ").append(refSyntax(i.ref)).append('.')
         if (single != null) {
           val isCurly = single.isCurlyBraced
           val useOuterSpace = isCurly && single
@@ -1152,6 +1152,10 @@ object OrganizeImports {
     def infoNoThrow(implicit doc: SemanticDocument): Option[SymbolInformation] =
       Try(symbol.info).toOption.flatten
   }
+
+  @inline
+  private def refSyntax(ref: Term.Ref)(implicit dialect: Dialect): String =
+    if (!ref.pos.isEmpty) ref.pos.text else treeSyntax(ref)
 
   @inline
   private def treeSyntax(tree: Tree)(implicit dialect: Dialect): String =
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -387,7 +387,7 @@ class OrganizeImports(
     type SortFunc = (StringBuilder, Boolean, List[Importee]) => Unit
     def sortSyntax(f: SortFunc)(imp: Importer): String = {
       implicit val sb = new StringBuilder
-      sb.append(treeSyntax(imp.ref)).append('.')
+      sb.append(refSyntax(imp.ref)).append('.')
       f(sb, imp.isCurlyBraced, imp.importees)
       sb.toString()
     }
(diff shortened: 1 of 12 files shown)


scalafix-tests/input/src/main/scala/test/organizeImports/Scala3SoftKeywordInScala2.scala
```diff
@@ -0,0 +1,10 @@
+/*
+rules = [OrganizeImports]
+OrganizeImports.removeUnused = false
+OrganizeImports.targetDialect = StandardLayout
+*/
+package test.organizeImports
+
+import test.organizeImports.QuotedIdent.`export`.{SimpleSpanProcessor, Other}
+
+object Scala3SoftKeywordInScala2
```
bjaglin: export doesn't seem to be a soft keyword? https://nightly.scala-lang.org/docs/reference/soft-modifier.html

bjaglin: also, could you exercise different keywords rather than just `export`?

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -1153,6 +1153,10 @@ object OrganizeImports {
       Try(symbol.info).toOption.flatten
   }
 
+  @inline
+  private def refSyntax(ref: Term.Ref)(implicit dialect: Dialect): String =
+    if (!ref.pos.isEmpty) ref.pos.text else treeSyntax(ref)
```
bjaglin: While this works for some cases, using the position to gate the prettyprinter on the entire qualifier is probably a partial fix, as `groupedImports = Merge` and `expandRelatives = true` will blur the lines on what's original or not.

Add expect tests with groupedImports & expandRelative and see how the method could deconstruct the ref and call itself.

scalafix-tests/input/src/main/scala/test/organizeImports/Scala3SoftKeywordInScala2.scala
bjaglin: there is something off with naming - export doesn't seem to be a soft keyword, see https://nightly.scala-lang.org/docs/reference/soft-modifier.html

scalafix-tests/shared/src/main/scala/test/organizeImports/QuotedIdent.scala
```diff
@@ -9,4 +9,8 @@ object QuotedIdent {
   }
 
   object `macro`
+  object `export` {
+    object Other
+    object SimpleSpanProcessor
+  }
```
bjaglin, line 16: Could you be more exhaustive and exercise other reserved keywords, like `given`? For given, it will be interesting to document how the keyword behaves vs a package/object `given`.

Given the implementation, sorting seem to happen on the non-escaped version, which is maybe not a problem per-se - when it doubt, test against 0.14.6 to preserve the historical behavior. We can think of a smarter solution with breaking changes, involving a new targetDialect, in a later PR.

bjaglin (commented):
Thanks for the PR. It doesn't look that trivial to get back the previous behavior where output would be compatible with both Scala 2 and Scala 3. I think this can be iterated on (see first comment), so we get a quick regression fix for 0.14.8 due in the coming day(s) as 3.9.0 is going to be announced soon. Then we can take a more serious stab at dialect selection.

subhranshudash13-dotcom:
Thanks for the review, @bjaglin
i have made a note of the changes to be made, I'll go through the `groupedImports = Merge` / `expandRelative = true` cases and see how `refSyntax` should handle those refs. I'll also add a few more cases like `given` and check the behavior against 0.14.6 to make sure we're not changing anything unnecessarily.

I'll update the PR once I've worked through those cases.

scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -387,7 +387,7 @@ class OrganizeImports(
     type SortFunc = (StringBuilder, Boolean, List[Importee]) => Unit
     def sortSyntax(f: SortFunc)(imp: Importer): String = {
       implicit val sb = new StringBuilder
-      sb.append(treeSyntax(imp.ref)).append('.')
+      sb.append(refSyntax(imp.ref)).append('.')
```
bjaglin: Follow up of https://github.com/scalacenter/scalafix/pull/2500#discussion_r3890669177: this is actually not needed, as it changes the order vs non-backticked imports compared to 0.14.6. I have no opinion on what's best, but I'd rather keep behavior consistent across versions.

As a result, I am reverting that change and adding a test to guard that in a separate commit.

bjaglin (approved):
Thanks!

PR #2504, opened 2026-09-02
Fix `RemoveUnused` / `OrganizeImports.removeUnused` for Scala 3.10+

### LINKED ISSUE #2503: Prefix matching rejects Scala 3.10+ for RemoveUnused / OrganizeImports.removeUnused
`RemoveUnused` and `OrganizeImports.removeUnused` do not work with Scala 3.10+ (as can be encountered with 3.10.0-RC1)
Scalafix incorrectly treats Scala **3.10+** as an old 3.1.x compiler that lacks SemanticDB unused diagnostics. 

## Reproducer

When running `./mill -i __.fix` against [Scala CLI](https://github.com/VirtusLab/scala-cli) modules cross-built with **Scala 3.10.0-RC1**, scalafix fails with:

```
[E0] You must use a more recent version of the Scala 3 compiler (3.3.4+)
[E1] "OrganizeImports.removeUnused" is not supported on 3.10.0-RC1 as the compiler is not providing enough information. Please upgrade the Scala compiler to 3.3.4 or greater. Otherwise, run the rule with "OrganizeImports.removeUnused" set to false to organize imports while keeping potentially unused imports.
```

## Root cause

`RemoveUnused` and `OrganizeImports` decide whether unused-warning diagnostics are available in SemanticDB using string prefix matching:

```scala
val diagnosticsAvailableInSemanticdb =
  Seq("3.0", "3.1", "3.2", "3.3.0", "3.3.1", "3.3.2", "3.3.3")
    .forall(v => !config.scalaVersion.startsWith(v))
```

Because `"3.10.0-RC1".startsWith("3.1")`, Scala 3.10 is classified as an old 3.1.x compiler. The same bug would affect any future double-digit minor version.

Gedochao:
Fixes https://github.com/scalacenter/scalafix/issues/2503

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
@@ -157,6 +157,17 @@ object ScalaVersion {
 
   private val yyyyMMdd = DateTimeFormatter.ofPattern("yyyyMMdd")
 
+  implicit val ordering: Ordering[ScalaVersion] =
+    Ordering.by(v =>
+      (v.major.value, v.minor.getOrElse(0), v.patch.getOrElse(0))
+    )
+
+  // unused warnings were only exposed to SemanticDB starting with Scala 3.3.4
+  def unusedDiagnosticsInSemanticdb(version: String): Boolean =
+    from(version).toOption.forall { v =>
+      v.isScala2 || ordering.gteq(v, Patch(Major.Scala3, 3, 4))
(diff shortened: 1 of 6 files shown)


bjaglin (approved):
Thanks! FTR, I will cut 0.14.8 shortly, along with the Scala 3.9.0 announcement.

Gedochao:
> FTR, I will cut 0.14.8 shortly, along with the Scala 3.9.0 announcement.

@bjaglin bump 😅

Issue #2473, opened 2026-07-07
`OrganizeImports` duplicates trailing comment on adjacent wildcard import when reordering


zarthross:
## Summary

`OrganizeImports` duplicates an inline/trailing comment when a wildcard import (`import x._`) that has a trailing comment needs to be reordered relative to an adjacent import (no blank line separating them). The comment ends up printed twice: once as a standalone line and once still attached to the wildcard import.

## Versions

- scalafix: 0.14.7
- Scala: reproduces on both 2.13 and 3.x

## Reproduction

Input:

```scala
/*
rules = [OrganizeImports]
OrganizeImports.removeUnused = false
 */
package test.organizeImports

import c._ // commentC
import b.B

object InlineCommentMoves
```

Expected output:

```scala
package test.organizeImports

import b.B
import c._ // commentC

object InlineCommentMoves
```

Actual output:

```scala
package test.organizeImports

// commentC
import b.B
import c._ // commentC

object InlineCommentMoves
```

## Notes

- This only reproduces when the wildcard import's comment and the next import are **adjacent** (no blank line between them). Adding a blank line between the two imports avoids the bug.
- A non-wildcard import in the same position (e.g. `import c.C // commentC`) does **not** trigger the duplication — only wildcard imports (`import c._`) seem to be affected.
- Likely related to #2267 / #2428 (comment preservation in `OrganizeImports`), possibly a regression or edge case not covered by that fix.

bjaglin:
Thanks for the report!

PR #2507, opened 2026-09-03
OrganizeImports: re-emit unchanged imports verbatim

mlachkar:
Since 0d232c9f, the rule deletes every import and re-inserts a pretty-print of its tree, whether or not it changed anything. It then has to recover, one by one, each formatting detail the printer discards: spaces inside curly braces (#2427), trailing commas (#2468), wrapped single importees (reported on #2468)... and there is always one more (#2500, backticks).

An import that is already organized and wrapped over several lines is now re-emitted from its source text, byte for byte. "Already organized" means: after applying the whole configuration (groupedImports, expandRelative, coalesceToWildcardImportThreshold, importSelectorsOrder, removeUnused...), the resulting importer is structurally identical to the source one -- same qualifier, same importees, same order -- and needs no syntax migration to the target dialect (`_` to `*`, `=>` to `as`). The configuration still decides the *structure* of every import; verbatim only preserves the *layout* of wrapped imports that are already in the configured structure. Single-line imports, and everything the rule rewrites, go through the pretty printer as before, so `import a.{ B }` is still normalized to `import a.B`.

Principle: an already-organized import must produce no diff. How an untouched import is formatted is scalafmt's business, not this rule's; the two must never fight over the same line.

For rewritten imports, the wrapping and trailing comma of the source import are kept when all importees come from a single wrapped source import, even if some importees are synthesized (coalesced wildcard).

What changes for users: a wrapped import the rule has no reason to touch keeps its exact layout (indentation, alignment, trailing comma, spacing). No existing expected output changes.

 Fixes the single-importee case reported on #2468 and the case reported in go through the pretty printer and need #2500 for their qualifier. #2473 (duplicated trailing comment on reorder) is unrelated and not addressed.

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -773,63 +773,105 @@ class OrganizeImports(
         def appendEndComment(pc: Tree.Comments): Unit =
           if (commentsPrinted.add(pc))
             pc.values.foreach(x => sb.append(' ').append(x.syntax))
+        def appendCurly(
+            isMultiline: Boolean,
+            useOuterSpace: Boolean,
+            trailingComma: Boolean
+        )(appendImportees: String => Unit): Unit = {
+          sb.append('{')
+          val sep = if (isMultiline) "\n  " else " "
+          if (isMultiline) sb.append(sep)
+          else if (useOuterSpace) sb.append(' ')
+          appendImportees(sep)
+          if (isMultiline) {
+            if (trailingComma) sb.append(',')
+            sb.append('\n')
+          } else if (useOuterSpace) sb.append(' ')
+          sb.append('}')
+        }
 
         val single =
           if (i.importees.lengthCompare(1) == 0) i.importees.head else null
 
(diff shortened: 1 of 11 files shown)


bjaglin:
Thanks for the PR! While the problem is real and the intent clear, I am wondering if this is the right fix. I would instead try to effectively revert https://github.com/scalacenter/scalafix/pull/2427 that seems to cause more troubles than good instead of trying to patch it. I am going to explore that.

bjaglin:
> Thanks for the PR! While the problem is real and the intent clear, I am wondering if this is the right fix. I would instead try to effectively revert #2427 that seems to cause more troubles than good instead of trying to patch it. I am going to explore that.

OK, my bad, after digging a bit more with Opus, I understood the implementation was what I was suggesting - use the scalameta prettyprinter only when needed.

So I ported your commit to https://github.com/scalacenter/scalafix/pull/2509:
- I rebased the implementation against https://github.com/scalacenter/scalafix/pull/2500 and a follow-up fix.
- I decided not to support the behavior tested by CoalesceMultiLineTrailingComma, which seemed like an edge case that didn't require the added complexity. I did keep the test with a different output file though, to document that choice.

Thanks again, I am now confident the upcoming release will bring back the 0.14.6 behavior, with improvements.

PR #2508, opened 2026-09-05
don't target all sources when --diff reports only deletions

### LINKED ISSUE #2505: --diff-base: empty diff (e.g. only deletions) disables diff filtering entirely, so all files get linted
## Describe the bug

When `--diff` / `--diff-base` is used and the git diff contains no added or modified files — e.g. a change that **only deletes files**, or no changes at all — scalafix lints (and rewrites) **all** files instead of none.

## Root cause

1. `JGitDiff` drops `DELETE` entries, so a delete-only diff produces `diffs = Nil`:
   https://github.com/scalacenter/scalafix/blob/v0.14.2/scalafix-cli/src/main/scala/scalafix/internal/jgit/JGitDiff.scala#L58
2. `FullDiffDisable.isEmpty` is defined as `diffs.isEmpty`, which makes an empty diff indistinguishable from "no diff requested" (`EmptyDiff`):
   https://github.com/scalacenter/scalafix/blob/v0.14.2/scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala#L29
3. `EscapeHatch.filter` short-circuits when `isEmpty` and returns all patches/diagnostics **without any diff-based filtering**:
   https://github.com/scalacenter/scalafix/blob/v0.14.2/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala#L57
   https://github.com/scalacenter/scalafix/blob/v0.14.2/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala#L119-L120

## Reproduction

```bash
# on a branch that only deletes a file relative to master
git checkout -b delete-only
git rm src/main/scala/some/pkg/Foo.scala
git commit -m "delete Foo"

scalafix --diff-base master SomeLintRule
```

**Expected:** nothing is linted — no file was added or modified, so every position in every file is "disabled" by the diff.

**Actual:** every file in the project is linted (and rewrite rules are applied everywhere), because the empty diff takes the `EscapeHatch.isEmpty` fast path that skips diff filtering entirely.

## Suggested fix

Distinguish "no diff requested" from "diff requested but empty". For example:

- make `FullDiffDisable.isEmpty` return `false` even when `diffs` is empty — the slow path in `EscapeHatch.filter` then correctly suppresses everything via `isDisabled`, or
- have `EscapeHatch.isEmpty` take into account whether a diff filter was requested at all.

## Additional context

- The same symptom was observed in #1714 (scalafix processing/editing files on a branch with no changes relative to the diff base).
- Hit in practice via sbt-scalafix 0.14.2 in CI: a pull request that only deleted one Scala file caused the entire repository to be linted, failing the build on pre-existing violations in untouched files.

SatvikMishra08:
## Problem

When `--diff` / `--diff-base` is used and the git diff contains no added or modified files (e.g. delete-only changes), Scalafix lints **all** files instead of none.

`JGitDiff` drops `DELETE` entries, so a delete-only diff produces `diffs = Nil`. `FullDiffDisable.isEmpty` was defined as `diffs.isEmpty`, making that case indistinguishable from `EmptyDiff` ("no diff requested"). `EscapeHatch.filter` then short-circuits on `isEmpty` and returns all patches/diagnostics without diff filtering.

Closes #2505

## Fix

Implement option 1 as endorsed by @bjaglin: make `FullDiffDisable.isEmpty` always return `false`, even when `diffs` is empty. `EmptyDiff` remains the only "no diff requested" sentinel. The slow path in `EscapeHatch.filter` then correctly suppresses everything via `isDisabled`.

## Tests

- `EscapeHatchSuite`: `FullDiffDisable(Nil)` is not empty, disables untouched inputs, and keeps `EscapeHatch` non-empty.
- `CliGitDiffSuite`: delete-only branch (`keep.scala` + `gone.scala`, then `rm gone`) under `--diff` does not report `keepVar` / lint all files.

Verified locally:

```text
unit2_13_18/testOnly scalafix.tests.core.EscapeHatchSuite scalafix.tests.cli.CliGitDiffSuite
# 21 tests, all passed (including both new cases)
```

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala
@@ -16,19 +16,17 @@ object DiffDisable {
 sealed trait DiffDisable {
   def isDisabled(position: Position): Boolean
   def isDisabled(file: Input): Boolean
-  def isEmpty: Boolean
+  def canSkip: Boolean
 }
 
 object EmptyDiff extends DiffDisable {
-  override def isEmpty: Boolean = true
+  override def canSkip: Boolean = true
   def isDisabled(position: Position): Boolean = false
   def isDisabled(file: Input): Boolean = false
 }
 
 class FullDiffDisable(diffs: List[GitDiff]) extends DiffDisable {
-  // Always non-empty: distinguish "diff requested but empty" (delete-only /
-  // no changes) from EmptyDiff ("no diff requested"). See #2505.
-  override def isEmpty: Boolean = false
+  override def canSkip: Boolean = false
   private val newFiles: Set[String] = diffs.collect { case NewFile(path) =>
     path.toAbsolutePath.toString
   }.toSet
--- a/scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala
@@ -26,7 +26,9 @@ object EmptyDiff extends DiffDisable {
 }
 
 class FullDiffDisable(diffs: List[GitDiff]) extends DiffDisable {
-  override def isEmpty: Boolean = diffs.isEmpty
+  // Always non-empty: distinguish "diff requested but empty" (delete-only /
+  // no changes) from EmptyDiff ("no diff requested"). See #2505.
+  override def isEmpty: Boolean = false
   private val newFiles: Set[String] = diffs.collect { case NewFile(path) =>
     path.toAbsolutePath.toString
   }.toSet
--- a/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/patch/EscapeHatch.scala
@@ -119,7 +119,7 @@ class EscapeHatch private (
   }
 
   def isEmpty: Boolean =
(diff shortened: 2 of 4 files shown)


scalafix-core/src/main/scala/scalafix/internal/diff/DiffDisable.scala
```diff
@@ -26,7 +26,9 @@ object EmptyDiff extends DiffDisable {
 }
 
 class FullDiffDisable(diffs: List[GitDiff]) extends DiffDisable {
-  override def isEmpty: Boolean = diffs.isEmpty
+  // Always non-empty: distinguish "diff requested but empty" (delete-only /
+  // no changes) from EmptyDiff ("no diff requested"). See #2505.
+  override def isEmpty: Boolean = false
```
bjaglin, line 31: I think this deserves a rename rather than a comment to clarify the fix. It's hard because the class name already carries negation (but I think a refactoring is beyond the scope of this PR). I would suggest something that reads better on the call site:
```suggestion
  override def canSkip: Boolean = false
```

bjaglin:
Thanks for the PR! Please make sure to run scalafmt before pushing.

bjaglin:
@SatvikMishra08 I will cut a release soon - any chance you can take the feedback into account?

SatvikMishra08:
Thanks for the ping — just pushed the rename: `isEmpty` → `canSkip` so a delete-only `--diff` isn't treated like `EmptyDiff`. Ready for the release whenever you are.

PR #2509, opened 2026-09-05
fix(organizeImports): preserve source spelling and layout of already-organized imports

bjaglin:
Since 0d232c9f (first released in 0.14.7), the rule deletes every import and re-prints it from the tree via `reprint()`, which prints from *structure* under a *dialect* and so discards anything the source had that the tree does not model — identifier spelling and layout. Each loss has been patched back one at a time (#2427 brace spacing, #2468 trailing commas, #2500 qualifier backticks…). This PR addresses the two remaining axes, keyed on a single invariant instead of more heuristics: **a node the pipeline rewrote is a `copy()` and carries no position; a node it left alone keeps its parsed position.** So position presence tells us, per node, when the source text can be re-emitted as is.

### 1. Emit escaped names from source, not reprint, everywhere

`refSyntax` (#2500) preserved backticks for an importer's *qualifier* only. This generalizes it to every emitted leaf name (`nameSyntax`), used by both the ref (`refSyntax`) and the importee (new `importeeSyntax`, covering `Rename`/`Unimport`/`Given` by rebuilding from parts rather than taking a composite span). `treeSyntax` (`reprint()`) stays for comparison keys only (dedup, grouping, sorting) — those must stay dialect-normalized so `A => B` and `A as B` dedup to one entry and sort order does not depend on backticks.

### 2. Re-emit already-organized wrapped imports verbatim

A wrapped import the rule leaves untouched is currently collapsed onto one line by the pretty-printer, fighting scalafmt over the same lines. Now such an import is re-emitted from its own source text, so its layout (indentation, alignment, trailing comma) is the formatter's business, not this rule's.

`isUnchangedFromSource` = has a range position (so it was not restructured — same qualifier, importees, order), spans multiple lines (single-line imports are still normalized, e.g. `import a.{ B }` -> `import a.B`), and needs no dialect migration (`=>`/`as`, `_`/`*`). No structural comparison, no `originalPrototype()` navigation. The reconstruction path is untouched: anything the rule rewrites — reordered, merged, coalesced, exploded, expanded — still goes through the pretty-printer as before (a coalesced multi-line import, for instance, re-flows to one line and the formatter re-wraps it).

Fixtures and code for this axis are adapted from #2507.

## New expect tests across releases

| expect test | 0.14.6 | 0.14.7 | this PR |
|---|---|---|---|
| CoalesceMultiLineTrailingComma | ✅ | ✅ | ✅ |
| SingleImporteeSplitInline | ✅ | ✅ | ✅ |
| SingleImporteeWrapped | ✅ | ❌ | ✅ |
| SingleImporteeWrappedNoComma | ✅ | ❌ | ✅ |
| NestedPackageWrapped | ❌ | ❌ | ✅ |
| VerbatimInteriorComment | ❌ | ❌ | ✅ |
| MergeScala3KeywordImportee | ❌ | ❌ | ✅ |

### DIFF
--- a/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
+++ b/scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
@@ -791,7 +791,7 @@ class OrganizeImports(
             sb.append('{')
             if (useOuterSpace) sb.append(' ')
           }
-          sb.append(treeSyntax(single))
+          sb.append(importeeSyntax(single))
           if (isCurly) {
             if (useOuterSpace) sb.append(' ')
             sb.append('}')
@@ -822,7 +822,7 @@ class OrganizeImports(
             if (sb.length > sblen) sb.append(',').append(sep)
             val proto = i2.originalPrototype()
             proto.begComment.foreach(appendBegComment)
-            sb.append(treeSyntax(i2))
+            sb.append(importeeSyntax(i2))
             proto.endComment.foreach(appendEndComment)
           }
           if (isMultiline) {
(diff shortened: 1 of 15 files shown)


scalafix-rules/src/main/scala/scalafix/internal/rule/OrganizeImports.scala
```diff
@@ -1152,22 +1152,72 @@ object OrganizeImports {
...
-      case Term.Select(qual, name) =>
-        refSyntax(qual) + "." + treeSyntax(name)
-      case _ =>
-        treeSyntax(ref)
+      case t: Term.Name => nameSyntax(t)
+      case Term.Select(qual, name) => refSyntax(qual) + "." + nameSyntax(name)
+      case _ => treeSyntax(ref)
     }
 
-  @inline
```
bjaglin, line 1168: we don't compile with optimizations so this hint was useless

PR #2512, opened 2026-09-08
Fix `ExplicitResultTypes` for Scala 3.10+

### LINKED ISSUE #2511: Prefix matching rejects Scala 3.10+ for `ExplicitResultTypes`
`ExplicitResultTypes` does not work with Scala 3.10+ (as can be encountered with 3.10.0-RC1).
Scalafix incorrectly treats Scala **3.10+** as an old 3.1.x compiler predating 3.3.0.
This is very similar to https://github.com/scalacenter/scalafix/issues/2503

## Reproducer

Given a project compiled with Scala 3.10.0-RC1 and the following `.scalafix.conf`:

```
rules = [
  ExplicitResultTypes
]
ExplicitResultTypes.fetchScala3CompilerArtifactsOnVersionMismatch = true
```

scalafix fails with:

```
error: The ExplicitResultTypes rule requires Scala 3 target sources to be compiled with Scala 3.3.0 or greater, but they were compiled with 3.10.0-RC1. To fix this problem, either remove ExplicitResultTypes from .scalafix.conf or upgrade the compiler in your build.
```

## Root cause

`ExplicitResultTypes` (Scala 3 variant) decides whether the target sources were compiled with a
sufficiently recent compiler using string prefix matching:

https://github.com/scalacenter/scalafix/blob/main/scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala#L63-L69

Because `"3.10.0-RC1".startsWith("3.1")`, Scala 3.10 is classified as an old 3.1.x compiler. The same
bug would affect any future double-digit minor version.

Gedochao:
Fixes https://github.com/scalacenter/scalafix/issues/2511

### DIFF
--- a/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
+++ b/scalafix-core/src/main/scala/scalafix/internal/config/ScalaVersion.scala
@@ -168,6 +168,11 @@ object ScalaVersion {
       v.isScala2 || ordering.gteq(v, Patch(Major.Scala3, 3, 4))
     }
 
+  def explicitResultTypesSupported(version: String): Boolean =
+    from(version).toOption.forall { v =>
+      !v.isScala3 || ordering.gteq(v, Patch(Major.Scala3, 3, 0))
+    }
+
   def from(version: String): Try[ScalaVersion] = {
     version match {
       case NightlyVersion(major, minor, patch, rc, date, shortSha1) =>
--- a/scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
+++ b/scalafix-rules/src/main/scala-3/scalafix/internal/rule/ExplicitResultTypes.scala
@@ -7,6 +7,7 @@ import scala.meta.*
 
 import dotty.tools.pc.ScalaPresentationCompiler
 import metaconfig.Configured
+import scalafix.internal.config.ScalaVersion
 import scalafix.internal.pc.PresentationCompilerTypeInferrer
 import scalafix.patch.Patch
 import scalafix.v1.*
@@ -61,7 +62,7 @@ final class ExplicitResultTypes(
           s"To fix this problem, either remove ExplicitResultTypes from .scalafix.conf or make sure Scalafix is loaded with $majorMinorScalaVersion."
       )
     } else if (
-      Seq("3.0", "3.1", "3.2").exists(v => config.scalaVersion.startsWith(v))
+      !ScalaVersion.explicitResultTypesSupported(config.scalaVersion)
     ) {
       Configured.error(
         s"The ExplicitResultTypes rule requires Scala 3 target sources to be compiled with Scala 3.3.0 or greater, but they were compiled with ${config.scalaVersion}. " +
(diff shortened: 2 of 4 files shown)


tgodzik (changes requested):
Looks like a wrong version of the presentation compiler is now used:
```
[error] (integration3_3_8 / Test / executeTests) java.lang.UnsupportedClassVersionError: dotty/tools/pc/ScalaPresentationCompiler has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error] (integration3_6_4 / Test / executeTests) java.lang.UnsupportedClassVersionError: dotty/tools/pc/ScalaPresentationCompiler has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error] (integration3_5_2 / Test / executeTests) java.lang.UnsupportedClassVersionError: dotty/tools/pc/ScalaPresentationCompiler has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error] (integration3_7_4 / Test / executeTests) java.lang.UnsupportedClassVersionError: dotty/tools/pc/ScalaPresentationCompiler has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[error] Total time: 563 s (0:09:23.0), completed Sep 8, 2026 12:06:39 PM
```
all of these should work on JDK 17-

tgodzik (approved):
LGTM!

Gedochao:
@bjaglin any chance for a 0.14.9 release soon? 🙃

bjaglin:
I just tagged 0.14.8, so I'd like to wait for a few pending fixes to land first. You could use a [snapshot](https://scalacenter.github.io/scalafix/docs/users/installation.html#snapshot) for the time being?

Are you blocked in https://github.com/VirtusLab/scala-cli/pull/4455? Is the impact on the build itself or passed down the clients because ExplicitResultResultTypes is ran by default on the fix command?

Gedochao:
The block on #4455 is passed to users. 
I could disable tests for `scala-cli fix` running this rule for Scala 3.10+ temporarily to work around it.
I mostly worry that fewer people will fully try Scala 3.10 RCs when they see scalafix failing their builds.

bjaglin:
I see... What about disabling ExplicitResultTypes per default in scala-cli on Scala 3.10+? If that's not possible, I can look into tagging a new version later this week.

Gedochao:
> What about disabling ExplicitResultTypes per default in scala-cli on Scala 3.10+?

I'd rather not do that, as I'd have to hardcode it in a given Scala CLI version... Sorry if it's a bother 😅 
I hope not to find any further problems like this (I have looked!)

bjaglin:
@Gedochao I'll cut 0.14.9 this week-end

bjaglin:
https://github.com/scalacenter/scalafix/releases/tag/v0.14.9

PR #2515, opened 2026-09-11
Add `--dry-run` to report linter errors without rewriting files

### LINKED ISSUE #956: Provide an option to look at linting errors without doing any rewrites
**Problem**

When writing linting rules which can possibly do a autofix/rewrite, it would be nice to have an option to run all linting rules to see what the errors were without doing a rewrite. And then run  scalafix again to rewrite the linting errors which have an autofix.

For e.g: Let's say I want to write a linter rule which enforces javaconverters over java conversions. It would be good if it let the dev know other helpful information in linting message and an option to autofix it. currently we either do a check  (--check flag) which prints patch to terminal (not pleasant) or automatic rewrite (with out --check flag)

**Proposal**

Add an option `--dry-run` (name upfor debate), which prints out only `Patch.Lint` messages  to terminal without doing any rewrites and Optionally, letting users know there is a possibility of fixing few linting errors.

jozanek:
Closes #956.

Reports linter errors without writing to files or printing a diff:

```sh
$ scalafix --dry-run
src/main/scala/Main.scala:4:18: error: [DisableSyntax.noSemicolons] semicolons are disabled
    println("ok");
                 ^
2 files can be fixed by running scalafix without --dry-run
```

Unlike `--check`, pending rewrites are summarised as a count rather than a diff and don't fail the run on their own. Linter errors still exit `LinterError`.

Also adds `ScalafixMainMode.DRY_RUN` for the Java interfaces.

### DIFF
--- a/docs/users/installation.md
+++ b/docs/users/installation.md
@@ -209,6 +209,22 @@ To automatically enforce that Scalafix has been run on all sources, use
 
 Use `scalafixAll --check`  to enforce Scalafix on your entire project.
 
+### Report linter errors only
+
+To look at linter errors without rewriting any file, use `scalafix --dry-run`.
+Unlike `--check`, it does not print the diff that a rewrite would produce, and
+files that could be fixed do not fail the invocation on their own - only linter
+errors do. Files with pending rewrites are reported as a count, so you know
+that running `scalafix` without `--dry-run` would change something.
+
+```sh
+$ scalafix --dry-run
+src/main/scala/Main.scala:4:18: error: [DisableSyntax.noSemicolons] semicolons are disabled
+    println("ok");
+                 ^
+2 files can be fixed by running scalafix without --dry-run
+```
+
 ### Cache in CI
 
 To avoid binary compatibility conflicts with the sbt classpath
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
@@ -71,6 +71,12 @@ case class Args(
     @Description("Print fixed output to stdout instead of writing in-place.")
     stdout: Boolean = false,
     @Description(
+      "Report linter errors without writing to files or printing a diff, " +
+        "reporting how many files could be fixed by a subsequent run. " +
+        "Unlike --check, auto-fixable violations alone do not fail the run."
+    )
+    dryRun: Boolean = false,
+    @Description(
       "If set, only apply scalafix to added and edited files in git diff against the master branch."
     )
     diff: Boolean = false,
--- a/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
+++ b/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
@@ -275,7 +275,13 @@ object MainOps {
         args.config.reporter.lint(diag)
       }
     }
-    if (args.args.check) {
+    if (args.args.dryRun) {
+      // Lint diagnostics were already reported above. Report nothing about
+      // rewrites, and leave it to `run` to count and clear this TestError:
+      // --dry-run reports, it does not assert like --check does.
+      if (fixed == input.text) ExitStatus.Ok
+      else ExitStatus.TestError
+    } else if (args.args.check) {
       if (fixed == input.text) {
         ExitStatus.Ok
       } else {
@@ -407,6 +413,7 @@ object MainOps {
     val N = files.length
     val width = N.toString.length
     var exit = discoveryError
+    var fixable = 0
 
     args.rules.rules.foreach(_.beforeStart())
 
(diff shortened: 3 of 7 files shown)


scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
```diff
@@ -417,12 +424,26 @@ object MainOps {
...
+      exit = ExitStatus.merge(exit, status)
     }
 
     args.rules.rules.foreach(_.afterComplete())
 
     val result = adjustExitCode(args, exit, files)
+    if (fixable > 0) {
+      val noun = if (fixable == 1) "file" else "files"
+      args.args.out.println(
+        s"$fixable $noun can be fixed by running scalafix without --dry-run"
```
bjaglin: Nit: I find "would" more aligned with the dry run concept
```suggestion
        s"$fixable $noun would be fixed by running scalafix without --dry-run"
```

bjaglin: Besides, I am concerned that the user might associate this line with the linter reports above if the number matches. Could we prepend a "Also, " when there were lint reported to reduce ambiguity? If so, docs should be updated accordingly.

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
```diff
@@ -70,6 +70,12 @@ case class Args(
     check: Boolean = false,
     @Description("Print fixed output to stdout instead of writing in-place.")
     stdout: Boolean = false,
+    @Description(
+      "Report linter errors without writing to files or printing a diff, " +
+        "reporting how many files could be fixed by a subsequent run. " +
```
bjaglin: I think this could be made more simple
```suggestion
      "Report linter errors and how many files would be fixed by a subsequent run. " +
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainMode.java
```diff
@@ -32,5 +32,14 @@ public enum ScalafixMainMode {
     /**
      * Use when the client triggers the run as a side effect of something else, as opposed to an explicit, interactive invocation. Write fixed contents in-place, with a custom configuration if it exists.
      */
-    IN_PLACE_TRIGGERED
+    IN_PLACE_TRIGGERED,
+
+    /**
+     * Report linter errors only, without printing a diff of the fixed contents.
```
bjaglin: For consistency with the CLI wording
```suggestion
     * Report linter errors and how many files would be fixed by a subsequent run.
```

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainMode.java
```diff
@@ -32,5 +32,14 @@ public enum ScalafixMainMode {
...
+     *
+     * Unlike {@link #CHECK}, files that could be fixed do not cause an error to be reported.
```
bjaglin: Nit: I find "would" more aligned with the dry run concept
```suggestion
     * Unlike {@link #CHECK}, files that would be fixed do not cause an error to be reported.
```

docs/users/installation.md
```diff
@@ -209,6 +209,22 @@ To automatically enforce that Scalafix has been run on all sources, use
...
+that running `scalafix` without `--dry-run` would change something.
+
+```sh
+$ scalafix --dry-run
+src/main/scala/Main.scala:4:18: error: [DisableSyntax.noSemicolons] semicolons are disabled
+    println("ok");
+                 ^
+2 files can be fixed by running scalafix without --dry-run
+```
+
```
bjaglin: I think the title is misleading because the purpose of this flag is to see how many rewrites we would get, summarized. I think semantics are so tricky that I wouldn't document it here (just like stdout) and just let power users discover it (via CTRL+F) through the CLI help at the end of this page.
```suggestion
```

bjaglin (commented):
Thanks for the PR. While I agree `dry-run` is the best of a bad bunch naming-wise, I think we should be very sharp on the docs of that flag.

scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala
jozanek: Done

scalafix-cli/src/main/scala/scalafix/internal/v1/Args.scala
jozanek: Simplified

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainMode.java
jozanek: Aligned

scalafix-interfaces/src/main/java/scalafix/interfaces/ScalafixMainMode.java
jozanek: Changed

docs/users/installation.md
jozanek: Removed

bjaglin:
Thanks!

PR #2519, opened 2026-09-18
fix(cli): support repeated classpath arguments (#2501)

### LINKED ISSUE #2501: cli: repeated --[tool-]classpath arguments are silently ignored
I have some source files, and a JAR with their compilation results (including `META-INF/semanticdb/...`). I can, therefore, run a semantic rule:
```
$ scalafix_2.13 --check --scalac-options=-Xsource:3 --rules ExplicitResultTypes --files common/secrets/Secret.scala --classpath bazel-bin/common/secrets/secrets.jar
```
This succeeds. If, however, I add a `--classpath` with any JAR that does **not** contain `META-INF/semanticdb/...` files, I get an error:
```
$ scalafix_2.13 --check --scalac-options=-Xsource:3 --rules ExplicitResultTypes --files common/secrets/Secret.scala --classpath bazel-bin/common/secrets/secrets.jar --classpath ~/.ivy2/cache/junit/junit/jars/junit-4.13.jar
error: SemanticDB not found: common/secrets/Secret.scala
```
This is surprising, because the `--help` says:
```
  --classpath Classpath (default: "<classpath>")
    Full classpath of the files to fix, required for semantic rules. The source
    files that should be fixed must be compiled with semanticdb-scalac.
    Dependencies are required by rules like ExplicitResultTypes, but the
    dependencies do not need to be compiled with semanticdb-scalac.
```

As I understand it, this fragment:

> the dependencies do not need to be compiled with semanticdb-scalac.

means that I don't have to provide SemanticDB resources for libraries that were not compiled from the inspected files.

There are reasons to provide non-SemanticDB-enabled JARs as `--classpath`. For example, getting `SymbolInformation` for a method seems to require the result type etc. in `--classpath`, even though that result may be a library file. The only workaround I can think of is recompiling all the libraries with SemanticDB support, which is... rough.

```
$ scalafix_2.13 --version
0.14.7
```

subhranshudash13-dotcom:
Fixes #2501

### Summary
When passing multiple `--classpath` or `--tool-classpath` arguments on the CLI, previous values were silently overwritten because `Classpath` and `URLClassLoader` are custom non-`Iterable` types and did not have the `@Repeated` annotation for Metaconfig's CLI parser.

This PR:
1. Adds `@Repeated` to `classpath` and `toolClasspath` in `Args.scala`.
2. Updates `classpathDecoder` to use `ConfDecoder[List[String]]`, accumulating paths from repeated arguments while preserving path separator splitting (`:` / `;`).
3. Adds unit and integration tests covering accumulated classpath resolution and semantic rule execution.

###Tests Added : 

Unit Tests (ArgsSuite.scala):
repeated --classpath arguments are accumulated: Validates that multiple --classpath arguments (e.g. --classpath a.jar --classpath b.jar) accumulate all entries in order into Args.classpath.
repeated --classpath arguments with path separator are accumulated: Validates that mixing multiple --classpath flags with path separators (e.g. --classpath a.jar:b.jar --classpath c.jar) correctly resolves all paths.
repeated --tool-classpath arguments are accumulated: Validates that multiple --tool-classpath flags produce a URLClassLoader containing URLs for all specified JARs.
Integration Test (ScalafixArgumentsSuite.scala):

repeated --classpath in withParsedArguments: Tests the full semantic rule lifecycle (RemoveUnused) by passing target classes (with SemanticDB) and dependency jars (scalaLibrary) via separate repeated --classpath arguments and asserting successful execution.

###Verification :
sbt "unit3_8_4 / testOnly scalafix.tests.config.ArgsSuite" (6/6 tests passed)
sbt "integration3_8_4 / testOnly scalafix.tests.interfaces.ScalafixArgumentsSuite" (21/21 tests passed)
./bin/scalafmt --test (Passed with 0 diffs)
sbt versionPolicyCheck (Passed across all Scala 2.12, 2.13, and 3.x modules)
sbt ci-docs (Documentation compiled successfully)


subhranshudash13-dotcom:
hey @bjaglin, pls review and let me know for any changes to be done from my side.

